PR #25, opened 2024-01-02
include go tests in search regex

0x2493:


budevg (approved):

0x2493:
Nested virtualization is disabled in nemax fic, so rdma-test is expected to fail

PR #85, opened 2024-01-09
NBS-4358: fix UnadjustedBlockCount; added test_override_storage_discovery_config

sharpeye:


cloud/blockstore/tests/storage_discovery/test.py
```diff
@@ -556,3 +557,67 @@ def test_change_layout(
...
+
+def test_override_storage_discovery_config(
+        nbs,
+        agent_id,
+        data_path,
+        disk_agent_configurator,
+        disk_agent_dynamic_config):
+
+    # global config
+    disk_agent_configurator.files["disk-agent"] = disk_agent_dynamic_config
```
budevg, line 570: it's not so clear that the config from disk_agent_dynamic_config is different then the config you override in cms ... maybe better generate the config explicitly in the test and then modify it and configure the cms with the changed version, this way whoever reads the test can notify the intention to override the dynamic config with the config in cms

budevg (changes requested):

budevg (approved):

umed (approved):

budevg (approved):

PR #94, opened 2024-01-10
NBSNEBIUS-79: Updating build documentation

dvrazumov:
Need to update build documentation & adjust the documentation for running the example

### DIFF
--- a/example/0-setup.sh
+++ b/example/0-setup.sh
@@ -36,18 +36,6 @@ for dir in $DATA_DIRS; do
     ln -svfT "$PERSISTENT_TMP_DIR/$dir" "$BIN_DIR/$dir"
 done
 
-<<<<<<< HEAD
-# check symlinks
-for bin in ydbd nbsd blockstore-nbd blockstore-client diskagentd
-do
-  if ! test -f "$BIN_DIR/$bin"; then
-    echo "$bin not found"
-    exit 1
-  fi
-done
-
-=======
->>>>>>> 53ebe7b85 (Updating build documentation)
 generate_cert() {
     local pass="pass123"
     local name="$1"
--- a/example/0-setup.sh
+++ b/example/0-setup.sh
@@ -9,27 +9,9 @@ find_bin_dir() {
     readlink -e `dirname $0`
 }
 
-find_blockstore_all_dir() {
-    readlink -e `dirname $0`/../cloud/blockstore/buildall
-}
-
 BIN_DIR=`find_bin_dir`
-BLOCKSTORE_ALL_DIR=`find_blockstore_all_dir`
 PERSISTENT_TMP_DIR=${PERSISTENT_TMP_DIR:-$HOME/tmp/nbs}
 source ./prepare_binaries.sh || exit 1
-
-BUILD_FILES=" \
-    contrib/ydb/apps/ydbd/ydbd \
-    cloud/blockstore/apps/server/nbsd \
-    cloud/blockstore/apps/disk_agent/diskagentd \
-    cloud/blockstore/apps/client/blockstore-client \
-    cloud/blockstore/tools/nbd/blockstore-nbd \
-    "
-
-# create symlinks
-for file in $BUILD_FILES; do
-    ln -svf $BLOCKSTORE_ALL_DIR/$file $BIN_DIR/
-done
 
 for dir in $DATA_DIRS; do
     mkdir -p "$PERSISTENT_TMP_DIR/$dir"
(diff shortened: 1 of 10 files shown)


example/5-create_and_attach_disk.sh
```diff
@@ -89,7 +89,7 @@ SOCK="$BIN_DIR/$id.sock"
 
 sudo modprobe nbd
 touch $SOCK
-sudo $BIN_DIR/$NBD --device-mode endpoint --disk-id $id --access-mode rw \
+sudo LD_LIBRARY_PATH=$LD_LIBRARY_PATH $BIN_DIR/$NBD --device-mode endpoint --disk-id $id --access-mode rw \
```
budevg, line 92: don't you need to set LD_LIBRARY_PATH previously to make this useful ...
probably something like
```
export LD_LIBRARY_PATH=$BIN_DIR
```

budevg (changes requested):

example/5-create_and_attach_disk.sh
dvrazumov: LD_LIBRARY_PATH is set here in [example/README.md](https://github.com/ydb-platform/nbs/pull/94/files#diff-bc501e6e1e9420017b2d954e834fd8c3afb7daf9e2eb780c46c491099a7c2f0eR22)

There are some difficulties here;
* ```ya``` and ```ninja``` produce output in different folders
* $BIN_DIR does not contain libraries, but only binaries
* There is no the only one library folder with ```ya``` tool, instead the libraries are copied alongside with binraies

If you have an idea how to make it more clear please post it here

budevg: we can also patch rpath in binaries instead of using the env variable
for example
```
apt install patchelf
NBSD=readlink -f ../cloud/blockstore/apps/server/nbsd
patchelf --set-rpath $(dirname $NBSD):'$ORIGIN' $NBSD
```

budevg (approved):

qkrorlqr (approved):

example/5-create_and_attach_disk.sh
dvrazumov: I decided to leave the option with LD_LIBRARY_PATH, but put it in a separate file. It now looks clear for me. @budevg , could you check please

example/prepare_binaries.sh
```diff
@@ -0,0 +1,34 @@
...
+do
+  if ! test -f $bin; then
+    echo "$bin not found, build all targets first"
+    return 1
+  fi
+done
+
+function ydbd () {
+  LD_LIBRARY_PATH=$(dirname $YDBD_BIN) $YDBD_BIN $@
+}
```
budevg, line 18: add new line to separate functions here and below

example/prepare_binaries.sh
```diff
@@ -0,0 +1,34 @@
...
+
+for bin in $YDBD_BIN $NBSD_BIN $BLOCKSTORE_CLIENT_BIN $DISK_AGENT_BIN $BLOCKSTORE_NBD_BIN
+do
+  if ! test -f $bin; then
+    echo "$bin not found, build all targets first"
+    return 1
+  fi
+done
+
+function ydbd () {
```
budevg, line 16: the correct syntax for bash function is either
```bash
function myfunc {
}
```
or
```bash
myfunc () {
}
```
no need to have both `function` and `()`

example/prepare_binaries.sh
```diff
@@ -0,0 +1,34 @@
+export NBS_ROOT=".."
```
budevg, line 1: no need to `export NBS_ROOT` 
you only export when you want to pass it as environment to some other program you invoke

by the way instead of redefining all possible binaries you can
`export LD_LIBRARY_PATH=...` once in this file (all the .so are copied to multiple location so it's enough to use some location for example the NBSD_BIN directory)

After that you source this file and you can invoke all the binaries with LD_LIBRARY_PATH passed correctly to the process and without adding the `LD_LIBRARY_PATH=...` prefix before the command.
For the sudo invoked process you can use `sudo -E <cmd>` to preserve environment and pass it to process invoked with sudo

example/0-setup.sh
```diff
@@ -35,6 +36,7 @@ for dir in $DATA_DIRS; do
     ln -svfT "$PERSISTENT_TMP_DIR/$dir" "$BIN_DIR/$dir"
 done
 
+<<<<<<< HEAD
```
budevg, line 39: looks like you have multiple merge conflicts in the commit
search for `<<<<<<<` all around

budevg (changes requested):

example/prepare_binaries.sh
dvrazumov: According to [manual](https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions): 
> This defines a shell function named fname. The reserved word function is optional. If the function reserved word is supplied, the parentheses are optional. 

So using both `function` word and parentheses is also correct syntax

example/0-setup.sh
dvrazumov: Sorry, fixed that

budevg (dismissed):

debnatkh (approved):

budevg (approved):

PR #115, opened 2024-01-11
NBS-4824: [Disk Manager] Add more logs for s3

BarkovBG:


cloud/disk_manager/internal/pkg/persistence/s3_metrics.go
```diff
@@ -51,6 +53,15 @@ func (m *s3Metrics) StatCall(
 		}
```
leftmain, line 53: let's add bucket and key to "S3 call hanging"/"S3 call error" logs, it'll help to identify hanging object

leftmain (approved):

SvartMetal (approved):

gy2411 (approved):

PR #121, opened 2024-01-11
[Disk Manager]: Do not hang on mkfs.ext4 in sync tests

jkuradobery:
Add -F parameter to the sync tests in order to avoid hanging on the disk reuse.


qkrorlqr (approved):

WilyTiger (approved):

jkuradobery:
There is no point in using the `-F` key. The explaination is the following:
mkfs.ext4 is pretty much mke2fs. And it checks for the file-system being present and asks for input afterwards. 
https://github.com/tytso/e2fsprogs/blob/master/misc/util.c#L96
In "right" tests we did not create pseudo-terminal, so mke2fs just creates the filesystem. To avoid affecting "right" tests, we should just change our ssh client implementation by adding the `-T` parameter. Will abandon this PR and create a new one.
@qkrorlqr @WilyTiger @SvartMetal

PR #185, opened 2024-01-19
NBS-179: support using multiple devices from the same rdma endpoint

budevg:
This pr fixes the problem with connecting to multiple devices on the same host using rdma compound storage. Before the fix same storage was reused to access all devices which resulted in data corruption

### DIFF
--- a/cloud/blockstore/libs/service_local/storage_rdma.cpp
+++ b/cloud/blockstore/libs/service_local/storage_rdma.cpp
@@ -470,12 +470,14 @@ class TRdmaStorageProvider final
 private:
     struct TEndpoint
     {
+        TString Uuid;
         TString Host;
         ui32 Port;
 
         bool operator<(const TEndpoint& other) const
         {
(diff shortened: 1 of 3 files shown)


tpashkin (approved):

cloud/blockstore/libs/service_local/storage_rdma.cpp
```diff
@@ -470,12 +470,14 @@ class TRdmaStorageProvider final
 private:
     struct TEndpoint
     {
+        TString Uuid;
         TString Host;
         ui32 Port;
 
         bool operator<(const TEndpoint& other) const
         {
-            return std::tie(Host, Port) < std::tie(other.Host, other.Port);
```
drbasic, line 478: Жду написания теста на эту машинерию

drbasic (changes requested):

drbasic (approved):

tpashkin (approved):

PR #189, opened 2024-01-20
NBSNEBIUS-82: not counting broken placement group partitions for non-HDD disks; slightly improved ShouldHandleCmsRequestsForHostsWithDependentHddDisks ut

qkrorlqr:
We should not allow maintenance if it affects a placement group which may have 2+ partitions with broken STORAGE_MEDIA_HDD_NONREPLICATED disks. But in reality we don't check disk media kind and may block maintenance for placement groups with STORAGE_MEDIA_SSD_NONREPLICATED disks. This PR adds the missing MediaKind check.


drbasic (approved):

WilyTiger (approved):

qkrorlqr:
этот пока смержу - локально тесты прогнал, в дальнейших ПР чеки уже должны быть починены

PR #204, opened 2024-01-22
[Disk Manager]: Fix eternal tests race condition

jkuradobery:
Implement file-locking mechanism for disk manager eternal tests. Disk manager eternal tests for 8TiB disks take more than 24hours to complete. Since they are started with cron, next test run starts before the previous run is finished. Since those tests use the same disk per suite but not per run, the same disk is detached from the previous machine, crashing tests for the previous run.

### DIFF
--- a/cloud/disk_manager/test/acceptance/test_runner/lib/helpers.py
+++ b/cloud/disk_manager/test/acceptance/test_runner/lib/helpers.py
@@ -1,8 +1,11 @@
+import contextlib
+import fcntl
 import logging
 import os
 
 import random
 import socket
+from pathlib import Path
 
 from typing import Callable, Protocol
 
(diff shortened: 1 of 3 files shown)


jkuradobery:
@SvartMetal Maybe we shall not detach disks for eternal tests at the test's start? But without the detach mechanism it would be unsafe, e.g. the test case is killed by OOM before disks were released?

cloud/disk_manager/test/acceptance/test_runner/lib/helpers.py
```diff
@@ -82,3 +85,12 @@ def size_prettifier(size_bytes: int) -> str:
         return '%sKiB' % (size_bytes // 1024)
     else:
         return '%sB' % size_bytes
+
+
+@contextlib.contextmanager
+def file_lock(name: str):
```
SvartMetal, line 91: Maybe it would be better to implement the same logic in the cron sript?

PR #221, opened 2024-01-24
[BB->GutHub sync] Libvhost support for MEM_SLOTS/VRING_ENABLE

EvgeniyKozev:
scripts/sync_with_arc: change the default sync branch
NBS no longer uses trunk to synchronize their changes to GitHub, so
switch the branch to the one they use now.

Add support for VHOST_USER_SET_VRING_ENABLE

aio_server: don't forget to free config strings on disk stop

vdev: advertise support for CONFIGURE_MEM_SLOTS
Now that we have all calls implemented we can finally advertise support
for this feature.

vhost_spec: add definition for VHOST_USER_PROTOCOL_F_CONFIGURE_MEM_SLOTS

Implement VHOST_REM_MEM_REG
This reuses existing logic from set_mem_table, with the only difference
being that instead of replacing the old mem table entirely we create a
copy of the old mem table with one memory region removed and then
pretend it's a SET_MEM_TABLE call.

Implement VHOST_USER_ADD_MEM_REG

memmap: add a vhd_memmap_dup helper
This will be used to implement region removal in the future commits.

Implement VHOST_GET_MAX_MEM_SLOTS
This simply uses the memmap API we added earlier.

memmap: add a vhd_memmap_max_memslots getter
This will be used to implemenet the GET_MAX_MEM_SLOTS vhost call.

pytest/test_disk: add missing disk proc attribute checking
Last patch missed a few extra places where this was used.

pytest/test_disk: don't start/stop backends without a proc
Not all backends have a proc attribute, e.g. the VirtioBlk backend only
has a raw file backing without an associated process.


qkrorlqr:
А нет какого-то более подробного описания изменений в исходных коммитах core?

EvgeniyKozev:
> А нет какого-то более подробного описания изменений в исходных коммитах core?

Есть. Забыл добавить. Мой косяк.

PR #235, opened 2024-01-24
NBSNEBIUS-89: [Tasks] extract scheduleInterval and maxTasksInflight into TaskSchedule struct; rename tasks/storage.schedule to scheduleState

SvartMetal:
In next PRs "cron tab" parameters will be added (and encapsulated inside TaskSchedule struct along with ScheduleInterval and MaxTasksInflight)

### DIFF
--- a/cloud/tasks/storage/storage_ydb_impl.go
+++ b/cloud/tasks/storage/storage_ydb_impl.go
@@ -657,8 +657,7 @@ func (s *storageYDB) createRegularTasks(
 	ctx context.Context,
 	session *persistence.Session,
 	state TaskState,
-	scheduleInterval time.Duration,
-	maxTasksInflight int,
+	schedule TaskSchedule,
 ) error {
 
 	state.Regular = true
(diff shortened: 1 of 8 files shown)


jkuradobery (commented):
I would rather rename Introduce TaskSchedule to Extract scheduleInterval and maxTasksInflight into TaskSchedule struct. This way it would be more clear that you do not introduce new functionality, just refactor the old one.

jkuradobery:
> In next PRs "cron tab" parameters will be encapsulated inside TaskSchedule struct

Is it really required?

cloud/tasks/storage/storage_ydb_impl.go
```diff
@@ -686,14 +685,14 @@ func (s *storageYDB) createRegularTasks(
 	defer res.Close()
 
 	found := false
-	sch := schedule{}
+	schState := scheduleState{}
```
jkuradobery, line 688: Why not just create a ScheduleState variable, schState is understandable given the context, yet I think that we should measure up to higher expectations regarding the readability of the code.

jkuradobery (commented):
Otherwise the code seems to be OK. I see two refactorings here, and the first is a direct cause of the second. If you need it merged ASAP, DM me.

jkuradobery (approved):

BarkovBG (approved):

PR #266, opened 2024-01-26
[Disk Manager]: Snapshot: Fix inflight queue not closing in ShallowCopySnapshot()

jkuradobery:
InflightQueue was not closed after the shallowCopySnapshot was finished, which was causing a goroutine leak.

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/storage_ydb_impl.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/storage_ydb_impl.go
@@ -527,6 +527,7 @@ func (s *storageYDB) shallowCopySnapshot(
 		common.ChannelWithCancellation{}, // holeValues
 		s.shallowCopyInflightLimit,
 	)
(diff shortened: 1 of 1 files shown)


cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/storage_ydb_impl.go
```diff
@@ -528,6 +528,7 @@ func (s *storageYDB) shallowCopySnapshot(
 		s.shallowCopyInflightLimit,
 	)
```
SvartMetal, line 529: should be like this:
```
	inflightQueue := common.NewInflightQueue(
		common.Milestone{Value: milestoneChunkIndex},
		processedIndices,
		common.ChannelWithCancellation{}, // holeValues
		s.shallowCopyInflightLimit,
	)
	defer inflightQueue.Close()

        waitSaver := func() error { return nil }
```

SvartMetal (approved):

gy2411 (approved):

PR #313, opened 2024-02-01
NBSNEBIUS-71: make .proto config more compatible with previous .sc config

debnatkh:
In prefious config specification for eternal_tests (https://github.com/ydb-platform/nbs/blob/ac799e1fe11eaa3e3b31dc8bacad6d7010731037/cloud/blockstore/tools/testing/eternal_tests/eternal-load/lib/config.sc) in the situations, where `IoDepth` was less than the number of specified ranges, the config generator was supposed to generate missing ranges. This was the default behavior of the config generator when using the .sc specification and should stay this way after the switch to .proto specification.


sharpeye (approved):

WilyTiger (approved):

debnatkh:
cloud/blockstore/libs/encryption/ut_keyring/bin/unittest	
cloud/blockstore/tests/notify/test.py.test_notify[back-online]	

флапы

PR #336, opened 2024-02-04
fix deletion of rdma server session

budevg:
Disconnected session was not deleted and all other sessions were deleted
instead due to problem in rcu list predicate implementation.

Just deleting the server session and calling DestroyQP is not enough and will
lead to deadlock. We need to wait for all WR's to flush and only then it's safe
to delete Connection object (calling rdma_destroy_id)

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/rcu.h
+++ b/cloud/blockstore/libs/rdma/impl/rcu.h
@@ -38,7 +38,7 @@ public:
 
         with_lock (Lock) {
             for (const T& item: *List) {
-                if (predicate(item)) {
+                if (!predicate(item)) {
                     list->push_back(item);
                 }
             }
--- a/cloud/blockstore/libs/rdma/impl/server.cpp
+++ b/cloud/blockstore/libs/rdma/impl/server.cpp
@@ -292,6 +292,7 @@ private:
     TLockFreeList<TRequest> InputRequests;
     TSimpleList<TRequest> QueuedRequests;
     TEventHandle RequestEvent;
+    bool FlushStarted = false;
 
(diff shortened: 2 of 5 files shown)


cloud/blockstore/libs/rdma/impl/rcu.h
```diff
@@ -38,7 +38,7 @@ class TRCUList
 
         with_lock (Lock) {
             for (const T& item: *List) {
-                if (predicate(item)) {
+                if (!predicate(item)) {
```
qkrorlqr, line 41: на это явно нужен unittest

cloud/blockstore/libs/rdma/impl/server.cpp
```diff
@@ -640,11 +669,6 @@ void TServerSession::FreeRequest(TRequestPtr req, TSendWr* send) noexcept
 
 void TServerSession::RecvRequestCompleted(TRecvWr* recv, ibv_wc_status status)
 {
-    if (status == IBV_WC_WR_FLUSH_ERR) {
```
qkrorlqr, line 643: а вот это, получается, вообще всегда было false, т.к. мы в HandleCompletionEvent в случае этого статуса не звали RecvRequestCompleted?

tpashkin (dismissed):

cloud/blockstore/libs/rdma/impl/server.cpp
budevg: ага сюда код некогда не попадал

tpashkin (approved):

Sazonov99 (approved):

Issue #173, opened 2024-01-18
[Disk Manager] Use NBS and YDB binaries in tests


BarkovBG:
now we are building nbs and ydb from the source code but need to use binaries in ya.make in tests

this will allow us remove all dependencies on nbs source code

qkrorlqr:
If you do this, you need to make a short manual that would describe how to update these prebuilt blockstore-server, filestore-server and filestore-vhost versions

BarkovBG:
> If you do this, you also need to write a short manual that would describe how to update these prebuilt blockstore-server, filestore-server and filestore-vhost binaries

@jkuradobery, can you pls write documentation as the commit author

PR #359, opened 2024-02-07
[Disk Manager]: Tests: Use YDB binary from S3 instead of building it …

jkuradobery:
…locally

We were using an old YDB version which led to hangs in disk manager tests after the context cancellation.For problem details, see: https://github.com/ydb-platform/ydb-go-sdk/issues/1025


qkrorlqr (approved):

BarkovBG:
link this pr here - https://github.com/ydb-platform/nbs/issues/173

BarkovBG (approved):

PR #373, opened 2024-02-08
fix deletion of rdma server session (#336)

budevg:
* fix deletion of rdma server session

Disconnected session was not deleted and all other sessions were deleted
instead due to problem in rcu list predicate implementation.

Just deleting the server session and calling DestroyQP is not enough and will
lead to deadlock. We need to wait for all WR's to flush and only then it's safe
to delete Connection object (calling rdma_destroy_id)

* add rcu list unit tests

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/rcu.h
+++ b/cloud/blockstore/libs/rdma/impl/rcu.h
@@ -38,7 +38,7 @@ public:
 
         with_lock (Lock) {
             for (const T& item: *List) {
-                if (predicate(item)) {
+                if (!predicate(item)) {
                     list->push_back(item);
                 }
             }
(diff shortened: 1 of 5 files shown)


cloud/blockstore/libs/rdma/impl/rcu.h
```diff
@@ -38,7 +38,7 @@ class TRCUList
 
         with_lock (Lock) {
             for (const T& item: *List) {
-                if (predicate(item)) {
+                if (!predicate(item)) {
```
drbasic, line 41: как это вообще раньше работало?

budevg: Все сессии отклучались а потом опять подключались)

drbasic (approved):

Sazonov99 (approved):

Issue #95, opened 2024-01-10
[Filestore] Compaction/Cleanup/FlushBytes optimization


qkrorlqr:
- Implement lazy compaction map loading **DONE**
- Implement garbage-based compaction **DONE**
- Use blob patching in compaction **LATER**
- Think about some ways to optimize Cleanup and FlushBytes and the FreshBytes layer https://github.com/ydb-platform/nbs/issues/1129
- Global triggers for Compaction and Cleanup - triggers for total deletion marker count per fs, total blob count per fs, etc. **DONE**
- Add metrics for background ops, display more background ops info on the tablet monpage **DONE**

qkrorlqr:
> Use blob patching in compaction

won't use patching - it's too risky, we still keep finding bugs in it 2.5 years after it was first introduced

qkrorlqr:
Found a performance issue: Compaction copies blocks which are not actually needed but whose MaxCommitId is still > than MinCommitId (CommitIds are not rebased yet) to its dst blobs. Second compaction run is needed to get rid of this garbage.

qkrorlqr:
found another problem: CompactionThreshold is set in terms of blob count, max blob size is set in bytes (and the default is reasonable - 4MiB) which leads to nonstop recompactions for filesystems with large block sizes

e.g. BlockSize=128KiB, MaxBlobSize=4MiB, CompactionThreshold=20
if we have a "full" compaction range (having 64 blocks x 16 nodes == 1024 blocks) there will be 128MiB of data in it => 32 blobs which is greater than 20 which leads us to endless compactions

a possible solution is adjusting this threshold based on the BlockSize / DefaultBlockSize ratio (DefaultBlockSize == 4KiB => we will have min 1 blob per range if there is some data in this range - it's our baseline)

qkrorlqr:
we also need some metrics showing the number of compactions per second split by the reasons - launches per second due to high blob count per range / due to high overall blob count / due to high garbage percentage

and outputting current garbage percentage and average blob count per range to the tablet monpage would also be nice

qkrorlqr:
**From the description of this issue:**
> - Implement lazy compaction map loading
Done

> - Implement garbage-based compaction
Done

> - Use blob patching in compaction
Postponed

> - Think about some ways to optimize Cleanup and FlushBytes and the FreshBytes layer
TODO

> - Global triggers for Compaction and Cleanup - triggers for total deletion marker count per fs, total blob count per fs, etc.
Done

[...]

PR #381, opened 2024-02-08
issue #95: 1. deduplicating out-of-order compaction map chunk load requests in queue 2. limiting the number of out-of-order compaction map chunk load requests in queue

qkrorlqr:
Otherwise if we have a large FS and a lot of concurrent WriteData requests CompactionMap takes forever to load

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
@@ -244,6 +244,8 @@ void TIndexTabletActor::HandleWriteData(
     };
 
     if (!CompactionStateLoadStatus.Finished) {
+        const ui32 limitInQueue =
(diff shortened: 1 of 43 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
```diff
@@ -259,8 +261,29 @@ void TIndexTabletActor::HandleWriteData(
...
+                    if (!req.OutOfOrder) {
+                        continue;
+                    }
+
+                    if (req.FirstRangeId == rangeId) {
+                        shouldEnqueue = false;
+                        break;
+                    }
+
+                    if (++oooRequestsInQueue == limitInQueue) {
```
debnatkh, line 278: А тут не надо делать `oooRequestsInQueue + 1 == limitInQueue`?

qkrorlqr: не надо

PR #399, opened 2024-02-09
issue-324: DescribeData: 1. BlobOffsets should be in bytes, not blocks 2. returning FileSize 3. added a ut that checks that we don't return data outside of the aligned superrange that contains the requested range

qkrorlqr:
Some fixes and improvements for DescribeData


debnatkh (approved):

qkrorlqr:
Флапнули 2 теста ДМ

yegorskii (approved):

PR #443, opened 2024-02-14
tweak ShouldHandleMutlipleQueues: wait for stats sync

sharpeye:
sometimes ShouldHandleMutlipleQueues fails because it can't get the actual stats.

### DIFF
--- a/cloud/blockstore/vhost-server/server_ut.cpp
+++ b/cloud/blockstore/vhost-server/server_ut.cpp
@@ -171,6 +171,11 @@ public:
 
     TSimpleStats GetStats(ui64 expectedCompleted) const
     {
+        // Without I/O, stats are synced every second and only if there is a 
+        // pending GetStats call. The first call to GetStats might not bring the
+        // latest stats; therefore, you need at least two calls so that the AIO
+        // backend will sync the stats.
+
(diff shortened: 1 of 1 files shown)


budevg (approved):

qkrorlqr:
А это в итоге хотим заливать или не хотим? Как-то на костыль похоже. Почему иногда стату не можем получить?

budevg (approved):

cloud/blockstore/vhost-server/server_ut.cpp
```diff
@@ -168,6 +168,21 @@ struct TFixture
         Files.clear();
         Options.Layout.clear();
     }
+
+    TSimpleStats GetStats(ui64 expectedCompleted) const
```
budevg, line 172: let's add comment to this function that explains why we retry get stats.
Something similar to
"Without running io's stats are synced every second and only if there is a pending GetStats call. First call to GetStats might not bring the latest stats so we need at  least 2 calls so that the aio backend will sync the stats"

budevg (changes requested):

budevg (approved):

yegorskii (approved):

PR #449, opened 2024-02-15
fix rdma completions polling

budevg:
We need to empty the completion queue otherwise when using EWaitMode::Poll we
might not get additional notification on the event channel and the remaining
completions won't be consumed.


cloud/blockstore/libs/rdma/impl/verbs.cpp
```diff
@@ -133,22 +133,27 @@ struct TVerbs
     bool PollCompletionQueue(ibv_cq* cq, ICompletionHandler* handler) override
     {
         ibv_wc wc[MAX_COMPLETIONS_PER_POLL];
+        bool gotWorkCompeltions = false;
```
tpashkin, line 136: Compeltions -> Completions

tpashkin (approved):

Sazonov99 (approved):

PR #508, opened 2024-02-20
Prepare nbs to ydb#1784

Enjection:
We are going to disable PROTOC_TRANSITIVE_HEADERS in `ydb/core/protos` in https://github.com/ydb-platform/ydb/pull/1784 to make this transition smooth for you I added additional include which brings back old behavior.
In case you want to speed up your builds as well you can remove this include and replace it with includes really needed.


Enjection:
It's temporary on hold. Seem's that ymake don't understand conditional includes, so I have to think about different approach

Enjection:
Okay, I found workaround for this -- we will just ignore it with `Y_IGNORE`. It won't break anything as long as we ignore main header in same file

Enjection:
@qkrorlqr Could we discuss this change? It is blocking us

qkrorlqr:
> @qkrorlqr Could we discuss this change? It is blocking us

I don't see any results indicating that this activity improves something (build time?). Here https://github.com/ydb-platform/ydb/pull/1784 I see lots of failed tests and builds in most of those PR-checks - I don't understand, how can we draw conclusions from such unstable measurements.

qkrorlqr (approved):

PR #517, opened 2024-02-21
NBS-4488: [Disk Manager] remove unnecessary migration config setting

BarkovBG:


cloud/disk_manager/internal/pkg/services/disks/config/config.proto
```diff
@@ -7,6 +7,8 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 ////////////////////////////////////////////////////////////////////////////////
 
 message DisksConfig {
+    reserved 10;
```
SvartMetal, line 10: It's not needed, we do not persist config anywhere except local filesystem

SvartMetal (approved):

gy2411 (approved):

PR #542, opened 2024-02-22
NEBINC-82: [Snapshot] should use 'url source forbidden' message when encounter 403 http status code

SvartMetal:


cloud/disk_manager/internal/pkg/dataplane/url/common/http_client.go
```diff
@@ -167,18 +167,16 @@ func (c *httpClient) head(
 		return nil, errors.NewRetriableError(err)
 	}
 
+	if resp.StatusCode == 403 {
+		return nil, NewSourceForbiddenError("http code %d", resp.StatusCode)
+	}
+
 	if resp.StatusCode >= 400 && resp.StatusCode <= 499 {
-		return nil, NewSourceNotFoundError(
-			"url source not found: code %d",
```
SvartMetal, line 172: 'url source not found' is already printed there https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/dataplane/url/common/errors.go#L17

jkuradobery:
@SvartMetal is there any chance we can determine if the link has expired or something? Maybe from the S3 response body?

SvartMetal:
> @SvartMetal is there any chance we can determine if the link has expired or something? Maybe from the S3 response body?
Can't see any. And it would not be great to depend on the S3 response text

BarkovBG (approved):

PR #571, opened 2024-02-27
issue-487 NBS-4829 [disk manager]: make possible to add arbitrary logging field to context, write fields from context in ydb client logger

gy2411:
Added methods that allow to add logging fields to context. Used that methods to improve ydb client logging: now it writes field from context.

Examples of rows after these changes:
```
Feb 27 13:00:29 d9hnumvcrda87ck5kbpm disk-manager[305304]: ydb_logger.go:73 DEBUG => COMPONENT: YDB, YDB_LOG_NAME: ydb.driver.resolver.init, ENDPOINT: localhost:32648, DATABASE: /Root, SECURE: false => start
Feb 27 13:00:29 d9hnumvcrda87ck5kbpm disk-manager[305304]: factory.go:69 DEBUG => calling client.Close for &{0xc0005c2480 300000000000 500000000 0x17f5fe0 0xc0004b0260}, localhost:17106
```
(in 'real' logs there also will be request id, idempotency key and operation id fiedls)

PR #581, opened 2024-02-27
issue-579: introduced NegativeEntryTimeout for entry_timeout returned to the client in case of errors, passing this timeout to TFileSystemConfig via NProto::TFileStore::Features which gets the values from TStorageConfig

qkrorlqr:
By default the caching of negative fuse lookup (getattr) responses is disabled. Can be enabled again via TStorageConfig - either globally or for specific filesystems (via the changestorageconfig private api action). entry_timeout and attr_timeout for normal responses can be configured in the same manner now (if needed).

#571

### DIFF
--- a/cloud/filestore/public/api/protos/fs.proto
+++ b/cloud/filestore/public/api/protos/fs.proto
@@ -14,6 +14,9 @@ option go_package = "github.com/ydb-platform/nbs/cloud/filestore/public/api/prot
 message TFileStoreFeatures
 {
     bool TwoStageReadEnabled = 1;
+    uint32 EntryTimeout = 2;
+    uint32 NegativeEntryTimeout = 3;
+    uint32 AttrTimeout = 4;
 }
 
 message TFileStore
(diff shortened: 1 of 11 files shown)


cloud/filestore/public/api/protos/fs.proto
```diff
@@ -14,6 +14,9 @@ option go_package = "github.com/ydb-platform/nbs/cloud/filestore/public/api/prot
 message TFileStoreFeatures
 {
     bool TwoStageReadEnabled = 1;
+    uint32 EntryTimeout = 2;
+    uint32 NegativeEntryTimeout = 3;
```
debnatkh, line 18: Может подписать, что это ms?

debnatkh (approved):

yegorskii (approved):

cloud/filestore/public/api/protos/fs.proto
qkrorlqr: Ну, ms - это у нас в конфигах умолчание

PR #587, opened 2024-02-29
Remove unused files

drbasic:


cmake/protobuf.cmake
```diff
@@ -1,111 +0,0 @@
-include(common)
-
-function(target_proto_plugin Tgt Name PluginTarget)
-  set_property(TARGET ${Tgt} APPEND PROPERTY
```
budevg, line 4: can we remove the `./build/export_generators/hardcoded-cmake` as well ? 
the `./build/export_generators/hardcoded-cmake/generator.toml` for example references the `cmake/protobuf.cmake` which was just removed

budevg (changes requested):

budevg (approved):

Sazonov99 (approved):

PR #607, opened 2024-03-01
fix: vms whose creation was cancelled

librarian:
If you do not add always() condition, VM creation could be cancelled, and we would not get necessary id to remove it

Example here: https://github.com/ydb-platform/nbs/actions/runs/8030802857

New push cancelled workflow and `release runner` didn't have an id to remove and now we have vm using resources

Script here to help find and remove those vms if they will happen in the future, it also helps with finding workflows that may have caused this behaviour

### DIFF
--- a/.github/scripts/vm-watcher.py
+++ b/.github/scripts/vm-watcher.py
@@ -1,0 +1,123 @@
+import os
+import grpc
+import json
+import logging
+import argparse
+from github import Github, Auth as GithubAuth
+from datetime import datetime, timedelta, timezone
+from yandexcloud import SDK, RetryInterceptor
+from yandex.cloud.compute.v1.instance_service_pb2_grpc import InstanceServiceStub
+from yandex.cloud.compute.v1.instance_service_pb2 import (
+    ListInstancesRequest,
+    DeleteInstanceRequest,
+)
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s: %(levelname)s: %(message)s"
+)
+
+CACHE_VM_ID = "dp7329odurnhplpf5ff0"
+
+
+def find_workflows_containing_string(
+    client, specified_time, search_string, owner="ydb-platform", repo="nbs"
+):
+    repo = client.get_repo(f"{owner}/{repo}")
+
+    # Calculate start time for search (-10 minutes)
+    start_time = specified_time - timedelta(minutes=10)
+    end_time = specified_time + timedelta(minutes=10)
+
+    # Initialize list to keep track of matching runs
+    matching_runs_info = []
+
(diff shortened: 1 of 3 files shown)


.github/scripts/vm-watcher.py
```diff
@@ -0,0 +1,113 @@
...
+    DeleteInstanceRequest,
+)
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s: %(levelname)s: %(message)s"
+)
+
+CACHE_VM_ID = "dp7329odurnhplpf5ff0"
+
+def find_workflows_containining_string(client, specified_time, search_string, owner="ydb-platform", repo="nbs"):
```
qkrorlqr, line 21: containing

.github/scripts/vm-watcher.py
```diff
@@ -0,0 +1,113 @@
...
+
+    for run in repo.get_workflow_runs():
+        # Check if the run started within our time window
+        run_started_at = run.created_at
+        if start_time <= run_started_at <= end_time:
+            print("Workflow",run.name, run.created_at, run.html_url)
+            # Get jobs or the current workflow run
+            for job in run.jobs():
+
+                if job.name == "Start self-hosted runner":
```
qkrorlqr, line 39: а нельзя как-то не по названию, а по типу, который был бы каким-нибудь enum-ом?

.github/scripts/vm-watcher.py
```diff
@@ -0,0 +1,113 @@
...
+                    print("Job", job.name)
+                    # Attempt to get logs (note: this might require additional handling for large logs)
+                    try:
+                        logs = job.get_log()
+                        if search_string in logs:
+                            matching_runs_info.append({
+                                "run_id": run.id,
+                                "run_url": run.html_url,  # Link to the workflow run
+                                "job_name": job.name
+                            })
```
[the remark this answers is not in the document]
qkrorlqr: что-то этот бот разошелся
он тут херню какую-то предлагает

.github/scripts/vm-watcher.py
```diff
@@ -0,0 +1,123 @@
...
+        default="bjeuq5o166dq4ukv3eec",
+    )
+    parser.add_argument(
+        "--ttl", required=True, help="The TTL for the VMs", default=24, type=int
+    )
+    parser.add_argument("--apply", action="store_true", help="Apply the changes")
+
+    args = parser.parse_args()
+
+    treshold_time = datetime.now() - timedelta(hours=args.ttl)
```
qkrorlqr, line 83: threshold

qkrorlqr (approved):

drbasic (approved):

PR #622, opened 2024-03-02
[Disk Manager] support vhd format

BarkovBG:


cloud/disk_manager/internal/pkg/dataplane/url/vhd/image_map_reader.go
```diff
@@ -98,6 +98,10 @@ func (r *ImageMapReader) Read(ctx context.Context) ([]common.ImageMapItem, error
 	return items, nil
 }
 
+func (r *ImageMapReader) ReadFooter(ctx context.Context) error {
```
jkuradobery, line 101: Why do we need this? I do not see usage.

BarkovBG: https://github.com/ydb-platform/nbs/pull/622/files#diff-da8c05420b3b357ebce97d7ab6b84fd6100f90a9eeae4c6bf1c2efcfd985cd07R259

jkuradobery: I think the public function is not required, you should move  reader.ReadFooter inside NewImageMapReader.

jkuradobery:
Please, replace commit name with something more like: "enable vhd format" or something similar. Allow to use does not sound quite right.

cloud/disk_manager/internal/pkg/dataplane/url/vhd/image_map_reader.go
BarkovBG: we need it for ImageMapReader's Size method 

it called before Read and header/footer need to be already read when it called

I want to make it private too because ImageMapIntrface does not public methods like ReadFooter/ReadHeader.
But I think that it is more correct to do this refactoring in separate PR because it isn't minor change - we need to add ctx to constuctor (bad idea) or need to add readFooter into Size method (bad idea too) or ...

jkuradobery (approved):

gy2411 (approved):

Issue #592, opened 2024-02-29
[NBS] disk agent stuck during RdmaServer stop


budevg:
when restarting disk agent through `systemctl restart blockstore-disk-agent` sometimes the agent become stuck and won't stop gracefully.

```
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29-13-23-38 :BLOCKSTORE_SERVER INFO: cloud/blockstore/libs/disk_agent/bootstrap.cpp:554: Stopped Scheduler
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29-13-23-38 :BLOCKSTORE_SERVER INFO: cloud/blockstore/libs/disk_agent/bootstrap.cpp:556: Stopped FileIOService
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29T13:23:38.807405Z :BLOCKSTORE_DISK_AGENT INFO: Poisoned
Feb 29 13:23:38 host systemd[1]: Stopping Blockstore Disk Agent...
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29T13:23:38.807656Z :TX_PROXY WARN: actor# [50317:7339938168503293722:12] HANDLE TEvClientDestroyed from tablet# 72057594046447617
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29T13:23:38.807669Z :BLOCKSTORE_HIVE_PROXY ERROR: Pipe to hive72057594037968897 has been reset
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29T13:23:38.807670Z :TX_PROXY WARN: actor# [50317:7339938168503293722:12] HANDLE TEvClientDestroyed from tablet# 72057594046447619
Feb 29 13:23:38 host NBS_DISK_AGENT[274733]: 2024-02-29T13:23:38.807768Z :INTERCONNECT_NETWORK NOTICE: [50317 <-> 16] connection closed by peer
[...]

PR #625, opened 2024-03-03
NBS #592: Prevent RdmaServer deadlock during disk agent stopping

budevg:
Disk agent actor cannot be released safely while RdmaTarget, which is used by
RdmaServer, can still get incoming requests from the server and submit them to
RdmaTarget's queue.

The proper release sequence should first disconnect all sessions on RdmaServer
and, therefore, ensure that no incoming requests can arrive. Then the RdmaTarget
can be released, and later the RdmaServer can be stopped.

### DIFF
--- a/cloud/blockstore/libs/storage/disk_agent/rdma_target.cpp
+++ b/cloud/blockstore/libs/storage/disk_agent/rdma_target.cpp
@@ -944,6 +944,7 @@ public:
 
     void Stop() override
     {
+        Server->Stop();
         TaskQueue->Stop();
     }
 


cloud/blockstore/libs/rdma/impl/server_ut.cpp
```diff
@@ -115,6 +115,37 @@ Y_UNIT_TEST_SUITE(TRdmaServerTest)
 
         server->Stop();
     }
+
+    Y_UNIT_TEST(ShouldDisconnectAll)
```
qkrorlqr, line 119: выглядит так, как будто этот тест ничего не проверяет

cloud/blockstore/libs/storage/disk_agent/rdma_target.cpp
```diff
@@ -944,6 +944,7 @@ class TRdmaTarget final
 
     void Stop() override
     {
+        Server->DisconnectAll();
```
tpashkin, line 947: Why can't we just call Stop here? Server->Stop() in turn will Stop all attached pollers

tpashkin: You may argue that Server might potentially be used by someone else, not just TRdmaTarget, but closing all sessions/endpoints has pretty much the same effect in the end

tpashkin: In fact there seems to be a race between peers to create a new session (since ConnectionPoller is still active) and DiskAgent actor to stop

cloud/blockstore/libs/rdma/impl/server.cpp
```diff
@@ -1276,6 +1277,20 @@ class TCompletionPoller final
         }
     }
 
+    void DisconnectAllSessions()
+    {
+        auto sessions = Sessions.Get();
+
+        for (const auto& session: *sessions) {
+            session->Flush();
```
tpashkin, line 1285: This is counterintuitive, let's create a proxy method `Disconnect`

cloud/blockstore/libs/rdma/impl/server.cpp
```diff
@@ -1276,6 +1277,20 @@ class TCompletionPoller final
...
+        }
+
+        while (sessions->size()) {
```
tpashkin, line 1288: Comment explaining that we are waiting for a main loop to close flushed or timed out sessions would be nice

cloud/blockstore/libs/rdma/iface/server.h
```diff
@@ -68,6 +68,8 @@ struct IServer
         TString host,
         ui32 port,
         IServerHandlerPtr handler) = 0;
+
+    virtual void DisconnectAll() = 0;
```
tpashkin, line 72: May I suggest DisconnectSessions?

cloud/blockstore/libs/rdma/impl/server.cpp
```diff
@@ -1276,6 +1277,20 @@ class TCompletionPoller final
         }
     }
 
+    void DisconnectAllSessions()
```
tpashkin, line 1280: May I suggest DisconnectSessions?

cloud/blockstore/libs/storage/disk_agent/rdma_target.cpp
budevg: I stop all endpoints from listening so it's impossible to create new session.

Doing Server->Stop() seems like a better idea no need to add extra code.

cloud/blockstore/libs/rdma/impl/server_ut.cpp
budevg: мы решили исправить это другим способом, остановив RdmaServer

tpashkin (approved):

drbasic (approved):

PR #657, opened 2024-03-06
Remove endpoint from storage when client stop it. Fix typo.

sanek325:
When client stop endpoint we should remove this endpoint from endpoint storage.
It doesn’t matter whether this endpoint was persistent or not.
It also doesn’t matter if we have errors during stopping.

### DIFF
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -298,7 +298,7 @@ private:
         const NProto::TStartEndpointRequest& newRequest,
         const NProto::TStartEndpointRequest& oldRequest);
 
-    NProto::TError OpenEndpointSockets(
+    NProto::TError OpenAllEndpointSockets(
         const NProto::TStartEndpointRequest& request,
         const TSessionInfo& sessionInfo);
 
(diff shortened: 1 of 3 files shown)


cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -306,11 +306,8 @@ class TEndpointManager final
         const NProto::TStartEndpointRequest& request,
         const TSessionInfo& sessionInfo);
 
-    TVector<TFuture<NProto::TError>> CloseEndpointSockets(
-        const NProto::TStartEndpointRequest& startRequest);
-
-    TFuture<NProto::TError> CloseEndpointSocket(
-        const NProto::TStartEndpointRequest& startRequest);
+    void CloseEndpointSockets(const NProto::TStartEndpointRequest& startRequest);
+    void CloseEndpointSocket(const NProto::TStartEndpointRequest& startRequest);
```
sharpeye, line 310: Maybe something like `CloseMainEndpointSocket` ?

sanek325: переименовал в CloseAllEndpointSockets и CloseEndpointSocket

sharpeye (approved):

yegorskii (approved):

PR #698, opened 2024-03-11
support both white and blacklist at the same time; issue-476

tpashkin:
support both white and blacklist at the same time; issue-476

### DIFF
--- a/cloud/storage/core/libs/features/ya.make
+++ b/cloud/storage/core/libs/features/ya.make
@@ -2,6 +2,7 @@ LIBRARY()
 
 SRCS(
     features_config.cpp
+    filters.cpp
 )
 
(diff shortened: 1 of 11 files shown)


sharpeye (dismissed):

cloud/storage/core/libs/features/ya.make
```diff
@@ -2,6 +2,7 @@ LIBRARY()
 
 SRCS(
     features_config.cpp
+    filters.cpp
```
qkrorlqr, line 5: не вижу правки в CMakeLists

tpashkin: поправил

sharpeye (approved):

PR #752, opened 2024-03-15
tweak blockstore-nbd

sanek325:
blockstore-nbd tool has three mode of work. There were a few "if" in different methods in the code, and it was too difficult to read this code. So I moved this code to separated methods and replace "if" by "switch" in all places.


qkrorlqr (changes requested):
PR-ов с таким названием и пустым описанием быть не должно

sharpeye (approved):

yegorskii (approved):

qkrorlqr (approved):

PR #768, opened 2024-03-18
Disk Agent switches to the Idle state when it can't find devices

sharpeye:
The point of the Idle state is to not spam to DR with RegisterAgent and UpdateAgentStats events.


komarevtsev-d:
Запустим large тесты?

budevg (approved):

PR #770, opened 2024-03-18
issue-463: start nbd-endpoint with any available nbd-device

sanek325:
Add flag UseFreeNbdDevice to StartEndpoint request. 
NBS gets first available nbd device and connects it to started nbd endpoint.

### DIFF
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -25,6 +25,7 @@
 #include <util/generic/overloaded.h>
 #include <util/generic/set.h>
 #include <util/string/builder.h>
+#include <util/system/fs.h>
 
 namespace NCloud::NBlockStore::NServer {
 
@@ -109,7 +110,7 @@ bool CompareRequests(
     const NProto::TStartEndpointRequest& left,
     const NProto::TStartEndpointRequest& right)
 {
-    Y_DEBUG_ABORT_UNLESS(25 == GetFieldCount<NProto::TStartEndpointRequest>());
+    Y_DEBUG_ABORT_UNLESS(26 == GetFieldCount<NProto::TStartEndpointRequest>());
     return left.GetUnixSocketPath() == right.GetUnixSocketPath()
         && left.GetDiskId() == right.GetDiskId()
         && left.GetInstanceId() == right.GetInstanceId()
@@ -137,7 +138,8 @@ bool CompareRequests(
             right.GetClientCGroups().begin(),
             right.GetClientCGroups().end())
         && left.GetPersistent() == right.GetPersistent()
-        && left.GetNbdDeviceFile() == right.GetNbdDeviceFile();
+        && left.GetNbdDeviceFile() == right.GetNbdDeviceFile()
+        && left.GetUseFreeNbdDeviceFile() == right.GetUseFreeNbdDeviceFile();
 }
 
 bool CompareRequests(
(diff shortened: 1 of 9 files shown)


cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -1174,27 +1286,50 @@ NProto::TError TEndpointManager::SwitchEndpointImpl(
...
         return NBD::CreateDeviceConnectionStub();
     }
 
-    auto device = NBD::CreateDeviceConnection(
-        Logging,
-        TNetworkAddress(TUnixSocketPath(request->GetUnixSocketPath())),
-        request->GetNbdDeviceFile(),
-        TDuration::Days(1));
+    if (!restoring) {
+        if (!NbdDeviceManager.AcquireDevice(request->GetNbdDeviceFile())) {
```
sharpeye, line 1318: It will be nice if AcquireDevice returns NProto::Error with a specific errror message.

sanek325: поправил

sharpeye (approved):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -105,11 +106,27 @@ bool CompareRequests(
         && left.GetBurstTime() == right.GetBurstTime();
 }
 
+bool CompareNbdDevices(
+    const NProto::TStartEndpointRequest& lft,
+    const NProto::TStartEndpointRequest& rgt)
+{
+    if (lft.GetNbdDeviceFile() && rgt.GetNbdDeviceFile()) {
```
budevg, line 113: doesn't empty string evaluates to bool true and this condition will always be true

cloud/blockstore/libs/nbd/device.h
```diff
@@ -20,6 +20,17 @@ struct IDeviceConnection
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct IDeviceConnectionFactory
```
sharpeye, line 23: We can do without this interface:
```cpp
using TDeviceConnectionFactory = std::function<
    IDeviceConnectionPtr(TNetworkAddress connectAddress, TString deviceName)>;
```

budevg (changes requested):

cloud/blockstore/public/api/protos/endpoints.proto
```diff
@@ -99,6 +104,9 @@ message TStartEndpointResponse
 
     // Volume information.
     TVolume Volume = 2;
+
+    // The device file (e.g. "/dev/nbd0") which nbd-client connected to.
+    string NbdDeviceFile = 3;
```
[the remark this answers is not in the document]
sanek325: I think no. It's response. We can just leave this field empty.

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
sanek325: No, if one of NbdDeviceFiles will be empty then this condition will be false. There are tests for it.

budevg: in this case what if we have `lft = ("", true) rgt=("/dev/nbd0", false)`
first if is false since `lft.GetNbdDeviceFile()` false
second if is false
adn we return `true`

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -1174,27 +1286,50 @@ NProto::TError TEndpointManager::SwitchEndpointImpl(
...
+
+    if (request->HasUseFreeNbdDeviceFile() &&
+        request->GetUseFreeNbdDeviceFile())
     {
+        if (restoring) {
+            return MakeError(E_ARGUMENT,
+                "Forbidden 'FreeNbdDeviceFile' flag in restoring endpoints");
+        }
+
+        auto nbdDevice = NbdDeviceManager.GetFreeDevice();
```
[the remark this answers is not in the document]
budevg: If we assume that `UseFreeNbdDevice` can be used only with `Persistent = true` let's add and explicit check for this

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
sanek325: Yes, it's correct. User asked nbs to start endpoint with any free device, but nbs has already started the same endpoint with /dev/nbd0, so we return true here and user gets S_ALREADY in response.

budevg: that's a little confusing since in CompareXX the two requests are different but the compare returns true ... maybe you should move this logic outside like here https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/endpoints/endpoint_manager.cpp#L716
or at least put comment to explain the last comparison

sanek325: I've just refactored this place.

budevg (approved):

yegorskii (dismissed):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -642,8 +739,17 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
...
-        return AlterEndpoint(std::move(ctx), *request, startedEndpoint);
+        auto endpoint = it->second;
+        auto error = AlterEndpoint(std::move(ctx), *request, *endpoint.Request);
+        if (HasError(error)) {
+            return TErrorResponse(error);
+        }
+
+        NProto::TStartEndpointResponse response;
+        response.MutableError()->CopyFrom(error);
+        response.MutableVolume()->CopyFrom(endpoint.Volume);
```
budevg, line 750: here you return old volume which was saved when StartEndpoint was called but there might be change in volume attributes (for example resize)  and call to RefreshEndpoint

maybe better to call GetSession and use fresh volume value from there

budevg (changes requested):

budevg (approved):

yegorskii (approved):

cloud/blockstore/libs/nbd/device.h
sanek325: discussed in private, decided not to change

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
sanek325: I agree. Let's do it separately.

Issue #356, opened 2024-02-07
support multipartition base disks


BarkovBG:

BarkovBG:
TODO: https://github.com/ydb-platform/nbs/pull/662/files#r1523663953

PR #786, opened 2024-03-20
[Blockstore] do not calculate logical used blocks metric for proxy overlay disks

BarkovBG:
issue https://github.com/ydb-platform/nbs/issues/356

- LogicalUsedBlocks metric uses for overlay disks
- proxy overlay disks are temporary so we needn't calculate this metric

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
@@ -234,7 +234,7 @@ void TPartitionActor::CompleteLoadState(
             State->AccessStats().SetLogicalUsedBlocksCount(
                 State->GetLogicalUsedBlocks().Count()
             );
-        } else {
+        } else if (!partitionConfig.GetIsProxyOverlay()) {
             State->GetLogicalUsedBlocks().Update(State->GetUsedBlocks(), 0);
             State->AccessStats().SetLogicalUsedBlocksCount(
                 State->AccessStats().GetUsedBlocksCount()
@@ -287,11 +287,11 @@ void TPartitionActor::HandleGetUsedBlocksResponse(
 
         Suicide(ctx);
         return;
-    } else {
-        LOG_DEBUG(ctx, TBlockStoreComponents::PARTITION,
-            "[%lu] LoadState completed",
-            TabletID());
     }
+    
+    LOG_DEBUG(ctx, TBlockStoreComponents::PARTITION,
+        "[%lu] LoadState completed",
+        TabletID());
 
     for (const auto& block: msg->Record.GetUsedBlocks()) {
         State->GetLogicalUsedBlocks().Merge(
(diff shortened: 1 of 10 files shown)


cloud/blockstore/libs/storage/volume/partition_requests.h
```diff
@@ -57,7 +57,7 @@ bool ToPartitionRequests(
     Y_UNUSED(requests);
     Y_UNUSED(blockRange);
 
-    Y_ABORT_UNLESS(0);
+    ythrow TServiceError(E_NOT_IMPLEMENTED);
```
SvartMetal, line 60: `Y_ABORT_UNLESS(0, "Not implemented")`?

cloud/blockstore/libs/storage/protos/part.proto
```diff
@@ -73,6 +73,8 @@ message TPartitionConfig
 
     // Optional tablet id of base volume.
     uint64 BaseDiskTabletId = 17;
+
+    bool IsProxyOverlay = 18;
```
SvartMetal, line 77: "ProxyOverlay" is not Blockstore concept.
Maybe it is better to disable logical used blocks calculation when `IsSystem == true` and `!BaseDiskId.empty()`?
Or even when `IsSystem == true`.

And this fix looks like "kostyl", so write comment about that, please

cloud/blockstore/libs/storage/service/service_ut_create.cpp
```diff
@@ -700,6 +700,65 @@ Y_UNIT_TEST_SUITE(TServiceCreateVolumeTest)
         }
     }
 
+    Y_UNIT_TEST(ShouldCreateVolumeWithMultipartitionBaseDisk)
```
SvartMetal, line 703: ShouldCreateSystemVolumeWithMultipartitionedBaseDisk

cloud/blockstore/libs/storage/volume/partition_requests.h
BarkovBG: nbs craches on Y_ABORT_UNLESS but we want to get readable error

so ythrow is better solution here

cloud/blockstore/libs/storage/protos/part.proto
BarkovBG: i dont know that proxy overlay disks is system

fixed

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
```diff
@@ -228,7 +228,7 @@ void TPartitionActor::CompleteLoadState(
     Y_ABORT_UNLESS(State->GetGarbageQueue().AddNewBlobs(args.NewBlobs));
     Y_ABORT_UNLESS(State->GetGarbageQueue().AddGarbageBlobs(args.GarbageBlobs));
 
-    if (State->GetBaseDiskId()) {
+    if (State->GetBaseDiskId() && !partitionConfig.GetIsSystem()) {
```
SvartMetal, line 231: Need comment about why we disable logical used blocks calculation for "system" disks

SvartMetal: Still is not answered why we disable that for such disks

SvartMetal: And it would be better to explain what "proxy overlay" disks are used for

SvartMetal (approved):

yegorskii (approved):

PR #795, opened 2024-03-21
feature: omit yc-github-runner in favor of python script

librarian:
I replaced js github action with python one mostly because of that nobody knows js in the team and modifications and bugfixes are very difficult + python sdk can change endpoint without uploading the same code to npm repo.

Also I incorporated few useful features that before required additional steps in prepare stage:
1. Getting and adding ssh keys for github user (api calls and setting ssh keys at the creation of vm is faster)
2. Showing ip addresses is basically free now (it is in response of CreateInstanceRequest)

Future plans are:
1. Move it to separate repo and maybe create docker action
2. Test if we can launch runner as `github` user, not root (to remove annoying `sudo -H -E everywhere` and everything)

### DIFF
--- a/.github/actions/runner_remove/action.yaml
+++ b/.github/actions/runner_remove/action.yaml
@@ -1,0 +1,50 @@
+name: Delete runner VM
+description: Delete runner VM using create-vm.py
+inputs:
+  repo_owner:
+    required: true
+    description: "repo owner"
+  repo:
+    required: true
+    description: "repo name"
+  service_account_key:
+    required: true
+    description: "service account key"
+  token:
+    required: true
+    description: "GITHUB_TOKEN"
+  vm_id:
+    required: true
+    description: "vm id"
+
+
+runs:
+  using: composite
+  steps:
+    - name: write service account key to file
+      shell: bash
+      run: |
+        export SERVICE_ACCOUNT_KEY=$(mktemp)
+        echo "SERVICE_ACCOUNT_KEY=${SERVICE_ACCOUNT_KEY}" >> $GITHUB_ENV
+        echo "${key_contents}" > $SERVICE_ACCOUNT_KEY
+      env:
+        key_contents: ${{ inputs.service_account_key }}
+    - name: install dependencies
+      shell: bash
+      run: |
+        pip install yandexcloud==0.258.0 PyGithub==2.2.0
+    - name: remove vm
+      id: remove-vm
+      shell: bash
+      run: |
+        python .github/scripts/create-vm.py remove \
+          --github-repo-owner "${REPO_OWNER}" \
+          --github-repo "${REPO}" \
+          --service-account-key "${SERVICE_ACCOUNT_KEY}" \
+          --id "${VM_ID}" \
+          --apply
+      env:
+        GITHUB_TOKEN: ${{ inputs.token }}
+        REPO_OWNER: ${{ inputs.repo_owner }}
+        REPO: ${{ inputs.repo }}
+        VM_ID: ${{ inputs.vm_id }}
--- a/.github/actions/runner_remove/action.yaml
+++ b/.github/actions/runner_remove/action.yaml
@@ -1,5 +1,5 @@
 name: Delete runner VM
-description: Delete runner VM using create-vm.py
+description: Delete runner VM using manage-vm.py
 inputs:
   repo_owner:
     required: true
@@ -37,7 +37,7 @@ runs:
       id: remove-vm
       shell: bash
       run: |
-        python .github/scripts/create-vm.py remove \
+        python .github/scripts/manage-vm.py remove \
           --github-repo-owner "${REPO_OWNER}" \
           --github-repo "${REPO}" \
           --service-account-key "${SERVICE_ACCOUNT_KEY}" \
--- a/.github/scripts/create-vm.py
+++ b/.github/scripts/create-vm.py
@@ -30,7 +30,7 @@ logging.basicConfig(
 )
 
 
-def output(key, value, secret=False):
+def github_output(key, value, secret=False):
     GITHUB_OUTPUT = os.environ.get("GITHUB_OUTPUT")
 
     if GITHUB_OUTPUT:
@@ -66,13 +66,14 @@ def fetch_github_team_public_keys(gh, github_org, team_slug):
     logging.info(
         f"Fetching SSH keys for members: {[member.login for member in members]}"
     )
+    member_keys_count = 0
     for member in members:
-        member_keys = []
+
         for key in member.get_keys():
-            member_keys.append(key.key)
+            member_keys_count += 1
             ssh_keys.append(key.key)
 
-        logging.debug(f"Fetched {len(member_keys)} SSH keys for {member.login}")
+        logging.debug(f"Fetched {member_keys_count} SSH keys for {member.login}")
 
     logging.debug(f"Fetched SSH keys: {ssh_keys}")
     return ssh_keys
@@ -107,13 +108,15 @@ def generate_cloud_init_script(user, ssh_keys, owner, repo, token, version, labe
     ]
     cloud_init["runcmd"] = ["\n".join(script)]
     logging.info(
-        f"Cloud-init: \n{yaml.dump(cloud_init, default_flow_style=False, width=math.inf)}".replace(
+        f"Cloud-init: \n{yaml.safe_dump(cloud_init, default_flow_style=False, width=math.inf)}".replace(
             token, "****"
         )
     )
     return (
         "#cloud-config\n"
-        + yaml.dump(cloud_init, default_flow_style=False, width=math.inf),  # noqa: W503
+        + yaml.safe_dump(  # noqa: W503
+            cloud_init, default_flow_style=False, width=math.inf
+        ),  # noqa: W503
         ssh_keys,
     )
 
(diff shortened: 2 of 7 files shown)


.github/scripts/create-vm.py
```diff
@@ -0,0 +1,462 @@
...
+    DeleteInstanceMetadata,
+)
+from yandex.cloud.compute.v1.instance_pb2 import IpVersion
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s: %(levelname)s: %(message)s"
+)
+
+
+def output(key, value, secret=False):
```
jkuradobery, line 33: `append_output_file`? Also, upper-case if generally for constants in python, or for global variables, it is neither.

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,462 @@
...
+    logging.info(
+        f"Fetching SSH keys for members: {[member.login for member in members]}"
+    )
+    for member in members:
+        member_keys = []
+        for key in member.get_keys():
+            member_keys.append(key.key)
+            ssh_keys.append(key.key)
+
+        logging.debug(f"Fetched {len(member_keys)} SSH keys for {member.login}")
```
jkuradobery, line 75: `member_keys` should be `int`. It's content is not used, just the length.

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,462 @@
...
+
+    logging.debug(f"Fetched SSH keys: {ssh_keys}")
+    return ssh_keys
+
+
+def generate_cloud_init_script(user, ssh_keys, owner, repo, token, version, label):
+
+    script = [
```
jkuradobery, line 83: You can just simple go with a multiline string. No need for "\n".join(...).

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,462 @@
...
+    ]
+    cloud_init["runcmd"] = ["\n".join(script)]
+    logging.info(
+        f"Cloud-init: \n{yaml.dump(cloud_init, default_flow_style=False, width=math.inf)}".replace(
+            token, "****"
+        )
+    )
+    return (
+        "#cloud-config\n"
+        + yaml.dump(cloud_init, default_flow_style=False, width=math.inf),  # noqa: W503
```
jkuradobery, line 116: Why not yaml.safe_dump?

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,462 @@
...
+            "Authorization": f"Bearer {github_token}",
+            "Accept": "application/vnd.github+json",
+            "X-Github-Api-Version": "2022-11-28",
+        },
+    ).json()
+
+    token = result.get("token")
+    if token:
+        # Mask the token in the logs
+        print(f"::add-mask::{token}")
```
jkuradobery, line 134: Avoid print's if you use logging

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,462 @@
...
+    )
+
+    ssh_keys = fetch_github_team_public_keys(gh, args.github_org, args.github_team_slug)
+
+    # fmt: off
+    standard_v2_cpu = [
+        2, 4, 6, 8, 10, 12, 14, 16, 20, 24,
+        28, 32, 36, 40, 44, 48, 52, 56, 60
+    ]
+    # fmt: on
```
jkuradobery, line 157: What's the point of this?

.github/scripts/create-vm.py
librarian: This is not for debugging, it is control string to tell GA that this string is masked.

.github/scripts/create-vm.py
librarian: Validation of input

.github/scripts/create-vm.py
librarian: fixed

.github/scripts/create-vm.py
librarian: With multiline strings it is not easy to control indentation in strings and it is pretty important. If this script will grow bigger i would rather make it separate file and read it.

.github/scripts/create-vm.py
librarian: well it is a global variable in that regard, env variable for the whole process

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+            timeout=args.timeout,
+            logger=logging.getLogger(),
+        )
+        instance_id = result.response.id
+        name = result.response.name
+        primary_ipv4 = result.response.network_interfaces[0].primary_v4_address
+        external_ipv4 = primary_ipv4.one_to_one_nat.address
+        local_ipv4 = primary_ipv4.address
+        if instance_id:
+            logging.info(
```
jkuradobery, line 242: Use logging parametrization instead of an f string. I know it is not a big deal for now, but string-formatting-before logging is considered a bad practice overall. https://www.palkeo.com/en/blog/python-logging.html

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+            meta_type=CreateInstanceMetadata,
+            timeout=args.timeout,
+            logger=logging.getLogger(),
+        )
+        instance_id = result.response.id
+        name = result.response.name
+        primary_ipv4 = result.response.network_interfaces[0].primary_v4_address
+        external_ipv4 = primary_ipv4.one_to_one_nat.address
+        local_ipv4 = primary_ipv4.address
+        if instance_id:
```
jkuradobery, line 241: If you want to check for None - check explicitly. If the variable is bool you should rename it accordingly.

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+    ).get_self_hosted_runners()
+
+    runner_id = None
+    for runner in runners:
+        if runner.name == str(vm_id):
+            runner_id = runner.id
+            break
+
+    if runner_id is None:
+        logging.info(f"Runner with name {vm_id} not found")
```
jkuradobery, line 272: Same. you can do `logging.info("Runner with name %s not found", vm_id)`

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+            f"https://api.github.com/repos/{github_repo_owner}/{github_repo}/actions/runners/{runner_id}",
+            headers={
+                "Authorization": f"Bearer {github_token}",
+                "Accept": "application/vnd.github+json",
+                "X-Github-Api-Version": "2022-11-28",
+            },
+        ).status_code
+
+        if delete_status_code == 204:
+            return True
```
jkuradobery, line 286: Why return true? You do not use the return value.

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+    generated_string = "".join(
+        random.choices(string.ascii_lowercase + string.digits, k=8)
+    )
+    logging.info(f"Generated label: {generated_string}")
+    return generated_string
+
+
+class KeyValueAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):  # noqa: U100
+        kv_dict = {}
```
jkuradobery, line 53: Can be simplified to `kv_dict = dict([tuple(kv.split('=')) for kv in values.split(',')])`. But it is a bit more "pythonic" and not that readable as your version though.

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+        default="api.ai.nebius.cloud",
+        help="API Endpoint",
+    )
+
+    create = subparsers.add_parser("create", help="Create a new VM")
+    create.add_argument(
+        "--github-org",
+        required=True,
+        default="ydb-platform",
+        help="GitHub organization name (we will fetch SSH keys from this org)",
```
jkuradobery, line 334: What are those ssh keys used for?

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+            fp.write(f"{key}={value}\n")
+
+    logging.info(f"echo \"{key}={'******' if secret else value}\" >> $GITHUB_OUTPUT")
+
+
+def generate_github_label():
+    generated_string = "".join(
+        random.choices(string.ascii_lowercase + string.digits, k=8)
+    )
+    logging.info(f"Generated label: {generated_string}")
```
jkuradobery, line 47: Same as other comments on logging

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+        f"Fetching SSH keys for members: {[member.login for member in members]}"
+    )
+    member_keys_count = 0
+    for member in members:
+
+        for key in member.get_keys():
+            member_keys_count += 1
+            ssh_keys.append(key.key)
+
+        logging.debug(f"Fetched {member_keys_count} SSH keys for {member.login}")
```
jkuradobery, line 76: Same as other comments on logging

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+
+    logging.debug(f"Fetched SSH keys: {ssh_keys}")
+    return ssh_keys
+
+
+def generate_cloud_init_script(user, ssh_keys, owner, repo, token, version, label):
```
jkuradobery, line 82: Please, add a typehint

.github/actions/runner_remove/action.yaml
```diff
@@ -0,0 +1,50 @@
+name: Delete runner VM
+description: Delete runner VM using create-vm.py
```
jkuradobery, line 2: > Delete runner VM using create-vm.py
Maybe you should consider naming the file the other way, or using two different scripts?

.github/scripts/create-vm.py
```diff
@@ -0,0 +1,465 @@
...
+            runner_id = runner.id
+            break
+
+    if runner_id is None:
+        logging.info(f"Runner with name {vm_id} not found")
+        raise ValueError(f"Runner with name {vm_id} not found")
+
+    if apply:
+        delete_status_code = requests.delete(
+            f"https://api.github.com/repos/{github_repo_owner}/{github_repo}/actions/runners/{runner_id}",
```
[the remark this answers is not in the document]
librarian: Don't see the point of that

.github/scripts/create-vm.py
librarian: I understand that it is good practice, but I don't see why it would be necessary here. It is not really big script and for now it is not intended to use it outside of our workflows

jkuradobery (approved):

qkrorlqr (approved):

PR #811, opened 2024-03-22
NBSNEBIUS-105: remove sync. erase from ADD_DEVICE

sharpeye:
There is no longer a need to delay the response to the ADD_DEVICE request until all local disks are cleaned.
Instead, if there are dirty local disks on the host, ADD_HOST will return E_TRY_AGAIN.


qkrorlqr:
нужно описание

tpashkin (approved):

komarevtsev-d (approved):

PR #813, opened 2024-03-25
NBSNEBIUS-197: StartEndpointRequest creates socket file if needed

sanek325:
There is a case when kubelet removes a created socket and sends again StartEndpointRequest to NBS. Then NBS reply S_ALREADY, but don’t recreate socket.

### DIFF
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -582,6 +582,10 @@ private:
         NProto::TStartEndpointRequest newReq,
         NProto::TStartEndpointRequest oldReq);
 
+    NProto::TError RestartListenerEndpoint(
+        TCallContextPtr ctx,
+        const NProto::TStartEndpointRequest& request);
+
     NProto::TError OpenAllEndpointSockets(
         const NProto::TStartEndpointRequest& request,
         const TSessionInfo& sessionInfo);
@@ -748,6 +752,20 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
     auto it = Endpoints.find(socketPath);
     if (it != Endpoints.end()) {
         auto endpoint = it->second;
+
+        if (!NFs::Exists(socketPath)) {
+            // restart listener endpoint to recreate the socket
+            auto error = RestartListenerEndpoint(ctx, *endpoint.Request);
+            if (HasError(error)) {
+                return TErrorResponse(error);
+            }
+
+            if (!NFs::Exists(socketPath)) {
+                return TErrorResponse(E_INVALID_STATE, TStringBuilder()
+                    << "failed to recreate socket: " << socketPath.Quote());
+            }
+        }
+
         auto error = AlterEndpoint(std::move(ctx), *request, *endpoint.Request);
         if (HasError(error)) {
             return TErrorResponse(error);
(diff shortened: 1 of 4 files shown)


yegorskii (approved):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -748,7 +756,7 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
     auto it = Endpoints.find(socketPath);
     if (it != Endpoints.end()) {
         auto endpoint = it->second;
-        auto error = AlterEndpoint(std::move(ctx), *request, *endpoint.Request);
+        auto error = RestartEndpoint(std::move(ctx), *request, *endpoint.Request);
```
budevg, line 759: instead of refactoring AlterEndpoint and checking for existence of socket in the end.
Let's first ensure that socket exists (restarting endpoint if needed) and then invoke AlterEndpoint ... then no need to refactor AlterEndpoint at all
the code will look like this
```c++
    auto it = Endpoints.find(socketPath);
    if (it != Endpoints.end()) {
        auto endpoint = it->second;
	auto error = EnsureEndpointSocketExists(...)
        if (HasError(error)) {
            return TErrorResponse(error);
        }
				
        error = AlterEndpoint(std::move(ctx), *request, *endpoint.Request);
        if (HasError(error)) {
            return TErrorResponse(error);
        }
```

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -837,59 +845,130 @@ NProto::TError TEndpointManager::AlterEndpoint(
...
     auto alterFuture = listener->AlterEndpoint(
-        oldReq,
+        request,
         sessionInfo.Volume,
         sessionInfo.Session);
 
     return Executor->WaitFor(alterFuture);
 }
 
+NProto::TError TEndpointManager::RecreateSocket(
```
budevg, line 928: A bit of confusing to call this `RecreateSocket` where there is nothing socket related in the function. It's better to put a comment explaining that restarting listener endpoint should recreate the socket and it would be good to recheck that socket exists and if it doesn't exist fail the function

budevg (changes requested):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
sanek325: There is a comparison of requests inside AlterEndpoint and it's a problem, because if nbs gets StartEndpointRequest with the same socket path but with different other arguments then we should do nothing and just return error "endpoint has already been started with other args". 
That is why I decided to refactor AlterEndpoint, and first of all compare the requests and then do AlterEndpoint and RecreateSocket if needed.

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
sanek325: Yes, I agree. 
So I added a comment, added a recheck that socket exists and also renamed the method.

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
budevg: Yep but regardless of the results of alter endpoint, in common case socket path should not be deleted for started endpoint. And  so it should be safe to stop/start endpoint with old parameters if needed before applying the alter endpoint logic

budevg (approved):

yegorskii (approved):

Issue #196, opened 2024-01-22
[NBS] Ability to save block checksums in TBlobMeta to check them upon ReadBlob


qkrorlqr:
In case of data corruption bugs we need to be able to track the layer which causes the issue. This issue is about recording block checksums after WriteBlob in TBlobMeta (https://github.com/ydb-platform/nbs/blob/dd8b1e2aee138b3d6b027967c7ccbbed1a0fb722/cloud/blockstore/libs/storage/protos/part.proto#L148) to check that the data read via ReadBlob is the same data that was written upon WriteBlob. Since it implies storing a lot of metadata it will cause a performance hit => we can't enable this logic by default, need to be able to enable it via features config.

qkrorlqr:
> need to be able to enable it via features config.

there is actually a lot of value in enabling this by default
but I will limit the block range which is affected and this block range will be configurable via TStorageServiceConfig

qkrorlqr:
Basic implementation done. TODO:
* implement checksums for unconfirmed blobs
* implement checksums for patched blobs

qkrorlqr:
> implement checksums for unconfirmed blobs

done

qkrorlqr:
the only thing that's left to do is:
* implement checksums for patched blobs

everything else is implemented, tested, deployed and seems to work fine

qkrorlqr:
patching-related thing moved here: https://github.com/ydb-platform/nbs/issues/1986

PR #825, opened 2024-03-26
issue-196: block checksums should be calculated after we copy the data from the client's buffer to our buffer, otherwise our client may modify the data between the checksum calculation stage and the writeblob stage

qkrorlqr:
#196

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_writemergedblocks.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_writemergedblocks.cpp
@@ -278,6 +278,9 @@ void TWriteMergedBlocksActor::NotifyCompleted(
     const NProto::TError& error)
 {
     using TEvent = TEvPartitionPrivate::TEvWriteBlocksCompleted;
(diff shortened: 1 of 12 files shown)


cloud/blockstore/libs/storage/partition/part_actor_writemergedblocks.cpp
```diff
@@ -528,7 +559,7 @@ void TPartitionActor::WriteMergedBlocks(
         requestInBuffer.Data.ReplyLocal,
         shouldAddUnconfirmedBlobs,
         std::move(requestInBuffer.Data.Handler),
-        checksumsEnabled
+        checksumsEnabled ? State->GetBlockSize() : 0
```
SvartMetal, line 562: Maybe annotate this argument, for better reading?

qkrorlqr: I think that here the meaning of this argument is obvious

SvartMetal (approved):

yegorskii (approved):

PR #853, opened 2024-03-28
fix: ternary-like expressions in github actions are weird. Also fix for missing rebase expressions

librarian:
In github actions there is a ternary-like operator, but it is not really [ternary](https://docs.github.com/en/actions/learn-github-actions/expressions#example).

>  It is important to note that the first value after the && must be truthy. Otherwise, the value after the || will always be returned.

What means truthy?

> Note that in conditionals, falsy values (false, 0, -0, "", '', null) are coerced to false and truthy (true and other non-falsy values) are coerced to true.

And bug occurs because 0 is casted, for some reason, to false

### DIFF
--- a/.github/workflows/build_and_test_cmake.yaml
+++ b/.github/workflows/build_and_test_cmake.yaml
@@ -53,12 +53,18 @@ jobs:
       with:
         submodules: true
         ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: ${{ !contains(github.event.pull_request.labels.*.name, 'rebase') && 1 || 0 }}
+    - name: Rebase PR
+      if: ${{ github.event.pull_request.head.sha != '' && contains(github.event.pull_request.labels.*.name, 'rebase') }}
+      shell: bash
+      run: |
+        git fetch origin ${{ github.event.pull_request.base.ref }}
+        git rebase origin/${{ github.event.pull_request.base.ref }}
     - name: Checkout
       uses: actions/checkout@v4
       if: github.event.pull_request.head.sha == ''
       with:
         submodules: true
-
     - name: Prepare s3cmd
       uses: ./.github/actions/s3cmd
       with:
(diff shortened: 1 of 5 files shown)


.github/workflows/build_and_test_cmake.yaml
```diff
@@ -53,12 +53,18 @@ jobs:
       with:
         submodules: true
         ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: ${{ !contains(github.event.pull_request.labels.*.name, 'rebase') && 1 || 0 }}
+    - name: Rebase PR
```
budevg, line 57: looks like this commit not just about fixing ternary operator but also about doing rebase ... better to split into separate commits

.github/workflows/build_and_test_cmake.yaml
```diff
@@ -53,12 +53,18 @@ jobs:
...
+      if: ${{ github.event.pull_request.head.sha != '' && contains(github.event.pull_request.labels.*.name, 'rebase') }}
+      shell: bash
```
budevg, line 59: Is re-basing during PR supposed to protect us from merging code which was not tested after rebase to latest main branch ? It won't help if all tests pass but user will decide to press the merge button after several days.

We could enforce `--ff-only` merges. Not sure if github supports it out of the box

budevg (changes requested):

.github/workflows/build_and_test_cmake.yaml
librarian: Idea is that it will lover chances of merging such code. We may implement on: push checks later

.github/workflows/build_and_test_cmake.yaml
librarian: This is just fixing places where rebase should happen, but was missed in previous iteration. It will work weird if you split it in separate changes.

librarian: Do you mean move this thing to separate commit inside this PR? It will be squashed afterwards though

budevg: maybe just  change commit name to `adding missing rebase and fixing ternary expressions` or something when you merge so it will be less confusing.
In general you don't have to squash you can "rebase and merge" and preserve two commits if it makes sense to preserve them. Anyway I don't insist on changing this PR. But would be great to have separate commits in the future

budevg (approved):

PR #876, opened 2024-04-02
fix: skip runner removal from GH and add wait for runner registration

librarian:


.github/scripts/manage-vm.py
```diff
@@ -32,7 +33,7 @@
 logger = logging.getLogger(__name__)
 
 
-def github_output(key, value, secret=False):
+def github_output(key: str, value: str, secret: bool = False):
```
jkuradobery, line 36: use_secret/is_secret? Just a secret sounds like you want to pass a secret.

.github/scripts/manage-vm.py
```diff
@@ -144,7 +155,49 @@ def get_runner_token(github_repo_owner, github_repo, github_token):
...
+    github_repo_owner: str,
+    github_repo: str,
+    sleep_interval: int = 10,
+    intervals: int = 60,
+) -> bool:
+    logger.info(
+        "Waiting for runner registration for %d*%d seconds", intervals, sleep_interval
+    )
+
+    for i in range(intervals):
```
jkuradobery, line 170: If you want to wait for a specific event to occur, the preferred solution would be:
```
finished_at = time.monotonic() + intervals
while True:
    if find_runner_by_name(...):
        return True
    if time.monotonic() > finished_at:
         return False
    time.sleep(max(min(sleep_interval, finished_at - time.monotonic()), 0))
```

jkuradobery: But if you do not care much for sleeping a bit more, then bob's your uncle, the for approach works fine.

.github/scripts/manage-vm.py
```diff
@@ -144,7 +155,49 @@ def get_runner_token(github_repo_owner, github_repo, github_token):
...
 
 
-def create_vm(sdk, args):
+def wait_for_runner_registration(
+    client: Github,
+    vm_id: str,
+    github_repo_owner: str,
+    github_repo: str,
+    sleep_interval: int = 10,
+    intervals: int = 60,
```
jkuradobery, line 164: Let's just call em timeout_sec

.github/scripts/manage-vm.py
```diff
@@ -144,7 +155,49 @@ def get_runner_token(github_repo_owner, github_repo, github_token):
...
+def find_runner_by_name(
+    client: Github, github_repo_owner: str, github_repo: str, vm_id: str
+) -> str:
+    runners = client.get_repo(
+        f"{github_repo_owner}/{github_repo}"
+    ).get_self_hosted_runners()
+
+    runner_id = None
+    for runner in runners:
+        if runner.name == str(vm_id):
```
jkuradobery, line 188: Why cast vm_id to string, is the type hint wrong?

.github/scripts/manage-vm.py
```diff
@@ -144,7 +155,49 @@ def get_runner_token(github_repo_owner, github_repo, github_token):
...
+    github_repo: str,
+    sleep_interval: int = 10,
+    intervals: int = 60,
+) -> bool:
+    logger.info(
+        "Waiting for runner registration for %d*%d seconds", intervals, sleep_interval
+    )
+
+    for i in range(intervals):
+        if find_runner_by_name(client, github_repo_owner, github_repo, vm_id):
```
jkuradobery, line 171: You should check for `is None` explicitly.

.github/scripts/manage-vm.py
```diff
@@ -253,30 +306,37 @@ def create_vm(sdk, args):
             github_output("local-ipv4", local_ipv4)
             if external_ipv4:
                 github_output("external-ipv4", external_ipv4)
+
+            logger.info("Waiting for VM to be registered as Github Runner")
+            runner_id = wait_for_runner_registration(
+                gh, instance_id, args.github_repo_owner, args.github_repo, 10
+            )
+            if runner_id is not None:
```
jkuradobery, line 314: You can just raise an exception while waiting for the runner registration. And log runner_id there also.

jkuradobery: I do not see runner_id used in other places, but if it is not used except for logging you can return None.

librarian: I, initally wanted to call it in `remove_vm` also, so I decided to move this decision, to throw exception or not, to different level.

.github/scripts/manage-vm.py
librarian: I thought about that, but for me more important to have specific number of retries, than to wait some amount of time.

yegorskii (approved):

dvrazumov (approved):

.github/scripts/manage-vm.py
```diff
@@ -144,7 +156,52 @@ def get_runner_token(github_repo_owner, github_repo, github_token):
...
+        "Waiting for runner registration every %d seconds x %d times",
+        timeout_sec,
+        retries,
+    )
+
+    for i in range(retries):
+        runner_id = find_runner_by_name(client, github_repo_owner, github_repo, vm_id)
+        if runner_id is not None:
+            return runner_id
+        else:
```
jkuradobery, line 177: You do not need "else" here. Flat code is better then nested.

.github/scripts/manage-vm.py
```diff
@@ -144,7 +156,52 @@ def get_runner_token(github_repo_owner, github_repo, github_token):
...
+        f"{github_repo_owner}/{github_repo}"
+    ).get_self_hosted_runners()
+
+    runner_id = None
+    for runner in runners:
+        if runner.name == vm_id:
+            runner_id = runner.id
+            break
+
+    if runner_id is None:
```
jkuradobery, line 196: Just log the vm_id and runner_id and return runner_id. The logging severity is the same. No need for additional checks/logs.

.github/scripts/manage-vm.py
```diff
@@ -253,30 +310,37 @@ def create_vm(sdk, args):
...
-            break
+    runner_id = find_runner_by_name(gh, github_repo_owner, github_repo, vm_id)
 
     if runner_id is None:
-        logger.info("Runner with name %s not found", vm_id)
-        raise ValueError("Runner with name %s not found", vm_id)
+        # this is not critical error, just log it and be done with it,
+        # removing the VM is more important
+        logger.info("Runner with name %s not found, skipping", vm_id)
+        return
```
jkuradobery, line 343: return None then

PR #921, opened 2024-04-07
NEBDUTY-970: refactor rdma test to be more robust to network errors (#882)

budevg:
- check network before running rdma test
- double check network on error during rdma test
- increase extra execute timeout from 5 secs to 10

### DIFF
--- a/cloud/blockstore/tests/rdma/rdma-test/test.py
+++ b/cloud/blockstore/tests/rdma/rdma-test/test.py
@@ -1,17 +1,27 @@
 import logging
+import time
 
 from cloud.blockstore.tests.python.lib.rdma import setup_rdma
 
 _logger = logging.getLogger(__name__)
 
+EXECUTE_EXTRA_TIMEOUT_SECS = 10
 
-def test_rdma():
-    rdma_port = 9999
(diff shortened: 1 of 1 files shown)


Sazonov99 (approved):

cloud/blockstore/tests/rdma/rdma-test/test.py
```diff
@@ -29,13 +40,32 @@ def test_rdma():
...
+
+    _logger.debug("starting rdma-test")
+
+    h1, h2 = setup_rdma()
+    check_network_connectivity(h1, h2)
+    try:
+        check_rdma_connectivity(h1, h2, timeout=test_duration_sec)
+    except Exception as e:
+        _logger.error("check_rdma_connectivity failed %s", e)
+        check_network_connectivity(h1, h2)
```
tpashkin, line 70: А зачем второй раз?

budevg: We had rdma test failures where the connection was stuck. There was a suspicion that this is some network issue. This is why in case of error we want to double check the connectivity

tpashkin (approved):

PR #933, opened 2024-04-09
NBS-1526 [disk manager] call mlock on dm process start

gy2411:
Lock process memory ranges containing binary and .so files when process starts. Similar to nbs: https://github.com/ydb-platform/nbs/blob/main/cloud/storage/core/libs/daemon/mlock.cpp

After memory is locked, /proc/<pid>/status and /proc/<pid>/maps for dm process look like this:
[dm_memory_state.txt](https://github.com/ydb-platform/nbs/files/15027268/dm_memory_state.txt) 
As we see, 39572 kB are locked.

NOTE! Before starting dm with this change, LimitMEMLOCK parameter in [Service] section of `/usr/lib/systemd/system/yc-disk-manager.service` should be set. Otherwise the process will fail to lock memory and disk manager will not start. I suggest `LimitMEMLOCK=268435456` (256 MiB).

### DIFF
--- a/cloud/disk_manager/internal/pkg/util/mlock.go
+++ b/cloud/disk_manager/internal/pkg/util/mlock.go
@@ -1,10 +1,78 @@
 package util
 
-type MemRange struct {
-	start  uint64
-	length uint64
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"syscall"
+
+	"github.com/ydb-platform/nbs/cloud/tasks/errors"
+)
+
+type memRange struct {
+	start uintptr
+	end   uintptr
 }
 
-func parseMemRange(line string) (MemRange, error) {
+type memItem struct {
+	memoryRange memRange
+	permissions string
+	offset      uintptr
+	device      string
+	inode       uint64
+	pathname    string
+}
 
+func parseMemRange(line string) (*memItem, error) {
+	var item memItem
+	_, err := fmt.Sscanf(
+		line,
+		"%x-%x %s %x %s %d %s",
+		&item.memoryRange.start,
+		&item.memoryRange.end,
+		&item.permissions,
+		&item.offset,
+		&item.device,
+		&item.inode,
+		&item.pathname,
+	)
+	return &item, err
+}
+
+func mlock(memoryRange memRange) error {
+	len := memoryRange.end - memoryRange.start
+	_, _, errno := syscall.Syscall(syscall.SYS_MLOCK, memoryRange.start, len, 0)
+	if errno != 0 {
+		return errors.NewNonRetriableErrorf("mlock syscall failed with errno %d", errno)
+	}
+	return nil
+}
+
+func shouldLockRange(item *memItem) bool {
+	return item.inode != 0 && item.permissions[0] == 'r'
+}
+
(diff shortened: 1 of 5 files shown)


cloud/disk_manager/internal/pkg/util/mlock.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"syscall"
+
+	"github.com/ydb-platform/nbs/cloud/tasks/errors"
+)
```
SvartMetal, line 11: ////...

like here https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/pkg/app/controlplane.go#L37

cloud/disk_manager/internal/pkg/util/mlock_test.go
```diff
@@ -0,0 +1,99 @@
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
```
SvartMetal, line 7: ///...

cloud/disk_manager/internal/pkg/util/mlock_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	for _, testCase := range cases {
+		item, err := parseProcMapsLine(testCase.line)
+		require.NoError(t, err)
+		shouldLock := shouldLockRange(item)
+		require.EqualValues(t, testCase.shouldLockExpected, shouldLock)
+	}
+}
+
+func TestLockBinaryDoesNotFail(t *testing.T) {
```
SvartMetal, line 96: TestLockProcessMemory

cloud/disk_manager/internal/pkg/util/mlock_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	for _, testCase := range cases {
+		item, err := parseProcMapsLine(testCase.line)
+		require.NoError(t, err)
+		require.EqualValues(t, testCase.itemExpected, *item)
+	}
+
+	_, err := parseProcMapsLine("02443000-02444000 r--p 02242000 fd:11")
+	require.Error(t, err)
+
```
SvartMetal, line 37: unnecessary \n

cloud/disk_manager/internal/pkg/util/mlock.go
```diff
@@ -0,0 +1,82 @@
...
+		"%x-%x %s %x %s %d %s",
+		&item.memRange.start,
+		&item.memRange.end,
+		&item.permissions,
+		&item.offset,
+		&item.device,
+		&item.inode,
+		&item.pathname,
+	)
+	if err != nil && (err != io.EOF || n < 6) {
```
SvartMetal, line 40: Why checking io.EOF is needed here?

Do we have tests where io.EOF happens while scanning here?

SvartMetal: Can we separate this if into some, and write comment about every case?

SvartMetal (approved):

BarkovBG (approved):

PR #1063, opened 2024-04-25
should mount proxy overlay disk locally

BarkovBG:
We do not use optimization that provided by proxy overlay disks because of remote mounting 

data path was:
Snapshot ->  nbs control -> nbs control where disk is mounted -> nbs control -> snapshot 

now it will be: 
snapshot -> nbs control (proxy overlay disk will be mounted on it locally) -> snapshot 

Also we can probably improve load balancing on the nbs control using this approach

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/nbs/source.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/nbs/source.go
@@ -249,11 +249,14 @@ func NewDiskSource(
 	dontReadFromCheckpoint bool,
 ) (dataplane_common.Source, error) {
 
-	if len(proxyDiskID) == 0 {
-		proxyDiskID = diskID
+	var session *nbs.Session
+	var err error
+	if len(proxyDiskID) != 0 {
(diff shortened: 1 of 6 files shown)


cloud/disk_manager/internal/pkg/dataplane/nbs/source.go
```diff
@@ -249,11 +249,14 @@ func NewDiskSource(
 	dontReadFromCheckpoint bool,
 ) (dataplane_common.Source, error) {
 
-	if len(proxyDiskID) == 0 {
-		proxyDiskID = diskID
+	var session *nbs.Session
+	var err error
+	if len(proxyDiskID) != 0 {
```
SvartMetal, line 254: Here we pass proxyDiskID == diskID https://github.com/ydb-platform/nbs/blob/b4ff2c5e13e4a2a55cb15064785ce82688e27df3/cloud/disk_manager/internal/pkg/dataplane/transfer_from_disk_to_disk_task.go#L59

Looks like it should be fixed

SvartMetal (approved):

gy2411 (approved):

PR #1091, opened 2024-04-29
fix: don't trigger DiskRegistryAgentNotFound when the agent is not in AgentRegInfo

sharpeye:
The bug: if an agent does not connect after restarting DR, then dependent disks will not be transferred to the RO state by TEvSwitchAgentDisksToReadOnlyRequest.

### DIFF
--- a/cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_switch_agent_to_read_only.cpp
+++ b/cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_switch_agent_to_read_only.cpp
@@ -20,16 +20,12 @@ void TDiskRegistryActor::HandleSwitchAgentDisksToReadOnly(
 {
     auto* msg = ev->Get();
 
-    auto it = AgentRegInfo.find(msg->AgentId);
-    if (it == AgentRegInfo.end()) {
-        ReportDiskRegistryAgentNotFound(
-            TStringBuilder()
-            << "HandleSwitchAgentDisksToReadOnly:AgentId: " << msg->AgentId);
-        return;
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_switch_agent_to_read_only.cpp
```diff
@@ -20,16 +20,12 @@ void TDiskRegistryActor::HandleSwitchAgentDisksToReadOnly(
...
-    if (it == AgentRegInfo.end()) {
-        ReportDiskRegistryAgentNotFound(
-            TStringBuilder()
-            << "HandleSwitchAgentDisksToReadOnly:AgentId: " << msg->AgentId);
-        return;
-    }
-
-    if (it->second.Connected) {
-        LOG_INFO(ctx, TBlockStoreComponents::DISK_REGISTRY,
+    if (auto it = AgentRegInfo.find(msg->AgentId);
```
dvrazumov, line 23: What if `SwitchAgentDisksToReadOnly` event was scheduled, and shortly after that the agent was removed from configuration? ExecuteTx<TSwitchAgentDisksToReadOnly>will be triggered on nonexistent agent?

sharpeye: Yes, but it's not a problem: SwitchAgentDisksToReadOnly returns E_NOT_FOUND

dvrazumov (approved):

komarevtsev-d (approved):

PR #1130, opened 2024-05-06
[Disk Manager] fix namings in nbs client tests; should not unmount nrd disks during test because it causes the volume tablet restart

BarkovBG:


cloud/disk_manager/internal/pkg/clients/nbs/tests/client_test.go
```diff
@@ -151,9 +151,20 @@ func writeRandomBlocks(
 	require.NoError(t, err)
 	defer session.Close(ctx)
 
+	writeBlocksToMountedDisk(t, ctx, session, startIndex, blockCount)
+}
+
+func writeBlocksToMountedDisk(
```
SvartMetal, line 157: writeBlocksToSession

or we can inline this method

SvartMetal (approved):

leftmain (approved):

PR #1163, opened 2024-05-10
Speed up checks

librarian:
Overview of what has been done:
1. Increased quotas for our cloud and since we have allocated hosts I increased VM memory to its limit and therefore we can increase number of test threads. This gave biggest performance boost: for nightly build, _build_ time decreased by 1hr. But if we upload .ya dir it takes this time back (archiving and uploading takes ~ 1hr)
2. Added task to build VM image on schedule. In result we shave off approximately 2-3 minutes each time, saving on creating VM (before we needed to unpack and update github runner software) and installing necessary packages (or checking that we installed them).
3. Said VM image can contain .ya dir with built code and downloaded artifacts. If we use network-ssd with overlayed VM image we usually shave off 1-2m. But we use network-ssd-nonreplicated which don't have this feature yet, so it is temporarily disabled. It is mostly limited by host network. And it is basically the same as using network cache machine (which we use by default). Additional benefit that we still will have this bonus if cache machine will die for some reason

### DIFF
--- a/.github/scripts/manage-images.py
+++ b/.github/scripts/manage-images.py
@@ -36,9 +36,9 @@ def status_to_string(image: Image) -> str:
 
 def convert_size(size_bytes):
     if size_bytes == 0:
-        return "0B"
-    size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
-    i = int(math.floor(math.log(size_bytes, 1024)))
+        return "0 B"
+    size_name = ("B", "KB", "MB", "GB", "TB")
+    i = math.floor(math.log(size_bytes, 1024))
     p = math.pow(1024, i)
     s = round(size_bytes / p, 2)
     return f"{s} {size_name[i]}"
@@ -71,9 +71,9 @@ def main(
 
     if update_image_id:
         variable.edit(value=new_image_id)
-        logger.info("%s (new) = %s" % (image_variable_name, new_image_id))
+        logger.info("%s (new) = %s", image_variable_name, new_image_id)
     else:
-        logger.info("Would set %s (new) = %s" % (image_variable_name, new_image_id))
+        logger.info("Would set %s (new) = %s", image_variable_name, new_image_id)
 
     request = ListImagesRequest(
         folder_id=folder_id,
@@ -104,24 +104,22 @@ def main(
             candidate_images.append(image)
 
         logger.info(
-            "Found image%s: %s %s %s %s %s %s %s %s"
-            % (
-                # fmt: off
-                prefix, image.id, image.name, status, created_at,
-                image.family, storage_size, min_disk_size,
-                ', '.join(f"{k}={v}" for k, v in image.labels.items())
-                # fmt: on
-            )
+            "Found image%s: %s %s %s %s %s %s %s %s",
+            # fmt: off
+            prefix, image.id, image.name, status, created_at,
+            image.family, storage_size, min_disk_size,
+            ', '.join(f"{k}={v}" for k, v in image.labels.items())
+            # fmt: on
         )
-    logger.info(f"Total images: {len(response.images)}")
+    logger.info("Total images: %d", len(response.images))
 
     for image in candidate_images[images_to_keep:]:
         if remove_old_images:
-            logger.info(f"Removing image {image.id}")
+            logger.info("Removing image %s", image.id)
             client.Delete(DeleteImageRequest(image_id=image.id))
-            logger.info(f"Image {image.id} removed")
+            logger.info("Image %s removed", image.id)
         else:
-            logger.info(f"Would remove image {image.id}")
+            logger.info("Would remove image %s", image.id)
 
 
 if __name__ == "__main__":
--- a/.github/scripts/manage-images.py
+++ b/.github/scripts/manage-images.py
@@ -91,16 +91,17 @@ def main(
         min_disk_size = convert_size(image.min_disk_size)
         prefix = " (PROTECTED)"
         postfix = ""
+
+        if image.id == new_image_id:
+            postfix = " (NEW)"
+
         if (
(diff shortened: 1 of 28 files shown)


.github/scripts/manage-images.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def convert_size(size_bytes):
+    if size_bytes == 0:
+        return "0B"
+    size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
+    i = int(math.floor(math.log(size_bytes, 1024)))
+    p = math.pow(1024, i)
+    s = round(size_bytes / p, 2)
+    return f"{s} {size_name[i]}"
```
jkuradobery, line 44: Why 0B but "2 GB"?

.github/scripts/manage-images.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def status_to_string(image: Image) -> str:
+    return Image.Status.Name(image.status)
+
+
+def convert_size(size_bytes):
+    if size_bytes == 0:
+        return "0B"
+    size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
```
jkuradobery, line 40: Do we really need PB/EB/ZB given u use this code only in this script and it won't be used outside?

.github/scripts/manage-images.py
```diff
@@ -0,0 +1,209 @@
...
+    i = int(math.floor(math.log(size_bytes, 1024)))
```
jkuradobery, line 41: From the doc: 
Help on built-in function floor in module math:

floor(x, /)
    Return the floor of x as an Integral.
    
    This is the largest integer <= x.
(END)
You probably forgot to remove int()

.github/scripts/manage-images.py
```diff
@@ -0,0 +1,209 @@
...
+            prefix = ""
+            if image.id == new_image_id:
+                postfix = " (NEW)"
+            if image.family == image_family_name:
+                postfix += " (SELECTED)"
+            candidate_images.append(image)
+
+        logger.info(
+            "Found image%s: %s %s %s %s %s %s %s %s"
+            % (
```
jkuradobery, line 108: Why string formatting? Just pass arguments.

.github/scripts/manage-images.py
```diff
@@ -0,0 +1,209 @@
...
+                # fmt: off
+                prefix, image.id, image.name, status, created_at,
+                image.family, storage_size, min_disk_size,
+                ', '.join(f"{k}={v}" for k, v in image.labels.items())
+                # fmt: on
+            )
+        )
+    logger.info(f"Total images: {len(response.images)}")
```
jkuradobery, line 116: Same

.github/scripts/manage-images.py
```diff
@@ -0,0 +1,207 @@
...
+    response = client.List(request)
+    candidate_images = []
+    for image in response.images:
+        status = status_to_string(image)
+        created_at = datetime.fromtimestamp(image.created_at.seconds, tz=timezone.utc)
+        storage_size = convert_size(image.storage_size)
+        min_disk_size = convert_size(image.min_disk_size)
+        prefix = " (PROTECTED)"
+        postfix = ""
+        if (
```
jkuradobery, line 94: > Flat is better than nested.
```
if image.id in PROTECTED_IMAGE_IDS:
    continue
if image.family != image_family_name:
    continue
if image.id == new_image_id:
   continue
```

jkuradobery: Also it seems like here you can log before those conditions

.github/scripts/manage-images.py
```diff
@@ -0,0 +1,207 @@
...
+        ):
+            prefix = ""
+            if image.id == new_image_id:
+                postfix = " (NEW)"
+            if image.family == image_family_name:
+                postfix += " (SELECTED)"
+            candidate_images.append(image)
+
+        logger.info(
+            "Found image%s: %s %s %s %s %s %s %s %s",
```
jkuradobery, line 107: In logging you can name parameters via %(name)s

.github/scripts/manage-vm.py
```diff
@@ -98,36 +98,70 @@ def generate_cloud_init_script(
...
+    echo "$((date)) [$i] config.sh exited (or timed-out) with code $exit_code"
+    [ $exit_code -eq 0 ] || find /actions-runner -name *.log -print -exec cat {{}} \; # noqa: W605
+done
+# exit code 0 to skip the error and to boot vm correctly
+./run.sh || exit 0
+"""
 
     cloud_init = {}
-    cloud_init["ssh_pwauth"] = False
+    # cloud_init["ssh_pwauth"] = False
```
jkuradobery, line 150:
```
**(ssh_keys and {"ssh_keys": ssh_keys} or {})
```
Also do not need to assign ["run_cmd"]. You can make it a single dict literal.

.github/scripts/manage-images.py
librarian: This will break cycle and we will not see those images in logs. But there is still an error however. image.id == new_image_id should be outside of this if statement

.github/scripts/manage-images.py
librarian: To do that you need to pass huge dict with keys and all, and it doesn't look good and looks excsessive.

.github/scripts/manage-vm.py
librarian: I am not really familiar with this syntax and i try not to use esoteric syntax when I can, because it complicates things for people who are not familiar with it either (ie c++ developers who may not be familiar with python at all). And it is not some cpu/memory intensive application to go full swing for those optimisations with memory allocation where readability is more important.

jkuradobery (approved):

EvgeniyKozev (approved):

PR #1200, opened 2024-05-14
[Disk Manager] add feature for disabling memory locking (currently we are facing crashes on start inside containers)

SvartMetal:
Fix after #933


cloud/disk_manager/pkg/app/run.go
```diff
@@ -83,17 +82,22 @@ func run(
...
+		}
 	}
 
+	var err error
+
 	if len(hostname) == 0 {
 		hostname, err = os.Hostname()
 		if err != nil {
 			logging.Error(ctx, "Failed to get hostname from OS: %v", err)
-			return errors.NewNonRetriableError(err)
```
SvartMetal, line 96: Not needed

jkuradobery (approved):

gy2411 (approved):

PR #1229, opened 2024-05-17
NBSNEBIUS-193: mount volumes in local mode

sanek325:


dvrazumov (approved):

dvrazumov:
Maybe it would be useful to forward this option as a parameter with default value as "local"

sharpeye (approved):

PR #1271, opened 2024-05-24
[Disk Manager] should be ablet to delete filesystem without specifying zone ID (as we do for disks)

SvartMetal:


cloud/disk_manager/internal/pkg/services/filesystem/delete_filesystem_task.go
```diff
@@ -64,20 +59,30 @@ func (t *deleteFilesystemTask) deleteFilesystem(
...
 		return errors.NewNonCancellableErrorf(
 			"id %v is not accepted",
-			t.request.Filesystem.FilesystemId,
+			filesystemID,
 		)
 	}
 
-	err = client.Delete(ctx, t.request.Filesystem.FilesystemId)
+	zoneID := filesystemMeta.ZoneID
+	if len(zoneID) == 0 {
```
SvartMetal, line 67: By analogy with https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/services/disks/delete_disk_task.go#L73

jkuradobery (approved):

BarkovBG (approved):

PR #1339, opened 2024-06-04
Add metrics to csi-driver

dvrazumov:


cloud/blockstore/tools/csi_driver/internal/driver/driver.go
```diff
@@ -27,6 +30,15 @@ func getEndpoint(unixSocket string, port uint) string {
 	}
 }
 
+func getCsiMethodName(fullMethod string) string {
+	// take the last part from "/csi.v1.Controller/ControllerPublishVolume"
+	index := strings.LastIndex(fullMethod, "/")
```
jkuradobery, line 35: I think you should consider rewriting this as string.Split() and accessing the last element if we have > 1element in the split.

dvrazumov: Why? It's neither more efficient nor shorter

jkuradobery: I would consider this more self-explanatory though. Although, you might be right, it still looks ok, closing

jkuradobery (approved):

gy2411 (approved):

PR #1576, opened 2024-07-10
[Disk Manager] acceptance tests: support different types of disks, add flag to skip images

gy2411:
We want to run acceptace tests for disk registry based disks. Now acceptance tests have flag that specifies disk type, but we can not run tests for disk registry based based for 2 reasons:

1) We can not create images from dr based disks yet.
2) Names of resources created by tests consist of test type, disk size and disk block size (e.g. acceptance-test-disk-eternal-8tib-4kib-1720126075). These names do not contain disk type. This is crucial for eternal tests, because they reuse disks from previous runs and choose the disk to reuse according to its name.

This pr fixes both problems:
1) Add flag that skip all work with images (so we are going to use this flag test cases with dr based disks).
2) Change name format: add disk type to all names of disks and other resources.

Note: when this pr is commited, eternal tests will not remove outdated resources with old name format. So one has to remove resources with old name format manually.

UPD
Turned out that all names of disks/snapshots/images must not contain more than 63 characters. So i had to shorted names, that's why i use 'acc' prefix instead of 'acceptance-test'.

### DIFF
--- a/cloud/disk_manager/test/acceptance/main.go
+++ b/cloud/disk_manager/test/acceptance/main.go
@@ -399,6 +399,7 @@ func mainTest(
 	suffix string,
 	outputDiskIDs string,
 	outputSnapshotIDs string,
+	skipImages bool,
 ) error {
 
 	var snapshotIDs []string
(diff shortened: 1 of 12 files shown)


cloud/disk_manager/test/acceptance/main.go
```diff
@@ -49,7 +49,7 @@ func createSnapshotFromDisk(
 	suffix string,
 ) (string, error) {
 
-	name := createUniqueName("acceptance-test-snapshot", suffix)
+	name := createUniqueName("acc-snapshot", suffix)
```
gy2411, line 52: Have to shorten name because it mast not be longer than 63 characters

cloud/disk_manager/test/acceptance/test_runner/eternal_acceptance_test_runner.py
```diff
@@ -33,9 +32,8 @@ def __init__(self, ycp: YcpWrapper, args: argparse.Namespace):
         super(EternalTestCleaner, self).__init__(ycp, args)
         test_type = args.test_type
         disk_name_pattern = re.compile(
-            fr'^acceptance-test-{test_type}-'
```
jkuradobery, line 36: You do not need to remove pattern for old disks then. Otherwise you are introducing breaking changes.

jkuradobery (approved):

jkuradobery (approved):

BarkovBG (approved):

PR #1713, opened 2024-08-02
[Snapshot] should return success only for 2xx statuses when creating image from URL

SvartMetal:


cloud/disk_manager/internal/pkg/dataplane/url/common/http_client.go
```diff
@@ -112,6 +112,35 @@ func (l *loggerWithURLReplaced) Error(msg string, keysAndValues ...interface{})
 
 ////////////////////////////////////////////////////////////////////////////////
 
+func checkFor4xxOr5xxStatusCode(statusCode int) error {
```
budevg, line 115: maybe instead of `checkFor4xxOr5xxStatusCode` you can name it `checkHttpStatus` or `checkHttpErrorStatus`

cloud/disk_manager/internal/pkg/dataplane/url/common/http_client.go
```diff
@@ -174,19 +203,16 @@ func (c *httpClient) head(
...
-	if resp.StatusCode >= 400 && resp.StatusCode <= 499 {
-		return nil, NewSourceNotFoundError("http code %d", resp.StatusCode)
+	err = checkFor4xxOr5xxStatusCode(resp.StatusCode)
+	if err != nil {
+		return nil, err
 	}
 
-	if resp.StatusCode >= 500 && resp.StatusCode <= 599 {
-		return nil, errors.NewRetriableErrorf("http code %d", resp.StatusCode)
+	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
```
budevg, line 211: why not do this check inside `checkHttpStatus` and return nil if status is 2xx and error in other cases

budevg (changes requested):

cloud/disk_manager/internal/pkg/dataplane/url/common/http_client.go
SvartMetal: good point

fixed

budevg (approved):

BarkovBG (approved):

budevg (approved):

PR #1790, opened 2024-08-13
issue-1588: fix and enable csi sanity tests for NodeGetVolumeStats

antonmyagkov:


cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -687,17 +687,62 @@ func logVolume(volumeId string, format string, v ...any) {
 	log.Printf("[v=%s]: %s", volumeId, msg)
 }
 
+func (s *nodeService) doesVolumeExist(
```
budevg, line 690: better name could be `hasVolume` or `volumeExists`

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -687,17 +687,62 @@ func logVolume(volumeId string, format string, v ...any) {
...
+	ctx context.Context,
+	volumeId string) bool {
+
+	describeVolumeRequest := &nbsapi.TDescribeVolumeRequest{
+		DiskId: volumeId,
+	}
+
+	_, err := s.nbsClient.DescribeVolume(ctx, describeVolumeRequest)
```
budevg, line 698: better check specific error for disk not exist and return error if there is unrelated error

there is function `IsDiskNotFoundError` here that might help https://github.com/ydb-platform/nbs/blob/users/myagkov/issue-1588-fix-csi-sanity-tests/cloud/disk_manager/internal/pkg/clients/nbs/client.go#L358

budevg (changes requested):

cloud/blockstore/tools/csi_driver/internal/driver/node.go
antonmyagkov: I decided to remove this function(and similar function in controller.go). it returns true/false however it's a three possible states in fact: volume exists, volume doesn't exist and unknown.
It's a bit more code now but handling of the error is not subtle anymore

cloud/blockstore/tools/csi_driver/internal/driver/node.go
antonmyagkov: Done

1. I moved IsDiskNotFoundError implementation to cloud/blockstore/public/sdk/go/client/client.go as I don't want to spread "hacky" implementation of this function in different modules.
2. cloud/disk_manager/internal/pkg/clients/nbs/client.go contains wrapper for IsDiskNotFoundError as I don't want to expose import of cloud/blockstore/public/sdk/go/client/client.go in other disk manager modules(e.g. cloud/disk_manager/internal/pkg/services/pools/rebase_overlay_disk_task.go)

budevg (approved):

drbasic (approved):

Issue #1350, opened 2024-06-05
[Filestore] Multitablet filesystems


qkrorlqr:
Right now one FS == one IndexTablet which is a bottleneck for:
* fetching and updating block layer index info upon read and write requests - severely limits our read/write IOPS
* fetching and updating byte layer data upon unaligned read and write requests
* processing file metadata requests like open/close/stat/create/delete

It's also a limiting factor for max FS size because a single tablet needs to be able to store the whole block layer index - this index can't be too big. In fact handling a 100+TiB FS is a challenge already.

We need to be able to provide linear scalability for FS size and single file-level ops:
* any kind of file IO
* open/close/stat calls

The suggested solution is to make N + 1 tablets for a single FS where N will be determined based on the FS size upon FS creation. First versions may even require manual creation of the additional N tablets. 1 tablet would store the directory inodes and node refs. The refs which point to file inodes would point to other N tablets which would store all files directly under the root, the names of the node refs in their root directories may simply be guids.

Technically, file creation and deletion would cause a multi-tablet transaction, but it won't be hard to implement - we don't need a full 2PC here and we can also keep a cache of pre-created 0-size files to be able to serve creation requests without a multi-tablet transaction. Deletion can be served asynchronously - the client wouldn't be able to find a file which was deleted after we delete its last node ref (if there are no open file handles) - so there is no need for a real synchronous multi-tablet transaction here either. Again, the first version doesn't need to have those optimizations and can simply handle multi-tablet transactions in the following way:
1. in addition to modifying the node refs table we should also add an entry about the requested op (create/delete node) to a log table
2. as long as we have op entries in the log table we should repeatedly retry those ops - as soon as an op completes, it should be deleted from that table.
3. the client (TStorageServiceActor) should return proper error code if it sees a mismatch between a node ref and a node (e.g. a ref to a not-yet-created node should be treated as if the file doesn't exist - ENOENT)

What needs to be done in the first version:
1. support external node references - node refs may point to either node ids local to the tablet or pairs <tabletId, nodeName> for external node references
2. support external inodes in TStorageServiceActor - it should be ready to receive a node ref in the form of <tabletId, nodeName> in response to its requests which work with file inodes and perform a second request to tabletId - handleId -> <tabletId, nodeName (or nodeId)> cache should be implemented in the first version I suppose
3. support tablet relations configuration - there should be an ability to tell a tablet to create file inodes in a bunch of other tablets
4. support the ability to create/delete external inodes

[...]

PR #1801, opened 2024-08-15
issue-1350: restoring follower sessions after follower restarts

qkrorlqr:
There is an object called "session" which corresponds to an active FS endpoint of a single VM - the session object stores that VM's locks and open file descriptors (file handles). A session exists in the leader tablet for each VM that has this FS attached and a session exists for each shard tablet as well. If a tablet restarts, it restores all sessions from persistent storage and launches a 5 minute timer waiting for a heartbeat (TCreateSessionRequest) from the client - if the heartbeat doesn't arrive, the session is deleted. If the leader restarts, each client reestablishes the connection to the leader and sends that heartbeat which causes all sessions - in the leader and in the shards - to be renewed. But if a shard restarts, the session in the shard won't receive a heartbeat until either the leader restarts or the client issues a request (any request) to that shard. If a client doesn't issue any requests to some shard within 5 minutes from the shard's restart, its session is deleted and all file descriptors for that client are closed causing subsequent EBADFs (E_FS_BADHANDLE).

The leader now regularly describes follower sessions. If some sessions are not found in some follower, the leader recreates those sessions in that follower. Session sync period is equal to IdleSessionTimeout / 3. I will make a separate config setting for this timeout in my next PR.

#1350

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_cleanupsessions.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_cleanupsessions.cpp
@@ -1,5 +1,8 @@
 #include "tablet_actor.h"
 
+#include <cloud/filestore/libs/storage/api/tablet_proxy.h>
+#include <cloud/filestore/private/api/protos/tablet.pb.h>
+
 #include <contrib/ydb/library/actors/core/actor_bootstrapped.h>
 
 namespace NCloud::NFileStore::NStorage {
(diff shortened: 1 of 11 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_cleanupsessions.cpp
```diff
@@ -11,13 +14,220 @@ namespace {
...
+        LogTag.c_str(),
+        msg->Info.FollowerId.c_str(),
+        msg->Info.SessionCount,
+        FormatError(msg->GetError()).c_str());
+
+    if (HasError(msg->GetError())) {
+        Error = msg->GetError();
+    }
+
+    Follower2Info[msg->Info.FollowerId] = std::move(msg->Info);
```
debnatkh, line 167: TABLET_VERIFY на индекс?

cloud/filestore/libs/storage/tablet/tablet_private.h
```diff
@@ -25,6 +26,7 @@ namespace NCloud::NFileStore::NStorage {
 #define FILESTORE_TABLET_REQUESTS_PRIVATE_ASYNC(xxx, ...)                      \
     xxx(Compaction,                             __VA_ARGS__)                   \
     xxx(CollectGarbage,                         __VA_ARGS__)                   \
+    xxx(SyncSessions,                        __VA_ARGS__)                      \
```
debnatkh, line 29: поехало

debnatkh (approved):

yegorskii (approved):

PR #1808, opened 2024-08-16
[Blockstore] make 'ReadBlocks transaction was interrupted' errors silent

SvartMetal:
Decided to mute those errors because they may easily happen if Blob Storage group write latency is slightly increased and disk has 'reads immediately follow writes' load pattern.
If those errors constitute a performance problem - it can be observed via 'max time' and other latency alerts

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_readblocks.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_readblocks.cpp
@@ -1059,10 +1059,15 @@ void TPartitionActor::CompleteReadBlocks(
     RemoveTransaction(*args.RequestInfo);
 
     if (args.Interrupted) {
+        ui32 flags = 0;
+        SetProtoFlag(flags, NProto::EF_SILENT);
+        auto error = MakeError(
+            E_REJECTED,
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/partition/part_actor_readblocks.cpp
```diff
@@ -1059,10 +1059,15 @@ void TPartitionActor::CompleteReadBlocks(
     RemoveTransaction(*args.RequestInfo);
 
     if (args.Interrupted) {
+        ui32 flags = 0;
+        SetProtoFlag(flags, NProto::EF_SILENT);
```
SvartMetal, line 1063: Here we also mute errors https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/storage/partition/part_actor_readblocks.cpp#L704 and https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/storage/partition/part_actor_readblocks.cpp#L747.

I believe it is OK, because those errors happen when `TGuardedSgList` is empty which and that means that corresponding request is already canceled

drbasic (approved):

sharpeye (approved):

PR #1869, opened 2024-08-26
issue-1350: fixed ListNodes -> UnlinkNode -> GetNodeAttr race which lead to E_IO due to having empty node attrs in the listing results

qkrorlqr:
In multitablet filestores we first send ListNodesRequest to the leader, then send GetNodeAttr / GetNodeAttrBatch requests to the followers where the files are stored. If an UnlinkNode request happens between these two steps, GetNodeAttr / GetNodeAttrBatch will return ENOENT. Our current code detects this but does nothing and sends empty node attr for such nodes to the layer above it (vfs_fuse) which detects this and responds to the client with E_IO errors. This PR removes such nodes from the ListNodesResponse. NodeNotFoundInFollower is reported only if such nodes are still present in the leader fs (checked via a bunch of GetNodeAttr requests to the leader). E_IO is returned only if the resulting ListNodesResponse is empty and all the nodes are "lost" (not found in the followers and at the same time the leader still has the entries for these nodes and those entries point to the same nodes in the followers).

#1350

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_listnodes.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_listnodes.cpp
@@ -3,6 +3,7 @@
 #include <cloud/filestore/libs/diagnostics/profile_log_events.h>
 #include <cloud/filestore/libs/storage/api/tablet.h>
 #include <cloud/filestore/libs/storage/api/tablet_proxy.h>
+#include <cloud/filestore/libs/storage/model/utils.h>
 #include <cloud/filestore/libs/storage/tablet/model/verify.h>
 
 #include <util/string/join.h>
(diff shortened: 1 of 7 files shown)


cloud/filestore/libs/storage/service/service_actor_listnodes.cpp
```diff
@@ -397,26 +414,159 @@ void TListNodesActor::HandleGetNodeAttrResponse(
...
+
+        // Here we need to remove the Names and Nodes corresponding to the
+        // missing nodes from the response. Plain Erase/EraseIf don't seem to
+        // be applicable since we need to do this operation over 2 repeated
+        // fields - not over a single stl-like container.
+        auto& nodes = *Response.MutableNodes();
+        auto& names = *Response.MutableNames();
+
+        ui32 j = 0;
+        for (ui32 i = MissingNodeIndices[0]; i < Response.NodesSize(); ++i) {
```
debnatkh, line 534: Кажется, надо посортировать MissingNodeIndices

qkrorlqr: fixed

debnatkh (approved):

PR #1941, opened 2024-09-03
check netlink return codes; issue-1832

tpashkin:


tpashkin:
Everything is ok on netlink level. Actual error https://github.com/ydb-platform/nbs/pull/1900/files

PR #1966, opened 2024-09-06
[Filestore] add information about backpressure to the monpage

debnatkh:
Adding information about backpressure state to the monpage for better observability:

---

image

---

image

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_state.h
+++ b/cloud/filestore/libs/storage/tablet/tablet_state.h
@@ -714,8 +714,11 @@ public:
         }
     };
 
+    using TBackpressureValues = TBackpressureThresholds;
+
     bool IsWriteAllowed(
         const TBackpressureThresholds& thresholds,
+        const TBackpressureValues& values,
         TString* message) const;
 
     //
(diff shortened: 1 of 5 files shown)


qkrorlqr (approved):

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -714,8 +714,11 @@ FILESTORE_DUPCACHE_REQUESTS(FILESTORE_DECLARE_DUPCACHE)
         }
     };
 
+    using TBackpressureValues = TBackpressureThresholds;
```
ya-ksgamora, line 717: I don't think that this type alias is adding to readability

Comparing variables `values` and `threshold` of the same type is more natural to me
Also it takes more time to find where this other type `TBackpressureValues` comes from

ya-ksgamora (changes requested):

cloud/filestore/libs/storage/tablet/tablet_state.h
debnatkh: The semantics of TBackpressureValues and TBackpressureThresholds are different: one is supposed to represent the current state, the other stands for limits. Using different types highlights semantics and allows for those types to actually differ. For example, we may need to pass something like a flag `DisableBackpressure`, which will be present in TBackpressureValues, but not TBackpressureThresholds

ya-ksgamora (approved):

PR #1992, opened 2024-09-10
moved parameters for registration from server to storage config + add…

yegorskii:
…ed registration tests for nbs (#1859)

* moved parameters for registration from server to storage config + added registration tests for nbs

* update

* update

* update

* fix ut

* update

* update

* update

* fix ut

* remove extra new lines


yegorskii:
added missed commit

PR #2182, opened 2024-09-30
fix: revert wait_for_block_device_to_appear implementation

antonmyagkov:
revert to previous wait_for_block_device_to_appear implementation https://github.com/ydb-platform/nbs/commit/98ac178cc7fc1cf886f38dae7ae5367780629686#diff-536ed3f8ea3b7c4f96c4e0fa2c894a1ee506c2da759d2ee1c77a334199779477

### DIFF
--- a/cloud/disk_manager/test/acceptance/test_runner/lib/helpers.py
+++ b/cloud/disk_manager/test/acceptance/test_runner/lib/helpers.py
@@ -37,10 +37,9 @@ def wait_for_block_device_to_appear(
     disk_id: str,
     module_factory,
     ssh_key_path: str | None = None,
-):
-    device_name = f'/dev/disk/by-id/virtio-{disk_id}'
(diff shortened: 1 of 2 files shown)


jkuradobery:
I think that we should instead of this revert, we should do `readlink -f` for the `/dev/disk/by-id/virtio-b*`.  This comment should be resolved. I do not think that the change from using lsblk/udevadm to symlinks in /dev/disk would fix the issue with disks being not found for 120 seconds though.

jkuradobery (approved):

antonmyagkov:
> I think that we should instead of this revert, we should do `readlink -f` for the `/dev/disk/by-id/virtio-b*`. This comment should be resolved. I do not think that the change from using lsblk/udevadm to symlinks in /dev/disk would fix the issue with disks being not found for 120 seconds though.

Let's fix tests first and return to proper "wait block device" implementation later.

Issue #2033, opened 2024-09-14
[Filestore] Detect RequestId collisions


qkrorlqr:
Sometimes guest OS might send us a duplicate fuse_req->unique (aka RequestId in our protocol) field for a new request. We end up looking this RequestId up in the tablet DupCache and if we still haven't evicted the entry for the old request we end up returning the old response (which is not the response to the actual new request). It seems to be a bug in the guest but it's much easier to make a workaround on our side than to fix this behaviour in Linux (and possibly other OS kernels). The workaround is to compare the response that we found in DupCache to the actual request - if they don't match, evict this entry from the cache and treat this request as just a new request.

To make proper comparison, we need to add more data to the responses (which are stored in DupCache). It seems that adding ParentNodeId + Name will be enough to make the probability of a collision indistinguishable from 0. For CreateHandle we can add another check to discard the entries that were persistently stored in DupCaches before this fix - we can simply compare the NodeId from the request to the NodeId from the response for requests without child name (open() by node_id, not by parent_node_id + name).

PR #2266, opened 2024-10-12
issue-2033: sending ResetSession requests to shards upon ResetSession to leader (in the same way it's done for Create/DestroySession); extracted common code for this thing from Create/DestroySession

qkrorlqr:
We need to do this in order to reset DupCache in shards upon ResetSession. We don't really need SessionState in shards (but it won't hurt to have it there as well).

#2033


debnatkh (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
```diff
@@ -1,6 +1,6 @@
 #include "tablet_actor.h"
-
-#include <cloud/filestore/libs/storage/api/tablet_proxy.h>
+#include "cloud/filestore/private/api/protos/tablet.pb.h"
```
debnatkh, line 2:
```suggestion
#include <cloud/filestore/private/api/protos/tablet.pb.h>
```

yegorskii (approved):

debnatkh (approved):

PR #2268, opened 2024-10-12
configure rdma target thread pool size; issue-2267

tpashkin:
Also move DiskAgent.RdmaTarget to RdmaConfig.DiskAgentTarget for convenience

### DIFF
--- a/cloud/blockstore/config/rdma.proto
+++ b/cloud/blockstore/config/rdma.proto
@@ -20,12 +20,12 @@ message TRdmaServer
     uint32 Backlog = 1;
     uint32 QueueSize = 2;
     uint32 MaxBufferSize = 3;
-    uint64 KeepAliveTimeout = 4;           // in milliseconds
+    uint64 KeepAliveTimeout = 4;            // in milliseconds
     EWaitMode WaitMode = 5;
     uint32 PollerThreads = 6;
-    uint32 MaxInflightBytes = 7;           // per client
-    uint64 AdaptiveWaitSleepDelay = 8;     // in microseconds
-    uint64 AdaptiveWaitSleepDuration = 9;  // in microseconds
+    uint32 MaxInflightBytes = 7;            // per client
+    uint64 AdaptiveWaitSleepDelay = 8;      // in microseconds
+    uint64 AdaptiveWaitSleepDuration = 9;   // in microseconds
     bool AlignedDataEnabled = 10;
 }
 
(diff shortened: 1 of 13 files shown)


cloud/blockstore/config/rdma.proto
```diff
@@ -42,8 +42,11 @@ message TRdmaClient
 
 message TRdmaTarget
 {
-    TRdmaEndpoint Endpoint = 1;
-    TRdmaServer Server = 2;
+    TRdmaEndpoint Endpoint = 1;            // deprecated
+    TRdmaServer Server = 2;                // deprecated
+    string Host = 3;
+    uint32 Port = 4;
+    uint32 Threads = 5;
```
budevg, line 49: Let's have all rdma related config in rdma.txt file which has the `TRdmaConfig` message serialized to it.
This `TRdmaTarget` message is actually deprecated and we should not extend it. I left it since the old config was using it [here](https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/config/disk.proto#L204). But the idea was to move all required params to the TRdmaConfig struct.

So in you PR let's just add the `RdmaServerNumHandlerThreads` (I think it's better then just Threads since we might have some other threads we want to configure) to the TRdmaConfig.
And we need to pass this config to the rdmaTargetConfig [here](https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_init.cpp#L41) by getting it from RdmaConfig that already present in the disk agent actor.

Also let's not refactor the host/port in this PR. I noticed that the port is still coming from disk-agent.txt config
[here](https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/storage/disk_agent/disk_agent_state.cpp#L427)
So either we leave it in the `NProto::TDiskAgentConfig` or move it to the rdma.txt as well. I will be easier to understand if it will be a separate PR

budevg (changes requested):

cloud/blockstore/config/rdma.proto
tpashkin: Moved `TRdmaTarget` to `TRdmaConfig`. Now everything except blockstore-server's rdma endpoint (which isn't currently used) is located in nbs-rdma.txt

Deployment plan:
1. Merge these changes. If `RdmaTarget` is present in `DiskAgentConfig`, it will also be enabled in `RdmaConfig` to smooth out the transition
2. Enable `DiskAgent` in `RdmaConfig`
3. Remove `RdmaTarget` from `DiskAgentConfig`

cloud/blockstore/libs/storage/disk_agent/rdma_target.h
```diff
@@ -40,9 +63,11 @@ struct IRdmaTarget: IStartable
 using TStorageAdapterPtr = std::shared_ptr<TStorageAdapter>;
 using IRdmaTargetPtr = std::shared_ptr<IRdmaTarget>;
 
+
+
```
budevg, line 67: remove two extra lines

budevg (changes requested):

budevg (approved):

drbasic (approved):

PR #2276, opened 2024-10-14
deleting explicit ActivityType setting - 1. it's broken 2. the default value is better anyway

qkrorlqr:
Otherwise we get activity=='-' for these actors in monitoring whereas the default value for activity is the name of the actor (which is good)


debnatkh (approved):

SvartMetal (approved):

yegorskii (approved):

qkrorlqr:
> cloud/blockstore/tests/csi_driver/test.py.test_nbs_csi_driver_mounted_disk_protected_from_deletion

флап

Issue #1199, opened 2024-05-14
[Disk Manager] Tracing with Open Telemetry


gy2411:
Traced should be collected for all requests to Disk Manager should be traced. Given request_id, we need opportunity to see what was this request doing (which tasks scheduled, which queries to nbs/nfs/ydb/s3 made, when exactly it happened and how much time did different steps take).

We are going to use Open Telemetry framework https://opentelemetry.io/. Namely, we should add [instrumentation](https://opentelemetry.io/docs/languages/go/instrumentation/#propagators-and-context) to Disk Manager code for creating spans and events. Also Disk Manager must be able to use trace context passed by caller (like trace id and parent span id) if it is present. Also Disk Manager must be able to export traces to a backend (in order to store and visualize traces).

gy2411:
Further improvements of tracing in Disk Manager:
- Do not sample spans for big task generation
- Handle retriable errors in spans (decide should we set Erorr status for such spans or not)
- Shorten error message of non-retryable errors in spans (do not write stacktrace, do not write retriable errors from error chain).
- Add request id attribute to spans of tasks. Also think should we add any other attributes, events for error handling spans of tasks.
- Add events for retriable errors in blockstore go SDK.

PR #2474, opened 2024-11-12
[Disk Manager] Add parent span for regular tasks, add parent span rotation for long-running regular tasks

gy2411:
#1199

Now spans for regular are inconvenient. Tasks have no common parent. For example, if a regular tasks had several generations, each generation will create spans in separate trace.

In this pr we create common parent span for spans created by a regular task. Also, we rotate the parent span if the task runs for long. This prevents from obtaining too large trace and from losing all spans due to trace expiration.

For example, assume that some task ran for 7 days, trace ttl is 5 days and parent rotation period is 3 days. And assume that we want to look at the traced of this task just after it's completion. Then the task will have three traces, the oldest one will be already expired, but we will be able to see the remaining two traces. Without rotation, we wouldn't be able to see anything.

Note: parent span rotation will not happen if a task is runned for long with one and the same generation.

### DIFF
--- a/cloud/tasks/config/config.proto
+++ b/cloud/tasks/config/config.proto
@@ -56,4 +56,7 @@ message TasksConfig {
     // Needed for tracing.
     // Spans of tasks with greater generation id will not be sampled.
     optional uint64 MaxSampledTaskGeneration = 32 [default = 100];
+    // Needed for tracing.
+    // After this timeout regular task's snaps will be created in another trace.
+    optional string RegularTaskTraceExpirationTimeout = 33 [default = "72h"];
 }
--- a/cloud/tasks/storage/storage_ydb_impl.go
+++ b/cloud/tasks/storage/storage_ydb_impl.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ydb-platform/nbs/cloud/tasks/common"
 	"github.com/ydb-platform/nbs/cloud/tasks/errors"
+	"github.com/ydb-platform/nbs/cloud/tasks/headers"
 	"github.com/ydb-platform/nbs/cloud/tasks/logging"
 	"github.com/ydb-platform/nbs/cloud/tasks/persistence"
 	grpc_codes "google.golang.org/grpc/codes"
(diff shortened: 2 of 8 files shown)


cloud/tasks/storage/storage_ydb_impl.go
```diff
@@ -682,6 +683,104 @@ func (s *storageYDB) addRegularTasks(
...
+
+		metadataUpdates = append(
+			metadataUpdates,
+			persistence.StructValue(
+				persistence.StructFieldValue("id", persistence.UTF8Value(taskID)),
+				persistence.StructFieldValue(
+					"metadata",
+					persistence.BytesValue(common.MarshalStringMap(
+						newMetadata.Vals(),
+					)),
```
gy2411, line 738: Хм, а неверное здесь надо также поле UpdatedAt обновить?

cloud/tasks/config/config.proto
```diff
@@ -56,4 +56,7 @@ message TasksConfig {
     // Needed for tracing.
     // Spans of tasks with greater generation id will not be sampled.
     optional uint64 MaxSampledTaskGeneration = 32 [default = 100];
+    // Needed for tracing.
+    // Rotation period of parent span for regular tasks.
+    optional string RegularTasksTracesRotationPeriod = 33 [default = "72h"];
```
SvartMetal, line 61: RegularTaskTracesRotationPeriod

cloud/tasks/storage/storage_ydb_impl.go
```diff
@@ -682,6 +683,104 @@ func (s *storageYDB) addRegularTasks(
 	return s.updateTaskStates(ctx, tx, transitions)
 }
 
+func (s *storageYDB) updateTracingContextInRegularTasks(
```
SvartMetal, line 686: предлагаю пока сделать простую правку, а ротацию не делать

gy2411:
Решили не делать ротацию трейсов (нет существенной необходимости).
pr, в котором только добавляется радительский спан: https://github.com/ydb-platform/nbs/pull/2641

Issue #2077, opened 2024-09-19
[Blockstore] 2/3 reading for mirror-3 disk


sharpeye:
In addition to the regular reading operation, mirror-3 disk could read a checksum from another replica, compare checksums and, if the comparison fails, perform a resync

qkrorlqr:
> if the comparition fails, perform a resync.

this part will made the code more complex
I decided to simply return E_REJECTED
if the data is the same in 2 out of 3 replicas, one of the next retry attempts will hit those 2 replicas and succeed
there's no point optimizing this scenario since it should never happen (or almost never)
the discrepancy will be eventually fixed by the scrubbing process

PR #2545, opened 2024-11-20
issue-2077: read N/M for mirrored disks (1 ReadBlocks + N-1 ChecksumBlocks)

qkrorlqr:
If MirrorReadReplicaCount > 1, we read data from one of the replicas and we request checksums from the remaining replicas. If some of the checksums differ, we return E_REJECTED. For m3 disks and MirrorReadReplicaCount=2 one of the next retries will return S_OK if at least 2 replicas contain the same data. If data is different in all replicas, nothing can be done automatically and read request will hang (will be retried until someone manually fixes the discrepancy). 

If such a discrepancy upon read is detected, MirroredDiskChecksumMismatchUponRead critical event is raised if there were no concurrent writes whose ranges overlap the range of the read request.

In case of a 2/3 read and a discrepancy in checksums we could write some code to read the 3rd replica and, if the data in it matches the data in one of the other 2 replicas, we could return S_OK, but the code would be more complex. So for simplicity we just return E_REJECTED and one of the next retries will hit the 2 replicas which contain the same versions of the data so eventually we will return S_OK. There's no point optimizing this scenario since we hope that this scenario isn't just rare - we hope it never happens.

There is also a small chance of a false positive MirroredDiskChecksumMismatchUponRead critical event in the following scenario:
1. WriteBlocks request is processed, it successfully executes in some replicas, fails in other replicas, E_REJECTED is returned, backoff in DurableClient begins
2. ReadBlocks request comes, sees checksum discrepancy (due to the incomplete write), sees no dirty flag upon completion, reports MirroredDiskChecksumMismatchUponRead
3. WriteBlocks gets retried and restores consistency among the replicas

But it's pretty hard to fix this case at the storage libs level and we think that the probability of such a case is low enough because normally guests don't do concurrent overlapping reads+writes. Let's try to use the current logic and, if such false positives become a problem, let's just do overlapping read/write protection at the client stack level (we already have conceptually similar logic there - e.g. read-modify-write for unaligned writes).  

#2077

### DIFF
--- a/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor.h
+++ b/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor.h
@@ -58,6 +58,7 @@ private:
     ui64 NetworkBytes = 0;
     TDuration CpuUsage;
 
+    THashMap<ui64, bool> ReadRequestIdentityKey2DirtyFlag;
     TRequestsInProgress<ui64, TBlockRange64> RequestsInProgress{
         EAllowedRequests::ReadWrite};
     TDrainActorCompanion DrainActorCompanion{
@@ -116,6 +117,10 @@ private:
 
     void HandleWriteOrZeroCompleted(
         const TEvNonreplPartitionPrivate::TEvWriteOrZeroCompleted::TPtr& ev,
+        const NActors::TActorContext& ctx);
+
+    void HandleMirroredReadCompleted(
+        const TEvNonreplPartitionPrivate::TEvMirroredReadCompleted::TPtr& ev,
         const NActors::TActorContext& ctx);
 
     void HandleRWClientIdChanged(
--- a/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor.h
+++ b/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor.h
@@ -25,6 +25,7 @@
 #include <contrib/ydb/library/actors/core/mon.h>
 
 #include <util/generic/deque.h>
+#include <util/generic/hash_set.h>
 
 namespace NCloud::NBlockStore::NStorage {
 
@@ -58,7 +59,7 @@ private:
     ui64 NetworkBytes = 0;
     TDuration CpuUsage;
 
-    THashMap<ui64, bool> ReadRequestIdentityKey2DirtyFlag;
+    THashSet<ui64> DirtyReadRequestIds;
     TRequestsInProgress<ui64, TBlockRange64> RequestsInProgress{
         EAllowedRequests::ReadWrite};
     TDrainActorCompanion DrainActorCompanion{
--- a/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
+++ b/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
@@ -2,6 +2,7 @@
 
 #include "mirror_request_actor.h"
 
+#include <cloud/blockstore/libs/diagnostics/critical_events.h>
 #include <cloud/blockstore/libs/service/request_helpers.h>
 #include <cloud/blockstore/libs/storage/api/undelivered.h>
 #include <cloud/blockstore/libs/storage/core/probes.h>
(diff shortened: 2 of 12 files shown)


WilyTiger (dismissed):

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
```diff
@@ -20,8 +21,14 @@ void TMirrorPartitionActor::HandleWriteOrZeroCompleted(
     const TActorContext& ctx)
 {
     const auto requestIdentityKey = ev->Get()->RequestCounter;
-    RequestsInProgress.RemoveRequest(requestIdentityKey);
+    TBlockRange64 range;
+    RequestsInProgress.ExtractRequest(requestIdentityKey, &range);
     DrainActorCompanion.ProcessDrainRequests(ctx);
+    for (const auto& request: RequestsInProgress.AllRequests()) {
```
drbasic, line 27: сonst auto& [requestId, requestRange]

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
```diff
@@ -72,6 +98,11 @@ void TMirrorPartitionActor::MirrorRequest(
         }
         WriteIntersectsWithScrubbing = true;
     }
+    for (const auto& request: RequestsInProgress.AllRequests()) {
```
drbasic, line 101: сonst auto& [requestId, requestRange]

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_readblocks.cpp
```diff
@@ -224,23 +427,52 @@ void TMirrorPartitionActor::ReadBlocks(
...
 
-    TActorId replicaActorId;
-    const auto error = State.NextReadReplica(blockRange, &replicaActorId);
-    if (HasError(error)) {
-        Reply(
-            ctx,
-            *requestInfo,
-            std::make_unique<typename TMethod::TResponse>(error));
+    TVector<TActorId> replicaActorIds;
+    const ui32 readReplicaCount = Min<ui32>(
```
drbasic, line 431: используй `TSet<TActorId> replicaActorIds; `
кода получится меньше

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor.h
```diff
@@ -58,6 +58,7 @@ class TMirrorPartitionActor final
     ui64 NetworkBytes = 0;
     TDuration CpuUsage;
 
+    THashMap<ui64, bool> ReadRequestIdentityKey2DirtyFlag;
```
drbasic, line 61: лучше использовать `THashSet<ui64> DirtyReadRequests`; и помещать в него значение только если произошло расхождение

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_ut.cpp
```diff
@@ -1680,6 +1686,121 @@ Y_UNIT_TEST_SUITE(TMirrorPartitionTest)
         UNIT_ASSERT_VALUES_EQUAL(1, rangeResynced);
     }
 
+    Y_UNIT_TEST(ShouldRejectReadAndForceRangeScrubbingUponChecksumMismatchIfRead2IsEnabled)
```
drbasic, line 1689: а каким образом ForceRangeScrubbing происходит? или я не понял названия теста.

qkrorlqr: а, сорян, это я сначала так думал делать, но потом передумал, а тест не переименовал

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_readblocks.cpp
drbasic: или не делать это вручную 
https://github.com/ydb-platform/nbs/blob/main/library/cpp/containers/sorted_vector/sorted_vector.h#L410

WilyTiger (approved):

drbasic (approved):

PR #2583, opened 2024-11-27
Backport ydb-platform/ydb#11803 with compatibility patches

Enjection:
- add explicit dependency on ss ops
- backport ydb#11803


Enjection:
Agreed with @drbasic that it would arrive with another sync of ydb. We don't need to backport changes.

Issue #2171, opened 2024-09-30
[Filestore] False positive AppCriticalEvents/NodeNotFoundInFollower upon concurrent CreateHandle/CreateNode + ListNodes


qkrorlqr:
Race:
1. CreateHandle with CREATE flag / CreateNode
2. NodeRef added in Leader
3. ListNodes
4. GetNodeAttr to follower -> ENOENT => AppCriticalEvents/NodeNotFoundInFollower
5. Node created in follower

Nothing bad happens apart from the false positive monitoring event

PR #2611, opened 2024-12-03
[Filestore] issue-2171: Avoid listing nodes with pending create on shards

budevg:
References #2171


debnatkh (approved):

budevg:
failed tests doesn't look related to this patch.

Issue #2690, opened 2024-12-12
[NBS] RECVs completed with error may lead to WR leak


tpashkin:
For example:
```
2024-11-23T20:11:04.771731Z :BLOCKSTORE_RDMA INFO: [1245282948318403695] start endpoint XXX [send_magic=1148223E recv_magic=635E9C6F]
...
2024-11-25T22:44:39.878861Z :BLOCKSTORE_RDMA ERROR: [1245282948318403695] RECV 635E9C6F.0.FB: IBV_WC_RETRY_EXC_ERR
2024-11-25T22:44:39.878920Z :BLOCKSTORE_RDMA ERROR: [1245282948318403695] ibv_post_recv failed with error 12: 0<EF>(<8B>#032^?
2024-11-25T22:44:39.878924Z :BLOCKSTORE_RDMA INFO: [1245282948318403695] disconnect
...
2024-11-25T22:44:49.884336Z :BLOCKSTORE_RDMA ERROR: [1245282948318403695] flush timeout [send_queue.size=255 recv_queue.size=256]
```
RECV 635E9C6F.0.FB completed with IBV_WC_RETRY_EXC_ERR and called ibv_post_recv which in turn thrown an exception, because QP was in error state. This lead to SEND, which was already [collected by ibv_poll_cq](https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/impl/verbs.cpp#L141) to be leaked. This can be seen in the last "flush timeout" message – 1 send was missing and never got back into SendQueue

PR #2706, opened 2024-12-16
fix client wr leak #2690

tpashkin:


drbasic (approved):

cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -952,7 +951,21 @@ void TClientEndpoint::SendRequest(TRequestPtr req, TSendWr* send)
     requestMsg->Out = req->OutBuffer;
 
     RDMA_TRACE("SEND " << TWorkRequestId(send->wr.wr_id));
-    Verbs->PostSend(Connection->qp, &send->wr);
+
+    try {
+        Verbs->PostSend(Connection->qp, &send->wr);
```
budevg, line 956: lest's call `SendRequestCompleted(send, IBV_WC_FATAL_ERR)` if PostSend fails. This way we already invoke our handler for failed SendRequest and implicitly invoke `Disconnect`

Same thing with `PostRecv` where we can call `RecvResponseCompleted` with error status.

cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -952,7 +951,21 @@ void TClientEndpoint::SendRequest(TRequestPtr req, TSendWr* send)
...
+    } catch (const TServiceError& e) {
+        RDMA_ERROR(
+            "SEND " << TWorkRequestId(send->wr.wr_id) << ": " << e.what());
+        SendQueue.Push(send);
+        ReportRdmaError();
+        Disconnect();
```
budevg, line 962: What about the server.cpp it also has post_send and post_recv which can leak resources in the same way. Let's fix server.cpp as well 

for PostSend we can call ReadRequestDataCompleted/WriteResponseDataCompleted/SendResponseCompleted
for PostRecv the RecvRequestCompleted tries to PostRecv again so it's more tricky probably need to flush session

budevg (changes requested):

cloud/blockstore/libs/rdma/impl/client.cpp
tpashkin: I'd rather not. The way counters are calculated right now, it will lead to double decrement

cloud/blockstore/libs/rdma/impl/client.cpp
tpashkin: This PR is specifically about the client. Server fix will require more work and will probably come quite a bit later

budevg (approved):

PR #2746, opened 2024-12-20
[Blockstore] preventing the rejection of a significant fraction of Disk Agents in the cluster

sharpeye:
#2752
If the percentage of agents that did not reconnect after Disk Registry' reboot is higher than DiskRegistryInitialAgentRejectionThreshold, then the rejection of such agents does not occur - we assume a connectivity failure in the cluster.


qkrorlqr (changes requested):
#issue

komarevtsev-d (dismissed):

tpashkin (dismissed):

qkrorlqr (approved):

Issue #176, opened 2024-01-18
[NBS] NRD/Mirrored disk default encryption


qkrorlqr:
We can implement default encryption for STORAGE_MEDIA_SSD_{NONREPLICATED,MIRROR2,MIRROR3} both at rest and in transit by encrypting data at the endpoint level - just like user key-based encryption. The default key can be generated upon CreateVolume or upon AlterVolume and stored in NProto::TVolumeConfig and then propagated to the endpoint upon MountVolume. The only remaining problem is the performance hit which is currently mostly caused by the need to update UsedBlockMap in the volume database. But actually we don't need this bitmap. A good encryption algorithm should produce data that's indistinguishable from the data generated by a uniform random generator. Therefore we don't need this bitmap - we can simply check that the block that we have just read from the storage node contains only zeroes - if so, we should not decrypt this block, otherwise - decrypt. We don't even have to check the whole block - checking the first 128 bytes should be more than enough. And just in case (which should be literally impossible - P=1/2^128) an encrypted block contains zeroes in the first 128 bytes, we can check it upon write and report a E_IO or a E_ARGUMENT error to the client and raise a CriticalEvent. This endpoint-level encryption is also great because it won't impact the vhost-server-based fast rdma datapath - we will just disable it at the fastpath endpoint level.

qkrorlqr:
The same logic will work just fine for all media kinds actually (STORAGE_MEDIA_SSD, STORAGE_MEDIA_HYBRID, etc.). But it will be in conflict with https://github.com/ydb-platform/nbs/issues/9 (encrypted data cannot be compressed).

danmx:
@qkrorlqr what about compressing before encryption?

qkrorlqr:
right now encryption is implemented several layers closer to the client than the place where we can efficiently compress data to store it in a compressed form 

we can compress before encryption but the implementation will be more clumsy

it's of course doable - just harder

drbasic:
Нет смысла ограничиваться и проверять на нули только часть блока. Проще проверять целиком. Алгоритм все равно остановится на первом не нулевом байте и время на проверку будет одинаковое.

PR #2789, opened 2025-01-02
issue-176: disable used block tracking

sharpeye:
#176
Tracking of used blocks should be disabled in case of volumes with encryption at rest: assume that unencrypted blocks are completely filled with zeros.

### DIFF
--- a/cloud/blockstore/libs/storage/volume/volume_state.cpp
+++ b/cloud/blockstore/libs/storage/volume/volume_state.cpp
@@ -221,6 +221,26 @@ void TVolumeState::ResetThrottlingPolicy(
             ThrottlerConfig.UseDiskSpaceScore));
 }
 
+bool TVolumeState::ShouldTrackUsedBlocks() const
+{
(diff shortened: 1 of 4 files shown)


qkrorlqr (commented):
что за NBSNEBIUS-265?

drbasic (approved):

cloud/blockstore/libs/storage/volume/volume_state.cpp
```diff
@@ -221,14 +221,34 @@ void TVolumeState::ResetThrottlingPolicy(
...
+    }
+
+    const bool overlay = !GetBaseDiskId().empty();
+    if (overlay) {
+        return true;
+    }
+
+    // TODO(drbasic)
+    // For encrypted disk-registry based disks, we will continue to
+    // write a map of encrypted blocks for a while.
```
tpashkin, line 237: "for a while" – это сколько и чем определено?

tpashkin (approved):

Issue #2813, opened 2025-01-08
[Filestore] local filestore can't create file with non root permissions


budevg:
```
sudo touch -u user /mnt/virtiofs/1.txt
```
creates 1.txt with root permissions

PR #2814, opened 2025-01-08
[Filestore] issue-2813: use uid/gid from fuse context when creating node

budevg:
#2813

### DIFF
--- a/cloud/filestore/libs/service_local/lowlevel.cpp
+++ b/cloud/filestore/libs/service_local/lowlevel.cpp
@@ -594,12 +594,15 @@ UnixCredentialsGuard::UnixCredentialsGuard(uid_t uid, gid_t gid)
         return;
     }
 
-    int ret = setresgid(-1, gid, -1);
+    // use syscall directly to change uid/gid per thread instead of glibc
+    // version of setresgid/setresuid since they will change uid/gid for all
+    // threads
+    int ret = syscall(SYS_setresgid, -1, gid, -1);
     if (ret == -1) {
         return;
     }
 
-    ret = setresuid(-1, uid, -1);
+    ret = syscall(SYS_setresuid, -1, uid, -1);
     if (ret == -1) {
         setresgid(-1, Gid, -1);
         return;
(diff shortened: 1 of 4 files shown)


cloud/filestore/libs/service_local/lowlevel.cpp
```diff
@@ -578,4 +578,48 @@ bool Flock(const TFileHandle& handle, int operation)
...
+    // version of setresgid/setresuid since they will change uid/gid for all
+    // threads
+    int ret = syscall(SYS_setresgid, -1, gid, -1);
+    if (ret == -1) {
+        return;
+    }
+
+    ret = syscall(SYS_setresuid, -1, uid, -1);
+    if (ret == -1) {
+        setresgid(-1, Gid, -1);
```
drbasic, line 607: syscall(SYS_setresgid, -1, Gid, -1); ?

cloud/filestore/libs/service_local/lowlevel.h
```diff
@@ -13,6 +13,19 @@ namespace NLowLevel {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+class UnixCredentialsGuard {
+private:
+    uid_t Uid = -1;
```
drbasic, line 18: OriginalUid
OriginalGid

drbasic (dismissed):

cloud/filestore/libs/service_local/lowlevel.cpp
```diff
@@ -578,4 +578,48 @@ bool Flock(const TFileHandle& handle, int operation)
...
+        return;
+    }
+
+    ret = syscall(SYS_setresuid, -1, uid, -1);
+    if (ret == -1) {
+        syscall(SYS_setresuid, -1, OriginalGid, -1);
+        return;
+    }
+
+    IsRestoreNeeded = true;
```
antonmyagkov, line 611: Probably you need two flags: IsRestoreGidNeeded IsRestoreUidNeeded because setresuid can return -1 and setresgid return 0.

budevg: Actually we assume that if one of setresuid/setregid fails then the whole op fails and we don't try to change partially just for gid

antonmyagkov (approved):

drbasic (approved):

PR #2886, opened 2025-01-20
Fix flapping test filestore/tests/profile_log/qemu-local-test

proller:
With reducing ProfileLogTimeThreshold to 100ms
Also flush profile log on destruct

### DIFF
--- a/cloud/filestore/libs/daemon/server/config_initializer.cpp
+++ b/cloud/filestore/libs/daemon/server/config_initializer.cpp
@@ -33,6 +33,10 @@ void TConfigInitializerServer::InitAppConfig()
     }
 
     ServerConfig = std::make_shared<NServer::TServerConfig>(serverConfig);
+
+    if (Options->DiagnosticsConfig) {
+        InitDiagnosticsConfig();
+    }
 }
 
 void TConfigInitializerServer::ApplyServerAppConfig(const TString& text)
--- a/cloud/filestore/libs/daemon/vhost/config_initializer.cpp
+++ b/cloud/filestore/libs/daemon/vhost/config_initializer.cpp
@@ -38,6 +38,10 @@ void TConfigInitializerVhost::InitAppConfig()
 
     VhostServiceConfig = std::make_shared<TVhostServiceConfig>(
         AppConfig.GetVhostServiceConfig());
+
+    if (Options->DiagnosticsConfig) {
+        InitDiagnosticsConfig();
+    }
 }
 
 void TConfigInitializerVhost::ApplyVHostAppConfig(const TString& text)
--- a/cloud/filestore/libs/diagnostics/profile_log.cpp
+++ b/cloud/filestore/libs/diagnostics/profile_log.cpp
@@ -41,6 +41,11 @@ public:
     {
     }
 
+    ~TProfileLog() override
+    {
+        Flush();
+    }
+
 public:
     void Start() override;
     void Stop() override;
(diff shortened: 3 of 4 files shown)


cloud/filestore/libs/daemon/server/config_initializer.cpp
```diff
@@ -33,6 +33,10 @@ void TConfigInitializerServer::InitAppConfig()
     }
 
     ServerConfig = std::make_shared<NServer::TServerConfig>(serverConfig);
+
+    if (Options->DiagnosticsConfig) {
```
budevg, line 37: why do you need this ? 
DiagnostiConfigs already initialized in the base class by calling `TBootstrapCommon::InitCommonConfigs`

cloud/filestore/libs/daemon/vhost/config_initializer.cpp
```diff
@@ -38,6 +38,10 @@ void TConfigInitializerVhost::InitAppConfig()
 
     VhostServiceConfig = std::make_shared<TVhostServiceConfig>(
         AppConfig.GetVhostServiceConfig());
+
+    if (Options->DiagnosticsConfig) {
```
budevg, line 42: why do you need this ? 
DiagnostiConfigs already initialized in the base class by calling `TBootstrapCommon::InitCommonConfigs`

cloud/filestore/libs/diagnostics/profile_log.cpp
```diff
@@ -41,6 +41,11 @@ class TProfileLog final
     {
     }
 
+    ~TProfileLog() override
```
budevg, line 44: I think best practice is not to mark destructor with override
discussion [here](https://github.com/isocpp/CppCoreGuidelines/issues/721) and [here](https://github.com/rianquinn/CppCoreGuidelines/blob/8be1869663b4518bed0af0e5d9dfa053bb003bbc/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final)

cloud/filestore/libs/diagnostics/profile_log.cpp
```diff
@@ -41,6 +41,11 @@ class TProfileLog final
...
+    {
+        Flush();
```
budevg, line 46: maybe better do the flush here
```c++
void TProfileLog::ScheduleFlush()
{
    if (AtomicGet(ShouldStop)) {
```

cloud/filestore/tests/python/lib/daemon_config.py
```diff
@@ -453,9 +479,20 @@ def __init__(
 
         self.__local_service_port = self._port_manager.get_port()
 
+        self.__diag_config=self.__generate_diag_txt()
+        with open(self.diag_config_file_path, "w") as config_file:
+            if self.__diag_config:
+                config_file.write(MessageToString(self.__diag_config))
+                os.fsync(config_file)
```
budevg, line 486: use `config_file.flush()` instead

cloud/filestore/tests/python/lib/daemon_config.py
```diff
@@ -412,6 +425,17 @@ def __init__(
...
+
+    def __generate_diag_txt(self):
+        diag = TDiagnosticsConfig()
+        diag.ProfileLogTimeThreshold = 100
```
budevg, line 436: if you set ProfileLogTimeThreshold to 100 msec, why do you need the Flush changes in the c++ code. Just make sure the test waits extra cycle or two before collecting the logs.
I suggest to discard all c++ changes and just make sure that [sync](https://linux.die.net/man/8/sync) is called at the end of the test and that you wait extra time (ProfileLogTimeThreshold * 2)

budevg (changes requested):

budevg (approved):

drbasic (approved):

PR #2898, opened 2025-01-21
Flush profile logs on destruct

proller:
To avoid losing up to 15 seconds of logs on server exit or soft crash

### DIFF
--- a/cloud/blockstore/libs/diagnostics/profile_log.cpp
+++ b/cloud/blockstore/libs/diagnostics/profile_log.cpp
@@ -130,7 +130,7 @@ private:
 
 TProfileLog::~TProfileLog()
 {
-    Flush();
+    DoFlush();
 }
 
 void TProfileLog::Start()
--- a/cloud/blockstore/libs/diagnostics/profile_log.cpp
+++ b/cloud/blockstore/libs/diagnostics/profile_log.cpp
@@ -114,7 +114,7 @@ public:
     {
     }
 
-    ~TProfileLog();
+    ~TProfileLog() override;
 
 public:
     void Start() override;
--- a/cloud/blockstore/libs/diagnostics/profile_log.cpp
+++ b/cloud/blockstore/libs/diagnostics/profile_log.cpp
@@ -114,6 +114,8 @@ public:
     {
     }
 
+    ~TProfileLog();
+
 public:
     void Start() override;
     void Stop() override;
@@ -125,6 +127,11 @@ private:
     void ScheduleFlush();
     void DoFlush();
 };
+
+TProfileLog::~TProfileLog()
+{
+    Flush();
+}
 
 void TProfileLog::Start()
 {
--- a/cloud/blockstore/libs/diagnostics/profile_log_ut.cpp
+++ b/cloud/blockstore/libs/diagnostics/profile_log_ut.cpp
@@ -110,14 +110,12 @@ struct TEnv
 
     ~TEnv()
     {
-        if (ProfileLog) {
-            ProfileLog->Stop();
-        }
+        ProfileLog->Stop();
     }
 
-    void ProcessLog()
+    void ProcessLog(bool runScheduler = true)
     {
-        if (Scheduler) {
+        if (runScheduler) {
             Scheduler->RunAllScheduledTasks();
         }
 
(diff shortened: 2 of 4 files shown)


cloud/blockstore/libs/diagnostics/profile_log.cpp
```diff
@@ -114,6 +114,8 @@ class TProfileLog final
     {
     }
 
+    ~TProfileLog();
```
komarevtsev-d, line 117: override/final

cloud/filestore/libs/diagnostics/profile_log.cpp
```diff
@@ -41,6 +41,8 @@ class TProfileLog final
     {
     }
 
+    ~TProfileLog();
```
komarevtsev-d, line 44: override/final

komarevtsev-d (dismissed):

cloud/blockstore/libs/diagnostics/profile_log_ut.cpp
```diff
@@ -388,6 +392,27 @@ Y_UNIT_TEST_SUITE(TProfileLogTest)
...
+        env.ProfileLog->Write(
+            {"disk2",
+             TInstant::Seconds(3),
+             IProfileLog::TReadWriteRequest{
+                 EBlockStoreRequest::WriteBlocks,
+                 TDuration::MilliSeconds(300),
+                 TDuration::MilliSeconds(42),
+                 TBlockRange64::WithLength(10, 20),
+             }});
+        env.Scheduler = nullptr;
```
budevg, line 408: to make test more readable and explicitly state that you want to process log without running the scheduler you can change ProcessLog signature (instead of setting env.Scheduler to nullptr and then check it)
```c++
void ProcessLog(bool runScheduler =  true)
```

cloud/blockstore/libs/diagnostics/profile_log_ut.cpp
```diff
@@ -388,6 +392,27 @@ Y_UNIT_TEST_SUITE(TProfileLogTest)
...
+        env.ProfileLog = nullptr;
```
budevg, line 409: you can `ProfileLog.reset(CreateProfileLogStub())` then you don't need to add everywhere the 
```c++
if (ProfileLog) {
}
```

cloud/filestore/libs/diagnostics/profile_log_ut.cpp
```diff
@@ -368,6 +374,25 @@ Y_UNIT_TEST_SUITE(TProfileLogTest)
             "fs3\t6000000\t4\t300000\t0",
             EventProcessor.FlatMessages[5]
         );
+
+        // Test flush on destruct
+        ProfileLog->Write(
+            {"fs3",
+             TRequestInfoBuilder()
```
budevg, line 381: same remarks as in blockstore's profile_log_ut

cloud/blockstore/libs/diagnostics/profile_log.cpp
```diff
@@ -114,6 +114,8 @@ class TProfileLog final
     {
     }
 
+    ~TProfileLog() override;
```
budevg, line 117: no need to  mark destructor with override https://github.com/isocpp/CppCoreGuidelines/issues/721

cloud/filestore/libs/diagnostics/profile_log.cpp
```diff
@@ -41,6 +41,8 @@ class TProfileLog final
     {
     }
 
+    ~TProfileLog() override;
```
budevg, line 44: no need to mark destructor with override https://github.com/isocpp/CppCoreGuidelines/issues/721

cloud/blockstore/libs/diagnostics/profile_log.cpp
```diff
@@ -126,6 +128,11 @@ class TProfileLog final
     void DoFlush();
 };
 
+TProfileLog::~TProfileLog()
+{
+    Flush();
```
budevg, line 133: in blockstore for some reason the implementation of Flush looks like this
```c++
bool TProfileLog::Flush()
{
    if (AtomicGet(ShouldStop)) {
        return false;
    }

    DoFlush();
    return true;
}
```
so it won't do anything since on service stop ShouldStop is set ... you better call DoFlush() instead

budevg (changes requested):

budevg (approved):

komarevtsev-d (approved):

PR #2934, opened 2025-01-28
FIx restoration shadow disk from backup

drbasic:
When restoring from backup, shadow disks are deleted, this is normal, since the backup may be old.

1. corrected the warning message
2. The shadow disk in the volume will immediately go into error state as soon as there disk registry restarted

### DIFF
--- a/cloud/blockstore/libs/storage/disk_registry/actors/restore_validator_actor.cpp
+++ b/cloud/blockstore/libs/storage/disk_registry/actors/restore_validator_actor.cpp
@@ -545,7 +545,7 @@ void TRestoreValidationActor::HandleListVolumesResponse(
                     ctx,
                     Component,
                     RESTORE_PREFIX << " DiskID " << itr->GetDiskId().Quote()
-                                   << " is found in backup but not in SS");
+                                   << " is found in backup, but not in SS");
             }
             SetErrorDevicesInBackup(itr->GetDeviceUUIDs(), ctx.Now());
             DisksInBackup.erase(itr->GetDiskId());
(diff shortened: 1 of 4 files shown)


agalibin (approved):

cloud/blockstore/libs/storage/disk_registry/actors/restore_validator_actor.cpp
```diff
@@ -532,12 +532,21 @@ void TRestoreValidationActor::HandleListVolumesResponse(
...
+                    ctx,
+                    Component,
+                    RESTORE_PREFIX << " ShadowDisk " << itr->GetDiskId().Quote()
+                                   << " is found in backup");
+            } else {
+                LOG_WARN_S(
+                    ctx,
+                    Component,
+                    RESTORE_PREFIX << " DiskID " << itr->GetDiskId().Quote()
+                                   << " is found in backup but not in SS");
```
tpashkin, line 548: Nit. Shouldn't there be comma before "but"?

cloud/blockstore/libs/storage/volume/volume_ut_checkpoint.cpp
```diff
@@ -3725,6 +3725,60 @@ Y_UNIT_TEST_SUITE(TVolumeCheckpointTest)
...
+            0);
+        volume.AddClient(clientInfo);
+
+        // Create checkpoint.
+        volume.CreateCheckpoint("c1");
+
+        // Reconnect pipe since partition has restarted.
+        volume.ReconnectPipe();
+
+        // Shadow disk enter to the error state.
```
tpashkin, line 3775: Nit. I'm not quite sure what we are trying to say here. "Shadow disk entered the error state" maybe?

tpashkin (approved):

agalibin (approved):

Issue #2674, opened 2024-12-11
[Filestore] Implement directory creation in shards


qkrorlqr:
Right now sharding is done for regular inodes (i.e. files). Directories, symlinks, etc. are managed by a single tablet per single logical filesystem. It's a bottleneck for:
* directory listing
* file creation, renaming and unlinking
* inode lookup by parent id + child name

For example, regarding file creation - we are limited by numbers around 5-10k creations per second per single logical filesystem.

We need to implement sharding for directories. It will lead to the need to implement something like distributed transactions - most notably for the rename operation: either source directory shard or target directory shard may be unable to perform the operation and we need to either commit the operation in both source and target or not commit it anywhere. The most straightforward way to implement it is via 2PC. Considering the fact that all of our transactions will have at most 2 participants and the fact that each of those participants is reliable and has access to persistent storage, we can assign transaction coordinator role to one of the participants (e.g. source directory shard). Then 2PC actually transforms into something which is pretty simple to implement:
1. source directory shard receives the request, checks whether it can perform it - if it can, locks the directory for unlinking and locks the source name in the directory for any kind of ops
2. source directory shard sends the request to the target directory shard
3. if target directory shard returns a success code, source directory shard commits the change on its side and releases the locks
4. if target directory shard returns an error code, source directory shard just releases the locks

qkrorlqr:
TODO: safety check (if our guids are not gUids) - check that NodeAttr in TCreateNodeResponse from shard matches the corresponding TCreateNodeRequest

qkrorlqr:
**RenameNode implementation plan**

Important rules:
* directory creation in shard is allowed only if LastNodeId == 1 in the main tablet (i.e. main tablet doesn't manage ANY nodes) - by this we get rid of the case when we need to convert a local node into an external node
* all nodes are either nodes referenced by <RootNodeId, GUID> or are external nodes - serves the same goal plus simplifies the directory structure - it's max depth in each shard is now 1 or 2

RenameNodeRequest will be sent from TStorageServiceActor to the shard in charge of the source directory. This shard will play the role of the transaction coordinator.

If both source and destination directories are managed by the same tablet, RenameNode works in the same way it works right now. It's a simple way to achieve 2 things:
* keep RenameNode logic the same for shardless filesystems
* optimize the case when a file is moved within the same directory - should happen a lot, e.g. for the cases like "create tmp file" -> "populate tmp file" -> "m file.tmp file"

[...]

Issue #2667, opened 2024-12-10
[Filestore] Hardlink creation in sharded filesystems is not atomic


qkrorlqr:
Hardlink creation in sharded filesystems is done via this actor: https://github.com/ydb-platform/nbs/blob/5329d25754cd4c0a0fa28574131600a87b05e234/cloud/filestore/libs/storage/service/service_actor_createnode.cpp#L25

It first creates a hardlink in the shard which is responsible for the linked node and then creates a link in the main tablet. This whole operation is not atomic - the first step may succeed and the second may fail after that. We need to reimplement it in a way similar to regular node creation - first step should add a request to OpLog in the same transaction with the other logic of this step and the second step should be retried until it either succeeds or fails with a non-retriable error. The difference with regular node creation is that here the order is reversed - we first create a node in the shard and, second, we make modifications to the main filesystem. Those modifications may fail so extra logic may be needed - i.e. rollback of the first step in shard.

debnatkh:
Proposed plan:

1. External node refs can now use either external ShardNodeName + ShardId OR foreign childId
2. LinkNode will first increment the link count using the `LinkNodeInShard` op and only then commit nodeRef
3. UnlinkNode should be able to handle unlinking [ShardNodeName + ShardId] (the original file) as well as ONLY [foreign child id] (it will decrement shard's node link count but will not touch nodeRef)

PR #3014, opened 2025-02-11
issue-2674: add fs posix compliance test for sharded directories, fix hardlinks behavior

debnatkh:
References #2674

---

During the implementation of #2674 a new logic was introduced that basically created new types of shards: those that both contain files that are external to some other shard (i.e., they have some uuid instead of name) and also those that contain directories of the actual filesystem. For the files that are external to some other shard there is no need to use the dupcache logic. So the flag BehaveAsShard was introduced that disabled the dupcache logic:

https://github.com/ydb-platform/nbs/blob/565d6dedbed6765bcc082f91650a9c66ffb01153/cloud/filestore/libs/storage/tablet/tablet_actor.cpp#L1234

This flag is enabled if this is a main filesystem or if the client explicitly says so by enabling the `BehaveAsDirectoryTablet` flag. So, if, say, the client wants to perform `CreateHandle(nodeId)`, this request can be directly routed to the shard that contains the file and the dupcache logic must be used, because the client can retry this operation and we do not want to create two handles for the same file. But if a request `CreateNode(parentId, name)` is performed, this request must be routed to the shard that contains the directory. There the DupCache must be used, but this request produces a new request to the corresponding shard, which must not use the DupCache. So the flag `BehaveAsDirectoryTablet` is disabled for this reqeust:

https://github.com/ydb-platform/nbs/blob/565d6dedbed6765bcc082f91650a9c66ffb01153/cloud/filestore/libs/storage/service/service_actor_createhandle.cpp#L127

For hardlinks this logic must be patched: The hardlinks are created in a two-stage manner: first a hardlink is created in a shard, and then the nodeRef is added to the directory shard. The first stage must not use the DupCache, but the second stage must use it. So the flag `BehaveAsDirectoryTablet` must be disabled for the first stage and enabled for the second stage. This is the fix that is present in this PR. Note that the whole two-stage manner that the hardlinks are created is to be rewritten as it can potentially lead to orphaned files in shards and is non-atomic, see https://github.com/ydb-platform/nbs/issues/2667.


cloud/filestore/libs/storage/service/service_ut_sharding.cpp
```diff
@@ -4557,6 +4557,35 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
...
+        TServiceClient service(env.GetRuntime(), nodeIdx);
+        TFileSystemConfig fsConfig;
+        fsConfig.DirectoryCreationInShardsEnabled = true;
+        CreateFileSystem(service, fsConfig);
+
+        auto headers = service.InitSession(fsConfig.FsId, "client");
+
+        const auto dirId = service.CreateNode(
+            headers,
+            TCreateNodeArgs::Directory(RootNodeId, "dir"))->Record.GetNode().GetId();
```
WilyTiger, line 4578: слишком длинная строчка

WilyTiger (approved):

yegorskii (approved):

PR #3059, opened 2025-02-19
[Fielstore] issue-2995: fix failing local filestore possix compliance tests

budevg:
- return ENAMETOOLONG properly when node name is too long
- clear umask to pass permissions properly to local filesystem
  By default umask is set to 0022 which won't allow creating files with proper
  permissions
- update the ENAMETOOLONG test canondata to reflect the tests now passing
- update canondata for chown tests when run as non-root
  The chown tests fail when run as a non-root user because they can not change
  euid/egid. Update the expected results.

  link 00.t 69 not ok -- -- chown fstest_978244a1a6 65534 -1: EPERM expected 0
  link 00.t 76 not ok -- -- chown fstest_978244a1a6 65534 -1: EPERM expected 0
- add local noserver test


yegorskii (approved):

budevg:
broken tests is not related to this pr I will fix it in another pr

PR #3209, opened 2025-03-18
always dump dmesg and atop upon `ya test` completion

debnatkh:


debnatkh:
https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/Check-github-workflow/13928613341/1/nebius-x86-64/test_logs/index.html

Работает

ya-ksgamora (approved):

PR #3236, opened 2025-03-21
feat: add checking for run status for future merge requirement

librarian:
Proposed change after this commit is that we 

Example of how it works:

https://github.com/librarian-test/nbs/pull/100
afbeelding

We **MUST** have two conditions met:
1. At least two approvals per PR
2. Two checks are required to finish: 
afbeelding

"Check for correct approvals" is this one https://github.com/ydb-platform/nbs/blob/main/.github/workflows/approvals.yaml which checks that both right and left teams approved this change. With exception of superuser approval from two superusers.
"Check run status" is the new one where we query API for all checks within PR check run (excluding two)

Some improvements for the future:
* separate list of users whose comment will explicitly make "Check run status" as successful. Or, alternatively, add those people to "Bypass list" in settings. Which need some research in what this thing actually does
* add support for regexes to exclude "service" jobs that might fail for some external reason, like releasing of runner.
* add separate mute file for PR checks that will be ignored during nightly builds

### DIFF
--- a/.github/actions/final_check/action.yaml
+++ b/.github/actions/final_check/action.yaml
@@ -1,0 +1,27 @@
+name: Check job statuses
+description: Fails if any non-excluded jobs in a workflow run have failed
+
+inputs:
+  token:
+    required: true
+    description: "GitHub token"
+  excluded:
+    required: false
+    description: "Comma-separated list of job names to exclude from failure check"
+    default: ""
+  self:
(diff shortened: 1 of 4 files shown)


librarian:
We decided that we won't be going forward with this PR because as it is - it is not usable enough and problem it is intended to solve doesn't happen very often.

PR #3447, opened 2025-04-30
Spring cleanup in workflows/actions + fixes for lower memory presets

librarian:
1. Removed nebius param and non NEBIUS_ prefixed variables
2. Removed setting of oom score for ya make
3. Moved bazel password from process argument to process file -> ps aux is safe to print
4. Recalculated thread count
5. Removed swapfile in favor of zram swap which is MUCH faster


.github/workflows/github_actions_scripts.yaml
```diff
@@ -66,11 +66,8 @@ jobs:
       if: ${{ github.event.pull_request.head.sha != '' && contains(github.event.pull_request.labels.*.name, 'rebase') }}
       shell: bash
       run: |
-        set -x
-        # git config is needed for some reason, idk why it works without
-        # it on self-hosted runners
-        git config user.email "librarian@nebius.com"
-        git config user.name "Rebase Robotovich"
```
komarevtsev-d, line 73: Зачем переименовал? 😁

komarevtsev-d (approved):

.github/workflows/github_actions_scripts.yaml
librarian: Потому что зачем мой email тут? Сейчас весь CI от robot-nbs работает

PR #3461, opened 2025-05-05
Add ability to adjust actor system thread count according to available CPU cores percentage

SvartMetal:
Introducing `--actor-system-available-cpu-cores-percentage` cmd-line flag which adjusts `System`, `User`, `IC` thread pool sizes (if those sizes are equal) according to available CPU cores percentage

### DIFF
--- a/cloud/storage/core/libs/kikimr/config_initializer.cpp
+++ b/cloud/storage/core/libs/kikimr/config_initializer.cpp
@@ -58,8 +58,8 @@ void AdjustActorSystemThreadsAccordingToAvailableCpus(
     }
 
     const auto availableCores = TAffinity::Current().GetCores().size();
-    // 3 stands for number of thread pools (System, User, IC)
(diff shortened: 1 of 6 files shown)


cloud/storage/core/libs/kikimr/config_initializer.cpp
```diff
@@ -1,19 +1,78 @@
...
+    for (auto* executor: executors) {
+        if (executor->GetThreads() != threadCount) {
+            // each pool should have equal number of threads
+            return;
+        }
+    }
+
+    const auto availableCores = TAffinity::Current().GetCores().size();
+    // 3 stands for number of thread pools (System, User, IC)
+    const double scalingFactor = (1.0 * availableCpuCoresPercentage) / 100 / 3;
```
drbasic, line 62: тогда не нужен будет коммент про 3
```
(1.0 * availableCpuCoresPercentage) / (100 * executors.size())
```

drbasic (approved):

WilyTiger (approved):

PR #3586, opened 2025-05-28
[Disk Manager, SU] Include Shards service into disks service. 

Rattysed:
We introduce a new service to manage shard selection when creating disks. By default, disks are created in the first shard defined in the configuration.

Added features:

- ShardService: a new component used by DiskService to select the target shard before disk creation.
- Default selection strategy: always use the first shard from the configuration.


cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -176,6 +167,17 @@
     MinOptimizedPoolAge: "1s"
     BaseDiskIdPrefix: "{base_disk_id_prefix}"
 >
+ShardsConfig: <
```
Rattysed, line 170: Если в конфиге не будет указан ShardsConfig, то при почти любом создании диска будет паника. Не знаю, насколько это валидное поведение

cloud/disk_manager/internal/pkg/services/disks/service.go
```diff
@@ -165,24 +167,16 @@ func (s *service) prepareZoneId(
...
 	}
 
-	if diskMeta == nil {
-		return shards[0], nil
+	if diskMeta != nil {
+		return diskMeta.ZoneID, nil
 	}
 
-	return diskMeta.ZoneID, nil
+	return s.shardsService.SelectShard(ctx, disk), nil
```
BarkovBG, line 179: pickShard

PR #3626, opened 2025-06-04
Do not automatically set the  tag for encrypted volumes

sharpeye:


WilyTiger (approved):

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor.cpp
```diff
@@ -218,7 +218,10 @@ void TMirrorPartitionActor::CompareChecksums(const TActorContext& ctx)
             DiskId.c_str(),
             DescribeRange(GetScrubbingRange()).c_str());
 
-        if (Config->GetAutomaticallyEnableBufferCopyingAfterChecksumMismatch())
+        if (Config
+                ->GetAutomaticallyEnableBufferCopyingAfterChecksumMismatch() &&
+            // a temporary buffer is already used for an encrypted volumes
```
komarevtsev-d:
```suggestion
            // a temporary buffer is already used for encrypted volumes
```

PR #3627, opened 2025-06-04
don't alter volume in modify tags action if the tags haven't changed

WilyTiger:
Currently, if a mismatch is found during scrubbing and AutomaticallyEnableBufferCopyingAfterChecksumMismatch is enabled, the mismatch could never get fixed through resync. Adding a tag to the volume will trigger a restart of the partitions with an updated volume config, and scrubbing will start from the first range again.

### DIFF
--- a/cloud/blockstore/libs/storage/service/service_actor_actions_modify_tags.cpp
+++ b/cloud/blockstore/libs/storage/service/service_actor_actions_modify_tags.cpp
@@ -269,7 +269,7 @@ void TModifyTagsActionActor::HandleDescribeVolumeResponse(
     }
 
     if (!changed) {
-        LOG_DEBUG(
+        LOG_INFO(
             ctx,
             TBlockStoreComponents::SERVICE,
             "Skipping tag modification; tags are already as desired for volume "


sharpeye (dismissed):

cloud/blockstore/libs/storage/service/service_actor_actions_modify_tags.cpp
```diff
@@ -253,12 +253,30 @@ void TModifyTagsActionActor::HandleDescribeVolumeResponse(
...
     for (const auto& tag: Request.GetTagsToAdd()) {
-        tags.insert(tag);
+        auto [_, inserted] = tags.insert(tag);
+        if (inserted) {
+            changed = true;
+        }
+    }
+
+    if (!changed) {
+        LOG_DEBUG(
```
komarevtsev-d: Может info? Кажется лог полезный и не спамящий

komarevtsev-d (approved):

drbasic (dismissed):

komarevtsev-d (approved):

sharpeye (approved):

Issue #1751, opened 2024-08-08
[Filestore] writeback-cache for files which are open for writing by only one client


qkrorlqr:
If some other client tries to open the same file for RDWR as well, the tablet needs to notify the first client so that it would disable writeback caching, and only after that allow RDWR opening by another client.

fsync should cause cache flush of course

budevg:
Fuse already supports writeback cache on the [client side](https://github.com/libfuse/libfuse/blob/master/include/fuse_common.h#L331). I will add option to pass this flag during fuse session initialization

SvartMetal:
TODO:

* https://github.com/ydb-platform/nbs/pull/3395#discussion_r2097683180
* https://github.com/ydb-platform/nbs/pull/3395#discussion_r2098747395
* https://github.com/ydb-platform/nbs/pull/3395#discussion_r2103098410

e673:
Ideas:
* https://github.com/ydb-platform/nbs/blob/bbeaa11871d15020b0f3fce18b2ef733e782e371/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp#L235
Shrink the read range if the cached data lays on its boundaries.

PR #3628, opened 2025-06-04
issue-1751: simplify TSessionSequencer: don't wrap TPromise into shared_ptr

SvartMetal:
#1751


SvartMetal:
False positive:
Screenshot 2025-06-04 at 19 30 17

PR #3784, opened 2025-06-27
[Disk Manager] Limit the number of checkpoint creation reties when create image/snapshot from disk registry based disk

gy2411:
https://github.com/ydb-platform/nbs/issues/1950

Singe task should not retry checkpoint creation too many times. Otherwise, there is a risk of overloading disk agent with checkpoint creation/deletion requests.

Add parameter CheckpointIterationLimit to dataplane config (equals 10 by default).


cloud/disk_manager/internal/pkg/facade/with_broken_checkpoints_test/with_broken_checkpoints_test.go
```diff
@@ -364,3 +383,114 @@ func TestImageServiceCreateImageFromDiskWithFailedShadowDiskAndOperationCancel(
...
+
+	response := makeBlankResponse(isImage)
+	err = internal_client.WaitResponse(
+		ctx,
+		client,
+		operation.Id,
+		response,
+	)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "Non retriable error")
```
gy2411, line 490: `errors.Is(err, errors.NewEmptyNonRetriableError())` тут не получается использовать, так как тип ошибки до сюда не доходит.

PR #3878, opened 2025-07-09
TFileRingBuffer: increase reliability is a buffer is tampered while it is in use

e673:
Extracted from https://github.com/ydb-platform/nbs/pull/3790
Based on https://github.com/ydb-platform/nbs/pull/3817

Resolve TODOs in the code related to the corruption.


e673:
No more needed

Issue #3314, opened 2025-04-02
[Blockstore] add io_uring storage backend for DiskRegistry-based disks


SvartMetal:
By analogy with https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/tools/testing/rdma-test/storage_local_uring.cpp

budevg:
This will also help for filestore-local which is currently using libaio to read/write data to files [here](https://github.com/ydb-platform/nbs/blob/aca85ee8465b3dc8b28997903de6c48610ce6525/cloud/filestore/libs/service_local/fs_data.cpp#L111)
We just need to make sure to have the AsyncReadV/AsyncWriteV api as well

PR #3883, opened 2025-07-09
issue-3314: allow sharing kernel workers between io_uring services; add TIoUring that does all the io_uring stuff; make io_uring service thread-safe; stop the completion thread using NOP instead of eventfd

sharpeye:
#3314

### DIFF
--- a/cloud/storage/core/libs/io_uring/service.cpp
+++ b/cloud/storage/core/libs/io_uring/service.cpp
@@ -108,37 +108,14 @@ private:
     TThread CompletionThread;
 
 public:
-    explicit TIoUring(TIoUringServiceParams params, TIoUring* other = nullptr)
-        : SubmissionThread(CreateThreadPool(params.SubmissionThreadName, 1))
-        , CompletionThread(
-              [this, threadName = std::move(params.CompletionThreadName)]
-              {
-                  SetHighestThreadPriority();
-                  ::NCloud::SetCurrentThreadName(threadName);
-                  ProcessCompletionQueue();
-              })
-    {
-        const auto error = InitRing(
-            &Ring,
-            params.SubmissionQueueEntries,
-            other ? &other->Ring : nullptr);
+    explicit TIoUring(TIoUringServiceParams params)
+        : TIoUring(std::move(params), nullptr)
+    {}
 
-        Y_ABORT_IF(
-            HasError(error),
-            "can't initialize the ring: %s",
-            FormatError(error).c_str());
-
-        if (!other && (params.BoundWorkers || params.UnboundWorkers)) {
-            const auto error = SetMaxWorkers(
-                &Ring,
-                params.BoundWorkers,
-                params.UnboundWorkers);
-            Y_ABORT_IF(
-                HasError(error),
-                "can't set max workers: %s",
-                FormatError(error).c_str());
-        }
-    }
+    // Share kernel worker threads with `other`
+    TIoUring(TIoUringServiceParams params, TIoUring& other)
+        : TIoUring(std::move(params), &other)
+    {}
 
     ~TIoUring()
     {
@@ -200,6 +177,36 @@ public:
     }
 
 private:
+    TIoUring(TIoUringServiceParams params, TIoUring* other)
+        : SubmissionThread(CreateThreadPool(params.SubmissionThreadName, 1))
+        , CompletionThread(
+              std::bind_front(
+                  &TIoUring::CompletionThreadProc,
+                  this,
+                  std::move(params.CompletionThreadName)))
+    {
+        const auto error = InitRing(
+            &Ring,
+            params.SubmissionQueueEntries,
+            other ? &other->Ring : nullptr);
+
+        Y_ABORT_IF(
+            HasError(error),
+            "can't initialize the ring: %s",
+            FormatError(error).c_str());
+
+        if (!other && (params.BoundWorkers || params.UnboundWorkers)) {
+            const auto error = SetMaxWorkers(
+                &Ring,
+                params.BoundWorkers,
+                params.UnboundWorkers);
+            Y_ABORT_IF(
+                HasError(error),
+                "can't set max workers: %s",
+                FormatError(error).c_str());
+        }
+    }
+
     void SubmitIO(
         int op,
         int fd,
@@ -293,8 +300,11 @@ private:
         }
     }
 
-    void ProcessCompletionQueue()
+    void CompletionThreadProc(const TString& threadName)
     {
+        SetHighestThreadPriority();
+        NCloud::SetCurrentThreadName(threadName);
+
         for (;;) {
             io_uring_cqe* cqe = nullptr;
             const int ret = io_uring_wait_cqe(&Ring, &cqe);
@@ -320,14 +330,16 @@ private:
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TIoUringService final
+struct TIoUringServiceBase
     : public IFileIOService
 {
     TIoUring IoUring;
 
-    explicit TIoUringService(
-            TIoUringServiceParams params,
-            TIoUring* other = nullptr)
+    explicit TIoUringServiceBase(TIoUringServiceParams params)
+        : IoUring(std::move(params))
+    {}
+
+    TIoUringServiceBase(TIoUringServiceParams params, TIoUring& other)
         : IoUring(std::move(params), other)
     {}
 
@@ -340,6 +352,14 @@ struct TIoUringService final
     {
         IoUring.Stop();
     }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TIoUringService final
+    : public TIoUringServiceBase
+{
+    using TIoUringServiceBase::TIoUringServiceBase;
 
     void AsyncRead(
         TFileHandle& file,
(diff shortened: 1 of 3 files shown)


cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +48,236 @@ NProto::TError Submit(io_uring* ring)
...
+        params.flags |= IORING_SETUP_ATTACH_WQ;
+        params.wq_fd = other->ring_fd;
     }
 
-    ~TRing()
-    {
-        io_uring_queue_exit(&Ring);
+    int ret = io_uring_queue_init_params(entries, ring, &params);
+    if (ret == -EINVAL) {
+        // IORING_SETUP_SINGLE_ISSUER is not supported
```
tpashkin: Как такое возможно, старое ядро?

sharpeye: Да, для IORING_SETUP_SINGLE_ISSUER нужно 6.0 - на моей dev-тачке, например, 5.15

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +48,236 @@ NProto::TError Submit(io_uring* ring)
...
-        }
+        SubmissionThread->Start();
+
+        // Since IORING_SETUP_R_DISABLED and IORING_SETUP_SINGLE_ISSUER
+        // are set, we must enable the ring in the context of the
+        // submission thread.
+        SubmissionThread->ExecuteSimple(
+            [this]
+            {
+                const auto error = EnableRing(&Ring);
```
drbasic, line 165: чем гарантируется что this живой?

sharpeye: Тем, что SubmissionThread принадлежит TIoUring

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +48,236 @@ NProto::TError Submit(io_uring* ring)
...
     }
 
-    [[nodiscard]] NProto::TError Register(TRing& ring)
+    void Stop()
     {
-        io_uring_sqe* sqe = io_uring_get_sqe(ring);
-        if (!sqe) {
-            return MakeError(E_INVALID_STATE, "submission queue is full");
+        if (!CompletionThread.Running()) {
+            return;
```
drbasic, line 179: SubmissionThread->Stop(); ?

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +48,236 @@ NProto::TError Submit(io_uring* ring)
...
+            });
     }
 
-    [[nodiscard]] NProto::TError Register(TRing& ring)
+    void Stop()
     {
-        io_uring_sqe* sqe = io_uring_get_sqe(ring);
-        if (!sqe) {
-            return MakeError(E_INVALID_STATE, "submission queue is full");
+        if (!CompletionThread.Running()) {
```
drbasic: если дернуть Start - Stop, то появляется гонка

sharpeye: Ага, я поправил

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +50,243 @@ NProto::TError Submit(io_uring* ring)
...
 
-class TEvent
+class TIoUring
 {
 private:
-    const TFileHandle Fd;
-    ui64 Value = 0;
+    io_uring Ring = {};
+
+    ITaskQueuePtr SubmissionThread;
```
SvartMetal, line 112: SubmissionQueue?

sharpeye: Тут речь именно про единственный поток, который сабмитит запросы в кольцо - SubmissionThread

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +50,238 @@ NProto::TError Submit(io_uring* ring)
     }
 }
 
+NProto::TError SetMaxWorkers(io_uring* ring, ui32 bound, ui32 unbound)
```
budevg: Why do we need here the unbound workers at all. According to docs these async workers are for using io_uring with sockets and pipes. We intend to use io_uring for io so let's simplify the api and reduce optional configs that will never be used in our codebase

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +50,238 @@ NProto::TError Submit(io_uring* ring)
...
+    if (wqOwner) {
+        params.flags |= IORING_SETUP_ATTACH_WQ;
+        params.wq_fd = wqOwner->ring_fd;
+    }
+
+    for (;;) {
+        const int ret = io_uring_queue_init_params(entries, ring, &params);
+
+        if (ret == -EINVAL && params.flags & IORING_SETUP_SINGLE_ISSUER) {
+            // IORING_SETUP_SINGLE_ISSUER is not supported
```
budevg, line 78: If IORING_SETUP_SINGLE_ISSUER not supported by kernel do we want to implement our own user space mechanism to forbid submission to ring from different threads?

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -46,129 +50,238 @@ NProto::TError Submit(io_uring* ring)
...
+            return MakeSystemError(-ret, "io_uring_queue_init_params");
+        }
+
+        return {};
+    }
+}
+
+NProto::TError EnableRing(io_uring* ring)
+{
+    for (;;) {
```
budevg, line 93: I think we need to implement some kind of timeout here and in other places where similar pattern used. There is a chance we will enter infinite loop due to some bug and won't even know this happens

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -191,14 +304,14 @@ class TCompletionThread final: public ISimpleThread
...
+    void CompletionThreadProc(const TString& threadName)
     {
         SetHighestThreadPriority();
-        NCloud::SetCurrentThreadName(Name);
+        NCloud::SetCurrentThreadName(threadName);
 
         for (;;) {
             io_uring_cqe* cqe = nullptr;
-            const int ret = io_uring_wait_cqe(Ring, &cqe);
+            const int ret = io_uring_wait_cqe(&Ring, &cqe);
```
budevg, line 312: maybe better to use `io_uring_wait_cqe_timeout` and then check for thread should stop condition in case thread is stopped

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -404,53 +470,66 @@ class TIoUringServiceNull final
...
+                                      << Params.CompletionThreadName << index;
 
-        NSan::Release(completion);
+        auto service =
+            std::make_shared<TService>(std::move(params), WqOwner.get());
 
-        const auto error = Submit(Ring);
-        if (HasError(error)) {
-            completion->Func(completion, error, 0);
+        if (Params.ShareKernelWorkers && !WqOwner) {
```
budevg, line 505: not sure I understand the logic here.
The first call to CreateFileIOService will always create service without shared ring
and following calls will use the WqOwner ring if they use `ShareKernelWorkers` option
what if I want all 3 to share the asynchronous worker thread ?
do we have use case where we want to have 2 sets of rings where each set shares worker thread ?

cloud/storage/core/libs/io_uring/service.cpp
sharpeye: IORING_SETUP_SINGLE_ISSUER is used as an optimization, not as a mechanism to forbid anything.
Only two threads have access to the io_uring object: SubmissionThread (exclusive access to the sq part) and CompletionThread (exclusive access to the cq part). Thus, there are no other threads that can touch the ring.

cloud/storage/core/libs/io_uring/service.cpp
sharpeye: Why is checking a flag in a loop better than stopping by the specific completion?

Anyway, `io_uring_wait_cqe_timeout` uses the sq part of the ring to submit a timeout request, so it can't be used in the `CompletionThread`.

cloud/storage/core/libs/io_uring/service.cpp
sharpeye: The first service's thread pool is shared with other services created by the same service factory. Example:
```cpp
auto foo = CreateIoUringServiceFactory({.ShareKernelWorkers = true});
auto bar = CreateIoUringServiceFactory({.ShareKernelWorkers = true});

// a, b and c share one thread pool
auto a = foo->CreateFileIOService();
auto b = foo->CreateFileIOService();
auto c = foo->CreateFileIOService();

// x, y and z share another thread pool
auto x = bar->CreateFileIOService();
auto y = bar->CreateFileIOService();
auto z = bar->CreateFileIOService();

```

cloud/storage/core/libs/io_uring/service.cpp
sharpeye: > I think we need to implement some kind of timeout here and in other places where similar pattern used. 

You mean retry for a certain amount of time (like a minute), then return a fatal error?

> There is a chance we will enter infinite loop due to some bug and won't even know this happens

What kind of bug? When the OS keeps returning retriable errors?

budevg: > You mean retry for a certain amount of time (like a minute), then return a fatal error?

yep if we can't enable ring for such significant time it's better to fail and propagate the error so that our alerts will indicate some kind of trouble
> What kind of bug? When the OS keeps returning retriable errors?

Some of the errors like EBUSY can be continuously returned when no resources available. So there could be a use case when the os is in some condition where EBUSY will be returned without possibility to recover
For example [here](https://github.com/torvalds/linux/blob/v6.6/io_uring/io_uring.c#L2831) although this is some feature of io_uring that we don't use. The logic there is that if you already registred event_fd ... fail with EBUSY

sharpeye: I'll add this thing in the next PR

komarevtsev-d (approved):

PR #3916, opened 2025-07-15
[Blockstore] mark request to mirror disk as dirty when starting ReadBlocks, not when completing (Write|Zero)Blocks; add additional check when inserting request to DirtyReadRequestIds to prevent this structure's slow bloating

SvartMetal:
Fixes a possible bug where a read request may not be marked dirty (but it should). Scenario:
* Read request starts after all overlapping in-flight write/zero requests passed this point: https://github.com/ydb-platform/nbs/blob/11033ffccb3a538de2e9cd9c4c2cde74adaf9ca9/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp#L137
* Read request finishes before any overlapping in-flight write/zero request is completed, so this code is no-op: https://github.com/ydb-platform/nbs/blob/11033ffccb3a538de2e9cd9c4c2cde74adaf9ca9/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp#L37

### DIFF
--- a/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
+++ b/cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
@@ -30,13 +30,9 @@ void TMirrorPartitionActor::HandleWriteOrZeroCompleted(
     if (!completeRequest) {
         return;
     }
+
     DrainActorCompanion.ProcessDrainRequests(ctx);
-    auto [volumeRequestId, _, range] = completeRequest.value();
-    for (const auto& [id, request]: RequestsInProgress.AllRequests()) {
-        if (range.Overlaps(request.BlockRange)) {
-            DirtyReadRequestIds.insert(id);
-        }
-    }
+    auto [volumeRequestId, _, __] = completeRequest.value();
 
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_actor_mirror.cpp
SvartMetal, line 1: тесты

cloud/blockstore/libs/storage/partition_nonrepl/part_mirror_ut.cpp
```diff
@@ -2143,6 +2143,177 @@ Y_UNIT_TEST_SUITE(TMirrorPartitionTest)
...
+            [] (auto& runtime, const auto&, const auto& actorId)
+        {
+            runtime.EnableScheduleForActor(actorId);
+        });
+
+        TAutoPtr<IEventHandle> toSend;
+        bool interceptReadDeviceBlocks = false;
+        bool interceptWriteDeviceBlocks = false;
+
+        runtime.SetEventFilter([&] (auto&, auto& event) {
```
WilyTiger: clangformat

Issue #3703, opened 2025-06-18
[Filestore] missing posix compliance listing tests


debnatkh:
Right now we have a bunch of tests that test guest scenatrios, like mkdir/chown/touch, specifically, fs_posix_compliance: https://github.com/ydb-platform/nbs/tree/d3f0bb19a12ff4f4de19edded26b05bc0620713c/cloud/filestore/tools/testing/fs_posix_compliance/suite/tests. They use a fork of these tests: https://github.com/zfsonlinux/fstest/tree/master/tests

We need to test `ls` scenarios also. It can be achieved by either using a custom scenario like in this test: https://github.com/ydb-platform/nbs/tree/d3f0bb19a12ff4f4de19edded26b05bc0620713c/cloud/filestore/tests/fs_posix_compliance/acl-set-test, or by expanding this test suite: https://github.com/ydb-platform/nbs/tree/d3f0bb19a12ff4f4de19edded26b05bc0620713c/cloud/filestore/tools/testing/fs_posix_compliance/suite/tests with some additional tests

PR #3919, opened 2025-07-16
[Filestore] Add posix compliance tests for ls and opendir

neihar:
https://github.com/ydb-platform/nbs/issues/3703


cloud/filestore/tools/testing/fs_posix_compliance/suite/fstest.c
```diff
@@ -487,6 +490,17 @@ call_syscall(struct syscall_desc *scall, char *argv[])
 			return (i);
 		}
 		break;
+	case ACTION_OPENDIR:
```
WilyTiger, line 493: Может тогда добавим readdir и closedir?

cloud/filestore/tools/testing/fs_posix_compliance/suite/tests/ls/06.t
```diff
@@ -0,0 +1,151 @@
...
+dir=`dirname $0`
+. ${dir}/../misc.sh
+
+echo "1..50"
+
+n0=`namegen`
+n1=`namegen`
+n2=`namegen`
+
+# ls -t sorts by modification time (newest first) TODO uncomment after fix flop on qemu
```
WilyTiger, line 15: > TODO uncomment after fix flop on qemu

а про что тут имеется в виду?

cloud/filestore/tools/testing/fs_posix_compliance/suite/fstest.c
neihar: I'm not sure if it makes sense to include this, since we can't easily test it in the same way as the other commands (i.e., bash-style), because it requires a DIR pointer, which isn't available in Bash. We could implement some workarounds to make it work, but I’d suggest we avoid adding it to this test suite. Probably we can add in the future more complex tests as separate suite with python.

cloud/filestore/tools/testing/fs_posix_compliance/suite/tests/ls/06.t
neihar: According to POSIX, this sorting should behave as described in the test, but unfortunately it doesn't — and sometimes we get different results. I tested a few scenarios on macOS, and I didn't observe any inconsistencies there.  However, in the test suite, the order sometimes flips unexpectedly. So, more deep investigation is needed why sorting with '-t' does not work.

cloud/filestore/tools/testing/fs_posix_compliance/suite/fstest.c
WilyTiger: ok

cloud/filestore/tools/testing/fs_posix_compliance/suite/tests/ls/06.t
WilyTiger: ok

WilyTiger:
Давай еще проанализируем какие тесты упали(получили not ok), может быть у нас уже сейчас есть какие-то баги

neihar:
- opendir
   - mount-kikimr-test:
      -  01.t tests with permissions  lines 18 and 27
   - mount-local-test:
      -  01.t tests with permissions  lines 18 and 27
      -  01.t chown command lines 15 and 25 
   - qemu-local-nonserver-test
      - 01.t chown command lines 15 and 25
   - qemu-local-test
      - 01.t chown command lines 15 and 25
- ls
   - all
       -  09.t lines 19-21.   system doesn't allow to create filename with tab, but it should be possible according to posix
       -  06.t line 34  "ls -tr" problems with reverse time sort
   - + qemu-kikimr-multishard-test, mount-kikimr-test, qemu-kikimr-multishard2-test, qemu-kikimr-teest, qemu-kikimr-nemesis
       -  10.t line 28 - permission problem 
       
So, there are permissions problems with the chown/chmod tests. I don't think it's worth diving deeper into those. The reverse time sorting and the tab in the filename also look like they're not so important.

@WilyTiger

yegorskii (approved):

PR #3928, opened 2025-07-16
TWriteBackCache: read/write requests from/to TFileRingBuffer outside lock sections

e673:


e673:
Implemented in different PR

PR #3949, opened 2025-07-21
issue-3314: add deadline for retrying

sharpeye:


cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -36,28 +36,69 @@ bool IsRetriable(int error)
...
+
+        SpinLockPause();
+    }
+
+    return ret;
+}
+
+NProto::TError Submit(io_uring* ring)
+{
+    int ret = io_uring_submit(ring);
```
budevg: why not just call `Retry<io_uring_submit>(ring)` once and eliminate the second  call on line 70 ?

cloud/storage/core/libs/io_uring/service.cpp
```diff
@@ -232,9 +259,10 @@ class TIoUring
         NSan::Release(completion);
 
         const auto error = Submit(&Ring);
-        if (HasError(error)) {
-            completion->Func(completion, error, 0);
-        }
+        Y_ABORT_IF(
```
budevg: why ? we can fail to submit and it doesn't make sense to abort the process

budevg (changes requested):

cloud/storage/core/libs/io_uring/service.cpp
sharpeye: Once we have `sqe`, we can't just drop it - otherwise the submission queue will be in an inconsistent state. And returning an IO error to the client because of, for example, a host overload doesn't seem right.

cloud/storage/core/libs/io_uring/service.cpp
sharpeye: It's just a little optimisation, fast path: `Submit` is on the hot path.

budevg (approved):

PR #3951, opened 2025-07-21
issue-1751: TWriteBackCache - implement fallback to not using cache on failure

e673:


e673:
Concern: inability to reopen the cache file at session recreation may lead to data loss.

Currently, it is not possible to distinguish the following scenarios:
1. Cache file cannot be created (wrong path, no space etc).
2. Cache file was previously used but cannot be reopened.

Need to add information about actual cache to the session metadata.

PR #3968, opened 2025-07-22
add custom io_uring service to vhost local filestore

sharpeye:


cloud/filestore/libs/service_local/io_uring.cpp
```diff
@@ -0,0 +1,269 @@
...
+            Y_UNUSED(ptr);
+        }
+    }
+
+    bool HandleCompletion(const NProto::TError& error, ui32 bytesTransferred)
+    {
+        Y_ABORT_UNLESS(PendingSubRequestsCount > 0);
+        --PendingSubRequestsCount;
+
+        TotalTransferredBytes += bytesTransferred;
```
budevg, line 51: coud it be that we split into 16k request into 4 requests of 4k and request in the middle can only read 3k so we read 15k but the hole is in the middle and not the end. Hole in the end is ok since we can read less bytes then requested but reading less bytes in the middle is not ok

budevg (changes requested):

cloud/filestore/libs/service_local/io_uring.cpp
budevg: also performing operation out of order could be problematic there is IOSQE_IO_LINK which can help
https://unixism.net/loti/tutorial/link_liburing.html

PR #3992, opened 2025-07-25
use TBlockDataRef in IFileIOService instead of TArrayRef

sharpeye:


cloud/storage/core/libs/aio/service.cpp
```diff
@@ -194,19 +194,16 @@ class TAIOService final
...
-        }
+        auto req = std::make_unique<iocb>();
 
         io_prep_preadv(
             req.get(),
             file,
-            iov.data(),
-            iov.size(),
+            // TArrayRef has the same layout as iovec
+            std::bit_cast<iovec*>(buffers.data()),
```
budevg, line 205: looks very risky the layout of ArrayRef can change elsewhere the code will compile and we will pass invalid data.
Also I am not sure how much performance we save by eliminate copying the  iovec for small io's the iovec size is small for large io's it could be 1k size for 1M requests. But the overhead of creating iovec is very negligible relative to 1M buffer I/O

budevg (changes requested):

budevg (approved):

PR #4046, opened 2025-07-31
io_uring: nvme passthrough

sharpeye:


cloud/storage/core/libs/io_uring/context.cpp
```diff
@@ -19,6 +22,37 @@ const TDuration DefaultTimeout = TDuration::Minutes(1);
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TNvmeUringCmd
```
sharpeye, line 25: https://github.com/ydb-platform/nbs/blob/main/contrib/libs/linux-headers/linux/nvme_ioctl.h#L74

PR #4051, opened 2025-08-01
[Config Generator] Add default k8s path for kubernetes clusters

jkuradobery:


jkuradobery:
We have decided to use two separate folders for debian and k8s deployed clusters in our manifests & configs directory structure. 
Thus, this PR is not needed.

 “The greatest victory is that which requires no battle.”  ― Sun Tzu, The Art of War

PR #4058, opened 2025-08-04
Allow to set TOS for RDMA traffic

komarevtsev-d:
Добавляю в конфиг значение, которое прорастает в TypeOfService для RDMA траффика. Может быть использовано для ECN

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/verbs.cpp
+++ b/cloud/blockstore/libs/rdma/impl/verbs.cpp
@@ -227,13 +227,26 @@ struct TVerbs
     TConnectionPtr CreateConnection(
         rdma_event_channel* channel,
         void* context,
-        rdma_port_space ps) override
+        rdma_port_space ps,
+        ui8 tos) override
     {
         rdma_cm_id* id = nullptr;
 
         int res = rdma_create_id(channel, &id, context, ps);
         if (res < 0) {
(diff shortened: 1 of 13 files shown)


cloud/blockstore/libs/rdma/impl/verbs.cpp
```diff
@@ -236,6 +237,18 @@ struct TVerbs
             RDMA_THROW_ERROR("rdma_create_id");
         }
 
+        if (tos != 0) {
```
budevg, line 240: what about setting tos 0 or is is set by default ?
here it's documented that there are 4 priorities
https://docs.nvidia.com/networking/display/mlnxofedv496060lts/quality+of+service+(qos)#src-110005360_safe-id-UXVhbGl0eW9mU2VydmljZShRb1MpLVJvQ0VRdWFsaXR5b2ZTZXJ2aWNlTWFwcGluZw

komarevtsev-d: Zero is the default value as i understand it.
I have a dump of RDMA traffic (via tcpdump) without setting the ToS:
Screenshot 2025-08-05 at 16 16 07
First 6 bits of `Traffic Class` is ToS and it is zero.

budevg (approved):

PR #4076, opened 2025-08-06
[Blockstore] Add new parameter to corruption tests: path to file with list of compute nodes

gy2411:
Add new parameter to corruption tests: path to the file listing compute nodes. One of these nodes will be chosen to run the test on. We already have '--compute-node' parameter, but now we add ability to pass the list of nodes instead of single node.

To to this, we had to add the corresponding parameter to ycp client in blockstore/pylibs.

If the new parameter is not specified, behaviour will not change.


cloud/blockstore/pylibs/ycp/ycp.py
```diff
@@ -393,6 +396,7 @@ def create_instance(self, config: CreateInstanceConfig) -> Instance:
                 subnet_id=subnet_id,
                 folder_id=config.folder_id,
                 compute_node=config.compute_node,
+                compute_nodes_list=config.compute_nodes_list,
```
gy2411, line 399: Рендеринг шаблона от добавения нового параметра не ломается, даже если шаблон не поменялся.

v131v (approved):

PR #4120, opened 2025-08-12
[Blockstore] add unit tests for TNvmeManager

sharpeye:


cloud/blockstore/tests/recipes/nvme-loop/__main__.py
```diff
@@ -0,0 +1,23 @@
...
+from library.python.testing.recipe import declare_recipe, set_env
+
+
+NVME_LOOP_DEVICES_FOLDER = "/opt/nvme-loop/devices"
+NVME_LOOP_LOCKS_FOLDER = "/tmp/nvme-loop/locks"
+
+
+def start(argv):
+
+    os.makedirs(NVME_LOOP_LOCKS_FOLDER, exist_ok=True)
```
budevg, line 12: let's create all dirs relative to test directory
``` python
    dev_dir = os.path.join(yatest_common.output_path(), "nvme-loop/devices")
    os.makedirs(working_dir, exist_ok=True)
    locks_dir = os.path.join(yatest_common.output_path(), "nvme-loop/locks")
    os.makedirs(locks_dir, exist_ok=True)
```

cloud/blockstore/libs/nvme/nvme_linux.cpp
```diff
@@ -81,13 +83,17 @@ void NVMeFormatImpl(
 
     memcpy(&cmd.cdw10, &format, sizeof(ui32));
 
-    int err = ioctl(device, NVME_IOCTL_ADMIN_CMD, &cmd);
+    const int r = ioctl(device, NVME_IOCTL_ADMIN_CMD, &cmd);
```
budevg: `const int r` -> `int ret` or `const int ret`

cloud/blockstore/libs/nvme/nvme_linux.cpp
```diff
@@ -129,6 +135,19 @@ TResultOrError<bool> IsRotational(TFileHandle& device)
     return val != 0;
 }
 
+TFileHandle OpenDevice(const TString& path, EOpenMode mode)
```
budevg: You can use `TFile` instead and it will throw exception if it's unable to open

cloud/blockstore/tests/recipes/nvme-loop/__main__.py
budevg: Actually I am not sure how I am expected to run this ut from my development vm ... it will run from github ci where you setup the  nvme-loop devices, but who will setup the nvme-loop devices inside development vm ?

Maybe you need to setup loop devices in the recipe or signal to skip the tests if they don't exist

cloud/blockstore/libs/nvme/nvme_linux.cpp
sharpeye: WIll do it in the another PR. For now I have reverted changes from this file.

cloud/blockstore/tests/recipes/nvme-loop/__main__.py
sharpeye: Folder `nvme-loop/devices` must by global.
For now, you have to setup nvme-loop by yourself (or use real nvme) and run nvme tests with `ya:manual` tag. In the next iteration I'll try to solve this issue.

komarevtsev-d (approved):

Issue #4208, opened 2025-08-24
[Filestore] increasing MaxHandlePerSessionCount in local fs config should increase the table on disk


budevg:
Right now `TPersistentTable` which was create with specific `initialRecordCount` will keep using it event if we change config to increase the `initialRecordCount`. 
This doesn't allow us to increase `MaxHandlePerSessionCount` and roll out it to all nodes since the old config will be used for all old session untill they reset.

We want to allow growing the `TPersistentTable` when the `initialRecordCount` increased

If we configure the `TPersistentTable` with smaller size we want to shrink only if after compaction of table all elements fit the new size

PR #4209, opened 2025-08-24
[Filestore] issue-4208: Allow to increase/decrease TPersistentTable size on disk if possible

budevg:
#4208

This will allow us to increase MaxHandlePerSessionCount in config and make
table on disk larger. It also allows to decrease MaxHandlePerSessionCount in
config and make the table on disk shorter if all elements fit the new size

### DIFF
--- a/cloud/storage/core/libs/common/persistent_table.h
+++ b/cloud/storage/core/libs/common/persistent_table.h
@@ -246,7 +246,7 @@ private:
             header->RecordSize,
             sizeof(TRecord));
 
-        const auto initalRecordCount = RecordCount;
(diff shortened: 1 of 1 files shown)


cloud/storage/core/libs/common/persistent_table.h
```diff
@@ -246,13 +246,26 @@ class TPersistentTable
             header->RecordSize,
             sizeof(TRecord));
 
-        RecordCount = header->RecordCount;
+        const auto initalRecordCount = RecordCount;
```
SvartMetal: in*i*tialRecordCount

neihar (approved):

PR #4286, opened 2025-09-08
fix flake8 error

yegorskii:


yegorskii:
https://github.com/ydb-platform/nbs/pull/4285

PR #4283, opened 2025-09-05
[Filestore] fix double-free while stopping filestore-vhost when 'Forget' request is in-flight

SvartMetal:
ASAN report:
```
[...]

PR #4313, opened 2025-09-10
[Filestore] fix linkage error after #4283

SvartMetal:
Error was introduced in #4283.
When building with `-D FORCE_STATIC_LINKING=yes` we get:
```
ld.lld: error: duplicate symbol: fuse_reply_none
>>> defined at fuse_lowlevel.c:304 (/home/svartmetal/nbs-github2/contrib/libs/virtiofsd/fuse_lowlevel.c:304)
>>>            fuse_lowlevel.c.o:(fuse_reply_none) in archive contrib/libs/virtiofsd/static/libvirtiofsd.a
>>> defined at fuse_virtio.c:373 (/home/svartmetal/nbs-github2/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:373)
>>>            fuse_virtio.c.o:(.text.fuse_reply_none+0x0) in archive cloud/filestore/libs/vfs_fuse/vhost/liblibs-vfs_fuse-vhost.a
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fuse.h
+++ b/cloud/filestore/libs/vfs_fuse/fuse.h
@@ -37,7 +37,10 @@ int fuse_cancel_request(
     fuse_req_t req,
     enum fuse_cancelation_code code);
 
-// 'overrides' fuse_reply_none
+// 'overrides' fuse_reply_none, needed for VIRTIO-specific request completion
+// handling.
+// See https://github.com/ydb-platform/nbs/pull/4283
+// and https://github.com/ydb-platform/nbs/pull/4313
 void fuse_reply_none_override(fuse_req_t req);
 
 #if defined(__cplusplus)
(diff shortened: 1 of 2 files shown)


cloud/filestore/libs/vfs_fuse/fuse.cpp
```diff
@@ -16,3 +16,8 @@ int fuse_cancel_request(
     (void)code;
     return fuse_reply_err(req, EINTR);
 }
+
+void fuse_reply_none_override(fuse_req_t req)
```
budevg, line 20: since it became virtio specific function lets' just call it `virtio_reply_none`

SvartMetal: I prefer to keep the current naming because virtio implementation is switchable using compile-time flag `FUSE_VIRTIO`, we have classic FUSE implementation when this flag is off

SvartMetal: Added comment in code

budevg (approved):

PR #4320, opened 2025-09-11
[Blockstore] convert E_BS_INVALID_SESSION to E_BS_MOUNT_CONFLICT for NRD/Mirror when DevicesAcquire request fails

sharpeye:
E_BS_MOUNT_CONFLICT is more suitable than E_BS_INVALID_SESSION when a new user attempts to mount a disk that has not been released by the previous user for some reason (e.g., due to a delayed Unpublish request from the CSI).

### DIFF
--- a/cloud/blockstore/libs/storage/disk_agent/model/device_client_ut.cpp
+++ b/cloud/blockstore/libs/storage/disk_agent/model/device_client_ut.cpp
@@ -229,7 +229,7 @@ Y_UNIT_TEST_SUITE(TDeviceClientTest)
             TAcquireParamsBuilder()
                 .SetUuids({"uuid2"})
                 .SetClientId("another_client"));
-        UNIT_ASSERT_VALUES_EQUAL(E_BS_INVALID_SESSION, error.GetCode());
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/volume/volume_actor_addclient.cpp
```diff
@@ -289,13 +289,17 @@ void TVolumeActor::HandleAcquireDiskResponse(
...
 }
 
 void TVolumeActor::HandleDevicesAcquireFinishedImpl(
-    const NProto::TError& error,
-    const NActors::TActorContext& ctx)
+    const NActors::TActorContext& ctx,
+    NProto::TError error)
 {
+    if (error.GetCode() == E_BS_INVALID_SESSION) {
+        error.SetCode(E_BS_MOUNT_CONFLICT);
```
SvartMetal: поменяем тут на E_BS_MOUNT_CONFLICT? 

https://github.com/ydb-platform/nbs/blob/100ed4d62092edaff933c6b31e40a019a43514b0/cloud/blockstore/libs/storage/disk_agent/model/device_client.cpp#L145

PR #4376, opened 2025-09-18
release fresh barrier in add blobs not at flush completion

yegorskii:
This is a follow up for https://github.com/ydb-platform/nbs/pull/4266

Currently, fresh blob barriers are released at flush completion. This is fine, but we calculate TrimFreshLogToCommitId for saving in meta during AddBlobs handling. This means that TrimFreshLogToCommitId is lower than the commit ID of the recently flushed data. If the tablet restarts before trimming the fresh log, it will reload all the fresh blobs from the recent flush again.

To solve this inefficiency, we can release fresh blob barriers during the execute phase of the AddBlobs transaction. This way, the TrimFreshLogToCommitId saved in meta will match the volatile TrimFreshLogToCommitId in state

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
@@ -51,46 +51,6 @@ bool HasDuplicates(const TVector<ui32>& items)
 
 ////////////////////////////////////////////////////////////////////////////////
 
-TFlushedCommitIds BuildFlushedCommitIdsFromChannel(
-    const TVector<TAddFreshBlob>& blobs)
-{
-    TFlushedCommitIds result;
-    TVector<ui64> commitIds;
-
-    for (const auto& blob: blobs) {
-        for (const auto& block: blob.Blocks) {
-            if (!block.IsStoredInDb) {
-                commitIds.push_back(block.CommitId);
-            }
-        }
-    }
-
-    if (!commitIds) {
-        return {};
-    }
-
-    Sort(commitIds);
-
-    ui64 cur = commitIds.front();
-    ui32 cnt = 0;
-
-    for (const auto commitId: commitIds) {
-        if (commitId == cur) {
-            ++cnt;
-        } else {
-            result.emplace_back(cur, cnt);
-            cur = commitId;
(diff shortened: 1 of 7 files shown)


agalibin (dismissed):

agalibin (approved):

cloud/blockstore/libs/storage/partition/part_tx.cpp
```diff
@@ -0,0 +1,41 @@
+#include "part_tx.h"
+
+namespace NCloud::NBlockStore::NStorage::NPartition {
+
+////////////////////////////////////////////////////////////////////////////////
+
+void TTxPartition::TAddBlobs::BuildFlushedCommitIdsFromChannel()
```
qkrorlqr, line 7: let's put this func somewhere else - somewhere in partition/model? we'll be able to write unit tests for this function then

cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -156,6 +156,12 @@ class TAddBlobsExecutor
             UpdateCompactionCounters(blob);
         }
 
+        for (const auto& i: Args.FlushedCommitIdsFromChannel) {
+            State.GetTrimFreshLogBarriers().ReleaseBarrierN(
```
qkrorlqr, line 160: We have a guarantee that fresh blobs are garbage collected only up to `TrimFreshLogToCommitId` and that we never ever have concurrently running `Flush` and `TrimFreshLog` ops? Otherwise we can have the following race condition:
1. `AddBlobs` executes - changes `TrimFreshLogToCommitId`
2. concurrently running `TrimFreshLog` picks up the new `TrimFreshLogToCommitId` value and deletes fresh blobs up to this commit id
3. `AddBlobs` fails to commit => we've just lost our fresh blobs

qkrorlqr: My concern is that GC runs "outside" of the transaction mechanism - so GC by itself is able to run concurrently with transactions, there's no global order

SvartMetal:
@yegorskii is it still actual?

Issue #3738, opened 2025-06-23
[DM] Enhance estimated times system by tracking spent time in statuses


iamnalinor:
# Introduction

<details>

## How task statuses work

<details>
<summary>Workflow diagram</summary>
<img src="https://github.com/user-attachments/assets/9151e9a3-7d71-4508-93f3-125b97194011" />
</details>

There are 2 types of runners:

1.  Regular: This runner picks up tasks in the `ready_to_run` (and `ready_to_cancel`) status. It begins execution, transitioning the task's status to `running` or `cancelling` respectively.

2.  Stalker: This runner handles tasks that appear to be stalled. It picks up tasks that are in the `running` (or `cancelling`) state if the time since their last update (`modified_at`) exceeds the `TaskStallingTimeout` (currently 10 seconds). Then, Stalker re-executes the task.

A task can be picked up by Stalker in two cases:

- The task genuinely stalled
- The task interrupted its execution by returning `InterruptExecutionError` (e.g. [pools.RebaseOverlayDisk](https://github.com/ydb-platform/nbs/blob/5fffc0f/cloud/disk_manager/internal/pkg/services/pools/rebase_overlay_disk_task.go#L78)). This is asynchronous waiting: the task waits for some condition by restarting itself every 10 seconds. The task remains in `running` status.

## Types of waiting

There are 2 types of waiting: synchronous and asynchronous. Key differences:

- Synchronous: the task is waiting for another task to complete. The dependent changes its state from `running` to `waiting_to_run` and, when the dependency is resolved, to `ready_to_run`.
- Asynchronous: the task is waiting for a certain condition by checking it every 10 seconds. This is implemented via stalking runner mechanism: if the condition is not met, we interrupt the task execution, so the stalker can pick it up 10 seconds later.

## Estimated times

`tasks` table has `EstimatedTime` field. An estimated time is a moment in the future by which the task should be finished (statuses `finished` or `cancelled`), otherwise it will be marked as hanging. The presense of hanging tasks, except for excluded task types, will make HangingTasks alert fire.

Default estimated time is +1h from task creation time. Some tasks have custom estimated times calculated by the size of the work, such as disk size, image size, bytes-to-copy count, and so on.

</details>

# The problem

Currently, estimated times include the time a task has spent in all statuses — not only `running`, but also in asynchronous and synchronous waiting. This can lead to problems in cases when the task itself finishes quickly, but waits for another task to finish. 

[...]

PR #4384, opened 2025-09-19
[DM] Add Size() method to Source interface and use it for estimating

iamnalinor:
Part of #3738 

Add `(Source).Size(ctx context.Context)` method that represents the size of the data would be read.

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/common/transfer.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/common/transfer.go
@@ -52,7 +52,11 @@ type Source interface {
 	baseSource
 
 	Read(ctx context.Context, chunk *Chunk) error
+	// Represents total chunk count in the source.
 	ChunkCount(ctx context.Context) (uint32, error)
+	// Represents the actual data size to read. May be less
+	// than ChunkCount * chunkSize. Useful for estimating.
+	Size(ctx context.Context) (uint64, error)
 }
 
 type ShallowSource interface {
--- a/cloud/disk_manager/internal/pkg/dataplane/common/transfer_test.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/common/transfer_test.go
@@ -52,6 +52,11 @@ func (m *sourceMock) ChunkCount(ctx context.Context) (uint32, error) {
 	return args.Get(0).(uint32), args.Error(1)
 }
 
+func (m *sourceMock) Size(ctx context.Context) (uint64, error) {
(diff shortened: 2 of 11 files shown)


cloud/disk_manager/internal/pkg/dataplane/common/transfer.go
```diff
@@ -52,7 +52,11 @@ type Source interface {
 	baseSource
 
 	Read(ctx context.Context, chunk *Chunk) error
+	// Represents total chunk count in the source.
 	ChunkCount(ctx context.Context) (uint32, error)
+	// Represents the actual data size to read. May be less
```
jkuradobery: Use Returns instead of represents. Represents is a weak wording. 
Returns data size to read, might be less than ChunkCount * chunkSize due to ...

cloud/disk_manager/internal/pkg/dataplane/common/transfer.go
```diff
@@ -52,7 +52,11 @@ type Source interface {
...
+	// than ChunkCount * chunkSize. Useful for estimating.
```
jkuradobery: Useful is not the correct word here. Use a full sentence, like:
Used to calculate the estimated transfer time. 
I am not insisting on the exact wording here, try using AI for better wording maybe.

cloud/disk_manager/internal/pkg/dataplane/common/transfer.go
iamnalinor: Поправил коммент b6917fece2b0e665e81b089a6d995e7f202004c9

cloud/disk_manager/internal/pkg/dataplane/common/transfer.go
```diff
@@ -52,7 +52,11 @@ type Source interface {
 	baseSource
 
 	Read(ctx context.Context, chunk *Chunk) error
+	// Returns total chunk count in the source.
 	ChunkCount(ctx context.Context) (uint32, error)
+	// Returns data size to read. Might be less than ChunkCount * chunkSize.
+	// Used to calculate the estimated transfer time.
+	Size(ctx context.Context) (uint64, error)
```
SvartMetal: непонятное название

можно `EstimatedBytesToTransfer`, например

iamnalinor: Поменял на EstimatedBytesToRead() 6b804a0

SvartMetal (approved):

PR #4391, opened 2025-09-22
dump more information about unhandled exceptions

yegorskii:
Sometimes we see the following messages in our integration tests where is not completely clear where in the code such exception was raised
```
uncaught exception:
    address -> 0x51a00000be90
    what() -> "library/cpp/json/writer/json_value.cpp:457: Not a string"
    type -> NJson::TJsonException
```
There is a library located in library/cpp/terminated_handler which allows to collect stack trace during unwinding and dump this information upon call to std::terminate.

So far library is enabled for non-release builds since we see such crashes usually in integration tests

### DIFF
--- a/cloud/blockstore/apps/disk_agent/ya.make
+++ b/cloud/blockstore/apps/disk_agent/ya.make
@@ -21,6 +21,10 @@ PEERDIR(
 
 IF (BUILD_TYPE != "PROFILE" AND BUILD_TYPE != "DEBUG" AND BUILD_TYPE != "RELWITHDEBINFO")
     SPLIT_DWARF()
+ELSE()
+    PEERDIR(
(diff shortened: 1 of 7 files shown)


cloud/blockstore/apps/disk_agent/ya.make
```diff
@@ -21,6 +21,10 @@ PEERDIR(
 
 IF (BUILD_TYPE != "PROFILE" AND BUILD_TYPE != "DEBUG" AND BUILD_TYPE != "RELWITHDEBINFO")
     SPLIT_DWARF()
+ELSE()
+    PEERDIR(
+        library/cpp/terminate_handler
+    )
```
SvartMetal, line 27: а почему безусловно не включить в PEERDIR выше?

PR #4417, opened 2025-09-26
Fix test name in make

agalibin:


cloud/filestore/tests/fio/ya.make
```diff
@@ -5,7 +5,7 @@ RECURSE_FOR_TESTS(
     qemu-kikimr-newfeatures-test
     qemu-kikimr-test
     qemu-kikimr-zero-copy-write-test
-    qemu-kikimr-zero-copy-write-fallback-test
```
neihar, line 8: why not fixing it vice versa in https://github.com/ydb-platform/nbs/blob/e635484791fb1bf335b46ece65b7eed71a111a71/cloud/filestore/tests/fio/qemu-kikimr-zero-copy-write-test-fallback/ya.make#L20 ?

It looks more consistent to end with ‘-test.’

agalibin: if @antonmyagkov does not have any objections, I could change test name.

drbasic (approved):

cloud/filestore/tests/fio/ya.make
SvartMetal: `qemu-kikimr-zero-copy-write-fallback-test` - looks like better name

drbasic (approved):

PR #4434, opened 2025-09-30
users/evgenybud/fs expiration time metrics

budevg:
- **enable io uring kernel polling**
- **[CSI] issue-4433: report fs mount expiration times in monitoring metrics**
- **fix spelling mistake**


cloud/blockstore/tools/csi_driver/internal/driver/external_fs.go
```diff
@@ -45,3 +47,23 @@ func LoadExternalFsOverrides(filePath string) (ExternalFsOverrideMap, error) {
 
 	return overrideMap, nil
 }
+
+func (om ExternalFsOverrideMap) GetMountExpirationTimes() map[string]int64 {
```
antonmyagkov, line 51: lets replace 'om' with meaningful name. 'externalFsOverrideMap' is fine for me

budevg: receiver names in go are usually one or two letters by convention
https://go.dev/wiki/CodeReviewComments#receiver-names

I used `om` as shortcut for override map let's keep with golang convention for this

neihar (approved):

PR #4470, opened 2025-10-07
[Disk Manager] Fix TestPrivateServiceOptimizeBaseDisks test

jkuradobery:
If some AcquireBaseDisks tasks work slowly (they need to create a new disk, it might be slow), and the first image is quickly created - the first task can expire and be removed and re-accessed afterwards, which causes an error.

### DIFF
--- a/cloud/disk_manager/internal/pkg/facade/private_service_test/private_service_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/private_service_test/private_service_test.go
@@ -353,8 +353,8 @@ func TestPrivateServiceOptimizeBaseDisks(t *testing.T) {
 
 	// Should wait for first disk creation in order to ensure that pool is
 	// created.
-	firstCreateDiskOperationId := operations[0].Id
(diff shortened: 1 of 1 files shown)


BarkovBG (approved):

cloud/disk_manager/internal/pkg/facade/private_service_test/private_service_test.go
```diff
@@ -351,7 +353,8 @@ func TestPrivateServiceOptimizeBaseDisks(t *testing.T) {
 
 	// Should wait for first disk creation in order to ensure that pool is
 	// created.
-	err = internal_client.WaitOperation(ctx, client, operations[0].Id)
+	firstCreateDiskOperationId := operations[0].Id
```
SvartMetal: firstCreateDiskOperationID

у правой части линтер будет ругаться, да?

@BarkovBG

jkuradobery: Логично, забыл.

PR #4471, opened 2025-10-08
[Filestore] add sync ping  before async request

neihar:
Test fails periodically:

https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/Nightly-build-(asan)/18215119928/1/nebius-x86-64-asan/logs/1/cloud/filestore/libs/server/ut/test-results/unittest/chunk2/testing_out_stuff/TServerTest.ShouldGetSessionEventsStream.err

### DIFF
--- a/cloud/filestore/libs/server/server_ut.cpp
+++ b/cloud/filestore/libs/server/server_ut.cpp
@@ -439,17 +439,31 @@ Y_UNIT_TEST_SUITE(TServerTest)
 
     Y_UNIT_TEST(ShouldGetSessionEventsStream)
     {
-        TAtomicStorage<IResponseHandlerPtr<NProto::TGetSessionEventsResponse>> storage;
+        TAtomicStorage<IResponseHandlerPtr<NProto::TGetSessionEventsResponse>>
+            storage;
 
         TBootstrap<TServerSetup> bootstrap;
         bootstrap.Service->GetSessionEventsStreamHandler =
-            [&] (auto callContext, auto request, auto responseHandler) {
-                Y_UNUSED(callContext, request);
-                storage.Set(responseHandler);
-            };
+            [&](auto callContext, auto request, auto responseHandler)
+        {
+            Y_UNUSED(callContext, request);
+            storage.Set(responseHandler);
+        };
+
+        bootstrap.Service->PingHandler = [](auto, auto)
+        {
+            return MakeFuture<NProto::TPingResponse>();
+        };
 
         bootstrap.CreateClient();
         bootstrap.Start();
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/server/server_ut.cpp
```diff
@@ -439,18 +439,32 @@ Y_UNIT_TEST_SUITE(TServerTest)
...
         bootstrap.Service->GetSessionEventsStreamHandler =
-            [&] (auto callContext, auto request, auto responseHandler) {
-                Y_UNUSED(callContext, request);
-                storage.Set(responseHandler);
-            };
+            [&](auto callContext, auto request, auto responseHandler)
+        {
+            Y_UNUSED(callContext, request);
+            storage.Set(responseHandler);
+        };
```
ruslan-gar, line 451: The code in lines 442, 447-451 has not actually changed. We usually try not to change formatting only

ruslan-gar (changes requested):

cloud/filestore/libs/server/server_ut.cpp
neihar: Yes, I applied formatting to the entire test to make gradual progress toward our clang-format style and avoid manually maintaining all the spacing. I think it is okay to do it that way.

ruslan-gar: In this case changes are minor, but, as I understand, our policy is not change formatting only

ruslan-gar (approved):

cloud/filestore/libs/server/server_ut.cpp
neihar: From my point of view, we should not change just formatting, but with reasonable scope around changed code it is okay.

cloud/filestore/libs/server/server_ut.cpp
```diff
@@ -439,18 +439,32 @@ Y_UNIT_TEST_SUITE(TServerTest)
 
     Y_UNIT_TEST(ShouldGetSessionEventsStream)
```
SvartMetal, line 440: в соседних тестах нет такой же проблемы?

тут упал не только этот тест https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/Nightly-build-(asan)/18336285397/1/nebius-x86-64-asan/summary/1/ya-test.html#FAIL

neihar: We talked about it in the meeting; we’ll take care of those tests separately.

Issue #4495, opened 2025-10-13
[NBS] lwtrace Fork/HasShuttles data race


tpashkin:
https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/PR-check/18383804197/1/nebius-x86-64-tsan/summary/1/ya-test.html#FAIL
```
2025-10-09 18:33:46,420 - DEBUG - ya.test - process_recipe_err_files: Processing recipe error files: ['/home/github/.ya/build/build_root/r3xz/002e19/cloud/filestore/tests/loadtest/service-kikimr-newfeatures-test/test-results/py3test/chunk2/testing_out_stuff/filestore-server.err']
2025-10-09 18:33:46,428 - ERROR - ya.test - process_recipe_err_files: b'ThreadSanitizer' sanitizer found errors:
[...]

Issue #2289, opened 2024-10-16
[Filestore] race in TIndexTabletActor LWTrace usage


qkrorlqr:
https://github-actions-s3.website.nemax.nebius.cloud/ydb-platform/nbs/PR-check/11254462513/1/nebius-x86-64-tsan/test_data/actions-runner/_work/nbs/nbs/cloud/filestore/tests/fio_index/qemu-kikimr-multishard-nemesis-test/test-results/py3test/chunk3/testing_out_stuff/filestore-server.err

```
WARNING: ThreadSanitizer: data race (pid=253865)
  Write of size 8 at 0x7b2400024270 by thread T56:
    #0 NSwapCheck::TSwapSelector<NLWTrace::IShuttle *, void>::Swap /actions-runner/_work/nbs/nbs/util/generic/utility.h:80:15 (filestore-server+0x1eb04c9a) (BuildId: 014f78b3943678e6d2427c5e81f15e8a61c8038e)
    #1 DoSwap<NLWTrace::IShuttle *> /actions-runner/_work/nbs/nbs/util/generic/utility.h:104:5 (filestore-server+0x1eb04c9a)
    #2 TIntrusivePtr<NLWTrace::IShuttle, TDefaultIntrusivePtrOps<NLWTrace::IShuttle> >::Swap /actions-runner/_work/nbs/nbs/util/generic/ptr.h:564:9 (filestore-server+0x1eb04c9a)
    #3 NLWTrace::IShuttle::Detach /actions-runner/_work/nbs/nbs/library/cpp/lwtrace/shuttle.h:122:20 (filestore-server+0x1eb04c9a)
    #4 NLWTrace::IShuttle::Drop /actions-runner/_work/nbs/nbs/library/cpp/lwtrace/shuttle.h:117:20 (filestore-server+0x1eb04c9a)
    #5 NLWTrace::TOrbit::ForEachShuttle(unsigned long, (lambda at /actions-runner/_work/nbs/nbs/library/cpp/lwtrace/shuttle.h:259:38) &&)::(anonymous class)::operator() /actions-runner/_work/nbs/nbs/library/cpp/lwtrace/shuttle.h:250:35 (filestore-server+0x1eb04c9a)
[...]

PR #4518, opened 2025-10-19
fix data race in LWTrace

neihar:
https://github.com/ydb-platform/nbs/issues/2289
https://github.com/ydb-platform/nbs/issues/4495 (duplicate)

### DIFF
--- a/library/cpp/lwtrace/shuttle_race_ut.cpp
+++ b/library/cpp/lwtrace/shuttle_race_ut.cpp
@@ -44,10 +44,10 @@ protected:
     }
 };
 
-Y_UNIT_TEST_SUITE(TOrbitThreadSafety)
+Y_UNIT_TEST_SUITE(TOrbitMultithreadedUsage)
 {
-    // Test concurrent HasShuttles() calls while modifying orbit
-    Y_UNIT_TEST(ConcurrentHasShuttlesAndAddShuttle)
+    // Test HasShuttles() calls while modifying orbit
+    Y_UNIT_TEST(HasShuttlesAndAddShuttle)
     {
         TOrbit orbit;
         std::atomic<bool> stopFlag{false};
@@ -98,8 +98,9 @@ Y_UNIT_TEST_SUITE(TOrbitThreadSafety)
         UNIT_ASSERT(orbit.HasShuttles());
     }
 
-    // Test the race condition from the tests crash: Fork() vs HasShuttles()
-    Y_UNIT_TEST(ConcurrentForkAndHasShuttles)
+    // Test the race condition from the tsan tests crash: Fork() vs
+    // HasShuttles()
+    Y_UNIT_TEST(ForkAndHasShuttles)
     {
         TOrbit orbit;
 
@@ -156,7 +157,7 @@ Y_UNIT_TEST_SUITE(TOrbitThreadSafety)
     }
 
     // Test the Serialize() race condition
-    Y_UNIT_TEST(ConcurrentSerializeAndHasShuttles)
+    Y_UNIT_TEST(SerializeAndHasShuttles)
     {
         TOrbit orbit;
 
--- a/library/cpp/lwtrace/ut/ya.make
+++ b/library/cpp/lwtrace/ut/ya.make
@@ -2,9 +2,14 @@ UNITTEST_FOR(library/cpp/lwtrace)
 
 FORK_SUBTESTS()
 
+IF (SANITIZER_TYPE == "thread")
+    SRCS(
+        shuttle_race_ut.cpp
+    )
+ENDIF()
+
 SRCS(
     log_ut.cpp
-    shuttle_race_ut.cpp
     trace_ut.cpp
 )
 


library/cpp/lwtrace/shuttle.h
```diff
@@ -198,9 +198,19 @@ namespace NLWTrace {
...
         // Checks if there is at least one shuttle in orbit
-        // NOTE: used by every LWTRACK macro check, so keep it optimized - do not lock
-        bool HasShuttles() const {
-            return HeadNoLock.Get();
+        // Uses atomic read to synchronize with NotConcurrent() and avoid data
+        // races, still optimized and no lock
+        bool HasShuttles() const
+        {
+            static_assert(sizeof(HeadNoLock) == sizeof(TAtomic));
+            const TAtomic* headPtr =
```
tpashkin, line 206: There are no direct uses of `HeadNoLock` left, so these casts are pretty pointless

library/cpp/lwtrace/shuttle.h
```diff
@@ -198,9 +198,19 @@ namespace NLWTrace {
         }
 
         // Checks if there is at least one shuttle in orbit
-        // NOTE: used by every LWTRACK macro check, so keep it optimized - do not lock
-        bool HasShuttles() const {
-            return HeadNoLock.Get();
+        // Uses atomic read to synchronize with NotConcurrent() and avoid data
+        // races, still optimized and no lock
```
tpashkin, line 202: I'm not sure I agree with this change. This definitely fixes the "race" reported by tsan, but linked issues actually look like false positives – as far as I can tell, we never really access the orbit concurrently. To me this looks more like a matter of strategically placing memory barriers when tsan is enabled

Besides, `TOrbit` was specifically made to be thread unsafe to avoid costs associated with atomic operations and make tracing as fast as possible. If we make this change, I'd like to see some benchmarks showing that the difference is negligible

library/cpp/lwtrace/shuttle_race_ut.cpp
```diff
@@ -0,0 +1,282 @@
...
+        return true;
+    }
+
+    bool DoJoin(const TShuttlePtr&) override
+    {
+        return true;
+    }
+};
+
+Y_UNIT_TEST_SUITE(TOrbitThreadSafety)
```
tpashkin: `TOrbit` was intentionally made unsafe to make it's contribution to measured latency as low as possible

library/cpp/lwtrace/shuttle.h
neihar: As I see it, our general usage pattern is:
```
LWTRACK(start)
process something
LWTRACK(end)
```
Given this, atomic reads for Intel/AMD shouldn't add much beyond preventing reordering and optimizations (cause reading is atomically from HW), which shouldn't affect our use cases. For ARM, we might lose a few cycles, but that shouldn't be critical either.So unless you see any holes in my logic, I don't see why we should leave undefined behavior in place for a questionable performance gain. Yes, we're already reading 64 bits atomically at the hardware level, but we're not protected from register caching and compiler optimizations. While this may not be an issue in most of our cases, it's still the right call to make the compiler aware of what we're doing.

Regarding benchmarks, I have a strong opinion that we can proceed without them and instead monitor our most latency-sensitive and load-critical metrics after deployment. In the worst-case scenario, we might see a single-digit percentage increase in collected latencies (though I believe we won't).

tpashkin: > I don't see why we should leave undefined behavior in place for a questionable performance gain

You are missing the main point I've made – there are no concurrent operations going on, these operations are sequentially consistent. The pattern is simple – one actor receives a message, adds a probe and starts another actor, which continues to add probes to this orbit, there is an implicit barrier between passing execution between actors

I do not object in making this change in principle, but this will make our fork incompatible with the original yandex version and make it hard to pull fixes from it

> Regarding benchmarks, I have a strong opinion that we can proceed without them and instead monitor our most latency-sensitive and load-critical metrics after deployment

This will show nothing, because tracing is a special case – right now it's done selectively, not in the background

library/cpp/lwtrace/shuttle_race_ut.cpp
```diff
@@ -0,0 +1,282 @@
...
+        // probability
+        std::atomic<size_t> forkCount{0};
+        std::atomic<size_t> checkCount{0};
+
+        // Thread 1: Continuously calls Fork()
+        auto forker = [&]()
+        {
+            for (size_t i = 0; i < numForks; ++i) {
+                TOrbit tempOrbit;
+                if (orbit.Fork(tempOrbit)) {
```
tpashkin, line 127: This doesn't look right. Even if the Fork fails due to concurrent use, it will still return 0

See, that's exactly the problem I was talking about. Making `HasShuttles` atomic doesn't make `TOrbit` thread safe in any sensible manner, just places a memory fence which makes tsan happy

If anything, I'd like to see access pattern similar to the linked issues – multithreaded, but not concurrent

library/cpp/lwtrace/shuttle.h
tpashkin: > I do not object in making this change in principle, but this will make our fork incompatible with the original yandex version and make it hard to pull fixes from it

To be precise, your changes in `HasShuttles` are very small, so I guess we can live with that (even though I'd like to see it written more concisely – without casts, etc)

But the manner in which tests are written implies `TOrbit` has become thread safe, which it has not

neihar: Do we have 100% guarantee that we do not use LWTRACK on the same orbit from different threads? (around TCallContext).

So initial intention was to make HasShuttle quick. And I think it was wrongly assumed that atomic read will harm perfomance. So, yes, now it is thread safe to call HasShuttle, but if we don't have races with this check it will will zero cost. 

Can't we propagate our fix to the original lwtrace?

library/cpp/lwtrace/shuttle.h
```diff
@@ -198,9 +198,19 @@ namespace NLWTrace {
...
+        bool HasShuttles() const
```
tpashkin, line 203: We agreed to keep these changes to a minimum, so it would be easier to pull to and from other forks

tpashkin: @yegorskii FYI

PR #4531, opened 2025-10-21
be a little more verbose on rdma connection errors

yegorskii:


tpashkin (dismissed):

cloud/blockstore/libs/rdma/impl/verbs.cpp
```diff
@@ -193,7 +193,11 @@ struct TVerbs
             hints,
             &addr);
         if (res < 0) {
-            RDMA_THROW_ERROR("rdma_getaddrinfo");
+            RDMA_THROW_ERROR(
+                TStringBuilder()
+                    << "rdma_getaddrinfo("
+                    << "host='" << host
```
tpashkin: nit: we can use `TString::Quote` here

Issue #4541, opened 2025-10-23
[Filestore] support char/block devices in filestore (both ydb-based and local)


debnatkh:
Similar to this issue: https://github.com/ydb-platform/nbs/issues/4354

Aspirisha:
With kikimr-backed filestore changes applied, we have following results for this setup (again, used commit is [8a2adf0824e4f5c48cc27d2fa908cef2588391c3](https://github.com/pjd/pjdfstest/commit/8a2adf0824e4f5c48cc27d2fa908cef2588391c3))

```
$ cat results | grep -oE '^(not )?ok [0-9]+' | sed 's/[0-9]\+$//' | sort | uniq -c
     70 not ok 
   8719 ok 
```

Among these 70 failures:

* 41 are because we do not correctly process `[l]chown -1 -1`
* 8 tests are from tests/chown/00.t: these are all tests having todo comment `# TODO Linux doesn't clear the SGID/SUID bits for directories, despite the description noted`, like the [one](https://github.com/pjd/pjdfstest/blob/8a2adf0824e4f5c48cc27d2fa908cef2588391c3/tests/chown/00.t#L234) for instance
* 4 tests fail in tests/rename/24.t: [these two](https://github.com/pjd/pjdfstest/blob/8a2adf0824e4f5c48cc27d2fa908cef2588391c3/tests/rename/24.t#L31-L32) and [these two](https://github.com/pjd/pjdfstest/blob/8a2adf0824e4f5c48cc27d2fa908cef2588391c3/tests/rename/24.t#L51-L52)
* 1 test fails in tests/unlink/08.t : [the second one](http://github.com/pjd/pjdfstest/blob/master/tests/rename/24.t#L31-L32) 
* 2 tests from tests/utimensat/08.t: [these two](https://github.com/pjd/pjdfstest/blob/8a2adf0824e4f5c48cc27d2fa908cef2588391c3/tests/utimensat/08.t#L29-L30)
* 2 tests from tests/rmdir/00.t: [these](https://github.com/pjd/pjdfstest/blob/8a2adf0824e4f5c48cc27d2fa908cef2588391c3/tests/rmdir/00.t#L26-L28) two: ctime/mtime of a directory not updated after rmdir called on nested directory
[...]

PR #4590, opened 2025-10-31
issue-4541: [Filestore] support for mknod block and character devices in local filestore

Aspirisha:
The majority of `mknod` tests were not affected and still fail, because the test environment is prepared with the filestore daemon running without root privileges, while `mknod` with special devices requires those and otherwise fails with `EPERM`. 
However, there is a test that started to pass, although the reasons are not that obvious. Namely, the following setups:
* `fs_posix_compliance/mount-local-test`
* `fs_posix_compliance/qemu-local-test`
* `fs_posix_compliance/qemu-local-noserver-test`

got two tests from `mknod/08.t` passing now, specifically for `type={block, char}` the [line 20 here](https://github.com/ydb-platform/nbs/blob/77f157144aad1e2306bcdbe6e83eed923f7db049/cloud/filestore/tools/testing/fs_posix_compliance/suite/tests/mknod/08.t#L20) now succeeds. This happens because even though generally `mknod` with character/block devices requires root privileges, there is one exemption which we face here: when both major and minor device versions are 0, we get so called [whiteout device](https://elixir.bootlin.com/linux/v6.1.56/source/include/linux/fs.h#L209-L214), for which a character special file may be created via mknod [without root privileges](https://elixir.bootlin.com/linux/v6.1.56/source/fs/namei.c#L3893). Because of that, [line 19](https://github.com/ydb-platform/nbs/blob/77f157144aad1e2306bcdbe6e83eed923f7db049/cloud/filestore/tools/testing/fs_posix_compliance/suite/tests/mknod/08.t#L19) actually creates a file, and line 20 receives the expected `EEXIST`.

#4541

### DIFF
--- a/cloud/filestore/libs/service_local/lowlevel.h
+++ b/cloud/filestore/libs/service_local/lowlevel.h
@@ -96,7 +96,9 @@ struct TFileSystemStat
 struct TFileStatEx : public TFileStat {
     using TFileStat::TFileStat;
     ui64 Dev = 0;
-    static_assert(sizeof(dev_t) <= sizeof(ui64), "dev_t is larger than 64 bits");
+    static_assert(
+        sizeof(dev_t) <= sizeof(ui64),
+        "dev_t is larger than 64 bits");
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -122,10 +124,16 @@ TFileHandle OpenAt(
 void MkDirAt(const TFileHandle& handle, const TString& name, int mode);
 void MkSockAt(const TFileHandle& handle, const TString& name, int mode);
 void MkFifoAt(const TFileHandle& handle, const TString& name, int mode);
-void MkCharDeviceAt(const TFileHandle& handle, const TString& name,
-    int mode, dev_t dev);
-void MkBlockDeviceAt(const TFileHandle& handle, const TString& name,
-    int mode, dev_t dev);
+void MkCharDeviceAt(
+    const TFileHandle& handle,
+    const TString& name,
+    int mode,
+    dev_t dev);
+void MkBlockDeviceAt(
+    const TFileHandle& handle,
+    const TString& name,
+    int mode,
+    dev_t dev);
 
 void RenameAt(
     const TFileHandle& handle,


cloud/filestore/libs/service_local/lowlevel.h
```diff
@@ -91,9 +91,14 @@ struct TFileSystemStat
     i64 MountFlags = 0;     // Mount flags of filesystem
 };
 
+struct TFileStatEx : public TFileStat {
```
debnatkh, line 96: Maybe some comment is needed here? like

`TFileStat does not have all the fields of `struct stat` so bla-bla-bla`?

cloud/filestore/libs/service_local/lowlevel.h
```diff
@@ -114,6 +119,8 @@ TFileHandle OpenAt(
 void MkDirAt(const TFileHandle& handle, const TString& name, int mode);
 void MkSockAt(const TFileHandle& handle, const TString& name, int mode);
 void MkFifoAt(const TFileHandle& handle, const TString& name, int mode);
+void MkCharDeviceAt(const TFileHandle& handle, const TString& name, int mode, dev_t dev);
```
debnatkh: limit is 80 chars

cloud/filestore/libs/service_local/lowlevel.h
```diff
@@ -91,9 +91,14 @@ struct TFileSystemStat
     i64 MountFlags = 0;     // Mount flags of filesystem
 };
 
+struct TFileStatEx : public TFileStat {
+    using TFileStat::TFileStat;
+    ui64 Dev = 0;
```
debnatkh, line 98: Probably we are gonna need compile-time `static_assert(sizeof(dev_t) * 8 <= 64, "dev_t is larger than 64 bits");`

Aspirisha: Not that we have similar checks for `mode_t`, but I agree it may never hurt to be more cautious.

cloud/filestore/libs/service_local/lowlevel.h
Aspirisha: From what I see in sources, it is not. Also we have `ColumnLimit: 0` set in clang format config.
Anyway, I will try to respect the 80-character limit rule, did not know about that.

cloud/filestore/libs/service_local/lowlevel.h
Aspirisha: Added small comment regarding this

cloud/filestore/libs/service_local/lowlevel.h
```diff
@@ -119,8 +122,10 @@ TFileHandle OpenAt(
 void MkDirAt(const TFileHandle& handle, const TString& name, int mode);
 void MkSockAt(const TFileHandle& handle, const TString& name, int mode);
 void MkFifoAt(const TFileHandle& handle, const TString& name, int mode);
-void MkCharDeviceAt(const TFileHandle& handle, const TString& name, int mode, dev_t dev);
-void MkBlockDeviceAt(const TFileHandle& handle, const TString& name, int mode, dev_t dev);
+void MkCharDeviceAt(const TFileHandle& handle, const TString& name,
```
debnatkh: formatting

cloud/filestore/libs/storage/tablet/tablet_actor_createnode.cpp
```diff
@@ -37,6 +37,10 @@ NProto::TError ValidateRequest(const NProto::TCreateNodeRequest& request)
         }
     }
 
+    if (request.HasBlockDevice() || request.HasCharDevice()) {
```
debnatkh, line 41: Add `TODO(#4541)`, please

cloud/filestore/public/api/protos/node.proto
```diff
@@ -67,6 +69,8 @@ message TNodeAttr
     string ShardFileSystemId = 11;
     // Node name in the shard FS.
     string ShardNodeName = 12;
+
+    uint64 DevId = 13;
```
Aspirisha, line 73: Probably should be `DevID` or even better just `Device`?

PR #4618, opened 2025-11-06
Move device operation id generation to TDeviceOperationTracker

drbasic:
continue https://github.com/ydb-platform/nbs/pull/4575

To reuse tracking in RDMA, a small refactoring - moved the generation of the operation identifier to the TDeviceOperationTracker class.


cloud/blockstore/libs/storage/core/device_operation_tracker.cpp
```diff
@@ -10,8 +10,15 @@
 
 namespace NCloud::NBlockStore::NStorage {
 
+
```
SvartMetal, line 13: лишний \n

PR #4655, opened 2025-11-11
Do less unnecessary work in ADD_HOST cms action

komarevtsev-d:
This is an attempt to optimize the ADD_HOST cms action.
1) Added an early return from `TDiskRegistryState::RegisterUnknownDevices` to avoid copying possibly huge `CurrentConfig`. 
2) `TDiskRegistryState::CalcConfigUpdateEffect` had a bug that `affectedAgents` were populated with every non-known agent. It was fixed, plus some minor optimizations were added.

### DIFF
--- a/cloud/blockstore/libs/storage/disk_registry/disk_registry_state.cpp
+++ b/cloud/blockstore/libs/storage/disk_registry/disk_registry_state.cpp
@@ -3673,8 +3673,8 @@ auto TDiskRegistryState::CalcConfigUpdateEffect(
         }
     }
 
-    THashSet<TString> allKnownDevices;
-    allKnownDevices.reserve(DeviceList.Size() * 2);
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/disk_registry/benchmark/ya.make
```diff
@@ -1,9 +1,5 @@
 G_BENCHMARK()
 
-IF (SANITIZER_TYPE)
-    TAG(ya:manual)
-ENDIF()
```
komarevtsev-d, line 5: ниже есть `INCLUDE(${ARCADIA_ROOT}/cloud/storage/core/tests/recipes/large.inc)` 
Это уже выставляет флаг manual

drbasic (approved):

PR #4676, opened 2025-11-17
TRequestCounters: add IsBasic flag + refactoring

e673:


e673:
Replaced with ModuleMetrics

Issue #4201, opened 2025-08-22
[Filestore] Turning on and off the write-back cache


e673:
https://github.com/ydb-platform/nbs/issues/1751

Implement functionality to turn on and off the write-back cache.
Turning off write-back cache should drain it:

* FlushAll is requested.
* Moving pending requests to the cache is prohibited.
* New WriteData requests become pending.
* Pending request wait until all related write requests are flushed, then they go directly to the session.
* After flushing all write requests, the file-backed buffer is reset.
 
Possible implementation:
```
// The cache is being used, all ReadData and WriteData should go through it
bool IsActive() const;

// Start using cache
// Inability to create a file-backed buffer is not a fatal error
// Alternative implementation: return bool and report an error by callback?
TError Activate();

// Stop using cache - it will remain active until the cache is drained
// The future is completed when IsActive is set to false
// It can be interrupted by Activate()
NThreading::TFuture<TError> DeactivateAsync();
```

PR #4702, opened 2025-11-21
issue-1751: [Filestore] WriteBackCache should flush everything before the session is destoyed

e673:
https://github.com/ydb-platform/nbs/issues/4201

Destroying session with unflushed write requests will cause losing the data.
We should flush everything.

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
@@ -2923,7 +2923,7 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
                     // as pending and will wait instead.
 
                     // Cache fullness can be detected by non-zero WriteData
-                    // attempts from Flush that is triggered when there is not
+                    // attempts from Flush that is triggered when there is no
                     // space to store the request
 
                     const bool nonZeroPendingIsExpected =


cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1244,6 +1175,120 @@ class TFileSystemLoop final
...
+    void OnStop(TPromise<void> stopCompleted)
+    {
+        if (WriteBackCache && !WriteBackCache.IsEmpty()) {
+            STORAGE_INFO(
+                "[f:%s][c:%s] WriteBackCache is not empty, starting "
+                "FlushAllData",
+                Config->GetFileSystemId().Quote().c_str(),
+                Config->GetClientId().Quote().c_str());
+
+            WriteBackCache.FlushAllData().Subscribe(
```
e673, line 1186: Q: Should I introduce a separate `EFileStoreRequest` for this and report it via RequestStats in order to track FlushAllData? May be `callContext` should be an argument for `FlushAllData`.

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1243,6 +1174,120 @@ class TFileSystemLoop final
...
+                [w = weak_from_this(),
+                 s = std::move(stopCompleted)](const TFuture<void>& f) mutable
+                {
+                    f.GetValue();
```
neihar, line 1190: can it be a failure(exception)?

e673: FlushAllData is not expected to throw an exception, we can safely make this call.
I don't like this code and I'd prefer `Y_UNUSED(f)` but this pattern is commonly used in other places.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,72 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
         TestShouldSupportZeroCopyWriteByWriteBackCache(false);
         TestShouldSupportZeroCopyWriteByWriteBackCache(true);
     }
+
+    Y_UNIT_TEST(ShouldFlushAllRequestsBeforeSessionIsDestroyed)
+    {
+        // The idea is to fill WriteBackCache with requests and to stop session.
+        // It should flush both cached and pending requests befor session is
```
qkrorlqr: nit: befor -> before, destoyed -> destroyed, temporary -> temporarily

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,72 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+        features.SetServerWriteBackCacheEnabled(true);
+
+        TBootstrap bootstrap(
+            CreateWallClockTimer(),
+            CreateScheduler(),
+            features);
+
+        bootstrap.Service->WriteDataHandler = [&](auto, auto)
+        {
+            bootstrap.Timer->Sleep(TDuration::MilliSeconds(10));
```
qkrorlqr: why do we need to sleep here? why not just wait on some future and then signal this future at some later point in the test code when we are ready to complete the requests?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,72 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+
+        auto path = TempDir.Path() / "WriteBackCache" / FileSystemId /
+                    SessionId / "write_back_cache";
+
+        UNIT_ASSERT(path.Exists());
+
+        auto stopFuture = bootstrap.StopAsync();
+        UNIT_ASSERT(stopFuture.Wait(Timeout));
+
+        UNIT_ASSERT(!path.Exists());
```
qkrorlqr, line 2954: a check which would verify that all write requests have been processed seems to be missing here

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: I already tried this.
For an unknown reason, a deadlock happened in this case (and it was not related to WriteBackCache).
I tried to investigate it but without success.

qkrorlqr: this `Sleep` looks like a hack (the same applies to most of the other `Sleep`s in the other uts in this file actually)

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: Fixed

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2859,19 +2869,35 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
             reqWrite->In->Body.flags |= O_WRONLY;
-            bootstrap.Fuse->SendRequest<TWriteRequest>(reqWrite);
-
-            totalByteCount += byteCount;
+            bootstrap.Fuse->SendRequest<TWriteRequest>(reqWrite).Subscribe(
+                [&](const auto&) { completedRequestCount++; });
         }
 
+        // The requests are sent to Fuse in a thread pool - there is no reliable
+        // way to ensure that the tasks are executed, so just wait a bit
```
qkrorlqr: Why no way? Just wait on some cond var here and notify that cond var in the callback where you increment `completedRequestCount`

Otherwise the test will be unstable

You can have a hard timeout based on the timer (e.g. 15s) - in order not to hang up in case of bugs, but otherwise the test shouldn't depend on any wall clock timers

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
qkrorlqr: Ah, I see that you've already fixes this, thanks

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: The check was added

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: > Just wait on some cond var here and notify that cond var in the callback where you increment completedRequestCount

The callback will not be called because the progression of WriteData requests is disabled at that moment.
And we enable it only after calling StopAsync.

The test depends on when the callbacks are called by ThreadPool.
If we call StopAsync too early, then some of the requests will be rejected and the test will hang/fail.

Here is the call chain:
https://github.com/ydb-platform/nbs/blob/3d34af010b700ab0be3ed1dd172d6aee19145952/cloud/filestore/libs/vhost/client.h#L44-L57
...
https://github.com/ydb-platform/nbs/blob/3d34af010b700ab0be3ed1dd172d6aee19145952/cloud/storage/core/libs/vhost-client/vhost-buffered-client.cpp#L97

There is no notifier that the request has been accepted — only the notifier that the request has been completed.

qkrorlqr: I see this code:
```
        bootstrap.Timer->Sleep(TDuration::Seconds(1));

        UNIT_ASSERT_LT_C(
            0,
            completedRequestCount.load(),
            "There should exist completed requests");
```
It means that we expect at least one request to complete after this `Sleep`. Which will be unstable and it's better to signal some condvar or even alongside `completedRequestCount++` and then wait on this condvar/event instead of sleeping.

> The test depends on when the callbacks are called by ThreadPool.

And if you really need to verify that `SendRequestImpl` has been called for all your requests (and there's no such verification in your code right now), then you can expose `RequestId` into the client's public interface and spin-wait until you see that `RequestId` has been incremented by the right number of requests. But again, your current unit test code doesn't seem to contain such verification.

cloud/filestore/libs/vfs_fuse/loop.cpp
neihar: As I understand we agreed to fix it in separate issue. In that case you can resolve this topic.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
qkrorlqr: Just to be clear: I'm not being nitpicky here - I've spent a lot of time myself debugging various unstable unit tests in filestore and reliance on `Sleep()`s has been one of the most common causes of instability.

e673: I've fixed the test - now it is free of timers etc.

Btw, there is a similar issue with the another test
`TFileSystemTest.ShouldNotCrashWhileStoppingWhenForgetRequestIsInFlight`
https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/PR-check/19960232358/1/nebius-x86-64/summary/1/ya-test.html#FAIL
(I can fix it here or in a separate PR)

qkrorlqr: Let's do it in a separate PR

PR #4706, opened 2025-11-21
issue-4138: Shard channel count reculacation during resize added

ruslan-gar:
If there is a filesystem which has a shard that has TotalBytesCount larger than that of main FS,
during transition to StrictFileSystemSizeEnforcementEnabled we calculate less channels number for this shard
than it really has, and it leads to and error. 
It's hard (but theoretically possible) to have this kind of filesystem in real life but we have one in our test environment.
So, recalculation of channel count that takes into account real number of channels in a shard is added.

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
@@ -273,9 +273,9 @@ void TAlterFileStoreActor::HandleDescribeFileStoreResponse(
             FileStoreConfig.ShardConfigs[shardNo].GetBlocksCount();
         FileStoreConfig.ShardConfigs[shardNo] = currentConfig;
         FileStoreConfig.ShardConfigs[shardNo].SetBlocksCount(blocksCount);
-        // We need to reconfigure the shard as only at this point we know
-        // original number of channels and it may happen that previously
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -264,10 +264,25 @@ void TAlterFileStoreActor::HandleDescribeFileStoreResponse(
...
     // A shard is described
     if (ev->Cookie != MainFileStoreCookie) {
-        FileStoreConfig.ShardConfigs[ev->Cookie].SetVersion(
+        const ui64 shardNo = ev->Cookie;
+        const ui64 blocksCount =
+            FileStoreConfig.ShardConfigs[shardNo].GetBlocksCount();
+        FileStoreConfig.ShardConfigs[shardNo] = currentConfig;
+        FileStoreConfig.ShardConfigs[shardNo].SetBlocksCount(blocksCount);
+        // We need to reconfigure the shard as only at this point we know
+        // original number of channels and it may happen that previously
```
debnatkh: `We need to reconfigure the shard, as only at this point do we know the original number of channels, and the calculated number of channels may be less than the real.`

debnatkh (approved):

PR #4711, opened 2025-11-24
issue-1751: [Filestore] Restore and drain WriteBackCache at session recreation

e673:

Resolve data loss in the following scenario:
1. WriteBackCache was enabled and used.
2. ServerWriteBackCacheEnabled flag was disabled in the configuration.
3. nfs-vhost was restarted.

Actual behavior: nfs-vhost completely ignores the non-flushed requests remaining in the persistent queue.

Expected behavior: nfs-vhost should check for the presence of the persistent queue and initialize WriteBackCache if the persistent queue exists regardless of the value of ServerWriteBackCacheEnabled flag. If ServerWriteBackCacheEnabled is not set, new writes should not be processed until the queue is flushed.

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
@@ -485,7 +485,7 @@ void TFileSystem::DoWrite(
     ui64 size,
     fuse_file_info* fi)
 {
-    const auto wbcMode = GetServerWriteBackCacheState(fi);
+    const auto wbcState = GetServerWriteBackCacheState(fi);
 
     const auto handle = fi->fh;
     const auto reqId = callContext->RequestId;
@@ -500,7 +500,7 @@ void TFileSystem::DoWrite(
         const auto& response = future.GetValue();
         const auto& error = response.GetError();
 
-        if (wbcMode != EServerWriteBackCacheState::Enabled) {
+        if (wbcState != EServerWriteBackCacheState::Enabled) {
             self->FSyncQueue
                 ->Dequeue(reqId, error, TNodeId{ino}, THandle{handle});
         }
@@ -510,7 +510,7 @@ void TFileSystem::DoWrite(
         }
     };
 
-    if (wbcMode == EServerWriteBackCacheState::Enabled) {
+    if (wbcState == EServerWriteBackCacheState::Enabled) {
         WriteBackCache.WriteData(callContext, std::move(request))
             .Subscribe(std::move(callback));
         return;
@@ -518,16 +518,16 @@ void TFileSystem::DoWrite(
 
     FSyncQueue->Enqueue(reqId, TNodeId {ino}, THandle {handle});
 
-    if (wbcMode == EServerWriteBackCacheState::Disabled) {
+    if (wbcState == EServerWriteBackCacheState::Disabled) {
         Session->WriteData(callContext, std::move(request))
             .Subscribe(std::move(callback));
         return;
     }
 
     Y_ABORT_UNLESS(
-        wbcMode == EServerWriteBackCacheState::Draining,
+        wbcState == EServerWriteBackCacheState::Draining,
         "Invalid EServerWriteBackCacheState value = %d",
-        wbcMode);
+        wbcState);
 
     WriteBackCache.FlushNodeData(request->GetNodeId())
         .Subscribe(


cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -960,17 +1005,23 @@ void TFileSystem::FSyncDir(
...
+        return EServerWriteBackCacheMode::Disabled;
     }
 
-    return true;
+    if (Config->GetServerWriteBackCacheEnabled()) {
+        return EServerWriteBackCacheMode::Enabled;
+    }
+
+    return WriteBackCache.IsEmpty() ? EServerWriteBackCacheMode::Disabled
+                                    : EServerWriteBackCacheMode::Draining;
```
neihar: Draining state will persist until next restart, shouldn't we switch states after draining phase? I understand that there is no big harm to call FlushNodeData and immediately call session->WriteData, but still it looks not right.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,188 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+            UNIT_ASSERT(suspend.Wait(WaitTimeout));
+        }
+
+        // Since write-back cache was enabled, the actual write didn't happen
+        // and the request is stored in the persistent queue
+        UNIT_ASSERT_VALUES_EQUAL(0, writeDataCalled.load());
+
+        {
+            NProto::TFileStoreFeatures features;
+            features.SetServerWriteBackCacheEnabled(true);
```
neihar: true? can we set this false and reuse it for callback?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,188 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+            auto reqWrite = std::make_shared<TWriteRequest>(
+                nodeId,
+                handleId,
+                0,
+                CreateBuffer(4096, 'a'));
+            reqWrite->In->Body.flags |= O_WRONLY;
+            auto write = bootstrap.Fuse->SendRequest<TWriteRequest>(reqWrite);
+            UNIT_ASSERT_NO_EXCEPTION(write.GetValue(WaitTimeout));
+
+            // Cache is drained and disabled - new requests go directly
```
neihar, line 3167: currently we will stack in forever 'Draining' state

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,188 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+            };
+
+            UNIT_ASSERT_VALUES_EQUAL(0, writeDataCalled2.load());
+
+            auto flush =
+                bootstrap.Fuse->SendRequest<TFlushRequest>(nodeId, handleId);
+            UNIT_ASSERT_NO_EXCEPTION(flush.GetValue(WaitTimeout));
+
+            // cache should be flushed
+            UNIT_ASSERT_VALUES_EQUAL(0, writeDataCalled.load());
```
neihar: Why do we test writeDataCalled, it was created in another scope for another instance of bootstrap

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1243,6 +1174,120 @@ class TFileSystemLoop final
...
+                Config->GetFileSystemId().Quote().c_str(),
+                Config->GetClientId().Quote().c_str());
+
+            WriteBackCache.FlushAllData().Subscribe(
+                [w = weak_from_this(),
+                 s = std::move(stopCompleted)](const TFuture<void>& f) mutable
+                {
+                    f.GetValue();
+                    if (auto p = w.lock()) {
+                        p->OnFlushAllData(std::move(s));
```
neihar: Don't like this recursive stuff. But don't have better proposal for now. no-op comment

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -413,45 +413,67 @@ void TFileSystem::Write(
...
-            self->FSyncQueue->Dequeue(reqId, error, TNodeId {ino}, THandle {handle});
+    Y_ABORT_UNLESS(
+        wbcMode == EServerWriteBackCacheMode::Draining,
+        "Invalid EServerWriteBackCacheMode value = %d",
+        wbcMode);
 
-            if (CheckResponse(self, *callContext, req, response)) {
-                self->ReplyWrite(*callContext, error, req, size);
+    auto flushFuture = WriteBackCache.FlushNodeData(request->GetNodeId());
+    if (flushFuture.HasValue()) {
```
neihar: Do we really need to do this? Future should handle it in that way internally as I understand.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,188 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+                offset,
+                CreateBuffer(byteCount, 'a'));
+
+            reqWrite->In->Body.flags |= O_WRONLY;
+            bootstrap.Fuse->SendRequest<TWriteRequest>(reqWrite);
+
+            totalByteCount += byteCount;
+        }
+
+        auto stopFuture = bootstrap.StopAsync();
```
neihar, line 2951: can we get filepath where file should be stored on disk and check that it is empty?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2808,188 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+                handleId,
+                0,
+                CreateBuffer(4096, 'a'));
+            reqWrite->In->Body.flags |= O_WRONLY;
+            auto write = bootstrap.Fuse->SendRequest<TWriteRequest>(reqWrite);
+            UNIT_ASSERT_NO_EXCEPTION(write.GetValue(WaitTimeout));
+
+            // Cache is drained and disabled - new requests go directly
+            // to the session
+            UNIT_ASSERT_VALUES_EQUAL(2, writeDataCalled2.load());
```
neihar, line 3169: can we check that cache file is empty?

neihar:
> For the future, this PR could be split into two parts: one for draining the cache on shutdown, and another for restoring it after suspension.
I see, it is already split, just in this pr we have both commits:)

cloud/filestore/libs/vfs_fuse/fs_impl.h
```diff
@@ -57,6 +57,20 @@ struct TReleaseRequest
 
 ////////////////////////////////////////////////////////////////////////////////
 
+enum class EServerWriteBackCacheMode
```
neihar: EServerWriteBackCacheState?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1243,6 +1174,120 @@ class TFileSystemLoop final
...
+                "[f:%s][c:%s] WriteBackCache is not empty, starting "
+                "FlushAllData",
+                Config->GetFileSystemId().Quote().c_str(),
+                Config->GetClientId().Quote().c_str());
+
+            WriteBackCache.FlushAllData().Subscribe(
+                [w = weak_from_this(),
+                 s = std::move(stopCompleted)](const TFuture<void>& f) mutable
+                {
+                    f.GetValue();
```
neihar, line 1191: can it be a failure(exception)?

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -413,45 +413,67 @@ void TFileSystem::Write(
...
+        return;
+    }
+
+    flushFuture.Subscribe(
+        [ptr = weak_from_this(),
+         callback = std::move(callback),
+         callContext = std::move(callContext),
+         request = std::move(request)](const auto& f) mutable
+        {
+            f.GetValue();
```
neihar: if (can??) flush fails, we just proceed to write normally

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -413,45 +413,67 @@ void TFileSystem::Write(
     request->SetBuffer(alignedBuffer.TakeBuffer());
 
     const auto size = buffer.size();
+    const auto wbcMode = GetServerWriteBackCacheMode(fi);
 
-    if (ShouldUseServerWriteBackCache(fi)) {
+    const auto handle = fi->fh;
+    const auto reqId = callContext->RequestId;
+
+    auto callback = [=, ptr = weak_from_this()](const auto& future)
```
neihar, line 493: is it possible to make some extraction and reuse code in Write and WriteBuf?

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: FlushAllData is not expected to throw an exception, we can safely make this call.
I don't like this code and I'd prefer Y_UNUSED(f) but this pattern is commonly used in other places.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: Yes. Added checks in the corresponding PR:
https://github.com/ydb-platform/nbs/pull/4702

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: Some remnants of the old code - removed this check

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: WriteBackCache object is not accessible from this place.
And accessing the cache file directly is not reliable (this is a memory mapped file with no disk synchronization guarantee).
So the answer is: without writing tons of boilerplate code — no.

cloud/filestore/libs/vfs_fuse/fs_impl.h
e673: I wanted to name it `EServerWriteBackCacheNodeProcessingMode` (because it also depended on node/handle attributes) but this name was too long.
In the future, it will not depend on node so I renamed it to `EServerWriteBackCacheState`.

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: No, it will not persist — the state will change to `Disabled` once the cache is completely flushed.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: If we set this to false here, the test code will not set WriteBackCache parameters and the cache file will not be loaded.

Upd: fixed the test code and reused it

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: There is periodic flush — the cache will eventually become empty

cloud/filestore/libs/vfs_fuse/loop.cpp
e673: I've made this part a bit better

cloud/filestore/libs/vfs_fuse/loop.cpp
e673: No, FlushAllData doesn't throw an exception (or sets it to a promise).

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: Yes, the code has been deduplicated

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: This was just a micro-optimization in order to save a couple of memory copy operations.
Indeed, the Draining state is temporary so it is better to make code more readable.
Removed this code.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
neihar: I see, we just don't have static state and always calculate it dynamically.

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -412,48 +412,74 @@ void TFileSystem::Write(
...
+    if (wbcMode == EServerWriteBackCacheState::Disabled) {
+        Session->WriteData(callContext, std::move(request))
+            .Subscribe(std::move(callback));
+        return;
+    }
 
-            const auto& response = future.GetValue();
-            const auto& error = response.GetError();
-            self->FSyncQueue->Dequeue(reqId, error, TNodeId {ino}, THandle {handle});
+    Y_ABORT_UNLESS(
```
qkrorlqr, line 527: let's use STORAGE_VERIFY / STORAGE_VERIFY_C macros - they force the developer to provide the entity id (filesystem id / tablet id / client id / etc), otherwise we'll just see that some error happened without knowing anything about the entity that caused the problem (so the debugging process will take more time)

here we can use client id in the STORAGE_VERIFY message

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
neihar: it is quite easy, you can look here for example https://github.com/ydb-platform/nbs/blob/da489bf0895427ecc8c48aab2bce0eb8a0ee857b/cloud/filestore/libs/vfs_fuse/fs_ut.cpp#L1071

I agree that it is a little bit hacky, but allow you to verify file state.

e673: Are you sure that accessing the file while it is being used by WriteBackCache is reliable?

neihar: We can suspend and check, but even without that, as I understand it, we can assume the test is deterministic and that nothing (including WBC) touches the file at this point. If that’s not the case, then we probably have a bigger problem.

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
qkrorlqr: Can be done in a separate PR - there're many other ABORT_UNLESS usages in vfs_fuse which would be nice to replace with STORAGE_VERIFY(clientId)

qkrorlqr (dismissed):

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
neihar: I would say, that ABORT_UNLESS is the only usage for checks in vfs_fuse and filestore vhost for now. Wasn't able to find a single instance of Verify in those components

neihar (approved):

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: I've added the suggested check

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -2808,6 +2823,267 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+        NProto::TFileStoreFeatures features;
+        features.SetServerWriteBackCacheEnabled(true);
+
+        // Disable automatic flush
+        TBootstrap bootstrap(
+            CreateWallClockTimer(),
+            CreateScheduler(),
+            features,
+            /* handleOpsQueueSize= */ 1000,
+            /* writeBackCacheAutomaticFlushPeriodMs= */ 1000000000,
```
neihar: Do we have special value to disable it?

e673: We can use zero value but it may be treated as misconfiguration in the future.
I'd prefer setting a large enough interval.

neihar: From this perspective, we should remove the logic that handles the 0 case :)
So either we use it, or, if we think it’s misleading and shouldn’t be used, we should remove it. Not in this PR ofc

e673: Zero value is needed in units tests for `TWriteBackCache` in order the tests to be deterministic.
Since it is already used in other tests (for example, `ShouldNotUseServerWriteBackCacheForDirectHandles`), I'll switch that huge value to 0 for consistency.

But mind that these tests may break in the future.

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -474,48 +474,74 @@ void TFileSystem::Write(
...
+    ui64 size,
+    fuse_file_info* fi)
+{
+    const auto wbcMode = GetServerWriteBackCacheState(fi);
+
+    const auto handle = fi->fh;
+    const auto reqId = callContext->RequestId;
 
-    if (ShouldUseServerWriteBackCache(fi)) {
+    auto callback = [=, ptr = weak_from_this()](const auto& future)
```
neihar, line 493: I guess capture wbcState by value works, because we can only have changes from Draining to Disable and logic is the same for both of them in callback, but potentially it can lead to the problems in the future with code modifications

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -474,48 +474,74 @@ void TFileSystem::Write(
...
+
+void TFileSystem::DoWrite(
+    TCallContextPtr callContext,
+    fuse_req_t req,
+    fuse_ino_t ino,
+    std::shared_ptr<NProto::TWriteDataRequest> request,
+    ui64 size,
+    fuse_file_info* fi)
+{
+    const auto wbcMode = GetServerWriteBackCacheState(fi);
```
neihar: wbcState

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: I don't understand your point.
What problems may occur?

neihar: For example you captured wbcState as Drained, but at the time when callback is called it can be "Disabled".

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: Renamed

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: Current logic is fine.
I can introduce the variable `shouldUseFSyncQueue` and use it instead of checking `wbcState != EServerWriteBackCacheState::Enabled` but this looks unnesessary.

PR #4745, opened 2025-11-28
issue-1751: [Filestore] Report correct file size by GetAttr and ListNodes when using WriteBackCache

e673:

The issue resolves the problem that `GetAttr` returns node size without taking into account cached data.

```
$ strace -f -tttT -o strace-git-clone-error.txt git clone https://github.com/martinetd/UDR
Cloning into 'UDR'...
remote: Enumerating objects: 826, done.
remote: Counting objects: 100% (3/3), done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 826 (delta 0), reused 1 (delta 0), pack-reused 823 (from 1)
Receiving objects: 100% (826/826), 467.76 KiB | 426.00 KiB/s, done.
fatal: premature end of pack file, 69 bytes missing
fatal: fetch-pack: invalid index-pack output
```

Problem indicated by strace:
```
<... pread64 resumed>"", 69, 475369) = 0 <0.001746>
```

What happened:
1. User opened file for writing and wrote 478966 bytes.
2. Write-back cache wrote 475147 bytes.
3. User opened a file for reading, CreateHandle reported size 475147.
4. Write-back cache wrote the remaining bytes.
5. User tried to read file beyond offset 475147 and received 0 bytes.
6. User closed file for writing.

Profile logs:
```
CreateHandle	0.000960s	S_OK	{parent_node_id=2488515, node_name=, flags=1, mode=0, node_id=2488515, handle=29247412031098704, size=475147}
CreateHandle	0.000704s	S_OK	{parent_node_id=2488515, node_name=, flags=1, mode=0, node_id=2488515, handle=70244355433295593, size=475147}
GetNodeAttr	0.000383s	S_OK	{parent_node_id=2488515, node_name=, flags=0, mode=292, node_id=2488515, handle=29247412031098704, size=475147}
...
WriteData	0.001408s	S_OK	[{node_id=2488515, handle=38285405765139321, offset=475147, bytes=3819}]
...
ReadData	0.000573s	S_OK	[{node_id=2488515, handle=70244355433295593, offset=471040, bytes=8192, actual_bytes=7926}]
```

Proposed solution:
* Adjust size returned by GetAttr using WriteBackCache
* Flush cache before processing SetAttr.

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
@@ -857,7 +857,9 @@ public:
     void ReleaseNodeStateRef(ui64 refId)
     {
         with_lock (Lock) {
-            Y_DEBUG_ABORT_UNLESS(NodeStateRefs.erase(refId));
+            const bool erased = static_cast<bool>(NodeStateRefs.erase(refId));
+            Y_DEBUG_ABORT_UNLESS(erased);
+
             const ui64 minRefId =
                 NodeStateRefs.empty() ? Max<ui64>() : *NodeStateRefs.begin();
 
@@ -873,17 +875,16 @@ public:
     ui64 GetCachedNodeSize(ui64 nodeId) const
     {
         with_lock (Lock) {
-            auto it = NodeStates.find(nodeId);
-            return it != NodeStates.end() ? it->second->CachedNodeSize : 0;
+            const auto* nodeStatePtr = NodeStates.FindPtr(nodeId);
+            return nodeStatePtr ? (*nodeStatePtr)->CachedNodeSize : 0;
         }
     }
 
     void SetCachedNodeSize(ui64 nodeId, ui64 size)
     {
         with_lock (Lock) {
-            auto it = NodeStates.find(nodeId);
-            if (it != NodeStates.end()) {
-                it->second->CachedNodeSize = size;
+            if (auto* nodeStatePtr = NodeStates.FindPtr(nodeId)) {
+                (*nodeStatePtr)->CachedNodeSize = size;
             }
         }
     }
@@ -975,7 +976,8 @@ private:
     // should be protected by |Lock|
     void EraseNodeState(ui64 nodeId)
     {
-        Y_DEBUG_ABORT_UNLESS(NodeStates.erase(nodeId));
+        const auto erased = static_cast<bool>(NodeStates.erase(nodeId));
+        Y_DEBUG_ABORT_UNLESS(erased);
         Stats->DecrementNodeCount();
     }
 


e673:
ToDo: add a test to the suite

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
```diff
@@ -78,21 +78,44 @@ void TFileSystem::SetAttr(
...
+        Session->SetNodeAttr(callContext, std::move(request))
+            .Subscribe(std::move(callback));
+    } else {
+        flushFuture.Subscribe(
+            [ptr = weak_from_this(),
+             callContext = std::move(callContext),
+             callback = std::move(callback),
+             request = std::move(request)](const auto& future) mutable
+            {
+                future.GetValue();
```
qkrorlqr: why is it ```TFuture<void>``` and not something like ```TFuture<NProto::TError>```?

WriteData can return a fatal error - both formally (the API allows it) and in real life scenarios (e.g. we're out of space logically - ENOSPC)

WriteBackCache seems to retry failed writes indefinitely disregarding ErrorKind (whether the error is retriable or not)

it doesn't look correct

e673: I've added a concern to the main issue:
https://github.com/ydb-platform/nbs/issues/1751#issuecomment-2958793714

qkrorlqr: No, let's first fix this problem and only then proceed to spread WriteBackCache class usage across our fuse layer code. I see no point in adding usages of the current seemingly incorrect interface only to fix them later. Doing it the other way around looks more efficient.

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
```diff
@@ -78,21 +78,44 @@ void TFileSystem::SetAttr(
...
+        }
+    };
+
+    // Changing attributes interferes with WriteBackCache
+    auto flushFuture = WriteBackCache
+                           ? WriteBackCache.FlushNodeData(request->GetNodeId())
+                           : NThreading::MakeFuture();
+
+    if (flushFuture.HasValue()) {
+        Session->SetNodeAttr(callContext, std::move(request))
```
neihar: the same as in other PRs, do we really need this branch?

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
e673: We need a discussion about it then.
Having a non-retriable error inside Flush means that we have lost client data that is not acceptable.
On the other side, retrying after a non-retriable error will result in infinite loop without progress.

Anyways, this is to be resolved in a separate PR as it is unrelated to the GetAttr issue.

qkrorlqr: I'll approve this one but the next thing that should be done regarding the write-back-cache task should be a fix for this problem.

> Having a non-retriable error inside Flush means that we have lost client data that is not acceptable.

Returning a 0 error code from fsync(fd) or sync() and then dropping the data is not acceptable. Or dropping the data written by a direct write. But if we return an error from fsync/sync, then it's fine as long as we don't drop the content of the write-back-cache for the problematic node-id.

qkrorlqr (approved):

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
e673: I plan to rewrite this place in the future PR (about handling direct read/write requests) to something like:
```
WriteBackCache.ExecuteWithoutCache(...)
```
It will simplify the code even more.

So I don't think it is needed to remove this micro-optimization - it doesn't affect anything anyway.

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -140,9 +140,15 @@ bool TFileSystem::UpdateNodeCache(
         return false;
     }
 
+    const ui64 adjustedSize =
```
debnatkh: А в этом месте не надо тоже править атрибуты?

https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp#L269

e673: Нет, не надо - они далее игнорируются
https://github.com/ydb-platform/nbs/blob/1cde0455c8dd625a9388e0db537baee854f2bff0/contrib/libs/virtiofsd/fuse_lowlevel.h#L1454-L1455

P.S. Было бы неплохо их кэшировать и затем отдавать в GetAttr, но это не относится к этой задаче.

e673: UPD: При использовании дефайна `FUSE_VIRTIO` вызывается `fuse_add_direntry_plus` - да, туда атрибуты передаются.

e673: Q: Задаётся ли у нас флаг `FUSE_VIRTIO`?

debnatkh: Задаётся

e673: Добавил фикс размера и для ListNodes

neihar: It’s a tricky situation: without a snapshot commit, different ListDirs requests of the same handle may use different commitIds for the data. As a result, directory entries from different batches are not consistent with each other.

POSIX doesn’t require us to reflect the actual file sizes after the first readdir (after opendir or rewinddir). So in an ideal solution, we would flush the WBC, take a single commitId, and then perform all directory listings (until rewinddir or closedir) using that commitId, without worrying about the WBC.

In our current reality, a single WBC flush before the first listing is probably sufficient. After that, it doesn’t really make sense to try to maintain “actual” sizes from the WBC.

Correct me if I am wrong @debnatkh

neihar: But for simple scenarios like 'ls' it should be fine (updating size from WBC) and observed result will be better.

e673: I don't like the idea of flushing the cache before each directory listing because it will significantly increase directory listing time. Since WriteBackCache knows nothing about directories, it will have to flush everything. Also, POSIX doesn't require us to return size at all so we are not restricted and may return whatever we want.

Virtiofsd may cache the returned attributes and reuse it so we need to report the most recent attributes.

But there is a corner case:
1. User writtes some data.
2. User requests ListNodes
3. The data is flushed and removed from cache.
4. ListNodes request returns old file size.

neihar: I do agree about current limitations with flushing cache before every listing.

neihar: Do you expect performance degradation in listing due to often calls to AdjustNodeSize ?

e673: Resolved the problem also for the mentioned corner case.

e673: Would be nice: optimization - if a client non-stop lists directories, the nodes will not be deleted.
Upd: fixed

e673: > Do you expect performance degradation in listing due to often calls to AdjustNodeSize ?

No. This is simple lookup in THashSet.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -182,6 +183,17 @@ struct TWriteBackCache::TNodeState
     // All entries with RequestId <= |AutomaticFlushRequestId| are to be flushed
     ui64 AutomaticFlushRequestId = 0;
 
+    // Cached data extends the node size but until the data is flushed,
+    // the changes are not visible to the tablet. Calls to GetAttr, ListNodes
```
neihar: Probably makes sense to add all requests to enumeration

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -182,6 +183,17 @@ struct TWriteBackCache::TNodeState
...
+    // and other requests that return node attributes should have the node
+    // size adjusted to this value.
+    ui64 MinNodeSize = 0;
```
neihar: I don't understand why it is Min, I would go with just NodeSize

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -182,6 +183,17 @@ struct TWriteBackCache::TNodeState
...
+
+    // Non-zero value indicates that the node state is to be deleted but is
+    // still referenced from |TImpl::NodeStateRefs|. It can be deleted when all
+    // references with id less than |DeletionId| are released.
```
neihar: with which id(can be not so obvious for first time reader)? probably worth correcting?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -825,6 +844,50 @@ class TWriteBackCache::TImpl final
         return EmptyFlag.load();
     }
 
+    ui64 AcquireNodeStateRef()
+    {
+        with_lock (Lock) {
+            ui64 refId = NextRequestId++;
```
neihar, line 851: should we care here about overflow or we assume that it is impossible to reach Max?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -3189,6 +3189,385 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+                "Scenario 3 " + testName);
+
+            proceedServiceOperations.SetValue();
+            UNIT_ASSERT_VALUES_EQUAL_C(
+                size + 4,
+                sizeFuture.GetValue(WaitTimeout),
+                "Scenario 3 " + testName);
+        };
+
+        auto test = [&](const TString& testName, auto getSizeFunc)
```
neihar, line 3557: maybe it does make sense to make separate suite for WBC fs_ut?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -3189,6 +3189,385 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+                getSizeFunc().GetValue(WaitTimeout),
+                "Scenario 2 " + testName);
+        };
+
+        // Scenario 3:
+        // - Request and wait for WriteData without O_DIRECT flag
+        // - Initiate a request that returns node size
+        // - Request and wait for Flush
+        // - Proceed the request that returns node size
+        auto testScenario3 = [&](const TString& testName, auto getSizeFunc)
```
neihar, line 3530: to be honest, it is hard to follow  this lambda stuff, some simplification could be helpful here.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -399,6 +411,13 @@ class TWriteBackCache::TImpl final
     // WriteData entries and Flush states grouped by nodeId
     THashMap<ui64, std::unique_ptr<TNodeState>> NodeStates;
 
+    // References to node states that prevents their deletion from |NodeStates|.
+    // Used to keep MinNodeSize information for flushed nodes.
+    TSet<ui64> NodeStateRefs;
+
+    // Node states to be deleted but that are referenced from |NodeStateRefs|.
+    TMap<ui64, TNodeState*> DeletedNodeStates;
```
neihar: Do we really need pointer in value or NodeId is enough?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -887,24 +952,40 @@ class TWriteBackCache::TImpl final
...
+        if (!nodeState->CanBeDeleted()) {
+            return;
+        }
+
+        NodesWithNewCachedEntries.erase(nodeState->NodeId);
+        NodesWithNewEntries.erase(nodeState->NodeId);
+
+        if (NodeStateRefs.empty()) {
+            Y_DEBUG_ABORT_UNLESS(NodeStates.erase(nodeState->NodeId));
             Stats->DecrementNodeCount();
```
neihar: Probably makes sense to write separate function with these 2 lines and call it in 2 places.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: WriteBackCache knows nothing about the real node size (it may be either higher or lower than the maximal offset of cached data parts). Renaming it to `NodeSize` will make a confusion because a reader may think that it is the final node size.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: This is `ui64` so we assume that it is impossible.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: Do you have any suggestions how to simplify this?
I can extract the functionality to a class but it will look almost the same except the lambdas become methods.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: The comment has been corrected

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
e673: To be honest, the whole fs_ut needs massive rework.
I'm not sure it should be done here.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: The comment has been corrected

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Yes, NodeId is enough - changed it.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Done

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
neihar: MaxCachedNodeSize? CachedNodeSize?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
neihar: may be add abort unless just in case?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
neihar: okay, let's think separately about separation of wbc fs_ut tests. Not in this PR.

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
neihar: I would at least do a separate suite with helper functions and then use separate Test macros for detailed scenarios. But we can do this refactoring for tests in separate PR I guess.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: CachedNodeSize is ok for me - renamed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -887,24 +952,45 @@ class TWriteBackCache::TImpl final
...
+            return;
+        }
+
+        NodesWithNewCachedEntries.erase(nodeState->NodeId);
+        NodesWithNewEntries.erase(nodeState->NodeId);
+
+        if (NodeStateRefs.empty()) {
+            EraseNodeState(nodeState->NodeId);
+        } else {
+            nodeState->DeletionId = NextRequestId++;
```
debnatkh, line 999: Is it safe to increment ui64 here without a lock?

e673: It is safe.
`DeleteNodeStateIfNeeded` is always called inside a lock.

neihar (approved):

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Added comments
```
should be protected by |Lock|
```

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -825,6 +845,49 @@ class TWriteBackCache::TImpl final
...
+            ui64 refId = NextRequestId++;
+            NodeStateRefs.insert(refId);
+            return refId;
+        }
+    }
+
+    void ReleaseNodeStateRef(ui64 refId)
+    {
+        with_lock (Lock) {
+            Y_DEBUG_ABORT_UNLESS(NodeStateRefs.erase(refId));
```
sharpeye: This code won't be executed in release

e673: Yes, this behavior is expected.
I don't want to crash if for some reason the element is already erased.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -825,6 +845,49 @@ class TWriteBackCache::TImpl final
...
+
+    void SetCachedNodeSize(ui64 nodeId, ui64 size)
+    {
+        with_lock (Lock) {
+            auto it = NodeStates.find(nodeId);
+            if (it != NodeStates.end()) {
+                it->second->CachedNodeSize = size;
+            }
+        }
+    }
```
sharpeye, line 890: [TMapOps](https://github.com/ydb-platform/nbs/blob/main/util/generic/mapfindptr.h) has handy functions:
```
return NodeStates.Value(nodeId, ui64{});
...
if (auto* state = NodeStates.FindPtr(nodeId)) {
    state->CachedNodeSize = size;
}
```

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
sharpeye: Does it mean that in Release build NodeStateRefs will never be cleaned up?

e673: Hm, I was expecting that the code inside `Y_DEBUG_ABORT_UNLESS` always runs but it is not checked.
Yes, this should be fixed then.

e673: Thanks. Fixed.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Done

Issue #4748, opened 2025-12-01
[Filestore local] implement mechanism to enable features for specific filesystems


budevg:
We want to enable some feature for specific filesystem id or project id. There is existing [features configuration](https://github.com/ydb-platform/nbs/blob/a181ce9106e77fb69e85007bdc0e292424c7d139/cloud/storage/core/libs/features/features_config.cpp#L48) which we can use to achieve this.

Additionally we want to extend the `features` config to enable matching based on regexp and not only exact string match

PR #4750, opened 2025-12-01
[Filestore local] issue-4748: Enable features for filestore local and apply for inode timeout configs

budevg:
#4748

### DIFF
--- a/cloud/filestore/libs/service_local/fs_session.cpp
+++ b/cloud/filestore/libs/service_local/fs_session.cpp
@@ -27,9 +27,9 @@ NProto::TCreateSessionResponse TLocalFileSystem::CreateSession(
 
         *response.MutableFileStore() = Store;
 
-        auto& cloudId = Store.GetCloudId();
-        auto& folderId = Store.GetFolderId();
-        auto& fsId = Store.GetFileSystemId();
+        const auto& cloudId = Store.GetCloudId();
+        const auto& folderId = Store.GetFolderId();
+        const auto& fsId = Store.GetFileSystemId();
 
         auto* features = response.MutableFileStore()->MutableFeatures();
         features->SetDirectIoEnabled(Config->GetDirectIoEnabled());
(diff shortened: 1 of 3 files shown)


cloud/storage/core/libs/features/filters.cpp
```diff
@@ -1,7 +1,28 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+bool RegexpMatchItemInSet(const THashSet<TString>& set, const TString& item)
+{
+    return AnyOf(
+        set,
+        [&](const TString& setElem)
+        {
+            std::regex r{setElem.data(), setElem.size()};
```
debnatkh: std::regex can throw an exception

Maybe we should use pcre? https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/tools/fs/cleanup-ext4-meta/ext4-meta-reader.cpp#L3

budevg: pcre seems to throw exception as well https://github.com/ydb-platform/nbs/blob/main/library/cpp/regex/pcre/pcre.h#L72
I will fix the std::regex code

cloud/filestore/libs/service_local/fs_session.cpp
```diff
@@ -27,6 +27,10 @@ NProto::TCreateSessionResponse TLocalFileSystem::CreateSession(
 
         *response.MutableFileStore() = Store;
 
+        auto& cloudId = Store.GetCloudId();
```
drbasic: const

cloud/storage/core/libs/features/filters.h
```diff
@@ -15,6 +16,7 @@ class TFilters
     THashSet<TString> CloudIds;
     THashSet<TString> FolderIds;
     THashSet<TString> EntityIds;   // DiskIds or FsIds
+    NProto::EFilterMatchAlgorithm MatchAlgorithm;
 
 public:
     explicit TFilters(NProto::TFilters config);
```
drbasic: const NProto::TFilters& ?

cloud/storage/core/libs/features/filters.h
```diff
@@ -4,6 +4,7 @@
 
 #include <util/generic/hash_set.h>
 #include <util/generic/string.h>
+#include <util/generic/vector.h>
```
drbasic: зачем?

PR #4762, opened 2025-12-03
Introduce TBlockRangeList

drbasic:
TBlockRangeList is a class that manages a collection of block ranges (TBlockRange64) with efficient overlap checking capabilities. It's designed to store and query block ranges, particularly useful for determining if a given range overlaps with any of the stored ranges.

### DIFF
--- a/cloud/blockstore/libs/common/block_range_list.cpp
+++ b/cloud/blockstore/libs/common/block_range_list.cpp
@@ -1,0 +1,64 @@
+#include "block_range_list.h"
+
+#include <util/stream/str.h>
+
+namespace NCloud::NBlockStore {
+
+////////////////////////////////////////////////////////////////////////////////
+
+void TBlockRangeList::AddRange(TBlockRange64 range)
+{
+    MaxLength = Max(MaxLength, range.Size());
+    Ranges.insert(range);
+}
+
+bool TBlockRangeList::RemoveRange(TBlockRange64 range)
+{
+    auto it = Ranges.find(range);
+    if (it != Ranges.end()) {
+        Ranges.erase(it);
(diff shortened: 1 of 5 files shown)


cloud/blockstore/libs/common/block_range_list.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <util/stream/str.h>
+
+namespace NCloud::NBlockStore {
+
+////////////////////////////////////////////////////////////////////////////////
+
+void TBlockRangeList::AddRange(TBlockRange64 range)
+{
+    MaxLength = Max(MaxLength, range.Size());
```
qkrorlqr, line 11: What do you think of introducing some limit for MaxLength (e.g. set up via a ctor parameter) to ensure that we don't accidentally add a very wide range making subsequent `Overlaps()` calls extremely slow? If the caller tries to add a range which is too wide, we can either abort() or return an error / return false.

drbasic: 1. We have an agreement that the maximum request size is 1024 blocks. And they are even fixed in the code. 
2. This class is designed to replace all O(N). 
3. In the worst case, the algorithm will run in the same O(N) time.  Therefore, it definitely doesn't get worse. 
4. I don't want to throw any exceptions or do something else. At most, I can add a comment so that you don't add a huge number of identical ranges, but it's not clear if this will help.

qkrorlqr: I'm not suggesting any complex changes, I'm suggesting a simple check which would cause an error to be returned or something like that. I'm fine with this change overall.

PR #4772, opened 2025-12-04
refactooring of checkrange's unit tests

yudov-maksim:
Checkrange unit tests existed as 3 (for ydb based disks, nrd, mirror) almost identical test suits. Now it was moved into the one codebase as a part of volume_ut tests with disk's type as a parameter for test case.
Comparing to previous code, tests have only minor changes. Logically it is almost the same.
"Almost" - for example, we had test with logical error. "ShouldGetSameChecksumsWhileCheckRangeSimmilarDisks" was designed to check writing to 2 disks, but it was done into one disks with 2 clients. This testcase was replaced by different, but logically similar "DoTestShouldGetSameChecksumsWhileCheckRangeEqualBlocks" test.

// This refactoring has done as a part of checkrange utility improvements.

### DIFF
--- a/cloud/blockstore/libs/storage/volume/volume_ut.cpp
+++ b/cloud/blockstore/libs/storage/volume/volume_ut.cpp
@@ -3563,11 +3563,22 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
             auto response =
                 volume.RecvCheckRangeResponse(TDuration::Seconds(1));
             const auto& record = response->Record;
-            UNIT_ASSERT_VALUES_EQUAL(false, HasError(record.error()));
-            UNIT_ASSERT_VALUES_EQUAL(false, HasError(record.status()));
-            UNIT_ASSERT_VALUES_EQUAL(
+
+            std::string extended_msg =
+                "startIndex: " + std::to_string(startIndex) +
+                ", size: " + std::to_string(size);
+            UNIT_ASSERT_VALUES_EQUAL_C(
+                false,
+                HasError(record.error()),
+                extended_msg);
+            UNIT_ASSERT_VALUES_EQUAL_C(
+                false,
+                HasError(record.status()),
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/volume/volume_ut.cpp
```diff
@@ -3532,6 +3533,247 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
...
+            clientInfo.GetClientId(),
+            42);
+
+        const auto checkRange = [&](ui32 startIndex, ui32 size)
+        {
+            volume.SendCheckRangeRequest("disk-id", startIndex, size);
+            auto response =
+                volume.RecvCheckRangeResponse(TDuration::Seconds(1));
+            const auto& record = response->Record;
+            UNIT_ASSERT_VALUES_EQUAL(false, HasError(record.error()));
```
e673: Let's add more verbosity and include `startIndex` and `size` in the assert message.

cloud/blockstore/libs/storage/volume/volume_ut.cpp
```diff
@@ -3532,6 +3533,247 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
...
+                return TTestActorRuntime::DefaultObserverFunc(event);
+            });
+
+        const auto checkRange = [&](ui32 idx, ui32 size)
+        {
+            volume.SendCheckRangeRequest("disk-id", idx, size);
+            auto response =
+                volume.RecvCheckRangeResponse(TDuration::Seconds(1));
+            const auto& record = response->Record;
+            UNIT_ASSERT_VALUES_EQUAL(E_IO, record.GetStatus().code());
```
e673: Let's add more verbosity and include startIndex and size in the assert message.

drbasic (approved):

PR #4777, opened 2025-12-05
Introduce map for TBlockRange with identity key

drbasic:
TBlockRangeMap is a class that manages a collection of block ranges with identity keys with efficient overlap checking capabilities. It's designed to store and query block ranges, particularly useful for determining if a given range overlaps with any of the stored ranges.

### DIFF
--- a/cloud/blockstore/libs/common/block_range_map.h
+++ b/cloud/blockstore/libs/common/block_range_map.h
@@ -10,15 +10,11 @@ namespace NCloud::NBlockStore {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-class TEmptyType
-{
-};
-
 // TBlockRangeMap is a class that manages a collection of block ranges and it
 // key (ui64) with efficient overlap checking capabilities. It's designed to
 // store and query block ranges and it key, particularly useful for determining
 // if a given range overlaps with any of the stored ranges.
-template <typename TKey, typename TValue = TEmptyType>
+template <typename TKey, typename TValue>
 class TBlockRangeMap
 {
 public:
@@ -48,7 +44,7 @@ private:
     };
 
     ui64 MaxLength = 0;
-    TMultiSet<TItem, TRangeComparator> Ranges;
+    TSet<TItem, TRangeComparator> Ranges;
     THashMap<TKey, decltype(Ranges.begin())> RangeByKey;
 
 public:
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/common/block_range_map.h
```diff
@@ -0,0 +1,181 @@
...
+    {
+        bool operator()(const TItem& lhs, const TItem& rhs) const
+        {
+            return std::tie(lhs.Range.End, lhs.Range.Start, lhs.Key) <
+                   std::tie(rhs.Range.End, rhs.Range.Start, rhs.Key);
+        }
+    };
+
+    ui64 MaxLength = 0;
+    TMultiSet<TItem, TRangeComparator> Ranges;
```
e673: Why `TMultiSet` and not just `TSet`?
According to the logic, all items should have distinct `key` value.
So duplicate `TItem` cannot exist.

cloud/blockstore/libs/common/block_range_map.h
```diff
@@ -0,0 +1,181 @@
...
+            {
+                result = &item;
+                return EEnumerateContinuation::Stop;
+            });
+
+        return result;
+    }
+
+    // Enumerate all overlapped ranges.
+    void AllOverlaps(TBlockRange64 other, TEnumerateFunc f) const
```
e673: Name suggestion: `EnumerateOverlapping` — the naming should be consistent with `Enumerate` function.

cloud/blockstore/libs/common/block_range_map.h
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // Removes block range specified by Key from the collection. Returns false
+    // if the range was not found in the collection.
+    bool RemoveRange(TKey key)
+    {
+        return ExtractRange(key).has_value();
+    }
+
+    // Checks that the other range overlaps with any range in Ranges.
```
e673, line 91: The returned value should be explained in the comment.

cloud/blockstore/libs/common/block_range_map.h
```diff
@@ -0,0 +1,181 @@
...
+    [[nodiscard]] const TItem* Overlaps(TBlockRange64 other) const
```
e673: Name suggestion: `FindFirstOverlapping`

cloud/blockstore/libs/common/block_range_map.h
```diff
@@ -0,0 +1,181 @@
...
+            });
+
+        return result;
+    }
+
+    // Enumerate all overlapped ranges.
+    void AllOverlaps(TBlockRange64 other, TEnumerateFunc f) const
+    {
+        // 1. Find the range x which: x.end >= other.start in the sorted list
+        // (by
```
e673: Remove \n here

cloud/blockstore/libs/common/block_range_map.h
```diff
@@ -0,0 +1,181 @@
...
+// if a given range overlaps with any of the stored ranges.
+template <typename TKey, typename TValue = TEmptyType>
+class TBlockRangeMap
+{
+public:
+    struct TItem
+    {
+        TKey Key = {};
+        TBlockRange64 Range;
+        TValue Value = {};
```
e673, line 25: Do you really need `Value` here?
You don't use it and don't test it.
Looks like overengineering.

e673 (changes requested):

cloud/blockstore/libs/common/block_range_map.h
drbasic: I will need this value in the near future)

e673: Then I have a small suggestion: it would be nice if you dump it in `DebugPrint` if it is possible.

e673 (approved):

Issue #568, opened 2024-02-26
[Filestore] Log inode events to tablet profile log


qkrorlqr:
Right now it seems that only block/blob-level events are logged.

PR #4781, opened 2025-12-05
issue-4695: logging io requests to tablet profile-log

qkrorlqr:
#### Notes
Logging `WriteData`, `AddData`, `ReadData`, `DescribeData` to tablet profile log to at least be able to see what were the exact node-id, offset and length of those requests. 

#### Issue

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut.cpp
@@ -2814,7 +2814,7 @@ Y_UNIT_TEST_SUITE(TStorageServiceTest)
         UNIT_ASSERT_VALUES_EQUAL(data, readDataResult->Record.GetBuffer());
 
         UNIT_ASSERT_VALUES_EQUAL(
-            1,
+            2, // 1 from service, 1 from tablet
             profileLog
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/actors/tablet_adddata.cpp
```diff
@@ -90,6 +97,13 @@ void TAddDataActor::ReplyAndDie(
     const TActorContext& ctx,
     const NProto::TError& error)
 {
+    FinalizeProfileLogRequestInfo(
```
debnatkh, line 100: А это происходит не там, же где и Metrics.AddData, потому что надо передать владение ProfileLogRequest из TAddDataActor обратно в TIndexTabletActor?

qkrorlqr: 1. it's implemented in a similar way for the other requests that are logged by the tablet (e.g. compaction - in `TCompactionActor`)
2. yes, passing it to the tablet is inconvenient and verbose
3. `ProfileLog->Write(...)` is just a push into a lock-free stack, it's not expensive

debnatkh (approved):

Issue #2390, opened 2024-10-30
[Filestore] directory handles don't survive filestore-vhost restart


qkrorlqr:
Currently directory handles which are used in OpenDir, ReadDir, FsyncDir are stored in memory and don’t survive restart of filestore-vhost service.

Simple C program to reproduce the bug:
```
#include <stdio.h>
#include <dirent.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>

int main(int argc, char **argv) {
  int fd = open(argv[1], O_RDONLY);
  if (fd == -1) {
    perror("open");
    return 1;
  }
  printf("press enter: ");
	fflush(stdout);
  char buffer[1024];
  ssize_t bytes_read = read(STDIN_FILENO, buffer, sizeof(buffer));
  if (bytes_read == -1) {
    perror("read");
    return 1;
  }
  
  fsync(fd);
  DIR *dir = fdopendir(fd);
  if (dir == NULL) {
    perror("fdopendir");
    return 1;
  }
  struct dirent *entry;
  while ((entry = readdir(dir)) != NULL) {
    printf("%s\n", entry->d_name);
  }
  closedir(dir);
  return 0;
}
```

WilyTiger:
The problem is that we store DirectoryHandles https://github.com/ydb-platform/nbs/blob/7c667ab927fc10390b05f16eb4431fca0fb527c3/cloud/filestore/libs/vfs_fuse/fs_impl.h#L83 in memory. A possible solution could be to store information about the handles persistently on the host's tmpfs, similar to the approach used for AsyncClose (#1541), WriteBackCache (#1751), and LocalFilestore (#2199).

PR #4788, opened 2025-12-07
[Filestore] directory handles improvements

neihar:
#2390

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/directory_handles_storage.cpp
+++ b/cloud/filestore/libs/vfs_fuse/directory_handles_storage.cpp
@@ -96,8 +96,8 @@ void TDirectoryHandlesStorage::ResetHandle(ui64 handleId)
 {
     TGuard guard(TableLock);
     if (HandleIdToIndices.contains(handleId)) {
-        for (auto it = std::next(HandleIdToIndices[handleId].begin(), 1);
-             it != HandleIdToIndices[handleId].end();
+        for (auto it = HandleIdToIndices[handleId].rbegin();
+             it != std::prev(HandleIdToIndices[handleId].rend(), 1);
              ++it)
         {
             if (!Table->DeleteRecord(*it)) {
@@ -117,15 +117,22 @@ void TDirectoryHandlesStorage::LoadHandles(TDirectoryHandleMap& handles)
 {
     // Since we store data in chunks instead of a single block, in rare cases
     // a crash during the reset or removal process can lead to inconsistent
-    // chunks order. We detect this inconsistency during the load phase clean
-    // data for this handle.
+    // chunks order. We detect this inconsistency during the load phase and
+    // clean data for this handle.
     struct TUpdateVersionInfo
     {
         ui64 LargestUpdateVersion = 0;
         ui64 ChunksCount = 0;
     };
 
+    struct TChunkInfo
+    {
+        ui64 UpdateVersion = 0;
+        ui64 StorageIndex = 0;
+    };
+
     TMap<ui64, TUpdateVersionInfo> updateVersionInfo;
+    TMap<ui64, TVector<TChunkInfo>> chunksInfo;
 
     {
         TGuard guard(TableLock);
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/vfs_fuse/directory_handles_storage.cpp
```diff
@@ -117,13 +117,12 @@ void TDirectoryHandlesStorage::LoadHandles(TDirectoryHandleMap& handles)
 {
     // Since we store data in chunks instead of a single block, in rare cases
     // a crash during the reset or removal process can lead to inconsistent
-    // chunks order. We detect this inconsistency during the load phase and fix
-    // it.
+    // chunks order. We detect this inconsistency during the load phase clean
+    // data for this handle.
```
budevg: "clean data for this handle" -> "and clean data for this handle"

cloud/filestore/libs/vfs_fuse/directory_handles_storage.cpp
```diff
@@ -178,24 +176,18 @@ void TDirectoryHandlesStorage::LoadHandles(TDirectoryHandleMap& handles)
...
             ReportDirectoryHandlesStorageError(
                 TStringBuilder()
-                << "Total chunks count " << updateVersionInfo.ChunksCount
+                << "Corrupted data for handle " << handleId
+                << ": total chunks count " << updateVersionInfo.ChunksCount
                 << " is not equal to largest update version "
                 << updateVersionInfo.LargestUpdateVersion << " + 1");
-            ResetHandle(handleId);
+
+            RemoveHandle(handleId);
```
budevg, line 206: With this change if there is crash during `ResetHandle` you will remove the handle and the guest will get EBADF.
This only make sense if there is crush during `RemoveHandle`
Maybe you can change the code so that in `RemoveHandle` you will delete record 0 first (indicating this is the Remove Handle) and during the `ResetHandle` you will just delete the [1,...] records. Then during the load you will delete the handle only if the record 0 is not preset and remove the [1,..] records otherwise completing the `ResetHandle`

budevg (changes requested):

cloud/filestore/libs/vfs_fuse/directory_handles_storage.cpp
neihar: you are right. We previously discussed that tmpfs can't be corrupted, but I still consider that it can, in this scenario it can be a problem to distinguish underlying storage corruption and crash during reset (without additional journaling). So, In this solution, I decided that one rare EBADF is acceptable solution. 
After thinking a little bit longer, I found another solution which implemented and pushed, that should match correctness and avoid unneeded EBADF.

neihar: In this new solution, if we crash during ResetHandle, we will restore an arbitrary state of the handle (because we remove the cache from the end). The cache itself will always be structurally valid in terms of containing data.

From the data semantics point of view crashing during ResetHandle implies that the guest was executing rewinddir. If we crash during next readdir (after rewinddir), the guest will not receive the response to this command; as I understand it, the call will hang and eventually return EIO (I’ll double-check this behavior, because in my tests it always hangs).
After restart operation retried, and we complete ResetHandle. Everything is fine.

Even if we not hang and return some error to user the next readdir should also contain 0 offset that will retrigger ResetHandle.

neihar: So, checked locally with qemu and vhost. 
1. rewinddir() in glibc implemented as llseek (for [example](https://mirrors.git.embecosm.com/mirrors/glibc/-/blob/glibc-2.22/sysdeps/posix/rewinddir.c#L32) )
2. it goes to the kernel [1](https://github.com/torvalds/linux/blob/67a454e6b1c604555c04501c77b7fedc5d98a779/fs/read_write.c#L29) [2](https://github.com/torvalds/linux/blob/67a454e6b1c604555c04501c77b7fedc5d98a779/fs/read_write.c#L180) [3](https://github.com/torvalds/linux/blob/67a454e6b1c604555c04501c77b7fedc5d98a779/fs/read_write.c#L85) [4](https://github.com/torvalds/linux/blob/67a454e6b1c604555c04501c77b7fedc5d98a779/fs/read_write.c#L64) 
And internally the kernel sets the offset to 0, so it does not generate any FUSE request for rewinddir itself, it only prepares the offset for the next readdir.
On the next readdir, we receive offset 0 and perform the handle reset.

If we crash during this process, [vhost virtio](https://github.com/yandex-cloud/yc-libvhost-server/blob/d6b391251d4a5155ecb84d42e0ffc614a6434b24/virtio/virt_queue.c#L285) will retry all in-flight requests on its side, and we will simply continue the reset operation after the crash. This makes it safe to restore from a crash that occurs during ResetHandle, data consistency is preserved, and the resulting behavior remains correct.

neihar:
```
2025-12-07T20:26:23.698164Z :NFS_VHOST DEBUG: : virtq_inflight_resubmit:313: /tmp/vhost.sock[1]: cnt = 0 inflight requests should be resubmitted
2025-12-07T20:29:33.459003Z :NFS_VHOST DEBUG: : virtq_inflight_resubmit:313: /tmp/vhost.sock[0]: cnt = 0 inflight requests should be resubmitted
2025-12-07T20:29:33.459247Z :NFS_VHOST DEBUG: : virtq_inflight_resubmit:313: /tmp/vhost.sock[1]: cnt = 1 inflight requests should be resubmitted
2025-12-07T20:30:21.653721Z :NFS_VHOST DEBUG: : virtq_inflight_resubmit:313: /tmp/vhost.sock[0]: cnt = 0 inflight requests should be resubmitted
2025-12-07T20:30:21.653919Z :NFS_VHOST DEBUG: : virtq_inflight_resubmit:313: /tmp/vhost.sock[1]: cnt = 1 inflight requests should be resubmitted
2025-12-07T21:15:26.857279Z :NFS_VHOST DEBUG: : virtq_inflight_resubmit:313: /tmp/vhost.sock[0]: cnt = 0 inflight requests should be resubmitted
```

Issue #4664, opened 2025-11-13
[Filestore] get rid of buffer copying in TFileSystem::Read for YDB-based filestore


debnatkh:
Copying happens here:

https://github.com/ydb-platform/nbs/blob/381b410490cfff565e2ecaae9798360464a380ec/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp#L373

We can pre-allocate target buffers and direct `TReadDataActor`, spawned by the `TStorageServiceActor`, to direct the response in these pre-allocated iovecs.

PR #4790, opened 2025-12-07
issue-4664: eliminate redundant buffer allocation and data copying in TReadDataActor when reading into iovecs

antonmyagkov:
issue: #4664

Handle reading to buffer and iovecs in the same way by wrapping buffer/iovecs into the rope.

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_readdata.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_readdata.cpp
@@ -806,7 +806,7 @@ void TReadDataActor::ReplyTwoStageAndDie(const TActorContext& ctx)
     if (end <= OriginByteRange.Offset) {
         BlockBuffer->clear();
     } else {
-        auto length = end - OriginByteRange.Offset;
+        const ui64 length = end - OriginByteRange.Offset;
         if (ReadRequest.GetIovecs().empty()) {
             BlockBuffer->ReserveAndResize(length);
             response->Record.set_allocated_buffer(BlockBuffer.release());


cloud/filestore/libs/storage/service/service_actor_readdata.cpp
```diff
@@ -143,11 +145,16 @@ TReadDataActor::TReadDataActor(
...
-    BlockBuffer->ReserveAndResize(ReadRequest.GetLength());
+    if (ReadRequest.GetIovecs().empty()) {
+        // BlockBuffer should not be initialized in constructor, because
+        // creating a block buffer leads to memory allocation (and
+        // initialization) which is heavy and we would like to execute that on a
+        // separate thread (instead of this actor's parent thread)
+        BlockBuffer->ReserveAndResize(ReadRequest.GetLength());
+        TargetBuffers = CreateRope(*BlockBuffer);
+    } else {
+        TargetBuffers = CreateRope(ReadRequest.GetIovecs());
```
antonmyagkov, line 205: The main change is here: buffers or iovecs are now wrapped into a rope, eliminating block buffer allocation and the following data copy from block buffer into the iovecs.

cloud/filestore/libs/storage/service/rope_utils.cpp
```diff
@@ -0,0 +1,83 @@
+#include "rope_utils.h"
+
+namespace NCloud::NFileStore::NStorage {
+
+namespace {
+
+template <bool Owner = false>
+class TIovecContiguousChunk: public IContiguousChunk
```
antonmyagkov, line 10: TIovecContiguousChunk was moved from service_actor_writedata.cpp

cloud/filestore/libs/storage/service/service_actor_readdata.cpp
```diff
@@ -702,12 +707,15 @@ void TReadDataActor::ReplyAndDie(const TActorContext& ctx)
     if (end <= OriginByteRange.Offset) {
         BlockBuffer->clear();
     } else {
-        BlockBuffer->ReserveAndResize(end - OriginByteRange.Offset);
-        response->Record.set_allocated_buffer(BlockBuffer.release());
+        auto length = end - OriginByteRange.Offset;
+        if (ReadRequest.GetIovecs().empty()) {
+            BlockBuffer->ReserveAndResize(length);
+            response->Record.set_allocated_buffer(BlockBuffer.release());
```
qkrorlqr, line 812: interesting function:
```
  inline void TReadDataResponse::set_allocated_buffer(TProtoStringType* buffer) {                                                                                                                                                                             
    if (buffer != nullptr) {                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                              
    } else {                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                              
    }                                                                                                                                                                                                                                                         
    buffer_.SetAllocated(&::PROTOBUF_NAMESPACE_ID::internal::GetEmptyStringAlreadyInited(), buffer,                                                                                                                                                           
        GetArenaForAllocation());                                                                                                                                                                                                                             
  #ifdef PROTOBUF_FORCE_COPY_DEFAULT_STRING                                                                                                                                                                                                                   
    if (buffer_.IsDefault(&::PROTOBUF_NAMESPACE_ID::internal::GetEmptyStringAlreadyInited())) {                                                                                                                                                               
      buffer_.Set(&::PROTOBUF_NAMESPACE_ID::internal::GetEmptyStringAlreadyInited(), "", GetArenaForAllocation());                                                                                                                                            
    }                                                                                                                                                                                                                                                         
  #endif // PROTOBUF_FORCE_COPY_DEFAULT_STRING                                                                                                                                                                                                                
    // @@protoc_insertion_point(field_set_allocated:NCloud.NFileStore.NProto.TReadDataResponse.Buffer)                                                                                                                                                        
  }  
```
what is it supposed to do and why does it have that empty if-else statement in the beginning?

cloud/filestore/libs/storage/service/service_actor_readdata.cpp
qkrorlqr, line 205: how do we make sure that no one (e.g. dsproxy) continues to use this `TRope` after we are done with this request?

antonmyagkov: I don't pass rope anywhere, for now it's convenient to use it as we don't have any limitations on iovec sizes in our api and EvReadDataResponse represented as a rope too so it's easy to reuse RopeUtils::memcpy to copy data.

The main weakness of the rope is its iterator performance, since the underlying structure is a list. As a next step, I’ve already suggested restricting the sizes of iovecs in our API so they are the same size, with only the last one allowed to be smaller for the unaligned tail. This would allow us to replace the rope with a custom implementation of the IBlockBuffer interface.

cloud/filestore/libs/storage/service/service_actor_readdata.cpp
antonmyagkov: This is a standard protobuf method

`void set_allocated_foo(string* value): Sets the string object to the field and frees the previous field value if it exists. If the string pointer is not NULL, the message takes ownership of the allocated string object and has_foo() will return true. The message is free to delete the allocated string object at any time, so references to the object may be invalidated. Otherwise, if the value is NULL, the behavior is the same as calling clear_foo().`

Empty if-else statement looks weird and will probably disappear after protoc update(ydb 24.4/ydb 25.1).

cloud/filestore/libs/storage/service/rope_utils.cpp
```diff
@@ -0,0 +1,83 @@
+#include "rope_utils.h"
+
+namespace NCloud::NFileStore::NStorage {
+
+namespace {
+
```
qkrorlqr, line 6: ///...

cloud/filestore/libs/storage/service/rope_utils.cpp
```diff
@@ -0,0 +1,83 @@
...
+            TRope(TRcBuf(
+                MakeIntrusive<TIovecContiguousChunk<>>(
+                    iovec.GetBase(),
+                    iovec.GetLength()))));
+    }
+
+    return rope;
+}
+
+TRope CreateRope(const TString& buffer)
```
qkrorlqr: the signature of this function doesn't look correct - it accepts a const reference which implies that it's not going to modify anything but internally it just `reinterpret_cast`s the pointer to const contents of the buffer and creates an object which can be used to modify the data

let's change it to `TRope CreateRope(TString& buffer)` and also change `buffer.data()` to `buffer.begin()` to trigger `TString`'s COW

qkrorlqr: this fixes the following scenario:
```
TString s1 = "xxx";
TString s2 = s1;
TRope r = CreateRope(s2);
r.Begin().ContiguousDataMut()[0]= 'y';
... now s1 contains "yxx" ...
```

but doesn't fix this one:
```
TString s1 = "xxx";
TRope r = CreateRope(s1);
TString s2 = s1;
r.Begin().ContiguousDataMut()[0] = 'y';
... now s2 contains "yxx" ...
```

so actually it's better to completely avoid `TString` usage in such interfaces and use either `void*` + `size_t` or `TStringBuf`

antonmyagkov: Replaced with void* pointer and size

cloud/filestore/libs/storage/service/rope_utils.cpp
antonmyagkov: Fixed

cloud/filestore/libs/storage/service/service_actor_readdata.cpp
```diff
@@ -200,7 +200,9 @@ void TReadDataActor::Bootstrap(const TActorContext& ctx)
         // initialization) which is heavy and we would like to execute that on a
         // separate thread (instead of this actor's parent thread)
         BlockBuffer->ReserveAndResize(ReadRequest.GetLength());
-        TargetBuffers = CreateRope(*BlockBuffer);
+        TargetBuffers = CreateRope(
+            const_cast<char*>(BlockBuffer->data()),
```
qkrorlqr: `BlockBuffer->begin()` returns non-const `char*`

antonmyagkov: fixed

qkrorlqr (dismissed):

cloud/filestore/libs/storage/service/rope_utils.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+template <bool Owner = false>
+class TIovecContiguousChunk: public IContiguousChunk
+{
+private:
+    char* Base = 0;
+    ui64 Size = 0;
+
+public:
+    TIovecContiguousChunk(ui64 base, ui64 size)
```
drbasic, line 17: почему не char*, size_t ?

cloud/filestore/libs/storage/service/rope_utils.cpp
```diff
@@ -0,0 +1,85 @@
...
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <bool Owner = false>
+class TIovecContiguousChunk: public IContiguousChunk
+{
+private:
+    char* Base = 0;
+    ui64 Size = 0;
```
drbasic, line 14: size_t ?
какая-то путаница, где-то ui64, где-то size_t.

cloud/filestore/libs/storage/service/service_actor_readdata.cpp
```diff
@@ -801,12 +806,15 @@ void TReadDataActor::ReplyTwoStageAndDie(const TActorContext& ctx)
     if (end <= OriginByteRange.Offset) {
         BlockBuffer->clear();
     } else {
-        BlockBuffer->ReserveAndResize(end - OriginByteRange.Offset);
-        response->Record.set_allocated_buffer(BlockBuffer.release());
+        auto length = end - OriginByteRange.Offset;
```
drbasic: const ui64 lenght

cloud/filestore/libs/storage/service/rope_utils.cpp
antonmyagkov: Из протобуфа прорастает https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/public/api/protos/data.proto#L128

cloud/filestore/libs/storage/service/rope_utils.cpp
antonmyagkov: Из протобуфа прорастает https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/public/api/protos/data.proto#L128

cloud/filestore/libs/storage/service/service_actor_readdata.cpp
antonmyagkov: fixed

PR #4793, opened 2025-12-08
issue-1751: [Filestore] Handle errors received by WriteData requests in Flush

e673:

## Current behavior

Client `fsync` and `close` requests wait until the data is successfully flushed.
If the data cannot be flushed due an error, these calls hang indefinitely and WriteBackCache gets stuck.

## Expected behavior

Client `fsync` requests should fail if an attempt to flush data was unsuccessful. For example, `ENOSPC` error is expected here.

It is allowed to drop unflushed WriteData requests for a node when the all node handles are released and flush has failed.
In this case, `close` should return an error.

### FlushNodeData

Waits until all pending and unflushed WriteData requests (related to the specific node and entered WriteBackCache before FlushNodeData call are flushed.

Stops waiting and returns an error if any of the affected WriteData requests has failed during flush. The error returned is the same as the error returned by a failed WriteData request. If multiple WriteData requests have failed, returns an error from any of those requests.

Note: `fsync` failure does not lead to the data loss, WriteBackCache will continue trying to flush the data.

### FlushAll

Waits until all WriteData requests (both pending and cached) entered WriteBackCache before FlushNodeData call are flushed.

Stops waiting and returns an error if any of the affected WriteData requests for any node has failed during flush.

### ReleaseHandle

Waits until there are no pending and unflushed WriteData requests related to the specific handle in WriteBackCache. Unlike FlushNodeData, unsuccessful flush does immediately imply ReleaseHandle failure.

Returns an error when unflushed requests have been dropped.

Does not destroys handle — it should be done by the caller code after either successful or unsuccessful result of ReleaseHandle.

The requests are dropped when both conditions are met:
- the recent attempt to flush the data was unsuccessful;
- `close` has been called for all handles related to the unflushed data.

Note: `close` failure is an indicator of data loss.

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
@@ -572,34 +572,6 @@ void TFileSystem::DoWrite(
     FSyncQueue->Enqueue(reqId, TNodeId {ino}, THandle {handle});
 
     if (wbcState == EServerWriteBackCacheState::Disabled) {
-        if (WriteBackCache) {
-            // Sync writes must include all previously cached data for
-            // the same node, so flush the node cache first and only then submit
-            // this write directly to the session.
-            // For O_DIRECT same file can be opened without O_DIRECT so we try
-            // to flush here as well
-            WriteBackCache.FlushNodeData(request->GetNodeId())
-                .Subscribe(
-                    [ptr = weak_from_this(),
-                     callback = std::move(callback),
-                     callContext = std::move(callContext),
-                     request = std::move(request)](const auto& f) mutable
-                    {
-                        const NProto::TError& error = f.GetValue();
-                        if (HasError(error)) {
-                            // Propagate flush error to the WriteData response
-                            NProto::TWriteDataResponse response;
-                            *response.MutableError() = error;
-                            callback(MakeFuture(std::move(response)));
-                        } else if (auto self = ptr.lock()) {
-                            self->Session
-                                ->WriteData(callContext, std::move(request))
-                                .Subscribe(std::move(callback));
-                        }
-                    });
-            return;
-        }
-
         Session->WriteData(callContext, std::move(request))
             .Subscribe(std::move(callback));
         return;
@@ -610,6 +582,11 @@ void TFileSystem::DoWrite(
         "Invalid EServerWriteBackCacheState value = %d",
         wbcState);
 
+    // Sync writes must include all previously cached data for
+    // the same node, so flush the node cache first and only then submit
+    // this write directly to the session.
+    // For O_DIRECT same file can be opened without O_DIRECT so we try
+    // to flush here as well
     auto flushFuture = WriteBackCache.FlushNodeData(request->GetNodeId());
     flushFuture.Subscribe(
         [ptr = weak_from_this(),
@@ -1129,16 +1106,16 @@ EServerWriteBackCacheState TFileSystem::GetServerWriteBackCacheState(
         return EServerWriteBackCacheState::Disabled;
     }
 
+    if (!Config->GetServerWriteBackCacheEnabled()) {
+        return WriteBackCache.IsEmpty() ? EServerWriteBackCacheState::Disabled
+                                        : EServerWriteBackCacheState::Draining;
+    }
+
     if (fi->flags & (O_DIRECT | O_SYNC | O_DSYNC)) {
-        return EServerWriteBackCacheState::Disabled;
+        return EServerWriteBackCacheState::Draining;
     }
 
-    if (Config->GetServerWriteBackCacheEnabled()) {
-        return EServerWriteBackCacheState::Enabled;
-    }
-
-    return WriteBackCache.IsEmpty() ? EServerWriteBackCacheState::Disabled
-                                    : EServerWriteBackCacheState::Draining;
+    return EServerWriteBackCacheState::Enabled;
 }
 
 }   // namespace NCloud::NFileStore::NFuse
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
@@ -20,6 +20,7 @@ struct THandleState
     ui64 ActiveWriteDataRequestCount = 0;
 
     // The promise is fulfilled when |RequestCount| hits 0
+    // Not initialized by default unless ReleaseHandle is requested
     NThreading::TPromise<NCloud::NProto::TError> ReadyToReleasePromise;
 };
 
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
@@ -291,7 +291,7 @@ EFlushRetryStatus TWriteBackCacheState::FlushFailed(
 
     Y_ABORT_UNLESS(
         nodeState.FlushStatus == ENodeFlushStatus::FlushRequested,
-        "Flush wasn't requested for a node %lu",
+        "Flush wasn't requested for node %lu",
         nodeId);
 
     // Fail Flush and FlushAll requests
@@ -484,6 +484,7 @@ void TWriteBackCacheState::RemoveActiveRequestFromHandleState(
 
     if (--handleState.ActiveWriteDataRequestCount == 0) {
         if (handleState.ReadyToReleasePromise.Initialized()) {
+            // The promise is initialized when ReleaseHandle was requested
             Y_ABORT_UNLESS(nodeState.HandleToReleaseCount > 0);
             nodeState.HandleToReleaseCount--;
             QueuedOperations.CompleteFlushOrReleasePromise(
@@ -506,7 +507,7 @@ void TWriteBackCacheState::DropCachedData(
     }
 
     if (nodeState.Cache.HasUnflushedRequests()) {
-        // ToDo(#1751): Implement logging for dropped unflushed requests
+        // TODO(#1751): Implement logging for dropped unflushed requests
         // similar to STORAGE_WARN macro
     }
 
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
@@ -124,6 +124,10 @@ public:
 
 private:
     // Combines acquiring mutex and executing queued operations on mutex release
+    // TQueuedOperations has custom Release method that:
+    // 1. Copies the accumulated operations to a temporary vector.
+    // 2. Releases mutex.
+    // 3. Executes the operations from the temporary vector.
     TGuard<TQueuedOperations> LockStateAndPostponeQueuedOperations() const;
 
     NThreading::TFuture<NProto::TWriteDataResponse> AddRequest(
--- a/cloud/filestore/libs/vfs_fuse/fs_impl.h
+++ b/cloud/filestore/libs/vfs_fuse/fs_impl.h
@@ -60,14 +60,14 @@ struct TReleaseRequest
 
 enum class EServerWriteBackCacheState
 {
-    // Requests should bypass WriteBackCache and go directly to the session
-    // (even if WriteBackCache is initialized)
+    // WriteBackCache is turned off
     Disabled,
 
     // Requests should go to the WriteBackCache
     Enabled,
 
-    // WriteBackCache is in the transition from Enabled to Disabled state.
+    // WriteBackCache is being turned off or a request with
+    // O_DIRECT/O_SYNC/O_DSYNC is made.
     // Requests should wait until WriteBackCache is flushed and then go
     // directly to the session
     Draining


cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -228,6 +266,7 @@ void TFileSystem::Release(
     }
 
     if (WriteBackCache) {
+        // TODO
```
e673: Q: What should I do if FlushNodeData returned non-retriable error?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -1281,62 +1299,88 @@ class TWriteBackCache::TImpl final
...
             }
         }
 
-        state.WriteRequests.clear();
-
-        if (state.FailedWriteRequests.empty()) {
-            Stats->FlushCompleted();
-            CompleteFlush(nodeState);
+        if (nonRetriableError) {
+            // TODO(#1751): handle non-retriable errors
```
e673: Q: What should I do if it is not possible to flush?

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
qkrorlqr: I would say that `fsync(fd)` and `sync()` should return errors until the problematic fd is closed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
qkrorlqr: keep problematic fd's data in the cache until this fd is closed, then just discard its data

qkrorlqr: you can also experiment with ext4 and see what it does when it fails to flush (e.g. in case the fs is out of space)

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: Done

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
```diff
@@ -97,22 +94,44 @@ void TFileSystem::SetAttr(
         }
     };
 
-    // Changing attributes interferes with WriteBackCache
-    auto flushFuture = WriteBackCache
-                           ? WriteBackCache.FlushNodeData(request->GetNodeId())
-                           : NThreading::MakeFuture();
+    const bool setSize =
```
qkrorlqr: nit: `setSize` -> `shouldSetSize`? `setSize` sounds like a name for an `std::function` which would actually set the size

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -3588,6 +3642,11 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
         test("CachedRead", getSizeFromCachedRead);
         test("DirectRead", getSizeFromDirectRead);
     }
+
+    Y_UNIT_TEST(ShouldHandleFlushErrors)
```
qkrorlqr: x

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -129,22 +129,131 @@ struct TBufferWriter
...
+    {}
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <class TResult>
+struct TPromiseSetResult
+{
+    TPromise<TResult> Promise;
+    TResult Result = {};
```
qkrorlqr: looks weird (bc `TPromise<TResult>` already contains a result)

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -840,6 +964,44 @@ class TWriteBackCache::TImpl final
...
+        if (nodeState == nullptr) {
+            return MakeFuture<NProto::TError>();
+        }
+
+        auto* handleState = nodeState->Handles.FindPtr(handle);
+        if (handleState == nullptr) {
+            return MakeFuture<NProto::TError>();
+        }
+
+        Y_ABORT_UNLESS(handleState->RequestCount > 0);
```
qkrorlqr: looks like a race condition because `ReleaseHandle()` is called from `TFileSystem::Release()` which can be called concurrently by multiple threads

we need write-back-cache tests which use multiqueue (needs to be enabled not only at the vfs_fuse layer but in the guest as well, see https://github.com/ydb-platform/nbs/tree/main/cloud/filestore/tests/fio/qemu-kikimr-mq-test)

qkrorlqr: why is node / handle absence not a problem but `RequestCount == 0` - a problem?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -1303,13 +1486,9 @@ class TWriteBackCache::TImpl final
     void PrepareFlush(TNodeState* nodeState)
     {
         Y_ABORT_UNLESS(nodeState->FlushState.Executing);
-        Y_ABORT_UNLESS(nodeState->FlushState.WriteRequests.empty());
 
-        if (!nodeState->FlushState.FailedWriteRequests.empty()) {
+        if (!nodeState->FlushState.WriteRequests.empty()) {
             // Retry write requests failed at the previous Flush attempt
```
qkrorlqr: but retries should be done by another component - `TDurableClient` (and sometimes `TSession`)

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -1374,61 +1553,46 @@ class TWriteBackCache::TImpl final
...
+                NStorage::CalculateByteCount(*it.Request) -
+                it.Request->GetBufferOffset();
 
-            Session->WriteData(std::move(callContext), request)
+            Session->WriteData(std::move(callContext), it.Request)
                 .Subscribe(
-                    [nodeState,
-                     request = std::move(request),
-                     ptr = weak_from_this()](auto future)
+                    [nodeState, &it, ptr = weak_from_this()](auto future)
```
qkrorlqr: `&it` - this looks fragile, even if we currently have a guarantee that a write request cannot be deallocated or reallocated before `Session->WriteData()` completion, it's easy to accidentally break this guarantee and cause a use-after-free bug

qkrorlqr: especially considering the fact that each write-request is a shared_ptr - why not just pass it by value

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -1472,16 +1647,48 @@ class TWriteBackCache::TImpl final
...
             auto* entry = RemoveFrontCachedEntry(nodeState);
+            const auto handle = entry->GetHandle();
             entry->FinishFlush(this);
 
             nodeState->AllEntries.Remove(entry);
             AllEntries.Remove(entry);
+
+            auto& handleState = nodeState->Handles[handle];
+            Y_ABORT_UNLESS(handleState.RequestCount > 0);
+            handleState.RequestCount--;
```
qkrorlqr: again looks like a race with `TFileSystem::Release() -> ReleaseHandle()`

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -1498,13 +1705,109 @@ class TWriteBackCache::TImpl final
...
+        // Fail all FlushNodeData requests
+        for (auto& flushRequest: nodeState->FlushRequests) {
+            QueuedOperations.FlushOrReleaseCompleted.push_back(
+                {.Promise = std::move(flushRequest.Promise), .Result = error});
+        }
+
+        nodeState->FlushRequests.clear();
+
+        FlushFailureTracker.FlushFailed(nodeState->NodeId, NextRequestId++);
+        if (FlushFailureTracker.Size() + DeletedNodeStates.size() ==
```
qkrorlqr: let's add a comment explaining what's happening here

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -129,22 +129,131 @@ struct TBufferWriter
...
+////////////////////////////////////////////////////////////////////////////////
+
+class TFlushFailureTracker
+{
+private:
+    THashMap<ui64, ui64> NodeIdToFlushFailureRequestIdMap;
+    TSet<ui64> FlushFailureRequestIdMap;
+
+public:
+    void FlushSucceeded(ui64 nodeId) {
```
qkrorlqr: { to the next line

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -1498,13 +1705,109 @@ class TWriteBackCache::TImpl final
...
+        const bool shouldDropCache =
+            nodeState->ReleasingHandleCount == nodeState->Handles.size();
+
+        if (shouldDropCache) {
+            // Reset Flush state because failed requests will not be retried
+            nodeState->FlushState.Executing = false;
+            nodeState->FlushState.WriteRequests.clear();
+            nodeState->FlushState.AffectedWriteDataEntriesCount = 0;
+
+            DropNodeData(nodeState, error);
```
qkrorlqr: why are we not deleting failed flush request id from `FlushFailureTracker`?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
```diff
@@ -61,9 +61,38 @@ class TWriteBackCache final
         TCallContextPtr callContext,
         std::shared_ptr<NProto::TWriteDataRequest> request);
 
-    NThreading::TFuture<void> FlushNodeData(ui64 nodeId);
```
qkrorlqr, line 100: this cache looks like a class with multiple responsibilities, at the very least:
* it stores writes and is able to serve read requests for the data that it contains
* it implements data flushing - sends those cached writes to the layer below it and signals completion

these 2 parts can be implemented as 2 different entities - each of them would be simpler than the current `TWriteBackCache` monster class:
1. `TDataCache` - serves reads and writes, has a way to extract all the data currently cached for a node
2. `TFlusher` - serves flush-node / flush requests, takes the data from `TDataCache`, writes it to the underlying layer, signals completion
3. periodic flushing can be implemented as yet another entity which would just call `TFlusher::Flush()` from time to time 

also, after reading the code that we have in `write_back_cache.cpp` I'm starting to think that it's better to rewrite it in a simpler way - looking at this code I can't be confident in its correctness and probably no one else can be confident that it's correct

write-back-cache was initially supposed to be a simple queue:
* append new `TWriteDataRequest` upon `WriteData`
* periodically try to flush the front of the queue and pop-front upon success

parallelism can be achieved simply by having multiple such queues, the queue for each `TWriteDataRequest` can be selected simply in a round-robin manner or hashed by node-id + range-id pair if that makes ReadData implementation simpler (this can be improved later on - for example, we can select the queue with the most free space) 

out of space in all queues? just don't reply to this `TWriteDataRequest` and put it into some "pending" request queue __outside of the write-back-cache__ - i.e. make a separate component for it, it serves a different purpose, it's about the not-yet-acked requests and it can be a simple queue - we don't need to use it to serve reads (again - these requests are not acked yet)

we can have a separate sequencer for the `TWriteDataRequest`s which are sent into write-back-cache - upon `Flush` we can just wait until the processed request counter passes the max sequence number that we had at the moment when `Flush` was called 

no status tracking needed per request - each request which is still present in the write-back-cache queues is not persisted yet

so:
* a separate write-back-cache-queue with extremely simple flushing logic - flush only the front of the queue, flush-promise can be set as soon as its sequence number becomes smaller than the sequence number at the head of the queue
* a separate pending request queue implemented as a separate component which doesn't have `ReadData` interface
* a separate request sequencer
* a separate flusher (one per queue - for simplicity; flush-all can be implemented by waiting for all queues)

qkrorlqr (changes requested):
It's very hard to review such PRs. You could at least split it into 2:
* one with the `ReleaseHandle()` logic
* another one - about `Flush*()` error processing

but maybe we should do this completely differently, see my other (very long) comment

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: There is a lock guard in the beginning of `TWriteBackCache::ReleaseHandle`
https://github.com/ydb-platform/nbs/blob/73c96788d5099b21c88bb9853cd1d5da1f59ac5f/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp#L967-L969

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: But `TPromise<TResult>` does not contain a value here.
Please, elaborate.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: WriteBackCache holds only nodes/handles with unflushed requests.
Absence of node/handle here = no pending/cached requests, just proceed with handle release.
`RequestCount == 0` is a problem because handle state should be removed from WriteBackCache when all related WriteData requests are flushed.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Yes, this is true.
But what should we do here?
At the previous meeting we decided that we woud retry flush in any case.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: I will replace `&it` with an array index when my ivm get back online.

> each write-request is a shared_ptr - why not just pass it by value

I need also `.Response` field:
https://github.com/ydb-platform/nbs/blob/73c96788d5099b21c88bb9853cd1d5da1f59ac5f/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp#L1573

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: There is a lock guard
https://github.com/ydb-platform/nbs/blob/73c96788d5099b21c88bb9853cd1d5da1f59ac5f/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp#L1644

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Looks like I forgot about it - will fix soon.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
e673: I agree with the idea to split it into multiple classes in order to improve readability and understanding.

But I don't like the idea of rewriting the algorithm and starting from the beginning. The current complex logic is a result of the evolution of that simple logic after multiple discussions.

> parallelism can be achieved simply by having multiple such queues ...  flush only the front of the queue

What if a client has more concurrency than the number of these queues? It will decrease the performance.

> hashed by node-id + range-id pair 

The requests are not expected to be aligned and may overlap. And I don't like the idea of splitting requests. So the idea to distribute the requests for the same node over different queues doesn't seem reasonable.

In the current implementation:
- each node has a separate queue (but all queues share the same backing ring buffer);
- instead of processing the requests one-by-one, flush takes a batch;
- the batch is optimized: adjacent requests are merged, overlapping requests are combined.

> out of space in all queues? just don't reply to this TWriteDataRequest and put it into some "pending" request queue outside of the write-back-cache 

This is the current implementation.

> a separate pending request queue implemented as a separate component which doesn't have ReadData interface

But with a small nuance: we decided that Flush should wait also for pending requests. SequenceId acts as a barrier not only for cached requests but also for flushed requests.

> we can just wait until the processed request counter passes the max sequence number that we had at the moment when Flush was called ... flush-promise can be set as soon as its sequence number becomes smaller than the sequence number at the head of the queue

This is the current implementation.

But we also need to have the logic to fail Flush.

> flush-all can be implemented by waiting for all queues

It was implemented this way initially.
But then it was optimized to just watch at the sequence number in the global queue.

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
```diff
@@ -106,22 +103,44 @@ void TFileSystem::SetAttr(
...
+            // Propagate Flush error to SetAttr response
+            const NProto::TError& flushResult = future.GetValue();
+            if (HasError(flushResult)) {
+                NProto::TSetNodeAttrResponse response;
+                *response.MutableError() = flushResult;
+                callback(NThreading::MakeFuture(std::move(response)));
+                return;
             }
+
+            self->WriteBackCache.SetCachedNodeSize(
```
e673, line 138: Note: node size is updated no matter of SetNodeAttr was successful or not (like it was before).
This will be resolved in a separate PR - I will introduce a universal mechanism to execute requests like SetAttr, ReadData and WriteData with O_DIRECT/O_DSYNC.

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
e673: Renamed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: No more relevant

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: No more relevant

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673, line 184: This line looks strange: `ReplyError` looks at the last parameter and will return `S_OK` to a client.
Refactoring is needed (split into ReplyError and ReplySuccess).

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
e673: The refactoring is completed.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: Done - the data is stored in cache until client releases all handles.

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -210,17 +233,21 @@ void TFileSystem::ReleaseImpl(
...
-            const auto& response = future.GetValue();
-            if (CheckResponse(self, *callContext, req, response)) {
-                self->ReplyError(*callContext, response.GetError(), req, 0);
-            }
-        });
+                const auto& response = future.GetValue();
+                if (CheckResponse(self, *callContext, req, response) &&
+                    self->CheckError(*callContext, req, writeBackCacheError))
+                {
+                    self->ReplyError(*callContext, {}, req, 0);
```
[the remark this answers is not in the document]
e673: > Capturing a reference parameter this way stores the reference

That's wrong. All arguments are captured by value ([=]).

cloud/filestore/libs/vfs_fuse/write_back_cache/queued_operations.cpp
```diff
@@ -18,13 +18,37 @@ struct TWriteDataPromiseCompletedEvent
...
+struct TFlushOrReleasePromiseFailedEvent
+{
+    NThreading::TPromise<NCloud::NProto::TError> Promise;
+    const NCloud::NProto::TError& Error;
 
     void Invoke()
     {
-        Promise.SetValue();
+        Promise.SetValue(Error);
     }
```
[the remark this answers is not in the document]
e673: In all use cases `Error` field outlive the lock.
But I agree, the code looks fragile and this optimization looks unnecessary.
Changed to value.

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -847,11 +889,16 @@ void TFileSystem::Flush(
...
+        auto writeBackCacheFlushFuture = WriteBackCache.FlushNodeData(ino);
+        auto waitAll =
+            NWait::WaitAll(fsyncQueueFuture, writeBackCacheFlushFuture);
+        future = waitAll.Apply(
+            [=](const auto&)
+            {
+                const auto& flushError = writeBackCacheFlushFuture.GetValue();
+                const auto& fsyncQueueError = fsyncQueueFuture.GetValue();
+                return HasError(flushError) ? flushError : fsyncQueueError;
+            });
```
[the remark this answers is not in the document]
e673: > return the flush error when present

This is done exactly:

```
return HasError(flushError) ? flushError : fsyncQueueError;
```

cloud/filestore/libs/vfs_fuse/write_back_cache/queued_operations.cpp
```diff
@@ -18,13 +18,37 @@ struct TWriteDataPromiseCompletedEvent
...
 {
-    NThreading::TPromise<void> Promise;
+    NThreading::TPromise<NProto::TWriteDataResponse> Promise;
+    const NCloud::NProto::TError& Error;
+
+    void Invoke()
+    {
+        NProto::TWriteDataResponse response;
+        *response.MutableError() = Error;
+        Promise.SetValue(std::move(response));
```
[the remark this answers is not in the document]
e673: In all use cases Error field outlive the lock.
But I agree, the code looks fragile and this optimization looks unnecessary.
Changed to value.

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
```diff
@@ -106,22 +103,44 @@ void TFileSystem::SetAttr(
...
+                request->GetNodeId(),
+                request->GetUpdate().GetSize());
+
+            self->Session->SetNodeAttr(callContext, std::move(request))
```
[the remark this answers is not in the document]
e673: You are right.
This behavior was present initially and will be resolved in a separate PR.
https://github.com/ydb-platform/nbs/pull/4793#discussion_r2880873399

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
```diff
@@ -63,6 +76,15 @@ struct TNodeState
...
     TDeque<TFlushRequest> FlushRequests;
 
+    // Holds active request handles and tracks handle release
+    THashMap<ui64, THandleState> Handles;
+
+    // Number of handles that have been requested for release.
+    // When HandleReleaseCount == Handles.size(), the next flush failure
+    // will cause all pending requests to be failed and cached data to be
+    // dropped
+    size_t HandleReleaseCount = 0;
```
SvartMetal: HandlesToReleaseCount

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
```diff
@@ -63,6 +76,15 @@ struct TNodeState
     // strictly increasing so newer flush requests have larger SequenceId.
     TDeque<TFlushRequest> FlushRequests;
 
+    // Holds active request handles and tracks handle release
```
SvartMetal, line 80: Map from handle id? Add a comment please

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -244,6 +279,76 @@ void TWriteBackCacheState::FlushSucceeded(ui64 nodeId, size_t requestCount)
...
+    }
+
+    nodeState.FlushRequests.clear();
+    FlushAllRequestQueue.clear();
+
+    if (nodeState.Handles.size() != nodeState.HandleReleaseCount) {
+        // Keep status ENodeFlushStatus::FlushRequested if flush is retried
+        return EFlushRetryStatus::ShouldRetry;
+    }
+
```
SvartMetal, line 320: Let's add comment that all handles here are going to be released and we are allowed to drop the data from cache

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -244,6 +279,76 @@ void TWriteBackCacheState::FlushSucceeded(ui64 nodeId, size_t requestCount)
     EvictUnpinnedFlushedEntries(nodeId, nodeState);
 }
 
+EFlushRetryStatus TWriteBackCacheState::FlushFailed(
+    ui64 nodeId,
+    const TError& error)
+{
+    auto guard = Guard(QueuedOperations);
```
SvartMetal: Let's find a more readable name, it's not so obvious that this guard will call queued operations when being destroyed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -244,6 +279,76 @@ void TWriteBackCacheState::FlushSucceeded(ui64 nodeId, size_t requestCount)
...
+    // Drop node data - pending and unflushed requests
+    while (nodeState.Cache.HasPendingRequests()) {
+        auto request = nodeState.Cache.DequeuePendingRequest();
+        QueuedOperations.FailWriteDataPromise(
+            std::move(request->AccessPromise()),
+            error);
+    }
+
+    if (nodeState.Cache.HasUnflushedRequests()) {
+        ReportWriteBackCacheDataLossError(
```
SvartMetal: This should not be a critical event, if user's filesystem is out of space - it is normal to drop the data (and let the user know via an error code)

SvartMetal: But we would like to have the corresponding metric and log when drop happens

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -264,6 +370,7 @@ TFuture<TWriteDataResponse> TWriteBackCacheState::AddRequest(
     const ui64 nodeId = request->GetNodeId();
 
     auto& nodeState = Nodes.GetOrCreateNodeState(nodeId);
+    nodeState.Handles[request->GetHandle()].RequestCount++;
```
SvartMetal: IncrementActiveWriteDataRequestCount

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -381,4 +488,21 @@ void TWriteBackCacheState::EvictUnpinnedFlushedEntries(
     }
 }
 
+void TWriteBackCacheState::DecrementHandleCount(
```
SvartMetal: DecrementActiveWriteDataRequestCount

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
```diff
@@ -7,12 +7,24 @@
...
 #include <util/generic/set.h>
 
 namespace NCloud::NFileStore::NFuse::NWriteBackCache {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct THandleState
+{
+    // Number of pending and unflushed requests
+    ui64 RequestCount = 0;
```
SvartMetal: ActiveWriteDataRequestCount

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
SvartMetal: But actually this function does much more than a simple decrement...

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
```diff
@@ -24,13 +24,25 @@ namespace NCloud::NFileStore::NFuse::NWriteBackCache {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+enum class EFlushRetryStatus
+{
+    // Flusher should stop trying to flush data - the error is unrecoverable
+    ShouldNotRetry,
+
+    // Flusher should retry attempts to flush data
```
SvartMetal: It would be more helpful to describe situations when retry could happen and when not

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
```diff
@@ -1223,6 +1219,136 @@ Y_UNIT_TEST_SUITE(TWriteBackCacheTest)
             {.WithCacheRecreation = true, .ZeroCopyWriteEnabled = true});
     }
 
+    Y_UNIT_TEST(FullyRandomizedWithWriteFailures)
+    {
+        const ui32 NodeCount = 4;
+        const ui32 RecreationProbability = 10;
```
SvartMetal: Let's add comment about units (percentage)

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
```diff
@@ -1424,16 +1550,17 @@ Y_UNIT_TEST_SUITE(TWriteBackCacheTest)
...
         // to write data but fails
         UNIT_ASSERT_VALUES_EQUAL(1, writeAttempts.load());
-        UNIT_ASSERT(!flushFuture.HasValue());
+        UNIT_ASSERT(flushFuture.HasValue());
+        UNIT_ASSERT(HasError(flushFuture.GetValue()));
 
         // WriteData request from Flush succeeds after WriteAttemptsThreshold
         // attempts.
         for (int i = 1; i < WriteAttemptsThreshold; i++) {
+            b.Timer->Sleep(TDuration::Seconds(1));
```
SvartMetal: Can we get this value from config?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
```diff
@@ -1424,16 +1550,17 @@ Y_UNIT_TEST_SUITE(TWriteBackCacheTest)
...
             b.RunAllScheduledTasks();
         }
 
-        UNIT_ASSERT(flushFuture.HasValue());
-        UNIT_ASSERT_EQUAL(writeAttempts, WriteAttemptsThreshold);
+        UNIT_ASSERT_VALUES_EQUAL(writeAttempts.load(), WriteAttemptsThreshold);
```
SvartMetal: Expected vs actual?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1306,11 +1307,22 @@ class TFileSystemLoop final
...
-    void StopAsyncOnWriteBackCacheFlushed(TPromise<void> stopCompleted)
+    void StopAsyncOnWriteBackCacheFlushed(
+        TPromise<void> stopCompleted,
+        const NProto::TError& error)
     {
-        Y_ABORT_UNLESS(
-            WriteBackCache && WriteBackCache.IsEmpty(),
-            "WriteBackCache was not emptied after FlushAllData");
+        if (HasError(error)) {
+            ReportWriteBackCacheCreatingOrDeletingError(Sprintf(
```
SvartMetal: Isn't it a valid situation?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1306,11 +1307,22 @@ class TFileSystemLoop final
...
+                "[f:%s][c:%s] (DestroySession) WriteBackCache::FlushAllData "
+                "failed: %s",
+                Config->GetFileSystemId().Quote().c_str(),
+                Config->GetClientId().Quote().c_str(),
+                FormatError(error).c_str()));
+        } else {
+            Y_ABORT_UNLESS(
```
SvartMetal: Do we really need to crash here?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1379,9 +1391,14 @@ class TFileSystemLoop final
 
         // We need to cleanup WriteBackCache file and directories
         if (WriteBackCache) {
-            // Deleting file when it contains unflushed requests
-            // will result in data loss
-            Y_ABORT_UNLESS(WriteBackCache.IsEmpty());
+            if (!WriteBackCache.IsEmpty()) {
+                ReportWriteBackCacheDataLossError(Sprintf(
```
SvartMetal: Should it be a data loss?

It may be related to the guest problem but not ours

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
SvartMetal, line 138: We should discuss this more thoroughly 

Let me meditate about this code...

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
e673: Renamed to `HandleToReleaseCount` (we decided not to use plurals before Count).

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
e673: Added comment decribing key meaning

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Added comment and extracted data drop to a separate function

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Changed to `LockStateAndPostponeQueuedOperations()`

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Added a metric value.
Log message will be added in a separate logging-related PR
(we discusses with @qkrorlqr and want to implement unified macro similar to `STORAGE_WARN` everywhere in writebackcache).

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Added function `AddActiveRequestToHandleState`

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Renamed to `RemoveActiveRequestFromHandleState`.

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
e673: Done

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
e673: Comments added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
e673: Added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
e673: Extracted a constant value used in `TBootstrap`

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
e673: Switched

cloud/filestore/libs/vfs_fuse/loop.cpp
e673: Yes, it is valid. I've changed the code

cloud/filestore/libs/vfs_fuse/loop.cpp
e673: No. I've changed the code

cloud/filestore/libs/vfs_fuse/loop.cpp
e673: Removed critical error and replaced with log message

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
e673: I want to add the following logic:
- Add method `FlushAndLockCache` - it will ensure that the cache has no requests and all pending requests will wait until lock is released.
- Any operation like SetNodeAttr, WriteData with O_DIRECT/O_SYNC/O_DSYNC will be executed under this lock.

Alternative implementation: `EmptyAndLockCache` - it will wait until node cache is free of unflushed, flushed requests and state refs.

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
```diff
@@ -7,12 +7,24 @@
...
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct THandleState
+{
+    // Number of pending and unflushed requests associated with the handle
+    ui64 ActiveWriteDataRequestCount = 0;
+
+    // The promise is fulfilled when |RequestCount| hits 0
+    NThreading::TPromise<NCloud::NProto::TError> ReadyToReleasePromise;
```
qkrorlqr, line 24: nit: NCloud:: not needed

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
```diff
@@ -33,7 +45,8 @@ struct TFlushRequest
 
     // The promise is fulfilled when there are no pending or unflushed requests
     // with SequenceId <= TFlushRequest::SequenceId (for a node or global)
-    NThreading::TPromise<void> Promise = NThreading::NewPromise();
+    NThreading::TPromise<NCloud::NProto::TError> Promise =
```
qkrorlqr, line 49: nit: NCloud:: not needed

cloud/filestore/libs/vfs_fuse/write_back_cache/queued_operations.cpp
```diff
@@ -18,13 +18,37 @@ struct TWriteDataPromiseCompletedEvent
...
+    {
+        NProto::TWriteDataResponse response;
+        *response.MutableError() = std::move(Error);
+        Promise.SetValue(std::move(response));
+    }
+};
+
+struct TFlushOrReleasePromiseCompletedEvent
+{
+    NThreading::TPromise<NCloud::NProto::TError> Promise;
```
qkrorlqr, line 36: nit: NCloud:: not needed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -244,6 +281,53 @@ void TWriteBackCacheState::FlushSucceeded(ui64 nodeId, size_t requestCount)
...
+    ui64 nodeId,
+    const TError& error)
+{
+    auto guard = LockStateAndPostponeQueuedOperations();
+
+    auto& nodeState = Nodes.GetOrCreateNodeState(nodeId);
+
+    Y_ABORT_UNLESS(
+        nodeState.FlushStatus == ENodeFlushStatus::FlushRequested,
+        "Flush wasn't requested for a node %lu",
```
qkrorlqr: nit: "a node" -> "node"

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -381,4 +467,65 @@ void TWriteBackCacheState::EvictUnpinnedFlushedEntries(
...
+
+void TWriteBackCacheState::RemoveActiveRequestFromHandleState(
+    TNodeState& nodeState,
+    ui64 handle)
+{
+    auto& handleState = nodeState.Handles[handle];
+    Y_ABORT_UNLESS(handleState.ActiveWriteDataRequestCount > 0);
+
+    if (--handleState.ActiveWriteDataRequestCount == 0) {
+        if (handleState.ReadyToReleasePromise.Initialized()) {
```
qkrorlqr, line 486: in which case can `ReadyToReleasePromise` be not initialized?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -381,4 +467,65 @@ void TWriteBackCacheState::EvictUnpinnedFlushedEntries(
...
+{
+    while (nodeState.Cache.HasPendingRequests()) {
+        auto request = nodeState.Cache.DequeuePendingRequest();
+        QueuedOperations.FailWriteDataPromise(
+            std::move(request->AccessPromise()),
+            error);
+    }
+
+    if (nodeState.Cache.HasUnflushedRequests()) {
+        // ToDo(#1751): Implement logging for dropped unflushed requests
```
qkrorlqr: TODO

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
```diff
@@ -65,9 +82,13 @@ class TWriteBackCacheState
     NThreading::TFuture<NProto::TWriteDataResponse> AddWriteDataRequest(
         std::shared_ptr<NProto::TWriteDataRequest> request);
 
-    NThreading::TFuture<void> AddFlushRequest(ui64 nodeId);
+    NThreading::TFuture<NCloud::NProto::TError> AddFlushRequest(ui64 nodeId);
```
qkrorlqr, line 85: nit: NCloud:: not needed - here and in other places

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
```diff
@@ -95,7 +116,16 @@ class TWriteBackCacheState
...
     void FlushSucceeded(ui64 nodeId, size_t requestCount);
 
+    // Inform that the flush has failed - the error should be propagated to
+    // Flush, FlushAll and ReleaseHandle requests
+    EFlushRetryStatus FlushFailed(
+        ui64 nodeId,
+        const NCloud::NProto::TError& error);
+
 private:
+    // Combines acquiring mutex and executing queued operations on mutex release
```
qkrorlqr, line 126: > executing queued operations on mutex release

how? the function seems to be using a standard `Guard` or am I missing something?

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -573,19 +610,24 @@ void TFileSystem::DoWrite(
...
-             callContext = std::move(callContext),
-             request = std::move(request)](const auto& f) mutable
-            {
-                f.GetValue();
-                if (auto self = ptr.lock()) {
-                    self->Session->WriteData(callContext, std::move(request))
-                        .Subscribe(std::move(callback));
-                }
-            });
+    auto flushFuture = WriteBackCache.FlushNodeData(request->GetNodeId());
```
qkrorlqr, line 590: looks like the same piece of code as in the `wbcState == EServerWriteBackCacheState::Disabled` branch

if so, let's try to rewrite this a bit so that there's no duplication?

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
e673: Without NCloud:
`error: no member named 'TError' in namespace 'NCloud::NFileStore::NProto'; did you mean 'perror'?`

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
e673: Without NCloud:
error: no member named 'TError' in namespace 'NCloud::NFileStore::NProto'; did you mean 'perror'?

cloud/filestore/libs/vfs_fuse/write_back_cache/queued_operations.cpp
e673: Without NCloud:
error: no member named 'TError' in namespace 'NCloud::NFileStore::NProto'; did you mean 'perror'?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: fixed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Added comments.
It is not initialized unless ReleaseHandle is called

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Changed ToDo -> TODO

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
e673: Added a comment describing what it is doing

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: Code has been deduplicated

cloud/filestore/libs/vfs_fuse/fs_impl_attr.cpp
e673: An internal ticket has been created

PR #4803, opened 2025-12-09
issue-568: implemented block checksum calculation and logging to tablet profile log for ReadData and WriteData requests

qkrorlqr:
### Notes
Checksum calculation is disabled by default - can be enabled via `TStorageServiceConfig::BlockChecksumsInProfileLogEnabled`

For unaligned requests we calculate checksums only for the requested parts of the blocks. This makes it harder to compare what's being written to what's being read because at the time of read we might've converted unaligned writes to full blocks - to mitigate this (to some degree, not perfectly of course) the code doesn't calculate crc for the whole block but finds the last non-zero byte and calculates crc for the subrange up to this last non-zero byte. This should let us compare WriteData request checksums for unaligned appends to the checksums that we see for ReadData requests.

### Issue

### DIFF
--- a/cloud/filestore/libs/diagnostics/profile_log_events.cpp
+++ b/cloud/filestore/libs/diagnostics/profile_log_events.cpp
@@ -56,6 +56,13 @@ void InitProfileLogLockRequestInfo(
 
 ui32 CalculateChecksum(TStringBuf buf)
 {
+    //
+    // We calculate the checksum only up to the last non-zero byte, i.e. we
+    // discard zero suffixes. This is needed to simplify checksum comparisons
+    // between unaligned appends coming from the client and full block writes
+    // for the same block coming from the tablet's internal logic.
+    //
+
     ui64 len = buf.size();
     while (len > 0) {
         constexpr auto WordSize = sizeof(ui64);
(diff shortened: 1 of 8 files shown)


cloud/filestore/libs/diagnostics/profile_log_events.cpp
```diff
@@ -47,6 +52,48 @@ void InitProfileLogLockRequestInfo(
...
+    // We calculate the checksum only up to the last non-zero byte, i.e. we
+    // discard zero suffixes. This is needed to simplify checksum comparisons
+    // between unaligned appends coming from the client and full block writes
+    // for the same block coming from the tablet's internal logic.
+    //
+
+    ui64 len = buf.size();
+    while (len > 0) {
+        constexpr auto WordSize = sizeof(ui64);
+        if (len % WordSize == 0 && len >= WordSize) {
```
yegorskii, line 69: а точно нужно len % WordSize == 0? Если длина буфера не кратна WordSize мы сразу свалимся в побайтовое сравнение, хотя могли бы прочитать несколько WordSize. Ну типа длина буфера 20 байт и пара слов там точно лежит хотя длина и не кратна

qkrorlqr: > а точно нужно len % WordSize == 0? 

yes, otherwise we'll be reading unaligned words for the whole length of the buffer

> Если длина буфера не кратна WordSize мы сразу свалимся в побайтовое сравнение, хотя могли бы прочитать несколько WordSize. Ну типа длина буфера 20 байт и пара слов там точно лежит хотя длина и не кратна

we will only read 4 bytes byte-by-byte in this case, then `len` will become divisible by `WordSize` and we'll continue reading word-by-word

yegorskii (approved):

PR #4830, opened 2025-12-12
[Disk manager] correctly handle deletion of non-existent disk; improve test on concurrent overlay disk creation and deletion

gy2411:
In https://github.com/ydb-platform/nbs/pull/4339 there was a bug. This bug leads to a leaking base disk slot and, consequenly, to hanging retire base disk task.

Scenario:
- Overlay disk is creating
- Concurrenlty, this overlay disk is deleting
- Disk is deleted successfully
- CreateOverlayDiskTask successfully acquires base disk slot
- CreateOverlayDiskTask sees that disk is deleted and starts to cancel
- On cancel, CreateOverlayDiskTask sees that the disk is deleted and does not proceed with releasing base disk

In this pr we make the following changes:
- Task does not stop after storage.DeleteDisk call even if the is is already deleted.
- Improve the test using CheckBaseDiskSlotReleased.
- Increase DeletedDiskExpirationTimeout parameter in test config. Otherwise, the deleted disk gets cleaned up from the database too quickly and the test does not work.

Without the fix, the improve test fails with error:
```
common.go:711: 
                Error Trace:    /-S/cloud/disk_manager/internal/pkg/facade/testcommon/common.go:711
                                                        /-S/cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_test.go:1273
                Error:          Received unexpected error:
                                Non retriable error, Silent=false: internal inconsistency: slot &{overlayDiskID:TestDiskServiceDeleteDiskWhenCreationIsInFlightAAAA overlayDiskKind:0 overlayDiskSize:134217728 baseDiskID:base-6c1b5f58-9f2f-44fb-8f27-6856e31de178 imageID:TestDiskServiceDeleteDiskWhenCreationIsInFlightAAAA zoneID:zone-a allottedSlots:1 allottedUnits:1 releasedAt:{wall:0 ext:62135596800 loc:0x1fe8200} targetZoneID: targetBaseDiskID: targetAllottedSlots:0 targetAllottedUnits:0 generation:0 status:0} should be released
```

### DIFF
--- a/cloud/disk_manager/internal/pkg/services/disks/delete_disk_task.go
+++ b/cloud/disk_manager/internal/pkg/services/disks/delete_disk_task.go
@@ -72,14 +72,12 @@ func (t *deleteDiskTask) deleteDisk(
 		return err
 	}
 
+	if len(diskMeta.ZoneID) == 0 {
+		// Disk does not exist, nothing to do.
+		return nil
+	}
+
 	zoneID := diskMeta.ZoneID
-	if len(zoneID) == 0 {
-		// Should proceed deleting even if the disk is absent it the storage.
-		zoneID = t.request.Disk.ZoneId
-	}
-	if len(zoneID) == 0 {
-		return t.storage.DiskDeleted(ctx, diskID, time.Now())
-	}
 
 	taskID, err := t.scheduler.ScheduleTask(
 		headers.SetIncomingIdempotencyKey(
--- a/cloud/disk_manager/internal/pkg/services/disks/delete_disk_task_test.go
+++ b/cloud/disk_manager/internal/pkg/services/disks/delete_disk_task_test.go
@@ -223,10 +223,9 @@ func TestDeleteDiskTaskWithNonExistentDisk(t *testing.T) {
 	scheduler := tasks_mocks.NewSchedulerMock()
 	poolService := pools_mocks.NewServiceMock()
 	nbsFactory := nbs_mocks.NewFactoryMock()
-	nbsClient := nbs_mocks.NewClientMock()
 	execCtx := newExecutionContextMock()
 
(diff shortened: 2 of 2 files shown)


cloud/disk_manager/internal/pkg/services/disks/delete_disk_task.go
```diff
@@ -72,12 +72,14 @@ func (t *deleteDiskTask) deleteDisk(
...
-	}
-
 	zoneID := diskMeta.ZoneID
+	if len(zoneID) == 0 {
+		// Should proceed deleting even if the disk is absent it the storage.
+		zoneID = t.request.Disk.ZoneId
+	}
+	if len(zoneID) == 0 {
+		return t.storage.DiskDeleted(ctx, diskID, time.Now())
+	}
```
gy2411, line 78: Нужно продолжать удаление диска, даже если диска нет в сторадже.

В частности, таск удаления диска, отсутсвтующего в сторадже, должен отправить в NBS запрос на удаление этого диска.

Поэтому при пустой ZoneID не делаем return, а забираем зону из запроса (как мы это и делали до того, как посадили баг: https://github.com/ydb-platform/nbs/pull/4339/files#diff-1605caf3e850a302b598eb7fed17f3d13cae8a4defd1b6c2f18df892bcbf8060)

gy2411: Выделил эту правку в отдельный pr https://github.com/ydb-platform/nbs/pull/4910

gy2411: И всё же лучше это место вернуть, как было.

Issue #3792, opened 2025-06-30
TFileRingBuffer: do not crash when capacity is changed — resize buffer instead


e673:
https://github.com/ydb-platform/nbs/issues/1751

Currently, the code looks are follows:
https://github.com/ydb-platform/nbs/blob/6014fd79fb3aab3bd04a518954d5fb831e7f9338/cloud/storage/core/libs/common/file_ring_buffer.cpp#L120-L126

It means that if the configuration is changed, the server that is currently using `WriteBackCache` or `HandleOpsQueue` in some sessions will go into crash loop after restart. Buffer resize should happen in this case instead.

Buffer can be resized immediately when the following conditions are met:

1. `ReadPos <= WritePos` (no wrapping).
2. `WritePos <= TargetCapacity`.

Otherwise we should wait until it is poossible to resize buffer and prevent from writing new data:

1. Add field `TargetCapacity` to the header.
2. Prohibit PushBack if `TargetCapacity != Capacity` and any of the conditions above turn from `true` to `false` (but it is ok to write data when the conditions are `false`).

⚠️ There may exist references to the mapped memory. It is prohibited to unmap it unless the ring buffer is empty.
It would be simpler to prohibit PushBack until the buffer is empty, then resize it and enable PushBack.

PR #4833, opened 2025-12-12
issue-1751: Support metadata by TFileRingBuffer

e673:

Add a buffer for storing arbitrary metadata in the persistent storage.

Previous format:
```
[Header][Data]
```

New format:
```
[Header]...[Metadata]...[Data]
   │           △           △
   └───────────┴───────────┘
```

The metadata is to be used by `TWriteBackCache`:
- Store `FileSystemId` and `SessionId` — additional checks will prevent from accidental using of the wrong state file.
- Store information about `TWriteDataEntry` serialization methods in order to support backward compatibility (currently, release of a new version of nfs-vhost may corrupt the data).

The metadata capacity may grow — a client is guaranteed to allocate up to the requested capacity.

### DIFF
--- a/cloud/storage/core/libs/common/file_ring_buffer_ut.cpp
+++ b/cloud/storage/core/libs/common/file_ring_buffer_ut.cpp
@@ -616,7 +616,7 @@ Y_UNIT_TEST_SUITE(TFileRingBufferTest)
         UNIT_ASSERT_EQUAL(28, rb.GetMaxAllocationBytesCount());
     }
 
-    Y_UNIT_TEST(ShouldNotSupportMetadata)
+    Y_UNIT_TEST(ShouldGetAndSetMetadata_ZeroMetadataCapacity)
     {
         const auto f = TTempFileHandle();
         const ui32 len = 36;
@@ -633,7 +633,7 @@ Y_UNIT_TEST_SUITE(TFileRingBufferTest)
         UNIT_ASSERT_VALUES_EQUAL("AAA", rb->Front());
     }
 
-    Y_UNIT_TEST(ShouldSupportMetadata)
+    Y_UNIT_TEST(ShouldGetAndSetMetadata_NonZeroMetadataCapacity)
     {
         const auto f = TTempFileHandle();
         const ui32 len = 36;
--- a/cloud/storage/core/libs/common/file_ring_buffer.h
+++ b/cloud/storage/core/libs/common/file_ring_buffer.h
@@ -26,6 +26,19 @@ private:
     std::unique_ptr<TImpl> Impl;
 
 public:
+    /** Creates or opens an existing file ring buffer stored in the file.
+    *
+    * Argument dataCapacity specifies the size of the data area in bytes, it
+    * has effect only when creating a new buffer. When opening an existing
+    * buffer, the argument is ignored and the existing data capacity is used.
+    *
+    * Argument metadataCapacity specifies the size of the metadata area in
+    * bytes. If the existing buffer has different metadata capacity, the
+    * metadata area is resized to the specified capacity, preserving existing
+    * metadata. If the size of the existing metadata is greater than the
+    * specified capacity, the metadata area is shrunk to fit the existing
+    * metadata.
+    */
     TFileRingBuffer(
         const TString& filePath,
         ui64 dataCapacity,


cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -137,6 +140,46 @@ class TEntriesData
...
+    {
+        if (data.size() > sizeof(TEntryHeader)) {
+            Header = reinterpret_cast<TEntryHeader*>(data.data());
+            Body = data.subspan(sizeof(TEntryHeader));
+        }
+    }
+
+    TStringBuf Get() const
+    {
+        return Header && Header->DataSize > 0 && Header->DataSize <= Body.size()
```
neihar: Minor: Can we have invalid data here or we can always return TStringBuf(Body.data(), Header->DataSize) and abort if Header is null?

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -66,16 +78,13 @@ struct Y_PACKED TEntryHeader
...
+    std::span<char> Data;
 
     template <class T = char>
     T* GetPtr(ui64 pos, ui64 size = sizeof(T)) const
     {
-        char* begin = Begin + pos;
-        const char* end = begin + size;
-        return Begin <= begin && begin <= end && end <= End
-                   ? reinterpret_cast<T*>(begin)
+        return pos < Data.size() && size <= Data.size() - pos
```
neihar, line 94: "size <= Data.size() - pos" to "pos + size <= Data.size()"

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+
+        if (Map.Length() == 0) {
+            dataCapacity = AlignUp(dataCapacity, sizeof(ui64));
+            if (metadataCapacity > 0) {
+                metadataCapacity = AlignUp(
+                    metadataCapacity + sizeof(TEntryHeader),
+                    sizeof(ui64));
+            }
+            const ui64 realSize = sizeof(TExtendedHeader) + dataCapacity +
+                                  sizeof(TEntryHeader) + metadataCapacity;
```
neihar: add comment that " sizeof(TEntryHeader) + metadataCapacity" it is stored size of metadata

probably adding "using TMetadataHeader = TEntryHeader" will provide more clarity to the code in general

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+            Map.ResizeAndRemap(0, realSize);
+            Header()->Version = VERSION_EXT_HEADER;
+            Header()->HeaderSize = sizeof(TExtendedHeader);
+            Header()->DataCapacity = dataCapacity;
+            ExtendedHeader()->DataOffset =
+                sizeof(TExtendedHeader) + metadataCapacity;
+            ExtendedHeader()->MetadataCapacity = metadataCapacity;
+            ExtendedHeader()->MetadataOffset = sizeof(TExtendedHeader);
```
neihar: do we really need to store this? it is already stored in HeaderSize and will be always the same (during current version)

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+                    metadataCapacity + sizeof(TEntryHeader),
+                    sizeof(ui64));
+            }
+            const ui64 realSize = sizeof(TExtendedHeader) + dataCapacity +
+                                  sizeof(TEntryHeader) + metadataCapacity;
+            Map.ResizeAndRemap(0, realSize);
+            Header()->Version = VERSION_EXT_HEADER;
+            Header()->HeaderSize = sizeof(TExtendedHeader);
+            Header()->DataCapacity = dataCapacity;
+            ExtendedHeader()->DataOffset =
```
neihar: Do we need to store this? looks like it can be derived from other variables.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -364,7 +473,7 @@ class TFileRingBuffer::TImpl
         }
 
         const auto sz = data.size() + sizeof(TEntryHeader);
-        if (sz > Header()->Capacity) {
+        if (sz > Header()->DataCapacity) {
```
neihar, line 511: No action needed in this PR, but wouldn't be better for data locality to create variable as member in class and use it instead of pointer in mmap every time?

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
         }
     }
 
+    void Init()
+    {
+        std::span<char> map(static_cast<char*>(Map.Ptr()), Map.MappedSize());
```
neihar: I really like std::span, but don't like this mix of filemap and span in the code below :)

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+            case VERSION_NO_EXT_HEADER: {
+                Y_ABORT_UNLESS(
+                    sizeof(THeader) + Header()->DataCapacity <= map.size());
+                Data = TEntriesData(
+                    map.subspan(sizeof(THeader), Header()->DataCapacity));
+                Metadata = TMetadata();
+                break;
+            }
+            case VERSION_EXT_HEADER: {
+                Y_ABORT_UNLESS(sizeof(TExtendedHeader) <= Header()->HeaderSize);
```
neihar: Y_ABORT_UNLESS(sizeof(TExtendedHeader) <= map.size());
before Header()->HeaderSize()

neihar: Shouldn't it be == instead of '<=' ?
 if sizeof(TExtendedHeader) < Header()->HeaderSize, something not right.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+                Y_ABORT_UNLESS(Header()->HeaderSize <= map.size());
```
neihar: Will not be needed after implementing 2 checks above

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+                const auto& eh = *ExtendedHeader();
+
+                Y_ABORT_UNLESS(Header()->HeaderSize <= eh.MetadataOffset);
```
neihar: Currently metadataoffset and headersize store the same value of sizeof(TExtendedHeader)

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +373,94 @@ class TFileRingBuffer::TImpl
...
+                Y_ABORT_UNLESS(eh.MetadataOffset <= eh.DataOffset);
+                Y_ABORT_UNLESS(
+                    eh.MetadataCapacity <= eh.DataOffset - eh.MetadataOffset);
+                Y_ABORT_UNLESS(eh.DataOffset <= map.size());
+                Y_ABORT_UNLESS(eh.DataCapacity <= map.size() - eh.DataOffset);
```
neihar: I think it is better to check that overall size matches file size, instead of series "<="

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -137,6 +140,46 @@ class TEntriesData
...
+    }
+
+    bool Set(TStringBuf buf)
+    {
+        if (buf.size() > Body.size()) {
+            return false;
+        }
+        if (Header) {
+            Header->DataSize = buf.size();
+            Header->Checksum = Crc32c(buf.data(), buf.size());
```
neihar: Shouldn't we add to Validate() metadata checksum check or separate function "ValidateMetadata"?

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -545,14 +654,33 @@ class TFileRingBuffer::TImpl
...
+
 TFileRingBuffer::TFileRingBuffer(
-        const TString& filePath,
-        ui64 capacity)
-    : Impl(new TImpl(filePath, capacity))
+    const TString& filePath,
+    ui64 dataCapacity,
+    ui64 metadataCapacity)
+    : Impl(
+          metadataCapacity > 0
```
neihar: I really believe that we should avoid maintaining 2 versions of FileRingBuffer, especially in our case when it is very easy to have new version of FileRingBuffer with metadataCapacity == 0 with 0 cost for clients that don't need it.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -4,29 +4,41 @@
...
 
 struct THeader
 {
     ui32 Version = 0;
-    ui64 Capacity = 0;
+    ui32 HeaderSize = 0;
+    ui64 DataCapacity = 0;
     ui64 ReadPos = 0;
     ui64 WritePos = 0;
     ui64 LastEntrySize = 0;
```
neihar, line 36: why new data should be in separate extended header and not in the main one? In my opinion it is unnecessary complication. We should modify(extend) regular Header and make migration.   

I would strongly advise to add new fields to the end of THeader:
"HeaderSize", "MetadataCapacity" (you don't need to store other fields, they can be just class data members and derived from other data.
And do the proper migration.
In your case we will suffer in the future if we decide to add another extension. just imagine "ExtendedExtendedHeader" :)

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -137,6 +140,46 @@ class TEntriesData
 
 ////////////////////////////////////////////////////////////////////////////////
 
+class TMetadata
+{
+private:
+    TEntryHeader* Header = nullptr;
```
neihar: "using TMetadataHeader = TEntryHeader"

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I disagree.
`pos + size` may overflow while `Data.size() - pos` - not.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: There is a field `Header->Checksum` but currently it is not checked.
I'll add a check in the constructor.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Do you have a better idea?

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: The idea of maintaining two versions of FileRingBuffer is to support backward compatibility without making migration.

The idea of creating an old version of FileRingBuffer with metadataCapacity == 0 is a protection measure against release downgrade. Imagine we started a release then something went wrong in the middle and we decided to revert it — if we created files in a new format (or performed a migration), then we are screwed.

We can remove creation of the old file format after the next release.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: > why new data should be in separate extended header and not in the main one? 

This is due to the support of two file formats.


> We should modify(extend) regular Header and make migration.

I strongly object against migration — I explained it here:
https://github.com/ydb-platform/nbs/pull/4833#discussion_r2616333181

Also, migration isn't so simple and will require writing more code.


> In your case we will suffer in the future if we decide to add another extension. just imagine "ExtendedExtendedHeader" :)

Don't worry, eventually the old file format will be evicted.

neihar: Can you explain how migration will be done? from my understanding we still need to do this, but on the caller level, drain file usage and then switch to new version, which will require also some temporary code, on all callers side which is pretty intrusive. 
Because now we have only 2 callers and I don't see AsyncDestroyHandleEnabled in prod configs, probably we can just rollout new version without migration at all at this time and just clean caches in our test envs

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: I don't have, but probably code without span will be more readable in this case. Don't have strong opinion here, just left a notice for consideration.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Added method `ValidateMetadata`.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: This code is no more relevant

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I expect that metadata and data sections can be placed in any part of the file.
This may be helpful if we want to implement file resize.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Explicit is better than implicit.
Header size may change.
I expect that metadata and data sections can be placed in any part of the file.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I've added a method `GetMappedData`

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: No more relevant here

e673: Done

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: No more relevant here

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Now they don't.
HeaderSize value also used in migration.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Added function `ValidateMetadata`

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: We don't expect the file size to match data size.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I've added migration into the `TFileRingBuffer`. Since we don't use this on prod, no need to support reverting it back.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: No more relevant

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: I checked only nebo repo, I think we need confirmation from @WilyTiger that AsyncHandles wasn't enabled for some filesystem in prod via proto StorageConfig.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -4,29 +4,49 @@
...
 ////////////////////////////////////////////////////////////////////////////////
 
-constexpr ui32 VERSION = 2;
+constexpr ui32 VERSION_PREV = 2;
+constexpr ui32 VERSION = 3;
 constexpr ui64 INVALID_POS = Max<ui64>();
 
+// Reserve some space after header so adding new fields will not require data
+// migration
+constexpr ui64 DefaultDataOffset = 256;
```
neihar: I would make it explicit 
constexpr ui64 HeaderReserveSize = X;
constexpr ui64 MetadataOffset = sizeof(THeader) + HeaderReserveSize

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -4,29 +4,49 @@
...
     ui32 Version = 0;
-    ui64 Capacity = 0;
+    ui32 HeaderSize = 0;
+    ui64 DataCapacity = 0;
     ui64 ReadPos = 0;
     ui64 WritePos = 0;
     ui64 LastEntrySize = 0;
 };
 
+struct THeader: THeaderPrev
```
neihar, line 39: I would advise to keep it as single structure, to avoid header changes in the next PR where you will remove migration code. But it is up to you.

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: And I would not mix DataOffset and MetadataOffset meaning like below "res.MetadataOffset = DefaultDataOffset"

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +348,90 @@ class TFileRingBuffer::TImpl
...
+        if (Header()->Version == VERSION_PREV) {
+            auto header = InitHeader(Header()->DataCapacity, metadataCapacity);
+            Migrate(header);
         }
 
-        auto* begin = static_cast<char*>(Map.Ptr()) + sizeof(THeader);
-        Data = TEntriesData(begin, begin + capacity);
+        TString s;
+        auto map = GetMappedData();
+        const auto& eh = *Header();
```
neihar: why eh? eh -> header?

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +348,90 @@ class TFileRingBuffer::TImpl
...
+        Y_ABORT_UNLESS(eh.MetadataOffset <= map.size());
+        Y_ABORT_UNLESS(eh.MetadataCapacity <= map.size() - eh.MetadataOffset);
+
+        Y_ABORT_UNLESS(eh.HeaderSize <= eh.DataOffset);
+        Y_ABORT_UNLESS(eh.DataOffset <= map.size());
+        Y_ABORT_UNLESS(eh.DataCapacity <= map.size() - eh.DataOffset);
+
+        Y_ABORT_UNLESS(
+            eh.MetadataOffset + eh.MetadataCapacity <= eh.DataOffset ||
+            eh.DataOffset + eh.DataCapacity <= eh.MetadataOffset);
```
neihar: I don't understand why we can't just fixate metadata after header and before data. Can we choose this option? what is benefit to support 2 options?

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +348,90 @@ class TFileRingBuffer::TImpl
...
-                sizeof(THeader) + Header()->Capacity <=
-                static_cast<ui64>(Map.Length()));
-            Y_ABORT_UNLESS(Header()->Version == VERSION);
-        } else {
-            Header()->Capacity = capacity;
-            Header()->Version = VERSION;
+        if (Header()->Version == VERSION_PREV) {
+            auto header = InitHeader(Header()->DataCapacity, metadataCapacity);
+            Migrate(header);
         }
```
neihar, line 486: Should we somehow handle case if caller passes dataCapacity or metadataCapacity more, than existing file supports? abort/error/logging?

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Done

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: What's wrong with removing it later with the migration code?
I use this structure now.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: The name `header` is too long. The code becomes ugly because of 80-char limitation.
Renamed `eh -> h` to avoid confusion with `eh` that is used for `TEntryHeader`.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Resizing the buffer is a separate task (and most likely it will be wontfix), currently we ignore any changes in `dataCapacity`.
Regardless `metadataCapacity`, the file is resized.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Short answer: extensibility.
Possible scenario: add possibility to store metadata even if it exceeds the capacity. In this case, the file can be resized and metadata can be placed after the data.

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: I mean, currently if we already using version 3 and caller expands from its side one of these variables, we just silently ignore them, shouldn't we give response that it is incorrect behavior and abort operation?

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: but currently we don't have this functionality and metadata is fixed between header and dataOffset, if we add new functionality about storing metadata after the data, then we can add those checks. 
BTW, I don't like idea to split metadata in chunks before and after data block. But it is out of scope for current PR.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: 1. Abort will result in an unwanted crash loop. This is not an option.
2. There is no mechanism to give a response and there is no need for it — it is better to implement buffer resize instead (and now it will be possible because we may reference data at any part of the file).

neihar: It will only crash if caller updated parameters, but because didn't have this behavior before, I agree that we can ignore it for now and later do proper resizing instead of abort

e673: > but because didn't have this behavior before

https://github.com/ydb-platform/nbs/issues/3792

This check was removed intentionally.

neihar: Then it is questionable change :) I definitely agree that proper resizing is the only good solution. But don't agree that ignoring error is better, then explicitly say that we don't support expansion to the caller. Btw, because it was changed not in this PR, Let's omit it and discuss in the PR with proper resizing logic if needed this discussion.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I've added a functionality to resize metadata - now it can go after data section if it is not possible to resize it in-place.

neihar: In that case I don't understand, why we can't place metadata to the end of the file at  first place?

In my opinion if we started with |header, metadata, data| order, then we need to continue to maintain it this way (or vice versa). And in case of resize we can just do migration again. In you case it looks pretty safe and crash resilient (with headerSIze field as mark, in general case, we can make separate variable to track reallocation status). Also, if metadata planned to be accessed frequently, probably having it at the beginning is better.

e673: The reason why I prefer to see metadata at the beginning by default is easier file analysis by hexdump (you need to read the first ~1K and get all the information needed). Yes, sometimes I did this.

The reason why I don't want to check that metadata always goes before data is flexibility and extensibility. I wanted to place the data fragments at arbitrary locations in order to minimize moving the data and making resizing easier. I think that it is enough to check that file regions do not overlap. If we can support working with any order with zero-cost — why should we restrict this? I also don't exclude the scenario when metadata can be added manually (who knows what may happen...).

So, I've decided to implement file normalization. If the structure is different than expected (wrong order, different metadata capacity), the file is normalized.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -191,6 +202,28 @@ struct TEntryInfo
...
+    res.HeaderSize = sizeof(THeader);
+    res.MetadataOffset = HeaderReserveSize;
+    res.MetadataCapacity = metadataCapacity;
+    res.DataOffset =
+        AlignUp(res.MetadataOffset + res.MetadataCapacity, sizeof(ui64));
+    res.DataCapacity = dataCapacity;
+    return res;
+}
+
+ui64 GetMaxUsedOffset(const THeader& header)
```
neihar: GetEndOfFileOffset?

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +376,170 @@ class TFileRingBuffer::TImpl
...
+            const ui64 fileSize = GetMaxUsedOffset(header);
+            if (fileSize != static_cast<ui64>(Map.Length())) {
+                ResizeAndRemap(fileSize);
+            }
+            return;
+        }
+
+        const ui64 newFileSize = GetMaxUsedOffset(header);
+
+        if (Header()->MetadataOffset != header.MetadataOffset &&
```
neihar: if we change header size beyond reserved size it will also trigger this condition and code will require refactor

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +376,170 @@ class TFileRingBuffer::TImpl
...
+        }
+
+        if (Header()->DataOffset != header.DataOffset &&
+            Header()->DataOffset < newFileSize)
+        {
+            // Move data to the temporary place
+            const ui64 newDataOffset =
+                Max(newFileSize, GetMaxUsedOffset(*Header()));
+            ResizeAndRemap(newDataOffset + Header()->DataCapacity);
+            CopyMappedData(
```
neihar, line 445: if both metadata and data don't match offset, what will happen? Probably test can be good here.

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: Do we really need this, because as I understand for now we stick with |header, metadata, data| and can't have different order.

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar, line 485: can you add tests to ensure data consistency if corruption happens during reallocations? you can look for examples in persistent_table or [dynamic_persistent_table](https://github.com/ydb-platform/nbs/blob/16be0f350d48cda1acb12d99a9f8590abada6b64/cloud/storage/core/libs/common/dynamic_persistent_table_ut.cpp#L1009)

cloud/storage/core/libs/common/file_ring_buffer.cpp
neihar: I am more worried here about simplicity. Resizing metadata when needed and keeping it in the strong order as you designed initially much more simple. Migrations shouldn't be a often scenario. Even when they occur it will be one time submillisecond straightforward workload at the start of the service. So, agility is nice, but in our case I am not sure that it worth it. Current solution with strict "header, metadata, data" looks fine for me.

e673: Method `NormalizeAndResizeMetadataIfNeeded` will bring the file to the expected layout and order.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: We may have different order if we were interrupted during resize and are trying to resume now.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: The logic was changed - this code no more relevant.
Now I don't move metadata - only data, the code has become simpler.
And now I restrict MetadataOffset <= DataOffset (hope it will not be needed to change it in the future).

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: A test was added

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +362,140 @@ class TFileRingBuffer::TImpl
...
+        Header()->MetadataCapacity = header.MetadataCapacity;
+        Header()->MetadataSize = 0;
+        Header()->Version = VERSION;
+
+        ResizeAndRemap(newFileSize);
+    }
+
+    void ResizeMetadata(ui64 desiredMetadataCapacity)
+    {
+        Header()->MetadataCapacity =
```
neihar, line 423: if we crash after this point we will have changed metadata capacity stored in file, but shouldn't be a big deal.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -322,36 +362,140 @@ class TFileRingBuffer::TImpl
...
+        Header()->DataOffset = header.DataOffset;
+        Header()->MetadataOffset = header.MetadataOffset;
+        Header()->MetadataCapacity = header.MetadataCapacity;
+        Header()->MetadataSize = 0;
+        Header()->Version = VERSION;
+
+        ResizeAndRemap(newFileSize);
+    }
+
+    void ResizeMetadata(ui64 desiredMetadataCapacity)
```
neihar, line 421: probably add comment to the interface, that if capacity already exist we will make best efforts to shrink it, but without guarantee.

cloud/storage/core/libs/common/file_ring_buffer_ut.cpp
```diff
@@ -602,6 +615,218 @@ Y_UNIT_TEST_SUITE(TFileRingBufferTest)
         rb.PopFront();
         UNIT_ASSERT_EQUAL(28, rb.GetMaxAllocationBytesCount());
     }
+
+    Y_UNIT_TEST(ShouldNotSupportMetadata)
```
neihar: Looks like it is this same behavior as if metadataCapacity = 0. Because test names with "ShouldNotSupport and ShouldSupport" not very descriptive in my opinion, I would probably join these 2 tests under general "ShouldRespectMetadataSizeSetup" or something like this.

neihar (approved):

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: Added a comment in `file_ring_buffer.h`

cloud/storage/core/libs/common/file_ring_buffer_ut.cpp
e673: 1. Changed test names.
2. Joining two tests is not needed.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I wanted to unify logic for both shrinking and growing metadata capacity.
And yes, this is not a big deal.

PR #4836, opened 2025-12-13
issue-568: support for block checksum calculation and logging to profile log in TStorageServiceActor

qkrorlqr:
### Notes
Similar to https://github.com/ydb-platform/nbs/pull/4803, but at the StorageServiceActor layer. Needed for data corruption investigation. In case of data corruption investigation having checksums both at the tablet and at the service layers will make it easier to find the component that causes data corruption.

### Issue

### DIFF
--- a/cloud/filestore/libs/diagnostics/profile_log_events.cpp
+++ b/cloud/filestore/libs/diagnostics/profile_log_events.cpp
@@ -94,6 +94,38 @@ ui32 CalculateChecksum(TStringBuf buf)
     return 0;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+
+void CalculateIovecsChecksums(
+    const google::protobuf::RepeatedPtrField<NProto::TIovec>& iovecs,
+    ui32 blockSize,
+    NProto::TProfileLogRequestInfo& profileLogRequest)
+{
+
+    //
+    // Making a copy for simplicity. Checksum calculation is more expensive
+    // than memcpy anyway.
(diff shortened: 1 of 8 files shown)


cloud/filestore/libs/diagnostics/profile_log_events.cpp
```diff
@@ -789,4 +791,46 @@ void CalculateChecksums(
...
+
+    CalculateChecksums(buffer, blockSize, profileLogRequest);
+}
+
+void CalculateReadDataResponseChecksums(
+    const NProto::TReadDataResponse& response,
+    ui32 blockSize,
+    NProto::TProfileLogRequestInfo& profileLogRequest)
+{
+    TStringBuf buffer(response.GetBuffer());
```
debnatkh: Do you need iovec support here? Or maybe just TODO since @antonmyagkov is implementing this in #4664 at the moment

qkrorlqr: will add a TODO

qkrorlqr: there's another thing in this PR that I don't like - I pass ProfileLogRequest as a raw pointer to ReadDataActor

it will be safer to calculate checksums in `TStorageServiceActor::CompleteRequest` instead

but in this case my code cannot know for sure whether iovecs are used so I will need to implement checksum calculation for iovecs anyway it seems

PR #4865, opened 2025-12-17
Eternal-load: introduce simple scenarios

e673:
Introduce simple tests for the eternal-load without data validation:
* Sequential
* Random

Additional functionality:
* Creating test files of any size (use suffixes b, k, m, g).
* Run multiple test instances simultaneously.

Example:
```
eternal-load --config-type generated --scenario random --engine sync --no-direct --run-in-callbacks --file testfile{} --filesize 128m --read-size 1 --write-size 1 --write-rate 80 --debug --iodepth 1 --test-count 32
```
Create 32 files named testfile0, testfile1, ..., testfile31, 128MiB each, read and write with request size = 1 byte at random positions.

### DIFF
--- a/cloud/blockstore/tools/testing/eternal_tests/eternal-load/lib/test_scenarios/simple_test_scenario.cpp
+++ b/cloud/blockstore/tools/testing/eternal_tests/eternal-load/lib/test_scenarios/simple_test_scenario.cpp
@@ -60,7 +60,7 @@ public:
     TTestWorker(TSimpleTestScenario* testScenario, const TLog& log)
         : TestScenario(testScenario)
         , ReadBuffer(testScenario->MaxReadByteCount)
-        , WriteBuffer(testScenario->MaxWriteByteCount)
+        , WriteBuffer(testScenario->MaxWriteByteCount + (sizeof(ui64) * 2))
         , Log(log)
     {}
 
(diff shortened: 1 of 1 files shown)


e673:
For reviewer: the PR is split into several commits (by functionality)

cloud/blockstore/tools/testing/eternal_tests/eternal-load/bin/options.cpp
```diff
@@ -24,6 +29,12 @@ const THashMap<TString, EIoEngine> nameToEngine = {
     {"sync", EIoEngine::Sync}
 };
 
+const THashMap<char, ui64> suffixToMultiplier = {
```
qkrorlqr: looks a bit weird, I personally would expect to see a function with a switch inside it instead of a global hashmap

qkrorlqr: especially considering the fact that it's only needed inside TByteCountOption => it should be a private static method inside TByteCountOption

cloud/blockstore/tools/testing/eternal_tests/eternal-load/bin/test.cpp
```diff
@@ -118,21 +121,49 @@ TTestExecutorSettings TTest::ConfigureTest() const
 
     TTestExecutorSettings settings;
 
+    std::function<ITestScenarioPtr(IConfigHolderPtr)> scenarioFactory;
```
qkrorlqr: it looks more complex than it has to be, just put the switch that you have below into a private function and call it where you currently call scenarioFactory()

e673: Done

cloud/blockstore/tools/testing/eternal_tests/eternal-load/bin/options.cpp
e673: Done

qkrorlqr (approved):

yegorskii (approved):

PR #4868, opened 2025-12-17
DiskStateUpdate is only saved if the disk is in SupportedNotications

PeshkovMikhail:
Issue: https://github.com/ydb-platform/nbs/issues/4873

In some cases, for example, when a disk is deleted and its id is no longer present in SupportsNotifications, records of its updates appear in the database, but everything remains the same in DiskRegistryState.

To solve this problem DiskStateUpdate will be appeared in the database only if disk_id is specified in SupportsNotifications

### DIFF
--- a/cloud/blockstore/libs/storage/disk_registry/disk_registry_state_ut_updates.cpp
+++ b/cloud/blockstore/libs/storage/disk_registry/disk_registry_state_ut_updates.cpp
@@ -220,47 +220,57 @@ Y_UNIT_TEST_SUITE(TDiskRegistryStateUpdatesTest)
         const auto deviceId = "uuid-1.1";
         const auto diskId = "disk-1";
 
-        const TVector agents {
-            AgentConfig(1, {
+        const TVector agents{AgentConfig(
+            1,
+            {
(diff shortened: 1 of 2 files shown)


cloud/blockstore/libs/storage/disk_registry/disk_registry_state_ut_updates.cpp
```diff
@@ -214,6 +214,55 @@ Y_UNIT_TEST_SUITE(TDiskRegistryStateUpdatesTest)
...
+        
+        TVector<TDiskStateUpdate> dbUpdates;
+        executor.ReadTx([&](TDiskRegistryDatabase db) {
+            db.ReadDiskStateChanges(dbUpdates);
+        });
+
+        TVector<TDiskStateUpdate> updates = statePtr->GetDiskStateUpdates();
+        UNIT_ASSERT(dbUpdates.size() == updates.size());
+        for(size_t i = 0; i < dbUpdates.size(); ++i) {
+            UNIT_ASSERT(google::protobuf::util::MessageDifferencer::Equals(dbUpdates[i].State, updates[i].State));
```
komarevtsev-d: clang-format

Issue #3745, opened 2025-06-23
[Filestore] Do flock for all TFileRingBuffer and TPersistentTable files


WilyTiger:
We want to protect these files from being corrupted if two instances of filestore-vhost run at the same time

PR #4870, opened 2025-12-17
lock state files for session persistent tables

akilvein:
https://github.com/ydb-platform/nbs/issues/3745

Add option for TPersistentTable to lock the state file. Avoid state corruption if case of several vhosts accidentally start with the same parameters on the same machine.

### DIFF
--- a/cloud/filestore/libs/service_local/fs_session.cpp
+++ b/cloud/filestore/libs/service_local/fs_session.cpp
@@ -99,6 +99,16 @@ NProto::TCreateSessionResponse TLocalFileSystem::CreateSession(
     auto clientSessionStatePath = StatePath / ("client_" + clientId);
     clientSessionStatePath.MkDir();
 
+    auto lockPath = clientSessionStatePath / "session.lock";
+    lockPath.Touch();
+    auto fileLock = MakeHolder<TFileLock>(lockPath.GetPath());
+    if (!fileLock->TryAcquire()) {
+        return TErrorResponse(
+            E_FAIL,
+            TStringBuilder() << "Failed to lock session state file " << lockPath
+                             << " invalid session: " << sessionId.Quote());
+    }
+
     session = std::make_shared<TSession>(
         Store.GetFileSystemId(),
         RootPath,
@@ -110,7 +120,8 @@ NProto::TCreateSessionResponse TLocalFileSystem::CreateSession(
         Config->GetNodeCleanupBatchSize(),
         Config->GetSnapshotsDirEnabled(),
         Config->GetSnapshotsDirRefreshInterval(),
-        Logging);
+        Logging,
+        std::move(fileLock));
 
     session->Init(request.GetRestoreClientSession());
     session->AddSubSession(sessionSeqNo, readOnly);
--- a/cloud/filestore/libs/service_local/service_ut.cpp
+++ b/cloud/filestore/libs/service_local/service_ut.cpp
@@ -380,6 +380,12 @@ struct TTestBootstrap
         AIOService->Stop();
     }
 
+    void StopFileStore()
+    {
+        Store->Stop();
+        Store.reset();
+    }
+
     std::optional<TFsPath> GetFsStateDir(const TString& id)
     {
         TVector<TFsPath> dirs;
@@ -1248,6 +1254,8 @@ Y_UNIT_TEST_SUITE(LocalFileStore)
             bootstrap.UnlinkNode(dirNodes.back(), "deletedDir", true);
         }
 
+        bootstrap.StopFileStore();
+
         TTestBootstrap other(bootstrap.Cwd);
 
         auto id = other.CreateSession("fs", "client", "", false, 0, true)
@@ -1318,6 +1326,8 @@ Y_UNIT_TEST_SUITE(LocalFileStore)
                 bootstrap.DestroyHandle(handles[i]);
             }
         }
+
+        bootstrap.StopFileStore();
 
         TTestBootstrap other(bootstrap.Cwd);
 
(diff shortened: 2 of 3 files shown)


cloud/filestore/libs/service_local/session.h
```diff
@@ -99,9 +103,27 @@ class TSession
...
+    virtual ~TSession()
+    {
+        UnlockFile(HandlesFileLock);
+        UnlockFile(IndexFileLock);
+    }
+
+    NProto::TError TryInit(bool restoreClientSession)
     {
         auto handlesPath = StatePath / "handles";
+        auto indexPath = StatePath / "nodes";
```
budevg: You are leaking implementation details from `TLocalIndex` into `TSession`. TSession shouldn't know that there is a "nodes" file which is used by the `TIndex`

cloud/filestore/libs/service_local/session.h
```diff
@@ -99,9 +103,27 @@ class TSession
...
+
+        if (!TryLockFile(handlesPath, HandlesFileLock) ||
```
budevg: Locking handles here is premature because the file may be deleted later:
```
        if (isNewSession) {
            ...
            handlesPath.DeleteIfExists();
```

cloud/filestore/libs/service_local/fs_session.cpp
budevg, line 99: Actually simplest and safest fix is to create a dedicated lock file in `clientSessionStatePath` (e.g. session.lock or state.lock) and never delete it. Then lock that file to guard the whole session state.

cloud/filestore/libs/service_local/service_ut.cpp
```diff
@@ -1189,10 +1189,19 @@ Y_UNIT_TEST_SUITE(LocalFileStore)
 
     Y_UNIT_TEST(ShouldRecoverLocalStores)
     {
-        TTestBootstrap bootstrap;
-        bootstrap.CreateFileStore("fs", "cloud", "folder", 100500, 500100);
+        TTempDirectoryPtr bootstrapCwd;
```
budevg: Is there actually a new test that tests that it's forbidden to run two instances of  same TSession ?

cloud/filestore/libs/service_local/service_ut.cpp
```diff
@@ -1288,45 +1309,58 @@ Y_UNIT_TEST_SUITE(LocalFileStore)
 
     Y_UNIT_TEST(ShouldRecoverSessionHandles)
     {
-        TTestBootstrap bootstrap("fs", "client");
```
budevg, line 1291: To avoid extra refactoring in UTs, add `CloseStore` to `TTestBootstrap` and call it explicitly when needed.

cloud/filestore/libs/service_local/service_ut.cpp
```diff
@@ -1288,45 +1309,58 @@ Y_UNIT_TEST_SUITE(LocalFileStore)
...
         auto ctx = MakeIntrusive<TCallContext>();
 
         TVector<TString> files = {"file1", "file2", "file3", "file4", "file5"};
         TVector<ui64> handles;
         TString expectedData = "aaaabbbbcccccdddddeeee";
+        TTempDirectoryPtr bootstrapCwd;
+        TString expectedSessionId;
 
-        for (auto& file: files) {
-            auto handle =
```
budevg, line 1299: Need to test this in devlab as well. Run filestore-vhost-local create fs, start endpoint attach vm. Then run another `filestore-vhost-local`  process and see what happens

cloud/filestore/libs/service_local/service_ut.cpp
akilvein: ShouldFailToLockSessionStateTwice

cloud/filestore/libs/service_local/service_ut.cpp
akilvein: StopFileStore

cloud/filestore/libs/service_local/session.h
```diff
@@ -99,10 +102,26 @@ class TSession
               Log)
     {}
 
-    void Init(bool restoreClientSession)
+    virtual ~TSession()
+    {
+        UnlockFile(SessionFileLock);
```
budevg: Do we need destructor here  ? The  TFileLock will close the file when it's destroyed and this will unlock it automatically

cloud/filestore/libs/service_local/session.h
```diff
@@ -99,10 +102,26 @@ class TSession
...
+    }
+
+    NProto::TError TryInit(bool restoreClientSession)
```
budevg: we can keep calling the function Init() and just add return value

cloud/filestore/libs/service_local/fs_session.cpp
budevg, line 100: Actually let's just  move the whole session state lock logic here. Create TFileLock to guard the session state, lock it and move it to TSession. When the TSession object destroyed the file lock will be automatically released. Also let's do unit tests in session_ut.cpp

budevg (changes requested):

cloud/filestore/libs/service_local/fs_session.cpp
akilvein: session does not have any locking logic now :/

akilvein: so we need to keep service_ut modifications and test lock in service_ut

cloud/filestore/libs/service_local/service_ut.cpp
```diff
@@ -2394,6 +2406,28 @@ Y_UNIT_TEST_SUITE(LocalFileStore)
                 defaultTimeout);
         }
     }
+
+    Y_UNIT_TEST(ShouldFailToLockSessionStateTwice)
```
budevg, line 2410: `ShouldFailToLockSessionStateTwice` only checks that same client + fs can't create second session while the first is alive. You need to test the following use cases:
- same client+fs fails when creating second session
- same client+fs can create again after destroy session (lock is released on DestroySession)
- different clients can create sessions on the same fs concurrently.

Additionally Using `StopFileStore` to test this is misleading because it tears down the store and releases locks implicitly. Better is to destroy the session explicitly.

Issue #4879, opened 2025-12-18
Data race between process_request and do_init in fs_posix_compliance/qemu-kikimr-multishard-test


budevg:
https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/Nightly-build-(tsan)/20291244174/1/nebius-x86-64-tsan/summary/1/ya-test.html#FAIL

```
WARNING: ThreadSanitizer: data race (pid=373838)
  Read of size 8 at 0x7b60000803a0 by thread T146:
    #0 process_request /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:261:19 (filestore-vhost+0x27a06c45) (BuildId: fc98f390c127d9b6e795f31f727321956f1a8380)
    #1 virtio_session_loop /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:532:19 (filestore-vhost+0x27a06c45)
    #2 fuse_session_loop /actions-runner/_work/nbs/nbs/contrib/libs/virtiofsd/fuse_lowlevel.c:2701:12 (libvirtiofsd.so+0xb67e) (BuildId: 18f367562f7a7ba0fbb10025fe24587deccfb01b)
    #3 NCloud::NFileStore::NFuse::(anonymous namespace)::TFuseLoopThread::ThreadProc() /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/loop.cpp:533:9 (filestore-vhost+0x27936571) (BuildId: fc98f390c127d9b6e795f31f727321956f1a8380)
    #4 void* (anonymous namespace)::ThreadProcWrapper<ISimpleThread>(void*) /actions-runner/_work/nbs/nbs/util/system/thread.cpp:383:45 (filestore-vhost+0xfc35ea2) (BuildId: fc98f390c127d9b6e795f31f727321956f1a8380)
    #5 (anonymous namespace)::TPosixThread::ThreadProxy(void*) /actions-runner/_work/nbs/nbs/util/system/thread.cpp:244:20 (filestore-vhost+0xfc3937c) (BuildId: fc98f390c127d9b6e795f31f727321956f1a8380)

  Previous write of size 8 at 0x7b60000803a0 by thread T147 (mutexes: write M0):
[...]

PR #4880, opened 2025-12-18
issue-4879: fix data race during initialization/re-initialization of fuse session

budevg:
#4879 

The data race was between the early, lock free read of `se->bufsize` in
`process_request()` and the unconditional write back to `se->bufsize` at the end of
`do_init()`. Even when the negotiated buffer size didn't change, `do_init()` always
executed `se->bufsize = bufsize;`, creating a write that raced with concurrent
readers on other fuse threads.

### DIFF
--- a/contrib/libs/virtiofsd/fuse_lowlevel.c
+++ b/contrib/libs/virtiofsd/fuse_lowlevel.c
@@ -1882,7 +1882,6 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid,
     struct fuse_init_in *arg;
     struct fuse_init_out outarg;
     struct fuse_session *se = req->se;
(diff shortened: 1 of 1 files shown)


contrib/libs/virtiofsd/fuse_lowlevel.c
```diff
@@ -1882,7 +1882,6 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid,
     struct fuse_init_in *arg;
     struct fuse_init_out outarg;
     struct fuse_session *se = req->se;
-    size_t bufsize = se->bufsize;
     size_t outargsize = sizeof(outarg);
```
drbasic, line 1885: Дайте ПР на исходный коммит в апстрим?

PR #4890, opened 2025-12-22
Track latency of DiskRegistry devices on agent level.

drbasic:
For large disks, the table is too large, grouping by agents reduces the size of the table and it becomes more convenient to look at the data.

### DIFF
--- a/cloud/blockstore/libs/storage/core/device_operation_tracker.cpp
+++ b/cloud/blockstore/libs/storage/core/device_operation_tracker.cpp
@@ -16,13 +16,27 @@ namespace {
 
 std::atomic<ui32> TrackingFrequency = 0;
 
+TString EscapeAgentId(const TString& agentId)
+{
+    TString result;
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/core/device_operation_tracker.cpp
```diff
@@ -239,18 +225,27 @@ auto TDeviceOperationTracker::GetTimeBuckets() const -> TVector<TBucketInfo>
...
     return result;
 }
 
-TVector<TDeviceOperationTracker::TDeviceInfo>
-TDeviceOperationTracker::GetDeviceInfos() const
+TVector<TDeviceOperationTracker::TBucketInfo>
+TDeviceOperationTracker::GetAgents() const
 {
-    return DeviceInfos;
+    TVector<TBucketInfo> result;
```
agalibin, line 240: тут можно сразу место зарезервить и emplace_back использовать

PR #4904, opened 2025-12-24
Schedulled comparison between local Disk Registry State and local db

PeshkovMikhail:


cloud/blockstore/config/storage.proto
```diff
@@ -1392,4 +1392,6 @@ message TStorageServiceConfig
 
     // Number of fresh channels per partition for HDD disks.
     optional uint32 FreshChannelCountHDD = 473;
+
+    optional uint32 CompareDiskRegistryStateWithLocalDbInterval = 474;
```
vladstepanyuk, line 1396: добавить бы коментарий

PR #4912, opened 2025-12-25
issue-568: checksumming for profile log - ignoring buffer overflows for reads, not ignoring any kind of buffer overflows for writes, reporting debug-info in buffer overflow crit event

qkrorlqr:
### Notes
* ReadData: having a range which is either partially or completely beyond the end of the buffer is valid, the client can try to read beyond the end of the file, we shouldn't be reporting crit events in this case
* WriteData: any kind of buffer overflow - even partial - is not okay, we should report a crit event
* reporting debug info in the buffer-overflow crit event - fs-id, node-id, range, buffer-size 

### Issue

### DIFF
--- a/cloud/filestore/libs/diagnostics/profile_log_events.cpp
+++ b/cloud/filestore/libs/diagnostics/profile_log_events.cpp
@@ -96,9 +96,11 @@ ui32 CalculateChecksum(TStringBuf buf)
 
 ////////////////////////////////////////////////////////////////////////////////
 
-void CalculateIovecsChecksums(
+bool CalculateIovecsChecksums(
     const google::protobuf::RepeatedPtrField<NProto::TIovec>& iovecs,
     ui32 blockSize,
+    bool ignoreBufferOverflow,
+    TStringBuf fsId,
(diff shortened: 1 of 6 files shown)


yegorskii (approved):

agalibin (approved):

cloud/filestore/libs/diagnostics/profile_log_events.cpp
```diff
@@ -847,11 +902,21 @@ void CalculateReadDataResponseChecksums(
     if (iovecs.empty()) {
         TStringBuf buffer(response.GetBuffer());
         buffer.Skip(response.GetBufferOffset());
-        CalculateChecksums(buffer, blockSize, profileLogRequest);
+        CalculateChecksums(
+            buffer,
+            blockSize,
+            true /* ignoreBufferOverflow */,
```
drbasic, line 908: чтобы не писать коммент у каждого вызова проще сделать enum и передавать его

qkrorlqr: Ну, не знаю, дело вкуса

Issue #4782, opened 2025-12-05
[Filestore] Don't use TwoStageRead for small requests and/or when the tablet is far from overloaded


qkrorlqr:
We have an issue for 4KiB reads.

TwoStageRead:
1. DescribeData - 1 IC request+response
2. ReadBlob -> EvGet -> requests to 3 VDisks (1 to fetch the data, the other 2 - to 2 possible handoff VDisks)
Total: 4 IC requests+responses

Direct ReadData: just one IC request+response - ReadDataRequest + ReadDataResponse

So whenever the client (filestore-vhost) is the bottleneck, we should move the burden of sending VDisk requests to the tablets. I'll start with a simple setting which would let us disable TwoStageRead for small requests (default threshold == 0 in order not to change the behaviour in current clusters by default). Later on I'll probably change the code to decide which scheme (two- or one- stage read) to use depending on the cpu and network usage at the tablet side.

PR #4915, opened 2025-12-27
issue-4782: calculating backend-info with is-overloaded flag using cpu-wait and returning it from tablet to service in response-headers

qkrorlqr:
### Notes
Adding `NProto::TBackendInfo` to response-headers and calculating `TBackendInfo::IsOverloaded` at the tablet layer using `CpuLack` (calculated from cpu-wait counters). This is needed to give the service layer more information on whether to do some of the work at the service layer or to delegate it to the tablet layer - e.g. whether to ask the tablet to do blob reading or to do it at the service layer.

We can start using some network-related metrics and tablet executor cpu usage alongside cpu-lack to calculate is-overloaded in the future if needed.

Returning this information for all requests (because it's now a part of the tablet API) even though in the short-term the plan is to use it only for reads (to decide whether to use TwoStageRead or to delegate blob reading to the tablet). 

Will add `IsOverloaded` usage at the service layer in a separate PR.

### Issue
https://github.com/ydb-platform/nbs/issues/4782

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_update_stats.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_update_stats.cpp
@@ -67,10 +67,10 @@ void TStorageServiceActor::HandleUpdateStats(
         if (HasError(error)) {
             auto errorMessage =
                 ReportCpuWaitCounterReadError(error.GetMessage());
-                LOG_WARN_S(
-                    ctx,
-                    TFileStoreComponents::SERVICE,
-                    "Failed to get CpuWait stats: " << errorMessage);
+            LOG_WARN_S(
+                ctx,
+                TFileStoreComponents::SERVICE,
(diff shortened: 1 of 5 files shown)


cloud/filestore/libs/storage/service/service_actor_update_stats.cpp
```diff
@@ -62,31 +63,30 @@ void TStorageServiceActor::HandleUpdateStats(
...
-                TFileStoreComponents::SERVICE,
-                "Failed to get CpuWait stats: " << errorMessage);
-        }
-
         auto now = ctx.Monotonic();
-        if (LastCpuWaitTs < now) {
+        if (HasError(error)) {
+            auto errorMessage =
+                ReportCpuWaitCounterReadError(error.GetMessage());
+                LOG_WARN_S(
```
agalibin: тут что-то не то с выравниванием

PR #4921, opened 2025-12-30
Speed up RDMA 

drbasic:
1. Rename sentRequestId to clientRequestId
2. Generate 16-bit reqId with sequential number generator
3. History of completed, timed out, cancelled requests removed
4. Speed up PopTimedOutRequests()
image

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/client.cpp
+++ b/cloud/blockstore/libs/rdma/impl/client.cpp
@@ -114,6 +114,8 @@ public:
     [[nodiscard]] ui32 CreateId()
     {
         for (;;) {
+            Y_DEBUG_ABORT_UNLESS(Requests.size() < RDMA_MAX_REQID - 1);
+
             if (RequestIdGenerator >= RDMA_MAX_REQID) {
                 RequestIdGenerator = 0;
             }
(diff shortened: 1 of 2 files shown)


cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -109,23 +107,32 @@ struct TRequest
...
         for (;;) {
+            Y_DEBUG_ABORT_UNLESS(Requests.size() < RDMA_MAX_REQID - 1);
+
+            if (RequestIdGenerator >= RDMA_MAX_REQID) {
+                RequestIdGenerator = 0;
+            }
+            const ui32 reqId = ++RequestIdGenerator;
             // must be unique through all in-flight requests
-            ui32 reqId = RandomNumber<ui32>(RDMA_MAX_REQID);
-            if (reqId && Requests.find(reqId) == Requests.end()) {
```
tpashkin, line 123: На самом деле не уверен, что есть особый смысл скипать 0, но пусть будет – можно отличить непроинициализированный запрос, если `reqId=0`

cloud/blockstore/libs/storage/partition_nonrepl/part_nonrepl_rdma_actor_writeblocks_multi_agent.cpp
```diff
@@ -308,12 +308,10 @@ void TNonreplicatedPartitionRdmaActor::HandleMultiAgentWrite(
         deviceRequest.DeviceIdx,
         TDeviceOperationTracker::ERequestType::Write);
 
-    const auto sentRequestId =
-        ep->SendRequest(std::move(req), requestInfo->CallContext);
-
```
budevg: I think it's better to leave it separated to  `ep->SendRequest` and the building the context with result.
Doing this during initialization of `TRequestContext` is less verbose in my opinion and can lead to some confusion.

Issue #2566, opened 2024-11-25
[Filestore] Support filesystems bigger than 1PiB


qkrorlqr:
We currently encode shard number in the highest 8 bits of node id and handle id. We also reserve value 0 for the nodes and handles managed by the main tablet and value 255 - just in case we need to have some "invalid shard" value. Default max shard size is 4TiB. So max filesystem size is 1016TiB. We should either increase actual max shard size to some value which is greater than 4TiB (very simple) or we should store shard number in e.g. 10 highest bits or mb even 12 highest bits. 12 bits leaves us with 52 bits for node id which is 4.5 quadrillion inodes - which is certainly enough. Increasing bit count is harder than increasing actual shard size since we will need to save bit count somewhere in the main filesystem config to be backwards-compatible (to properly identify shard number bit count for the existing filesystems).

debnatkh:
Closing this issue as done.

Implemented solution:

* The first 8 bits of a node/handle identifier represented the shard number.
* The next 8 bits of node ID are always zero (since node ID is incremented starting from 1, it will take 2^48 operations to get to the second most significant byte). 
* We use the following 8 bits for higher-order shard bits.
* Methods `ExtractShardNo(ui64 id)` and `ShardedId(ui64 id, ui32 shardNo)` are adjusted accordingly
* So, from a value 0xABCDEFG, we should extract the shardNo as 0xBA
* The only problem is the existence of so-called legacy file handles (ones with non-zero 7th byte). Their existence prevents the filesystem from getting more shards than 254

PR #4927, opened 2026-01-02
issue-2566: cleanup legacy handle processing code

qkrorlqr:
### Notes
Ran into `shardNo` being incorrect before the `SelectShard()` call here: https://github.com/ydb-platform/nbs/pull/4925 and had to change `SelectShard`'s `shardNo` param to a reference to get the right `shardNo` value from the `SelectShard()` call. This felt ugly so I decided to clean this up:
* Completed a very old TODO item - raising a crit event and returning an error in case the `shardNo` that's extracted from `HandleId` or `NodeId` is greater than `shardCount`. This TODO item referred to ancient filesystems which already had main-tablet-generated handles at the moment when we decided to add shards to those filesystems. This TODO item was relevant somewhere around July-August 2024 but should not be relevant now anymore (https://github.com/ydb-platform/nbs/pull/1488).
* Cleaned up the logic that processes the case when the 7th byte in a `HandleId` is non-zero but the filesystem has <= 255 shards - moved this check from `SelectShard()` into `ExtractShardNoSafe()` so that all `shardNo`s used throughout service layer code are correct and don't need any extra patching.
* Added a couple new uts to make debugging direct shard requests a bit easier - the first version of the PR failed in `cloud/filestore/tests/client/test.py::test_multitablet_findgarbage*` integration tests because those tests use direct client requests to shards and those failed because of ShardNo vs ShardIds validation  

### Issue
https://github.com/ydb-platform/nbs/issues/2566

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
@@ -1251,6 +1251,41 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
     SERVICE_TEST_SID_SELECT_IN_LEADER(ShouldReturnErrorForInvalidShardNo)
     {
         config.SetMultiTabletForwardingEnabled(true);
+        TShardedFileSystemConfig fsConfig;
+        CREATE_ENV_AND_SHARDED_FILESYSTEM();
+
+        auto headers = service.InitSession(fsConfig.FsId, "client");
+
+        auto createHandleResponse = service.CreateHandle(
+            headers,
+            fsConfig.FsId,
+            RootNodeId,
+            "file1",
+            TCreateHandleArgs::CREATE_EXL)->Record;
+
+        auto data = GenerateValidateData(256_KB);
+
+        const ui64 nodeId1 = createHandleResponse.GetNodeAttr().GetId();
+        const ui64 handle1 = createHandleResponse.GetHandle();
+
+        service.SendWriteDataRequest(
+            headers,
+            fsConfig.FsId,
+            nodeId1,
+            ShardedId(handle1, 111),
+            0,
+            data);
+        auto writeDataResponse = service.RecvWriteDataResponse();
+        UNIT_ASSERT_VALUES_EQUAL_C(
+            E_INVALID_STATE,
+            writeDataResponse->GetStatus(),
+            writeDataResponse->GetErrorReason());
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/service/service_actor_writedata.cpp
```diff
@@ -741,7 +741,8 @@ void TStorageServiceActor::HandleWriteData(
 
     // In handleless IO mode, if the handle is not set, we use the nodeId to
     // infer the shard number
-    ui32 shardNo = ExtractShardNo(
+    const ui32 shardNo = ExtractShardNoSafe(
+        filestore,
```
SvartMetal, line 745: Do we need to apply the same logic to RenameNode as well?
Also, do we still need the ExtractShardNo function? The change would be easier to verify if ExtractShardNo were removed (or renamed)

ruslan-gar: ShardedId and ExtractShardNo pack and unpack a shard number into handle/nodeId. 
ExtractShardNoSafe takes into account filestore.GetShardFileSystemIds().size() and takes the least significant 
byte from the shardNo if the number of shards is less than 255.
We need the safe function in order to handle properly handles created by the cold code. 
It is possible to add a param to ExtractShardNo, but for me ExtractShardNoSafe looks better

debnatkh (approved):

cloud/filestore/libs/storage/service/service_actor_writedata.cpp
qkrorlqr: > Do we need to apply the same logic to RenameNode as well?

I thought about it but right now it's not required because `RenameNode` uses `ExtractShardNo` only if `DirectoryCreationInShardsEnabled == true` and I was planning to enable `DirectoryCreationInShardsEnabled` only for new filesystems (so no legacy handles).

> Also, do we still need the ExtractShardNo function? The change would be easier to verify if ExtractShardNo were removed (or renamed)

Initially I wanted to rename the current func to `ExtractShardNoUnsafe` and call it only from its unittests and from `ExtractShardNoSafe` but then decided that it's not required in all the places where we call `ExtractShardNo` and decided that I'd rather make this PR smaller by not doing that. But I don't have a firm opinion on this, I'm ok with both ways.

yegorskii (approved):

Issue #4547, opened 2025-10-24
Optimize TMixedBlocks::FindBlocks performance


SvartMetal:
Currently this function takes too much time to execute, especially for 1 MiB requests with deletion markers.
Benchmark results:
```
svartmetal@svartmetal-ivm:~/nbs-github/cloud/filestore/libs/storage/tablet/model/bench$ ./bench -b 120
----------- TMixedBlocks_FindBlocksWithoutDeletionMarkers ---------------
 samples:       1712
 iterations:    1811656
 iterations hr:    1.81M
 run time:      20.01999958
 per iteration: 26471.61988 (26.5K) cycles
----------- TMixedBlocks_FindBlocksWithOneDeletionMarker ---------------
 samples:       1680
 iterations:    1743778
 iterations hr:    1.74M
 run time:      20.01257271
 per iteration: 27492.67279 (27.5K) cycles
----------- TMixedBlocks_FindBlocksWithOneDeletionMarkerAtTheEnd ---------------
 samples:       1690
 iterations:    1764381
 iterations hr:    1.76M
 run time:      20.00665973
 per iteration: 27158.91053 (27.2K) cycles
----------- TMixedBlocks_FindBlocksWithMergedDeletionMarkers ---------------
 samples:       1666
 iterations:    1715878
 iterations hr:    1.72M
 run time:      20.01759897
 per iteration: 27949.69114 (27.9K) cycles
----------- TMixedBlocks_FindBlocksWithMergedDeletionMarkersAtTheEnd ---------------
 samples:       1674
 iterations:    1730730
 iterations hr:    1.73M
 run time:      20.00131041
 per iteration: 27679.46161 (27.7K) cycles
----------- TMixedBlocks_FindBlocksWithMixedDeletionMarkers ---------------
 samples:       1135
 iterations:    796953
 iterations hr:    797K
 run time:      20.02858092
 per iteration: 60217.861 (60.2K) cycles
----------- TMixedBlocks_FindSingleBlockWithoutDeletionMarkers ---------------
 samples:       3822
 iterations:    9020628
 iterations hr:    9.02M
 run time:      20.00593365
 per iteration: 5309.822041 cycles
----------- TMixedBlocks_FindSingleBlockWithMergedDeletionMarkers ---------------
 samples:       3021
 iterations:    5636403
 iterations hr:    5.64M
 run time:      20.00252201
 per iteration: 8503.220239 cycles
```

[...]

PR #4928, opened 2026-01-02
[Filestore] issue-4547: pass ReadBlobDisabled flag to service in CreateSessionResponse; service should support ReadBlobDisabled in two-stage-read mode; add ReportFakeBlobWas(Read|Written) critical events

SvartMetal:
#4547

To measure the benefits of DescribeData performance optimizations, it is crucial to be able to model the system without reading real data from BlobStorage. Instead, fake data should be used, making it possible to simulate hundreds of gibibytes of reads without requiring a large BlobStorage cluster.

To achieve this, we plan the following milestones:
* Support fake ReadBlob (`ReadBlobDisabled=true`) for two-stage reads - **_we are here_**
* Support fake DescribeData when `ReadBlobDisabled=true`. Fake DescribeData responses should contain non-empty data (for example, 4 blobs per 1 MiB request)
* Add configurable latency for fake DescribeData
* Add configurable latency for fake ReadBlob.

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_readdata.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_readdata.cpp
@@ -350,8 +350,7 @@ void TReadDataActor::ReadBlobsIfNeeded(const TActorContext& ctx)
     }
 
     if (ReadBlobDisabled) {
-        ReportFakeBlobWasRead(
-            LogTag + " fake blob was read (ReadBlobDisabled=true)");
+        ReportFakeBlobWasRead();
         ReplyAndDie(ctx);
         return;
     }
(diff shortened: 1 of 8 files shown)


cloud/filestore/libs/storage/service/service_actor_readdata.cpp
```diff
@@ -344,6 +349,13 @@ void TReadDataActor::ReadBlobsIfNeeded(const TActorContext& ctx)
         return;
     }
 
+    if (ReadBlobDisabled) {
+        ReportFakeBlobWasRead(
```
qkrorlqr: nit: we usually use TStringBuilder in such cases:
```
ReportXXX(TStringBuilder() << LogTag << " some message " << ...);
```

and we don't really need this part: " fake blob was read (ReadBlobDisabled=true)" - it's already present in the name of the critical event ("FakeBlobWasRead")

so just: ```ReportFakeBlobWasRead(LogTag);```

qkrorlqr: Maybe actually my idea about a critical event for this wasn't good - we'll be printing a log message upon every read request in this case. Or we need to make a separate ReportXXX() func version which would not log anything - would just increment the crit event counter (sensor).

SvartMetal: Thanks for the review! Fixed

yegorskii (approved):

PR #4930, opened 2026-01-05
issue-4547: tablet should support fake DescribeData requests with configurable latency

SvartMetal:
#4547

Follow-up for #4928

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
@@ -1,5 +1,6 @@
 #include "tablet_actor.h"
 
+#include <cloud/filestore/libs/diagnostics/critical_events.h>
 #include <cloud/filestore/libs/diagnostics/throttler_info_serializer.h>
 #include <cloud/filestore/libs/diagnostics/trace_serializer.h>
 #include <cloud/filestore/libs/storage/tablet/model/profile_log_events.h>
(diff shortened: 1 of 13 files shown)


cloud/filestore/public/api/protos/fs.proto
```diff
@@ -49,6 +49,7 @@ message TFileStoreFeatures
     bool BlockChecksumsInProfileLogEnabled = 34;
     uint32 RegularFileEntryTimeout = 35;
     bool ReadBlobDisabled = 36;
+    bool FakeDescribeDataEnabled = 38;
```
SvartMetal, line 53: Merge `WriteBlobDisabled = 37` first

cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
```diff
@@ -1132,4 +1140,79 @@ void TIndexTabletActor::CompleteTx_ReadData(
...
+    const ui32 latencyUs = ClampVal(
+        Config->GetFakeDescribeDataLatencyUs(),
+        0u,
+        MaxLatencyUs);
+    const auto deadlineCycles = NDateTimeHelpers::SumWithSaturation(
+        startCycles,
+        DurationToCyclesSafe(TDuration::MicroSeconds(latencyUs)));
+
+    // Spin until deadline
+    while (GetCycleCount() < deadlineCycles) {}
```
qkrorlqr, line 1213: is blocking `TIndexTabletActor` intentional? to simulate in-memory index lookup (which does its work synchronously on-cpu)

SvartMetal: Yes. When the index resides in the cache, handling a 1 MiB HandleDescribeData can take about 20–30 us. In this case, DescribeData request handling is completely synchronous 

image

drbasic (approved):

Issue #4961, opened 2026-01-06
Session is not marked as orphaned after client disconnection


WilyTiger:
We are ignoring disconnects from the pipe in the tablet
https://github.com/ydb-platform/nbs/blob/5fa1a1a1adf03cd13aca143392b283214d7e9f09/cloud/filestore/libs/storage/tablet/tablet_actor.cpp#L1137

Even though we have a handler for disconnects
https://github.com/ydb-platform/nbs/blob/5fa1a1a1adf03cd13aca143392b283214d7e9f09/cloud/filestore/libs/storage/tablet/tablet_actor.cpp#L711,
in the EvServerDisconnected event the Sender is not the id of the session owner stored in SessionByOwner, because the sender of the disconnect event is the pipe client, not the session owner.

As a result, the session may remain active even after the client has disconnected, until the tablet is rebooted. This can lead to the filesystem deletion hanging.

WilyTiger:
In this PR #4962 I tried to fix the original problem by making tablet pipe disconnect handling use the pipe server id instead of `ev->Sender`.

Initially, I thought the issue was only that `TEvTabletPipe::TEvServerDisconnected::Sender` is not the same actor id that the tablet stores as the session owner during `CreateSession`. If we simply enable the existing disconnect handler, `OrphanSession(ev->Sender, ...)` does not find the session.

The PR changes the tablet-side model to track the pipe server explicitly. During `CreateSession`, the tablet stores `ev->Recipient` as the pipe server id, because the request is delivered through tablet pipe and the pipe server is the actor that receives the request before forwarding it to the tablet. (https://github.com/ydb-platform/ydb-rfc/blob/main/ydb_book/05-Tablet-infrastructure/05.02-Tablet-Pipe.md)
`TSubSession` now stores both the original owner and the pipe server:
https://github.com/ydb-platform/nbs/blob/81d1a8e760/cloud/filestore/libs/storage/tablet/subsessions.h#L16
The tablet state keeps a mapping from pipe server id to session:
https://github.com/ydb-platform/nbs/blob/81d1a8e760/cloud/filestore/libs/storage/tablet/tablet_state_impl.h#L44
CreateSession and RecoverSession update this mapping:
https://github.com/ydb-platform/nbs/blob/81d1a8e760/cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp#L191
[...]

PR #4962, opened 2026-01-06
issue-4961: fix handling of TEvServerDisconnected in tablet

WilyTiger:
#4961

In the EvServerDisconnected event, the Sender is not the id of the session owner that we stored in SessionByOwner during session creation. That is why I introduced a new TSessionOwnerByPipeServerMap to keep the PipeServerId -> Owner mapping.

We need to store PipeServerId during session creation. The initial recipient of messages sent through the pipe is the pipe client. The client rewrites the recipient to the pipe server and sends the message via the actor system. The pipe server then forwards the message to the tablet using RecipientRewrite. As a result, the pipe server is the Recipient, while the tablet is specified in RecipientRewrite. That’s why I stored ev->Recipient as the ID of the pipe server.

This is described in the documentation:
https://github.com/ydb-platform/ydb-rfc/blob/main/ydb_book/05-Tablet-infrastructure/05.02-Tablet-Pipe.md

### DIFF
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -261,14 +261,8 @@ public:
         const NActors::TActorId& owner,
         const NActors::TActorId& pipeServer)
     {
-        auto result = SubSessions.UpdateSubSession(seqNo, readOnly, owner, pipeServer);
-        UpdateSeqNo();
-        return result;
-    }
-
-    ui32 DeleteSubSession(const NActors::TActorId& owner)
-    {
-        auto result = SubSessions.DeleteSubSession(owner);
+        auto result =
+            SubSessions.UpdateSubSession(seqNo, readOnly, owner, pipeServer);
         UpdateSeqNo();
         return result;
     }
@@ -280,7 +274,7 @@ public:
         return result;
     }
 
-    ui32 DeleteSubSession(ui64 sessionSeqNo)
+    std::optional<TSubSession> DeleteSubSession(ui64 sessionSeqNo)
     {
         auto result = SubSessions.DeleteSubSession(sessionSeqNo);
         UpdateSeqNo();
@@ -289,12 +283,22 @@ public:
 
     TVector<NActors::TActorId> GetSubSessions() const
     {
-        return SubSessions.GetSubSessions();
+        return SubSessions.GetSubSessionsOwner();
     }
 
     TVector<NActors::TActorId> GetSubSessionsPipeServer() const
     {
         return SubSessions.GetSubSessionsPipeServer();
+    }
+
+    std::optional<TSubSession> GetSubSessionBySeqNo(ui64 seqNo) const
+    {
+        return SubSessions.GetSubSessionBySeqNo(seqNo);
+    }
+
+    bool ReadyToDestroy(ui64 seqNo) const
+    {
+        return SubSessions.ReadyToDestroy(seqNo);
     }
 
     ui64 GenerateDupCacheEntryId()
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -440,6 +440,8 @@ struct TSessionHistoryEntry
 using TSessionList = TIntrusiveListWithAutoDelete<TSession, TDelete>;
 using TSessionMap = THashMap<TString, TSession*>;
 using TSessionOwnerMap = THashMap<NActors::TActorId, TSession*>;
+using TSessionOwnerByPipeServerMap =
+    THashMap<NActors::TActorId, NActors::TActorId>;
 using TSessionClientMap = THashMap<TString, TSession*>;
 using TSessionHistoryList = TDeque<TSessionHistoryEntry>;
 
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -255,7 +255,7 @@ public:
         SetMaxRwSeqNo(SubSessions.GetMaxSeenRwSeqNo());
     }
 
-    NActors::TActorId UpdateSubSession(
+    std::optional<TSessionPipeInfo> UpdateSubSession(
         ui64 seqNo,
         bool readOnly,
         const NActors::TActorId& owner,
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -258,9 +258,10 @@ public:
     NActors::TActorId UpdateSubSession(
         ui64 seqNo,
         bool readOnly,
-        const NActors::TActorId& owner)
+        const NActors::TActorId& owner,
+        const NActors::TActorId& pipeServer)
     {
-        auto result = SubSessions.UpdateSubSession(seqNo, readOnly, owner);
+        auto result = SubSessions.UpdateSubSession(seqNo, readOnly, owner, pipeServer);
         UpdateSeqNo();
         return result;
     }
@@ -268,6 +269,13 @@ public:
     ui32 DeleteSubSession(const NActors::TActorId& owner)
     {
         auto result = SubSessions.DeleteSubSession(owner);
+        UpdateSeqNo();
+        return result;
+    }
+
+    ui32 DeleteSubSessionByPipeServer(const NActors::TActorId& pipeServer)
+    {
+        auto result = SubSessions.DeleteSubSessionByPipeServer(pipeServer);
         UpdateSeqNo();
         return result;
     }
@@ -282,6 +290,11 @@ public:
     TVector<NActors::TActorId> GetSubSessions() const
     {
         return SubSessions.GetSubSessions();
+    }
+
+    TVector<NActors::TActorId> GetSubSessionsPipeServer() const
+    {
+        return SubSessions.GetSubSessionsPipeServer();
     }
 
     ui64 GenerateDupCacheEntryId()
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -255,7 +255,7 @@ public:
         SetMaxRwSeqNo(SubSessions.GetMaxSeenRwSeqNo());
     }
 
-    std::optional<TSessionPipeInfo> UpdateSubSession(
+    TSubSessionUpdateResult UpdateSubSession(
         ui64 seqNo,
         bool readOnly,
         const NActors::TActorId& owner,
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -272,13 +272,6 @@ public:
         return result;
     }
 
-    ui32 DeleteSubSession(const NActors::TActorId& owner)
-    {
-        auto result = SubSessions.DeleteSubSession(owner);
-        UpdateSeqNo();
-        return result;
-    }
-
     ui32 DeleteSubSessionByPipeServer(const NActors::TActorId& pipeServer)
     {
         auto result = SubSessions.DeleteSubSessionByPipeServer(pipeServer);
@@ -286,7 +279,7 @@ public:
         return result;
     }
 
-    ui32 DeleteSubSession(ui64 sessionSeqNo)
+    std::optional<TSubSession> DeleteSubSession(ui64 sessionSeqNo)
     {
         auto result = SubSessions.DeleteSubSession(sessionSeqNo);
         UpdateSeqNo();
@@ -295,12 +288,22 @@ public:
 
     TVector<NActors::TActorId> GetSubSessions() const
     {
-        return SubSessions.GetSubSessions();
+        return SubSessions.GetSubSessionsOwner();
     }
 
     TVector<NActors::TActorId> GetSubSessionsPipeServer() const
     {
         return SubSessions.GetSubSessionsPipeServer();
+    }
+
+    std::optional<TSubSession> GetSubSessionBySeqNo(ui64 seqNo) const
+    {
+        return SubSessions.GetSubSessionBySeqNo(seqNo);
+    }
+
+    bool ReadyToDestroy(ui64 seqNo) const
+    {
+        return SubSessions.ReadyToDestroy(seqNo);
     }
 
     ui64 GenerateDupCacheEntryId()
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -259,12 +259,14 @@ public:
         ui64 seqNo,
         bool readOnly,
         const NActors::TActorId& owner,
+        const NActors::TActorId& pipeServer,
         ui32 tabletGeneration)
     {
         auto result = SubSessions.UpdateSubSession(
             seqNo,
             readOnly,
             owner,
+            pipeServer,
             tabletGeneration);
         UpdateSeqNo();
         return result;
@@ -273,6 +275,13 @@ public:
     ui32 DeleteSubSession(const NActors::TActorId& owner)
     {
         auto result = SubSessions.DeleteSubSession(owner);
+        UpdateSeqNo();
+        return result;
+    }
+
+    ui32 DeleteSubSessionByPipeServer(const NActors::TActorId& pipeServer)
+    {
+        auto result = SubSessions.DeleteSubSessionByPipeServer(pipeServer);
         UpdateSeqNo();
         return result;
     }
@@ -287,6 +296,11 @@ public:
     TVector<NActors::TActorId> GetSubSessions() const
     {
         return SubSessions.GetSubSessions();
+    }
+
+    TVector<NActors::TActorId> GetSubSessionsPipeServer() const
+    {
+        return SubSessions.GetSubSessionsPipeServer();
     }
 
     ui64 GenerateDupCacheEntryId()
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -445,6 +445,8 @@ struct TSessionHistoryEntry
 using TSessionList = TIntrusiveListWithAutoDelete<TSession, TDelete>;
 using TSessionMap = THashMap<TString, TSession*>;
 using TSessionOwnerMap = THashMap<NActors::TActorId, TSession*>;
+using TSessionOwnerByPipeServerMap =
+    THashMap<NActors::TActorId, NActors::TActorId>;
 using TSessionClientMap = THashMap<TString, TSession*>;
 using TSessionHistoryList = TDeque<TSessionHistoryEntry>;
 
--- a/cloud/filestore/libs/storage/tablet/session.h
+++ b/cloud/filestore/libs/storage/tablet/session.h
@@ -272,38 +272,34 @@ public:
         return result;
     }
 
-    ui32 DeleteSubSessionByPipeServer(const NActors::TActorId& pipeServer)
+    TDeleteSubSessionResult DeleteSubSessionByPipeServer(
+        const NActors::TActorId& pipeServer)
     {
         auto result = SubSessions.DeleteSubSessionByPipeServer(pipeServer);
         UpdateSeqNo();
         return result;
     }
 
-    std::optional<TSubSession> DeleteSubSession(ui64 sessionSeqNo)
+    TDeleteSubSessionResult DeleteSubSession(ui64 sessionSeqNo)
     {
         auto result = SubSessions.DeleteSubSession(sessionSeqNo);
         UpdateSeqNo();
         return result;
     }
 
-    TVector<NActors::TActorId> GetSubSessions() const
+    TVector<NActors::TActorId> GetSubSessionOwnerIds() const
     {
-        return SubSessions.GetSubSessionsOwner();
+        return SubSessions.GetSubSessionOwnerIds();
     }
 
-    TVector<NActors::TActorId> GetSubSessionsPipeServer() const
+    TVector<NActors::TActorId> GetSubSessionPipeServerIds() const
     {
-        return SubSessions.GetSubSessionsPipeServer();
+        return SubSessions.GetSubSessionPipeServerIds();
     }
 
     std::optional<TSubSession> GetSubSessionBySeqNo(ui64 seqNo) const
     {
         return SubSessions.GetSubSessionBySeqNo(seqNo);
-    }
-
-    bool ReadyToDestroy(ui64 seqNo) const
-    {
-        return SubSessions.ReadyToDestroy(seqNo);
     }
 
     ui64 GenerateDupCacheEntryId()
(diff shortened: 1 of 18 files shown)


cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
```diff
@@ -305,7 +324,8 @@ void TIndexTabletState::OrphanSession(const TActorId& owner, TInstant deadline)
         session->Unlink();
         Impl->OrphanSessions.PushBack(session);
 
-        Impl->SessionByOwner.erase(it);
+        Impl->SessionByOwner.erase(sit);
```
qkrorlqr: the only `SessionByOwner` lookup that I see in the code is in `OrphanSession()` - do we still need `SessionByOwner`? maybe just having `SessionByPipeServerId` is enough?

cloud/filestore/libs/storage/tablet/tablet_ut_sessions.cpp
```diff
@@ -963,6 +963,51 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Sessions)
             tablet.DestroySession();
         }
     }
+
+    Y_UNIT_TEST(ShouldMarkSessionAsOrphanAfterPipeDisconnetion)
```
qkrorlqr: Disconnetion -> Disconnection (or, better - Disconnect)

cloud/filestore/libs/storage/tablet/tablet_ut_sessions.cpp
```diff
@@ -963,6 +963,51 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Sessions)
...
+        {
+            tablet.ReconnectPipe();
+            auto sessions = tablet.DescribeSessions();
+            UNIT_ASSERT_VALUES_EQUAL(sessions->Record.SessionsSize(), 1);
+            const auto& session = sessions->Record.GetSessions(0);
+            UNIT_ASSERT_VALUES_EQUAL(session.GetClientId(), "client");
+            UNIT_ASSERT_VALUES_EQUAL(session.GetSessionId(), "session");
+            UNIT_ASSERT_VALUES_EQUAL(session.GetIsOrphan(), true);
+        }
+
```
SvartMetal: excessive \n

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
WilyTiger: I see one place https://github.com/ydb-platform/nbs/blob/2ee3d7c2e9af373ea2798269e9d5bc3867902c15/cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp#L218 there Session deleted from SessionByOwner during RecoverSession for old owner. If I will replace SessionByOwner with SessionByPipeServer then I need also to store pipeServerId inside TSubSession https://github.com/ydb-platform/nbs/blob/2ee3d7c2e9af373ea2798269e9d5bc3867902c15/cloud/filestore/libs/storage/tablet/subsessions.h#L20 
I am not sure if its okay to replace completely Owner with PipeServer. Also there is killing of owner tablet in RecoverSession https://github.com/ydb-platform/nbs/blob/2ee3d7c2e9af373ea2798269e9d5bc3867902c15/cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp#L131

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
```diff
@@ -283,36 +298,38 @@ TSession* TIndexTabletState::FindSession(
...
-    auto it = Impl->SessionByOwner.find(owner);
-    if (it == Impl->SessionByOwner.end()) {
+    auto it = Impl->SessionByPipeServer.find(pipeServer);
+    if (it == Impl->SessionByPipeServer.end()) {
         return; // not a session pipe
     }
 
     auto* session = it->second;
-
+
```
SvartMetal: Extra whitespaces

cloud/filestore/libs/storage/tablet/session.h
```diff
@@ -286,7 +288,22 @@ struct TSession
 
     TVector<NActors::TActorId> GetSubSessions() const
     {
-        return SubSessions.GetSubSessions();
+        return SubSessions.GetSubSessionsOwner();
+    }
+
+    TVector<NActors::TActorId> GetSubSessionsPipeServer() const
```
qkrorlqr, line 294: GetSubSessionPipeServerIds

cloud/filestore/libs/storage/tablet/session.h
```diff
@@ -286,7 +288,22 @@ struct TSession
 
     TVector<NActors::TActorId> GetSubSessions() const
```
qkrorlqr, line 289: GetSubSessionOwnerIds

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -50,7 +51,7 @@ NActors::TActorId TSubSessions::AddSubSession(
     SubSessions.push_back(
         {seqNo,
          readOnly,
-         owner,
+         {owner, pipeServer},
```
qkrorlqr, line 54: please use designated initializers for aggregate initialization

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -85,26 +91,40 @@ NActors::TActorId TSubSessions::UpdateSubSession(
...
+        // Owner and PipeServer change independently on reconnect.
+        const bool ownerChanged = subsession->PipeInfo.Owner != owner;
+        const bool pipeServerChanged =
+            subsession->PipeInfo.PipeServer != pipeServer;
+        if (!ownerChanged && !pipeServerChanged) {
+            return {};
+        }
+
+        TSubSessionUpdateResult result;
+        result.StalePipeServer = subsession->PipeInfo.PipeServer;
```
qkrorlqr, line 104: but `pipeServerChanged` might be false here - is it a problem?

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -137,14 +157,7 @@ ui32 TSubSessions::DeleteSubSession(ui64 sessionSeqNo)
         });
 
     if (subsession == SubSessions.end()) {
-        return !ReadyToDestroy(sessionSeqNo);
```
qkrorlqr, line 140: why have these `ReadyToDestroy` checks disappeared?

cloud/filestore/libs/storage/tablet/subsessions.cpp
qkrorlqr, line 164: why aren't we setting it to the max value among the subsessions that are still present?

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -154,14 +167,25 @@ ui32 TSubSessions::DeleteSubSession(ui64 sessionSeqNo)
         MaxSeenSeqNo = MaxSeenRwSeqNo;
```
qkrorlqr, line 167: again not clear
this code definitely requires comments

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
```diff
@@ -153,7 +153,12 @@ void TIndexTabletActor::ExecuteTx_DestroySession(
             return;
         }
 
-        if (session->DeleteSubSession(args.SessionSeqNo)) {
+        if (!session->ReadyToDestroy(args.SessionSeqNo)) {
```
qkrorlqr, line 156: ok, I see it now

qkrorlqr, line 156: who's ready to destroy? or to be destroyed? this code requires comments

// hopefully this doesn't mean ready to destroy user's file descriptors

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
```diff
@@ -217,7 +222,6 @@ void TIndexTabletActor::CompleteTx_DestroySession(
         return;
     }
 
-    UnregisterSessionByPipeServer(args.SessionId);
```
qkrorlqr, line 220: hmm, why?

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
```diff
@@ -211,101 +223,42 @@ NActors::TActorId TIndexTabletState::RecoverSession(
...
-        Impl->SessionByOwner.erase(oldOwner);
 
-        LOG_INFO(*TlsActivationContext, TFileStoreComponents::TABLET,
-            "%s removed old owner for session c: %s, s: %s, owner: %s",
-            LogTag.c_str(),
-            session->GetClientId().c_str(),
-            session->GetSessionId().c_str(),
-            oldOwner.ToString().c_str());
+    if (update.StalePipeServer) {
+        Impl->SessionByPipeServer.erase(*update.StalePipeServer);
```
qkrorlqr, line 238: so probably if we set StalePipeServer to pipeServer it still works - we just erase it and then insert it back again

but still looks not very straightforward

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
```diff
@@ -211,101 +223,42 @@ NActors::TActorId TIndexTabletState::RecoverSession(
...
+    Impl->SessionByPipeServer[pipeServer] = session;
 
-    if (oldOwner != owner) {
-        session->InactivityDeadline = {};
-
-        session->Unlink();
-        Impl->Sessions.PushBack(session);
-
-        Impl->SessionByOwner.emplace(owner, session);
+    session->InactivityDeadline = {};
```
qkrorlqr, line 242: why do we not need the `oldOwner != owner` check anymore?

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
qkrorlqr: ready-to-destroy-subsession?

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
WilyTiger: SessionIdsByPipeServer and all its helper functions were deleted - it duplicates SessionByPipeServer map. SessionByPipeServer cleanup for this same destroy path already happens via RemoveSession()

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
WilyTiger: Ready to destroy full session, if we delete subsession with this seqNo

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: Yes, this is a confusing part of the code.

This update relies on the current subsession model: MaxSubSessions is 2, and there can be only one current rw seqNo for a session. So if we remove the subsession whose seqNo is equal to MaxSeenRwSeqNo, there is no other current rw seqNo to fall back to, and MaxSeenRwSeqNo is set to 0

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: This part also relies on the fact that we keep at most two subsessions, and only one of them can be rw.

If we remove the subsession whose seqNo is equal to MaxSeenSeqNo, the only other meaningful seqNo we can fall back to is MaxSeenRwSeqNo. We do not scan the remaining subsessions here

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1053,55 +1053,46 @@ void TIndexTabletActor::HandleTabletMetrics(
...
         LOG_INFO(
             ctx,
             TFileStoreComponents::TABLET,
             "%s Deleting unconfirmed data for session: sessionId=%s",
             LogTag.c_str(),
-            sessionId.Quote().c_str());
+            session->GetSessionId().Quote().c_str());
 
-        DeleteUnconfirmedDataForSession(sessionId, ctx);
+        DeleteUnconfirmedDataForSession(session->GetSessionId(), ctx);
```
neihar, line 1087: As I understand, 2 subsessions can have their own CreateSessionActor and pipe. So both pipes will be in SessionByPipeServer pointing at the same TSession. When either control pipe drops, DeleteUnconfirmedDataForSession will be called for this subsession and drop unconfirmed data.

Before changes we unregistered old pipe before registering new.

It can be as follows: Data move over per-node IndexTabletProxy pipe, but CreateSession uses per-subssession control pipe and control pipe disconnect on old sub-session rejects writes on new.

It shouldn't be a data corruption problem, but theoretically spikes in latency because of fallback.

So, probably 
1) we can leave it as is, if it is rare
2) add subsession to TTrackedUnconfirmedData to distinguish what was dropped
3) probably we can delete unconfirmed data only if it is last subsession for session (needs check)

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1667,9 +1657,7 @@ STFUNC(TIndexTabletActor::StateAdapter)
         HFunc(TEvents::TEvPoisonPill, HandlePoisonPill);
 
         IgnoreFunc(TEvTabletPipe::TEvServerConnected);
-        HFunc(
-            TEvTabletPipe::TEvServerDisconnected,
-            HandleSessionDisconnectedInWork);
+        HFunc(TEvTabletPipe::TEvServerDisconnected, HandleSessionDisconnected);
```
neihar, line 1669: now, that we use this function for Zombie as well, it is worth to consider how good to use transactions in this state. I do not remember other transactions that call ExecuteTx at Zombie state. If it works - okay, but needs to be checked. Probably it will not work or not as expected

WilyTiger: I think we can just switch to IgnoreFunc in zombie state: on tablet reboot, LoadSessions puts every session straight into OrphanSessions. Same for unconfirmed data - there's a separate recovery pass on boot (ReadUnconfirmedData/ConfirmBlobs) that doesn't depend on this handler in zombie state

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
WilyTiger: This is a real edge case, but a pretty narrow one: the only place we get two live subsessions on one session is client migration, and in that case one of them is always RO, it shouldn't issue writes at all. So the actual failure window is: the RO leftover's pipe dies while the RW subsession happens to have an unconfirmed write in flight at that exact moment - that write gets dropped and falls back to WriteData. Rare, and self-correcting (fallback is safe, just slower for that one request)

Also worth noting the old SessionIdsByPipeServer-based logic wasn't correct here either - UnregisterSessionByPipeServer(sessionId) on every CreateSession/RecoverSession completion wiped all pipe registrations for that session id regardless of which subsession's pipe it was, so a stale registration from the other subsession could just as easily go missing (under-delete) instead of over-delete. So this isn't a regression from the current change, just a different shape of a pre-existing gap in this area.

I'd leave it as is for now. Later we could teach TTrackedUnconfirmedData to track which subsession a write belongs to, so cleanup on disconnect only drops that subsession's data instead of the whole session's

neihar: I am okay with solution 1, probably we can create task to check that behaviour with big writes and link it to live migration task

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
neihar: From my side ignore looks more appropriate than calling executeTx, but I would prefer to have logs for such cases

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
WilyTiger: oldOwner != owner was always true: if UpdateSubSession actually changed the owner, it returns the old owner as oldOwner - which is different from the new owner by definition. If the owner did not change, it returns TActorId() instead - and since owner in RecoverSession arguments can't be empty, that's also != owner. So the check was true in both cases

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: Yes, it was missed, fixed this and added test there pipeServerChanged was false and ownerChanged was true

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: I moved the check back here and extracted the shared logic into a common helper (DeleteSubSessionIf). I also changed the subsession deletion functions to return a separate struct with clearly named fields. Initially, I wanted to move the ReadyToDestroy logic out of these deletion functions, but that turned out to make the code even harder to understand. The function now returns whether the entire session can be deleted after removing the subsession. If a subsession was deleted, it also returns information about that subsession

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: added comments

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: added comments

cloud/filestore/libs/storage/tablet/subsessions.cpp
WilyTiger: Correction to my earlier reply - "only one rw seqNo" isn't actually guaranteed. During migration there's a real window (between the new endpoint getting promoted to RW and the old one being stopped) where both subsessions can be readOnly=false at the same time - nothing here enforces single-writer

qkrorlqr (dismissed):

qkrorlqr (dismissed):

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1053,55 +1053,46 @@ void TIndexTabletActor::HandleTabletMetrics(
...
+        DeleteUnconfirmedDataForSession(session->GetSessionId(), ctx);
     }
 
     // msg.ServerId is the tablet-pipe server actor that received requests
     // from this client connection. Unconfirmed data keeps this actor id from
     // GenerateBlobIds, so clean it up when the pipe disconnects.
     DeleteUnconfirmedDataForPipeServer(msg.ServerId, ctx);
-    RemoveSessionByPipeServer(msg.ServerId);
+
+    OrphanSession(msg.ServerId, ctx.Now());
```
neihar: small: probably we can avoid double lookups in map as we already know the session.

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
```diff
@@ -153,7 +153,12 @@ void TIndexTabletActor::ExecuteTx_DestroySession(
             return;
         }
 
-        if (session->DeleteSubSession(args.SessionSeqNo)) {
+        auto result = session->DeleteSubSession(args.SessionSeqNo);
+        if (!result.SessionCanBeDestroyed) {
```
neihar: how it will work for RO sessions? will we endup in TIndexTabletState::RemoveSession(TSession* session) without actual subsession, because we remove it in DeleteSubSession above?

neihar: probably test for that?

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
```diff
@@ -153,7 +153,12 @@ void TIndexTabletActor::ExecuteTx_DestroySession(
...
+            if (result.Removed) {
+                RemovePipeServer(result.Removed->PipeInfo.PipeServer);
```
neihar: if we have live migration on the same host, will we have one pipe for both old and new seqno and removal here is wrong? Can we add test for that?

cloud/filestore/libs/storage/tablet/tablet_state_impl.h
```diff
@@ -31,20 +31,13 @@ namespace NCloud::NFileStore::NStorage {
...
-
 struct TIndexTabletState::TImpl
 {
     TSessionList Sessions;
     TSessionList OrphanSessions;
     TSessionMap SessionById;
-    TSessionOwnerMap SessionByOwner;
+    TSessionOwnerMap SessionByPipeServer;
     TSessionClientMap SessionByClient;
-    TSessionIdsByPipeServerMap SessionIdsByPipeServer;
```
neihar, line 47: for 1 tablet system it is okay to remove, but how it will behave for shards without map? we can have one pipe for multiple clients, probably we can also add test for that

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
neihar: do we want to use ctx.Now() here that will lead to expired right away, or add some idle time to tolerate flaps?

neihar (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
WilyTiger: Each endpoint (old seqNo and new seqNo) is its own TCreateSessionActor with its own pipe. Being on the same host doesn't merge them into one pipe, pipes are per-actor, not per-host. So old and new seqNo always end up on different PipeServer actors on the tablet side.
This PR doesn't change any migration logic - it only add SessionByPipeServer tracking and Orphan logic. Adding new tests for migration scenarios, or fixing bugs in the existing migration logic if any turn out to exist, is out of scope for this change

neihar: what about TShardRequestActor - > MakeIndexTabletProxyServiceId

cloud/filestore/libs/storage/tablet/tablet_actor_destroysession.cpp
WilyTiger: Yes, it was a problem and was a problem before my PR in old map, fixed and added tests

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
WilyTiger: Yeah, definitely missing "+ IdleSessionTimeout". Added test for that also

WilyTiger:
https://claude.ai/chat/0e9d1758-1bd4-42e9-a511-5d33aecb95dc

Issue #4941, opened 2026-01-06
Data Corruption Due to CommitId Overflow DestroyHandle


neihar:
cause similar to https://github.com/ydb-platform/nbs/issues/4937

look at ScheduleRebootTabletOnCommitIdOverflow(ctx, "AllocateData");

ScheduleRebootTabletOnCommitIdOverflow(ctx, "DestroyHandle");

this can cause having garbage after close(fd) for an fd referring to an already unlinked file

Issue #4939, opened 2026-01-06
Data Corruption Due to CommitId Overflow  in CreateHandle


neihar:
cause similar to https://github.com/ydb-platform/nbs/issues/4937

look at ScheduleRebootTabletOnCommitIdOverflow(ctx, "CreateHandle");

this won't cause data corruption but can cause EBADF

PR #4972, opened 2026-01-08
[Filestore] Correct silent failure during create and destroy handle when commit id overflows

neihar:
CreateHandle:
When commit id overflows during create handle we return E_REJECTED and client should retry operation instead of returning success without actual handle creation that leads to EBADF
DestroyHandle:
Returning E_REJECTED to client for retry instead of leaving garbage after unsuccessful operation.

#4941
#4939

### DIFF
--- a/cloud/filestore/libs/storage/testlib/helpers.cpp
+++ b/cloud/filestore/libs/storage/testlib/helpers.cpp
@@ -7,8 +7,7 @@ namespace NCloud::NFileStore::NStorage {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-std::function<
-    bool(NActors::TTestActorRuntimeBase&, TAutoPtr<NActors::IEventHandle>&)>
+NActors::TTestActorRuntimeBase::TEventFilter
 TTabletRebootTracker::GetEventFilter()
 {
     return [this](auto& runtime, auto& event)
--- a/cloud/filestore/libs/storage/testlib/helpers.h
+++ b/cloud/filestore/libs/storage/testlib/helpers.h
@@ -357,9 +357,7 @@ private:
     TSet<ui32> Generations;
 
 public:
-    std::function<
-        bool(NActors::TTestActorRuntimeBase&, TAutoPtr<NActors::IEventHandle>&)>
-    GetEventFilter();
+    NActors::TTestActorRuntimeBase::TEventFilter GetEventFilter();
 
     bool IsPipeDestroyed() const;
     void ClearPipeDestroyed();


cloud/filestore/libs/storage/tablet/tablet_actor_destroyhandle.cpp
```diff
@@ -97,8 +97,21 @@ void TIndexTabletActor::ExecuteTx_DestroyHandle(
...
+                << "DestroyHandle: " << args.SessionId
+                << ", Handle: " << args.Request.GetHandle()
+                << ", NodeId: " << args.Node->NodeId,
+                args.Node->NodeId);
+        }
+    };
+
     auto commitId = GenerateCommitId();
     if (commitId == InvalidCommitId) {
+        writeOrphanNode();
```
neihar: Just mark as Orphan and reboot, looks like in this case it is valid behavior. Not sure that we need to change it

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -492,6 +495,10 @@ void TIndexTabletActor::CompleteTx_CreateHandle(
         ctx);
 
     NCloud::Reply(ctx, *args.RequestInfo, std::move(response));
+
+    if (args.Error.GetCode() == E_REJECTED && args.WriteCommitId == InvalidCommitId) {
```
qkrorlqr: this condition is too fragile, let's actually use a separate dedicated flag - something like `args.CommitIdOverflowDetected`

cloud/filestore/libs/storage/tablet/tablet_actor_destroyhandle.cpp
```diff
@@ -112,11 +125,7 @@ void TIndexTabletActor::ExecuteTx_DestroyHandle(
             commitId);
 
         if (HasError(e)) {
-            WriteOrphanNode(db, TStringBuilder()
-                << "DestroyHandle: " << args.SessionId
-                << ", Handle: " << args.Request.GetHandle()
-                << ", RemoveNode: " << args.Node->NodeId
-                << ", Error: " << FormatError(e), args.Node->NodeId);
+            writeOrphanNode();
```
qkrorlqr: this lambda func checks the condition that's already checked by the `if` statement on line 118

this DestroyHandle operation won't succeed in the commit-id-overflow case and will be retried - no need to log orphan node upon commit-id-overflow

cloud/filestore/libs/storage/tablet/tablet_ut_data.cpp
```diff
@@ -7857,7 +7839,7 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Data)
 
         tablet.DestroyHandle(handle);
         UNIT_ASSERT_C(
-            generations.size() >= 2,
+            rebootTracker.GetGenerationsCount() >= 2,
```
qkrorlqr: nit: we usually call such things `GetGenerationCount()` (`GetEntityCount()` instead of `GetEntitiesCount()`)

`BlocksCount` is a historical exception

PS originally this rule comes from Yandex C++ styleguide

cloud/filestore/libs/storage/testlib/helpers.h
```diff
@@ -340,4 +345,52 @@ inline TString GetErrorReason(const std::unique_ptr<T>& response)
...
+class TTabletRebootTracker
+{
+private:
+    bool PipeDestroyed = false;
+    TSet<ui32> Generations;
+
+public:
+    auto GetEventFilter()
+    {
+        return [this](auto& runtime, auto& event)
```
qkrorlqr: let's put the implementation of all of the `TTabletRebootTracker` methods into `helpers.cpp`

cloud/filestore/libs/storage/tablet/tablet_actor_destroyhandle.cpp
```diff
@@ -97,8 +97,21 @@ void TIndexTabletActor::ExecuteTx_DestroyHandle(
...
+            WriteOrphanNode(db, TStringBuilder()
+                << "DestroyHandle: " << args.SessionId
+                << ", Handle: " << args.Request.GetHandle()
+                << ", NodeId: " << args.Node->NodeId,
+                args.Node->NodeId);
+        }
+    };
+
     auto commitId = GenerateCommitId();
     if (commitId == InvalidCommitId) {
```
qkrorlqr, line 98: another problem - we should perform this check __before__ we do `DestroyHandle(db, handle)`

otherwise we won't find this handle upon `DestroyHandle` op retry attempt

neihar: exactly, that's why I put orphan node. Do you think we need to change this behavior?

cloud/filestore/libs/storage/tablet/tablet_actor_destroyhandle.cpp
```diff
@@ -94,29 +94,27 @@ void TIndexTabletActor::ExecuteTx_DestroyHandle(
...
-            args.Node->MinCommitId,
-            commitId);
+        auto e = RemoveNode(db, *args.Node, args.Node->MinCommitId, commitId);
 
         if (HasError(e)) {
-            WriteOrphanNode(db, TStringBuilder()
-                << "DestroyHandle: " << args.SessionId
-                << ", Handle: " << args.Request.GetHandle()
-                << ", RemoveNode: " << args.Node->NodeId
-                << ", Error: " << FormatError(e), args.Node->NodeId);
```
qkrorlqr, line 119: `FormatError(e)` got lost

cloud/filestore/libs/storage/testlib/helpers.h
```diff
@@ -2,6 +2,14 @@
 
 #include <cloud/filestore/libs/service/filestore.h>
 
+#include <contrib/ydb/core/base/tablet.h>
```
qkrorlqr: includes for `tablet.h` and `tablet_pipe.h` can be moved to `helpers.cpp`

cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
```diff
@@ -148,6 +148,101 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Handles)
...
+                    handleResponse->GetError().GetCode());
+                continue;
+            }
+
+            ui64 handle = handleResponse->Record.GetHandle();
+
+            tablet.SendWriteDataRequest(handle, 0, block, 'a');
+            auto writeResponse = tablet.RecvWriteDataResponse();
+
+            UNIT_ASSERT_VALUES_EQUAL(S_OK, writeResponse->GetStatus());
```
SvartMetal: `tablet.WriteData(handle, 0, block, 'a');`

cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
```diff
@@ -148,6 +148,101 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Handles)
...
+                    E_REJECTED,
+                    destroyResponse->GetError().GetCode());
+                continue;
+            }
+
+            tablet.SendReadDataRequest(successfulHandles[i], 0, block);
+            auto readResponse = tablet.RecvReadDataResponse();
+            reconnectIfNeeded();
+
+            UNIT_ASSERT(S_OK != readResponse->GetStatus());
```
SvartMetal: UNIT_ASSERT_VALUES_UNEQUAL

cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
```diff
@@ -148,6 +148,101 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Handles)
...
+            auto writeResponse = tablet.RecvWriteDataResponse();
+
+            UNIT_ASSERT_VALUES_EQUAL(S_OK, writeResponse->GetStatus());
+
+            successfulHandles.push_back(handle);
+        }
+
+        UNIT_ASSERT_C(
+            rebootTracker.GetGenerationCount() >= 2,
+            "Expected at least 2 different generations due to tablet reboot");
```
SvartMetal: `"Expected at least 2 different generations due to tablet reboot got " << ...`

cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
neihar: Used more words for consistency within the test and explicit check . change to tablet.WriteData

cloud/filestore/libs/storage/testlib/helpers.cpp
```diff
@@ -1 +1,47 @@
...
+
+#include <contrib/ydb/core/base/tablet.h>
+#include <contrib/ydb/core/base/tablet_pipe.h>
+
+namespace NCloud::NFileStore::NStorage {
+
+////////////////////////////////////////////////////////////////////////////////
+
+std::function<
+    bool(NActors::TTestActorRuntimeBase&, TAutoPtr<NActors::IEventHandle>&)>
```
SvartMetal: `TTestActorRuntime::TEventFilter`?

Issue #4940, opened 2026-01-06
Data Corruption Due to CommitId Overflow in CreateCheckpoint


neihar:
cause similar to https://github.com/ydb-platform/nbs/issues/4937

look at ScheduleRebootTabletOnCommitIdOverflow(ctx, "CreateCheckpoint");

can cause errors in disk-manager after we start using checkpoints for filestore

PR #4989, opened 2026-01-12
[Filestore] return E_REJECTED instead of silently missing checkpoint

neihar:
#4940
Because of commit id overflow currently we silently missing checkpoint creation failure and that leads towards failure in the future when we attempt to use it.

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_ut_checkpoints.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_ut_checkpoints.cpp
@@ -333,8 +333,8 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Checkpoints)
 
     Y_UNIT_TEST(ShouldHandleCommitIdOverflowInCreateCheckpoint)
     {
-        const auto block = 4_KB;
-        const auto maxTabletStep = 8;
+        const ui32 block = 4_KB;
+        const ui32 maxTabletStep = 8;
 
         NProto::TStorageConfig storageConfig;
         storageConfig.SetMaxTabletStep(maxTabletStep);
@@ -355,6 +355,8 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Checkpoints)
         const auto id =
             CreateNode(tablet, TCreateNodeArgs::File(RootNodeId, "test"));
         ui64 handle = CreateHandle(tablet, id);
+
+        ui32 createCheckpointRejectedCount = 0;
 
         auto reconnectAndRecreateHandleIfNeeded = [&](bool createHandle = true)
         {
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/tablet/tablet_ut_checkpoints.cpp
```diff
@@ -329,6 +330,113 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Checkpoints)
...
+
+        TVector<std::pair<TString, char>> successfulCheckpoints;
+        const int targetSuccessfulCheckpoints = 4;
+
+        for (int i = 0; i < targetSuccessfulCheckpoints;) {
+            tablet.SendWriteDataRequest(handle, 0, block, 'a' + i);
+            auto writeResponse = tablet.RecvWriteDataResponse();
+            reconnectAndRecreateHandleIfNeeded();
+
+            if (FAILED(writeResponse->GetStatus())) {
```
qkrorlqr: better to use `HasError(error)` than `FAILED(status)` (macro vs func)

cloud/filestore/libs/storage/tablet/tablet_ut_checkpoints.cpp
```diff
@@ -329,6 +330,113 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Checkpoints)
         tablet.DestroyHandle(handle1);
         tablet.DestroyHandle(handle2);
     }
+
+    Y_UNIT_TEST(ShouldHandleCommitIdOverflowInCreateCheckpoint)
+    {
+        const auto block = 4_KB;
```
qkrorlqr: let's not use auto for primitive types, let's declare them explicitly (`const ui32 blockSize = 4_KB`, `const ui32 maxTabletStep = 8`)

cloud/filestore/libs/storage/tablet/tablet_ut_checkpoints.cpp
```diff
@@ -329,6 +330,113 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Checkpoints)
...
+            UNIT_ASSERT_BUFFER_CONTENTS_EQUAL(
+                buffer,
+                block,
+                successfulCheckpoints[i].second);
+
+            ++i;
+        }
+
+        UNIT_ASSERT_C(
+            rebootTracker.GetGenerationCount() >= 2,
```
qkrorlqr, line 438: this doesn't seem to check that we ever got E_REJECTED for a `CreateCheckpointRequest`, this assert condition can still be true if we only had reboots due to `CreateHandle`/`WriteData` requests

neihar: E_REJECTED also doesn't guarantee that we faced exactly commit id overflow, but I agree that it is narrowing the situations when we break test somehow. Added.

Issue #4990, opened 2026-01-12
[Filestore] setgid flag is not respected for directory in filestore


debnatkh:
To quote [Wikipedia](https://en.wikipedia.org/wiki/Setuid#When_set_on_a_directory):

>  Setting the setgid permission on a directory causes files and subdirectories created within to inherit its group ownership, rather than the primary group of the file-creating process. Created subdirectories also inherit the setgid bit. The policy is only applied during creation and, thus, only prospectively. Directories and files existing when the setgid bit is applied are unaffected, as are directories and files moved into the directory on which the bit is set.

--- 

This logic is implemented in the kernel driver here: https://github.com/torvalds/linux/blob/0f61b1860cc3f52aef9036d7235ed1f017632193/fs/inode.c#L2638-L2639. Since it is not implemented for the virtiofs, the same logic should be implemented either on vhost, or on the filestore-server side

PR #4991, opened 2026-01-12
issue-4990: introduce `GidPropagationEnabled` option that will respect `setgid` flag set on directories

debnatkh:
#4990


cloud/filestore/libs/storage/tablet/tablet_actor_createnode.cpp
```diff
@@ -628,6 +628,14 @@ bool TIndexTabletActor::PrepareTx_CreateNode(
         args.Error = ErrorIsNotDirectory(args.ParentNodeId);
         return true;
     }
+    if (Config->GetGidPropagationEnabled()) {
+        if (args.ParentNode->Attrs.GetMode() & S_ISGID) {
```
budevg, line 632: How will this logic work for request forwarded to shard ? The args.Attrs aren't copied to shard request and when running the CreateNode logic on shard it won't know that the parent has setgid set for it.

budevg (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor_createnode.cpp
debnatkh: Nice catch. You are right. This indeed will be a problem. Fixed by duplicating gid/mode adjusting in the original request also

PR #5017, opened 2026-01-15
Remove meaningless TReplicaDescriptor::ReplicaId and use DiskId instead

komarevtsev-d:
The `TReplicaDescriptor` structure contained a string field called `ReplicaId`, which was originally named incorrectly as it actually represented a disk id. Secondly, a vector of `TReplicaDescriptor` was created for resyncing and scrubbing, and each element in the vector had the same "replica id" value.

In this pull request, the `ReplicaId` field has been removed and replaced with passing the disk id alongside a vector.

### DIFF
--- a/cloud/blockstore/libs/storage/partition_nonrepl/resync_range_ut.cpp
+++ b/cloud/blockstore/libs/storage/partition_nonrepl/resync_range_ut.cpp
@@ -153,7 +153,8 @@ struct TTestEnv
                 actorId,
                 TActorSetupCmd(part.release(), TMailboxType::Simple, 0)
             );
-            Replicas.push_back({static_cast<ui32>(i), actorId});
+            Replicas.push_back(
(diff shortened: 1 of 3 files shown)


cloud/blockstore/libs/storage/partition_nonrepl/resync_range_ut.cpp
```diff
@@ -153,7 +153,7 @@ struct TTestEnv
                 actorId,
                 TActorSetupCmd(part.release(), TMailboxType::Simple, 0)
             );
-            Replicas.push_back({name, static_cast<ui32>(i), actorId});
+            Replicas.push_back({static_cast<ui32>(i), actorId});
```
drbasic: .ReplicaIndex = i, .ActorId=actorId

drbasic (dismissed):

Issue #4983, opened 2026-01-12
[Local Filestore] support weka snapshots


budevg:
When using a local filestore backed by the Weka filesystem, we want access to the hidden `.snapshots` directory that stores snapshots taken on Weka.

PR #5018, opened 2026-01-15
[Local Filestore] issue-4983: support weka snapshots

budevg:
#4983

### DIFF
--- a/cloud/filestore/libs/service_local/fs_node.cpp
+++ b/cloud/filestore/libs/service_local/fs_node.cpp
@@ -238,14 +238,17 @@ NProto::TListNodesResponse TLocalFileSystem::ListNodes(
     for (auto& entry: entries) {
         auto remapResult =
             session->RemapNode(parent->GetNodeId(), entry.second.INode);
-        if (remapResult) {
-            auto remappedNode = *remapResult;
-            if (!remappedNode) {
+        if (remapResult.HasRemap()) {
+            if (!remapResult.IsFound()) {
                 continue;
             }
             try {
-                entry.second = remappedNode->Stat();
+                entry.second = remapResult.GetNode()->Stat();
             } catch (...) {
+                STORAGE_ERROR(
+                    "Failed to stat remappedNode, parentNodeId="
+                    << Hex(parent->GetNodeId())
+                    << ", nodeId=" << Hex(entry.second.INode))
                 continue;
             }
         }
--- a/cloud/filestore/libs/service_local/index.cpp
+++ b/cloud/filestore/libs/service_local/index.cpp
@@ -255,8 +255,13 @@ public:
         fileId.LustreFid.Oid = nodeId & 0xffffff;
         fileId.LustreFid.Seq = (nodeId >> 24) & 0xffffffffff;
 
-        auto handle = fileId.Open(RootHandle, O_PATH);
-        return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        try {
+            auto handle = fileId.Open(RootHandle, O_PATH);
+            return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        } catch (...) {
+        }
+
+        return nullptr;
     }
 };
 
@@ -284,8 +289,13 @@ public:
         // https://github.com/torvalds/linux/blob/dd83757f6e686a2188997cb58b5975f744bb7786/fs/nfs/export.c#L98
         fileId.VastNfsInodeId.FileType = S_IFREG;
 
-        auto handle = fileId.Open(RootHandle, O_PATH);
-        return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        try {
+            auto handle = fileId.Open(RootHandle, O_PATH);
+            return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        } catch (...) {
+        }
+
+        return nullptr;
     }
 };
 
@@ -428,6 +438,9 @@ private:
         try {
             entries = SnapshotsDirNode->List(true /* ignore errors */);
         } catch (...) {
+            ReportLocalFsFailedToRefreshSnapshotsDir(
+                TStringBuilder()
+                << "List failed, exception=" << CurrentExceptionMessage());
         }
 
         for (auto& entry: entries) {
@@ -442,15 +455,19 @@ private:
                 newMap[fileId.WekaInodeId.AntiCollisionId] =
                     fileId.WekaInodeId.SnapViewId;
             } catch (...) {
+                ReportLocalFsFailedToRefreshSnapshotsDir(
+                    TStringBuilder()
+                    << "Open failed, name=" << entry.first
+                    << ", exception=" << CurrentExceptionMessage());
             }
         }
+
+        LastSnapshotsDirRefreshCycles = GetCycleCount();
 
         if (newMap != AntiCollisionToSnapViewMap) {
             AntiCollisionToSnapViewMap = std::move(newMap);
             OnSnapshotsChanged();
         }
-
-        LastSnapshotsDirRefreshCycles = GetCycleCount();
     }
 };
 
@@ -510,42 +527,63 @@ TString TNodeMapper::ToString() const
                             << Hex(SnapshotsNode->GetNodeId()) << ")";
 }
 
-std::optional<TIndexNodePtr> TNodeMapper::RemapNode(
+TNodeMapper::TRemapResult TNodeMapper::RemapNode(
     ui64 parentNodeId,
     const TString& name)
 {
     if (parentNodeId == RootNodeId && name == ".snapshots") {
-        return SnapshotsNode;
+        return TRemapResult::Remapped(SnapshotsNode);
     }
 
     if (parentNodeId != SnapshotsNode->GetNodeId()) {
-        return {};
+        return TRemapResult::NotNeeded();
+    }
+
+    NLowLevel::TFileStatEx stat;
+
+    try {
+        stat = SnapshotsNode->Stat(name);
+    } catch (...) {
+        // name doesn't exist in snapshots directory
+        return TRemapResult::NotFound();
+    }
+
+    if (!stat.IsDir()) {
+        // ignore non-directory nodes in snapshots directory
+        return TRemapResult::NotFound();
     }
 
     try {
-        auto stat = SnapshotsNode->Stat(name);
         TWriteGuard guard(NodeLoaderLock);
-        return {NodeLoader.LoadSnapshotDirNode(stat.INode)};
+        return TRemapResult::Remapped(
+            NodeLoader.LoadSnapshotDirNode(stat.INode));
     } catch (...) {
+        ReportLocalFsFailedToLoadSnapshotDir(
+            TStringBuilder()
+            << "name=" << name << ", exception=" << CurrentExceptionMessage());
     }
 
-    return {nullptr};
+    return TRemapResult::NotFound();
 }
 
-std::optional<TIndexNodePtr> TNodeMapper::RemapNode(
+TNodeMapper::TRemapResult TNodeMapper::RemapNode(
     ui64 parentNodeId,
     ui64 nodeId)
 {
     if (parentNodeId != SnapshotsNode->GetNodeId()) {
-        return {};
+        return TRemapResult::NotNeeded();
     }
 
     try {
         TWriteGuard guard(NodeLoaderLock);
-        return {NodeLoader.LoadSnapshotDirNode(nodeId)};
+        return TRemapResult::Remapped(
+            NodeLoader.LoadSnapshotDirNode(nodeId));
     } catch (...) {
+        ReportLocalFsFailedToLoadSnapshotDir(
+            TStringBuilder() << "nodeId=0x" << Hex(nodeId)
+                             << ", exception=" << CurrentExceptionMessage());
     }
 
-    return {nullptr};
+    return TRemapResult::NotFound();
 }
 }   // namespace NCloud::NFileStore
--- a/cloud/filestore/libs/service_local/index.h
+++ b/cloud/filestore/libs/service_local/index.h
@@ -4,6 +4,7 @@
 
 #include "lowlevel.h"
 
+#include <cloud/filestore/libs/diagnostics/critical_events.h>
 #include <cloud/filestore/libs/service/filestore.h>
 
 #include <cloud/storage/core/libs/common/persistent_table.h>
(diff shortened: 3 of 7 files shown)


cloud/filestore/config/server.proto
```diff
@@ -206,6 +206,14 @@ message TLocalServiceConfig
 
     // Enable zero-copy for read operations in the VFS FUSE layer
     optional bool ZeroCopyReadEnabled = 37;
+
+    // Propagate .snapshots dir from backend file system (currently supported
+    // only on weka)
+    optional bool SnapshotsDirEnabled = 38;
```
debnatkh, line 212: Probably, `.snapshot` path should be configurable, and empty might signify that it is disabled?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,272 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+    } catch (...) {
+    }
+
+    return {nullptr};
 }
 
+std::optional<TIndexNodePtr> TNodeMapper::RemapNode(
+    ui64 parentNodeId,
+    ui64 nodeId)
+{
```
debnatkh, line 572: Do you think it is reasonable to make RemapNode a configurable per-fs set of pairs {src, dst}?
Then this code can be done weka-specifics oblivious

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,272 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        // taking a handle for the mount root inode, but replace its
+        // AntiCollisionId and SnapViewId with those from the snapshot inode
+        // so the handle resolves to the snapshot view of fsdir.
+
+        // Copy InodeId and FsId from RootFileId
+        NLowLevel::TFileId snapshotFileId(RootFileId);
+
+        // Use AntiCollisionId from the snapshot dir inode
+        snapshotFileId.WekaInodeId.AntiCollisionId =
+            snapshotDirNodeId & 0xffff;
```
debnatkh: Can you clarify this one a bit?

cloud/filestore/config/server.proto
budevg: actually it's better to keep this as boolean. The name is hardcoded per fs type and if in the future we will have to support some external file system where the name will be something else we will need to implement this in per fs specific NodeMapper. Right now there is single node mapper and to keep it simple the hardcoded ".snapshots" is used

cloud/filestore/libs/service_local/index.cpp
budevg: I think it won't work well with set of pairs. We have infinite set of pairs for directories inside .snapshots (every sub snapshot) so it easier to implement the logic in code which makes node mapper less generic but it still encapsulates this logic

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        TVector<NLowLevel::TDirEntry> entries;
+
+        try {
+            entries = SnapshotsDirNode->List(true /* ignore errors */);
+        } catch (...) {
+        }
+
+        for (auto& entry: entries) {
+            try {
+                if (!entry.second.IsDir()) {
```
debnatkh, line 448: Is it possible for a client to do `cd .snapshot && touch filename`? Or are there guarantees that the contents are just .snap dirs?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        } catch (...) {
+        }
+
+        return nullptr;
+    }
+
+    TIndexNodePtr LoadSnapshotsDirNode() override
+    {
+        // We need a handle to the snapshots dir which resides at the root
+        // of weka file system. The mount root (fsdir) is also resides at this
```
debnatkh: is also -> also

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        // Example:
+        //   root:                  /mnt/weka/wekadir/fsdir
+        //   snapshots dir:         /mnt/weka/wekadir/.snapshots
+        //
+        // This is a special directory with InodeId == 0x2
+        if (!SnapshotsDirNode) {
+            NLowLevel::TFileId snapshotFileId(RootFileId);
+            snapshotFileId.WekaInodeId.InodeId = 0x2;
+            SnapshotsDirNode =
+                std::move(LoadNode(snapshotFileId.WekaInodeId.InodeContext));
```
debnatkh, line 356: Could LoadNode throw an exception? For example, when we hit the nr_open limit, or if the handle turns out to be incorrectly constructed?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+    }
+
+    if (parentNodeId != SnapshotsNode->GetNodeId()) {
+        return {};
+    }
+
+    try {
+        auto stat = SnapshotsNode->Stat(name);
+        TWriteGuard guard(NodeLoaderLock);
+        return {NodeLoader.LoadSnapshotDirNode(stat.INode)};
```
debnatkh: So, upon each stat, RemapNode will be called, and RefreshAntiCollisionToSnapViewMap will be called. Is that correct? Shouldn't it hinder performance?

cloud/filestore/libs/service_local/index.cpp
budevg: no. the directory is readonly with only snapshot directories inside

cloud/filestore/libs/service_local/index.cpp
budevg: Yes but the `RefreshAntiCollisionToSnapViewMap` has `SnapshotsDirRefreshInterval` which makes it nop most of the time.

if the user doing some restore he will
```
cd /mnt/fs/.snapshots/snap3
cp -r dir /mnt/fs/dir
```
once the snap3 inode will be resolved it will be cached and further access won't trigger `RefreshAntiCollisionToSnapViewMap` unless used explicitly stat /mnt/fs/.snapshots again

We also assume that the restore operation should not happen frequently and will happen only during emergencies

cloud/filestore/libs/service_local/index.cpp
budevg: yes but the `SnapshotsDirNode` is loaded immideatly on start of the filestore-vhost-local. So it's unlikely it will fail on nr_open limit. If it fails then the exception is propagated to TLocalIndex and it will start without snapshots support
```c++
            try {
                if (NodeLoader && SnapshotsDirEnabled) {
                    NodeMapper =
                        std::make_unique<TNodeMapper>(*NodeLoader, NodesLock);
                    STORAGE_INFO(
                        "Inititialize NodeMapper, NodeMapper="
                        << NodeMapper->ToString());
                }
            } catch (...) {
                STORAGE_ERROR(
                    "Failed to initialize NodeMapper" <<
                    ", Exception=" << CurrentExceptionMessage());
            }
```

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        {
+            return;
+        }
+
+        TMap<ui16, ui32> newMap;
+        TVector<NLowLevel::TDirEntry> entries;
+
+        try {
+            entries = SnapshotsDirNode->List(true /* ignore errors */);
+        } catch (...) {
```
SvartMetal, line 440: Maybe fire a critical event?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+    if (parentNodeId == RootNodeId && name == ".snapshots") {
+        return SnapshotsNode;
+    }
+
+    if (parentNodeId != SnapshotsNode->GetNodeId()) {
+        return {};
+    }
+
+    try {
+        auto stat = SnapshotsNode->Stat(name);
```
SvartMetal: Let's check that returned inode is a directory?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        TWriteGuard guard(NodeLoaderLock);
+        return {NodeLoader.LoadSnapshotDirNode(stat.INode)};
+    } catch (...) {
```
SvartMetal, line 560: Log error?
Or fire a critical event

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+    ui64 nodeId)
+{
+    if (parentNodeId != SnapshotsNode->GetNodeId()) {
+        return {};
+    }
+
+    try {
+        TWriteGuard guard(NodeLoaderLock);
+        return {NodeLoader.LoadSnapshotDirNode(nodeId)};
+    } catch (...) {
```
SvartMetal, line 581: Log error?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+            } catch (...) {
+            }
+        }
+
+        if (newMap != AntiCollisionToSnapViewMap) {
+            AntiCollisionToSnapViewMap = std::move(newMap);
+            OnSnapshotsChanged();
+        }
+
+        LastSnapshotsDirRefreshCycles = GetCycleCount();
```
SvartMetal: Usually, it is better not to modify class member variables after a callback has been invoked (for safety)

cloud/filestore/libs/service_local/fs_node.cpp
```diff
@@ -236,6 +236,20 @@ NProto::TListNodesResponse TLocalFileSystem::ListNodes(
...
+            session->RemapNode(parent->GetNodeId(), entry.second.INode);
+        if (remapResult) {
+            auto remappedNode = *remapResult;
+            if (!remappedNode) {
+                continue;
+            }
+            try {
+                entry.second = remappedNode->Stat();
+            } catch (...) {
+                continue;
```
SvartMetal, line 252: Should we log the error?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -209,34 +210,66 @@ void TIndexNode::RemoveXAttr(const TString& name)
...
+        NLowLevel::TFileId fileId(RootFileId);
         fileId.LustreFid.Oid = nodeId & 0xffffff;
         fileId.LustreFid.Seq = (nodeId >> 24) & 0xffffffffff;
-        break;
-    case NLowLevel::TFileId::EFileIdType::Weka:
-        fileId.WekaInodeId.Id = nodeId;
-        break;
-    case NLowLevel::TFileId::EFileIdType::VastNfs:
+
+        auto handle = fileId.Open(RootHandle, O_PATH);
```
SvartMetal: Should we catch exception here as we do for TWekaNodeLoader?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
         // there
         // https://github.com/torvalds/linux/blob/dd83757f6e686a2188997cb58b5975f744bb7786/fs/nfs/export.c#L98
         fileId.VastNfsInodeId.FileType = S_IFREG;
-        break;
-    default:
-        STORAGE_THROW_SERVICE_ERROR(E_FS_NOTSUPP);
+
+        auto handle = fileId.Open(RootHandle, O_PATH);
```
SvartMetal: Should we catch exception here as we do for TWekaNodeLoader?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+        auto snapViewId = GetSnapViewId(fileId.WekaInodeId.AntiCollisionId);
+        if (!snapViewId) {
+            return nullptr;
+        }
+        fileId.WekaInodeId.SnapViewId = *snapViewId;
+
+        try {
+            auto handle = fileId.Open(RootHandle, O_PATH);
+            return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        } catch (...) {
```
SvartMetal, line 336: Log the error?
Critical event?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -250,18 +283,269 @@ TIndexNodePtr TNodeLoader::LoadNode(ui64 nodeId) const
...
+                if (!entry.second.IsDir()) {
+                    continue;
+                }
+
+                auto handle =
+                    SnapshotsDirNode->OpenHandle(entry.first, O_RDONLY, 0);
+                TFileId fileId(handle);
+                newMap[fileId.WekaInodeId.AntiCollisionId] =
+                    fileId.WekaInodeId.SnapViewId;
+            } catch (...) {
```
SvartMetal, line 457: Log the error? 
Critical event?

cloud/filestore/libs/service_local/index.h
```diff
@@ -318,9 +375,21 @@ class TLocalIndex
...
-        Nodes.insert(root);
+            try {
+                if (NodeLoader && SnapshotsDirEnabled) {
+                    NodeMapper =
+                        std::make_unique<TNodeMapper>(*NodeLoader, NodesLock);
+                    STORAGE_INFO(
+                        "Inititialize NodeMapper, NodeMapper="
+                        << NodeMapper->ToString());
+                }
+            } catch (...) {
```
SvartMetal, line 427: Critical event?

cloud/filestore/libs/service_local/index.h
```diff
@@ -111,21 +111,36 @@ struct INodeLoader
...
-
-    [[nodiscard]] TIndexNodePtr LoadNode(ui64 nodeId) const;
-
+    explicit TNodeMapper(INodeLoader& nodeLoader, TRWMutex& nodeLoaderLock);
+
+    // Return value:
+    // - value not present - no need to remap
+    // - value present and nullptr - remap didn't find the node
+    // - value present and != nullptr - the node was remapped successfully
+    std::optional<TIndexNodePtr> RemapNode(
```
SvartMetal: struct TRemapResult would be more readable

cloud/filestore/libs/service_local/index.cpp
budevg: I'd rather not log here because the inode may already be deleted. The NodeLoader - LocalIndex contract is that the index attempts to load a node by inode if it's missing from cache. If that load fails, the upper layer will report an error or propagate it to the VM as ENOTFOUND.

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -255,8 +255,13 @@ class TLustreNodeLoader: public TNodeLoader
         fileId.LustreFid.Oid = nodeId & 0xffffff;
         fileId.LustreFid.Seq = (nodeId >> 24) & 0xffffffffff;
 
-        auto handle = fileId.Open(RootHandle, O_PATH);
-        return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        try {
+            auto handle = fileId.Open(RootHandle, O_PATH);
+            return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        } catch (...) {
```
debnatkh, line 261: Is it okay to drop exceptions here without logging/reporting?

cloud/filestore/libs/service_local/index.cpp
```diff
@@ -284,8 +289,13 @@ class TVastNfsNodeLoader: public TNodeLoader
         // https://github.com/torvalds/linux/blob/dd83757f6e686a2188997cb58b5975f744bb7786/fs/nfs/export.c#L98
         fileId.VastNfsInodeId.FileType = S_IFREG;
 
-        auto handle = fileId.Open(RootHandle, O_PATH);
-        return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        try {
+            auto handle = fileId.Open(RootHandle, O_PATH);
+            return std::make_shared<TIndexNode>(nodeId, std::move(handle));
+        } catch (...) {
```
debnatkh, line 295: Same question about dropping exception

budevg: yes the fileId.Open is invoking `open_by_handle_at` which may fail for various reasons. The handle can be invalid since the inode was deleted for example or we got some non existing inode number from guest kernel and want to resolve it. We ignore exception on purpose and return nullptr indicating that this inode can not be loaded from inode number. The higher layer using the TLocalIndex will raise the error which can be  E_NOENT or some other error depending on the filesystem operation performed

cloud/filestore/libs/service_local/index.cpp
budevg: same response as in WekaNodeLoader

Issue #4942, opened 2026-01-06
Data Corruption Due to CommitId Overflow DestroySession


neihar:
cause similar to https://github.com/ydb-platform/nbs/issues/4937

look at ScheduleRebootTabletOnCommitIdOverflow(ctx, "DestroySession");

this can cause the session and its handles to be not deleted - garbage will be left after unmount

PR #5027, opened 2026-01-17
[Filestore] Handle commit id overflow for Destroy session correctly and ut

neihar:
#4942


cloud/filestore/libs/storage/tablet/tablet_tx.h
```diff
@@ -549,6 +549,8 @@ struct TTxIndexTablet
         /* const */ NProtoPrivate::TDestroySessionRequest Request;
 
         TNodeSet Nodes;
+        ui64 CommitId = 0;
```
qkrorlqr: In all other transactions `CommitId` default value is `InvalidCommitId` - let's make the default here the same and introduce a `bool CommitIdOverflowDetected` flag

neihar: resolved

Issue #4945, opened 2026-01-06
Data Corruption Due to CommitId Overflow RenameNodeInDestination


neihar:
cause similar to https://github.com/ydb-platform/nbs/issues/4937

look at ScheduleRebootTabletOnCommitIdOverflow(ctx, "RenameNodeInDestination");

Issue #4937, opened 2026-01-06
Data Corruption Due to CommitId Overflow WriteData


neihar:
**Problem**: When the tablet's commit ID counter reached maximum value, write operations would fail without setting up the args.Error field in the transaction args and the tablet would schedule a reboot for itself. In this case if the tablet reaches the completion phase of the transaction earlier than the reboot signal is processed, we would return a successful result code to the client - i.e. would acknowledge the data that hasn't been persistently written.

**Root Cause**: On commit ID overflow, the tablet would reboot immediately via RebootTabletOnCommitOverflow() without sending error responses to clients.

**Solution**: Modified all write operations (WriteData, AddData) to return explicit E_REJECTED error ("CommitId overflow") before scheduling tablet reboot. Added MaxTabletStep configuration parameter and comprehensive unit test to verify proper error handling and data consistency.

PR #5028, opened 2026-01-17
issue-2674, issue-4945: added a simple RenameNodeInDestination ut and fixed CommitIdOverflow race for this method

qkrorlqr:
### Notes
* added `TIndexTabletTest_NodesInternal` test suite - currently it contains only two simple uts for `RenameNodeInDestination` but I'm planning to add a lot more
* fixed CommitIdOverflow vs Tx completion race in `RenameNodeInDestination` similar to https://github.com/ydb-platform/nbs/issues/4937 

### Issue
https://github.com/ydb-platform/nbs/issues/4945

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
@@ -14,13 +14,42 @@ using namespace NKikimr;
 
 Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
 {
-    // TODO:
-    // 1. check error when dest == local node
-    // 2. check CommitIdOverflow for otherwise correct request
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -0,0 +1,147 @@
...
+                rebootTracker.ClearPipeDestroyed();
+            }
+        };
+
+        ui32 failures = 0;
+
+        TVector<TString> names;
+        TVector<NProto::TNodeAttr> nodes;
+
+        for (ui32 i = 0; i < 10; ++i) {
```
neihar, line 96: Just FYI for future similar tests, we can make test more readable by using specific maxCommitId and loop for tested request. Yes, it is more fragile for future changes (if amount of commitIds before request changed).

PR #5035, opened 2026-01-19
issue-2674, issue-4946: PrepareRenameNodeInSource unit tests + CommitIdOverflow fix; added private methods for OpLog manipulation

qkrorlqr:
### Notes
* Fixed `PrepareRenameNodeInSource` CommitId overflow handling
* Added a unit test for `PrepareRenameNodeInSource` which tests its logic including `OpLog` check and the behaviour upon CommitId overflow
* Introduced `OpLog` manipulation private API and PrepareRenameNodeInSource private API - to use them in the unit tests related to `RenameNode`
* Moved `tablet_private.h/cpp` to a separate lib because they are used from the `tablet/actors` without a proper `PEERDIR` which doesn't let us use `tablet/protos` inside `tablet_private.h`  

### Issue
https://github.com/ydb-platform/nbs/issues/4946


cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -142,6 +142,148 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+
+            fileNames.push_back(fileName);
+            ++i;
+        }
+
+        //
+        // Calling PrepareRenameNodeInSource for each of those fake node-refs.
+        // Comparing the response to the expected values as well, not just
+        // checking CommitIdOverflow processing logic.
+        //
```
neihar, line 220: do we want to write comments in that style everywhere (with space before and after)? I saw that you started to use it in writedata 2 weeks ago, now you continue.

cloud/filestore/libs/storage/tablet/tablet_actor_oplog.cpp
```diff
@@ -144,6 +171,77 @@ void TIndexTabletActor::CompleteTx_DeleteOpLogEntry(
         "%s DeleteOpLogEntry completed (%lu)",
         LogTag.c_str(),
         args.EntryId);
+
+    if (args.RequestInfo) {
```
neihar, line 175: What do you think about emitting some generic event or consuming generic callback and pass request handling inside it. I know that my option will do technically exactly the same, but I don't like here explicit request that is needed only for tests

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -337,6 +357,27 @@ void TIndexTabletActor::CompleteTx_PrepareRenameNodeInSource(
...
 
+    Y_DEFER {
+        if (args.CommitId == InvalidCommitId) {
+            ScheduleRebootTabletOnCommitIdOverflow(
+                ctx,
+                "PrepareRenameNodeInSource");
+        }
+    };
+
+    if (args.IsExplicitRequest) {
```
neihar, line 368: The same as above regarding leaking this info to this layer, but probably in this case the removal from here not so simple. Maybe at least consider naming like "IsExplicit{test || debug || ...}Request"

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -337,6 +357,27 @@ void TIndexTabletActor::CompleteTx_PrepareRenameNodeInSource(
         args.Error = MakeError(E_INVALID_STATE, std::move(message));
     }
 
+    Y_DEFER {
+        if (args.CommitId == InvalidCommitId) {
```
neihar, line 361: better to use CommidIdOverflow flag

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -142,6 +142,148 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+
+        for (ui32 i = 0; i < iterations; ++i) {
```
neihar, line 222: If we fail, we still increment the counter. Is this the intended behavior? From test perspective it will assert anyway ofc.

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -142,6 +142,148 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+                dstRequest.GetNewName());
+            UNIT_ASSERT_VALUES_EQUAL(dstShardId, dstRequest.GetFileSystemId());
+            UNIT_ASSERT_VALUES_EQUAL(
+                shardId,
+                dstRequest.GetSourceNodeShardId());
+            UNIT_ASSERT_VALUES_EQUAL(
+                shardNodeName,
+                dstRequest.GetSourceNodeShardNodeName());
+
+            //
```
neihar, line 270: for comments other than "comments for requests" it looks strange and takes too much space

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -142,6 +142,148 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+            // Deleting OpLogEntry so that the tablet doesn't try to replay it
+            // upon restart. The request can't be replayed because dst shard
+            // doesn't exist.
+            //
+
+            tablet.DeleteOpLogEntry(response->OpLogEntryId);
+        }
+
+        UNIT_ASSERT_C(
```
neihar, line 279: at least 3, because you create node first

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -337,6 +357,27 @@ void TIndexTabletActor::CompleteTx_PrepareRenameNodeInSource(
...
+
+    if (args.IsExplicitRequest) {
+        //
+        // This branch is supposed to be used only in unit tests.
+        //
+
+        using TResponse =
+            TEvIndexTabletPrivate::TEvPrepareRenameNodeInSourceResponse;
+        auto response = std::make_unique<TResponse>(args.Error);
+        response->OpLogEntryId = args.OpLogEntry.GetEntryId();
```
neihar, line 376: if error, should we return EntryId?

cloud/filestore/libs/storage/tablet/tablet_tx.h
```diff
@@ -2330,17 +2334,39 @@ struct TTxIndexTablet
...
+            , EntryId(entryId)
+        {}
+
+        void Clear() override
+        {
+        }
+    };
+
+    //
+    // GetOpLogEntry
```
neihar, line 2353: Image

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -268,9 +289,8 @@ void TIndexTabletActor::ExecuteTx_PrepareRenameNodeInSource(
...
     args.CommitId = GenerateCommitId();
     if (args.CommitId == InvalidCommitId) {
-        return ScheduleRebootTabletOnCommitIdOverflow(
-            ctx,
-            "PrepareRenameNodeInSource");
+        args.Error = ErrorCommitIdOverflow();
+        return;
     }
 
     // OpLogEntryId doesn't have to be a CommitId - it's just convenient
```
neihar, line 296: now we know why it was InvalidNodeId by default

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
qkrorlqr: Yes, will refactor all such places in a single separate PR

cloud/filestore/libs/storage/tablet/tablet_actor_oplog.cpp
qkrorlqr: I'm not sure about a callback but probably we can implement some generic `if (args.RequestInfo) {Reply(...);}` code in the transaction definition macro and probably we can also add `if (args.CommitIdOverflowDetected) {ScheduleReboot(...);}` there, but let's do it separately

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
qkrorlqr: I want to test this specific transaction but there's no natural API to call it directly. So I needed to introduce some API - which is inevitably a request-response event pair. And then I need to somehow change the post-tx-commit behaviour - I could encapsulate it in a callback and pass a different callback in the test request event handler but IMO it would look more cumbersome.

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
qkrorlqr: the value of `OpLogEntryId` doesn't matter in the case when we return an error so I decided not to bother with an extra `if`

cloud/filestore/libs/storage/tablet/tablet_tx.h
qkrorlqr: Yes, I'm just following the style that's used in our `*_tx.h` files =/

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
qkrorlqr: IMO such style is easier to read when the comment is relatively long (multiple lines)

But I would rather not enforce any specific style

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
qkrorlqr: Yes, it is, it's ok that we skip some of the node refs

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
qkrorlqr: ok, will fix in one of the next PRs

PR #5050, opened 2026-01-21
Set ValidatedBlocksRatio from DiskAgent config

drbasic:


cloud/blockstore/libs/daemon/common/bootstrap.cpp
```diff
@@ -791,11 +791,14 @@ void TBootstrapBase::InitLocalService()
         CreateLocalStorageProvider(
             FileIOServiceProvider,
             NvmeManager,
-            {.DirectIO = false,
-             .UseSubmissionThread = false,
-             .DataIntegrityValidationPolicy =
-                 Configs->DiskAgentConfig
-                     ->GetDataIntegrityValidationPolicyForDrBasedDisks()}));
+            TLocalStorageProviderParams{
```
budevg, line 794: There are additional places where CreateLocalStorageProvider is used and you don't pass the parameter there

https://github.com/ydb-platform/nbs/blob/a182bc047eb7a9ad8684851fdee9e17ba47ab4c3/cloud/blockstore/libs/storage/disk_agent/model/bootstrap.cpp#L119

https://github.com/ydb-platform/nbs/blob/a182bc047eb7a9ad8684851fdee9e17ba47ab4c3/cloud/blockstore/libs/storage/disk_agent/model/bootstrap.cpp#L138

Actually I think the main place where storage provider initalized for disk agent is here https://github.com/ydb-platform/nbs/blob/a182bc047eb7a9ad8684851fdee9e17ba47ab4c3/cloud/blockstore/libs/storage/disk_agent/model/bootstrap.cpp#L182

PR #5053, opened 2026-01-21
fix data race in TDescribeVolumeTest::ShouldReplyRetriableErrorOnTimeout

yegorskii:
Fixes:
https://github-actions-s3.storage.eu-north1.nebius.cloud/ydb-platform/nbs/PR-check/21148543927/1/nebius-x86-64-tsan/logs/1/cloud/blockstore/libs/cells/impl/ut/test-results/unittest/chunk9/testing_out_stuff/TDescribeVolumeTest.ShouldReplyRetriableErrorOnTimeout.err

The problem is that scheduler in test is not stopped upon test completion (destructor is not called because there still active reference to it captured in currently executed lambda https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/cells/impl/describe_volume.cpp#L176). So lambda is executed and at the same time AtExit is called.

One way to fix this was to just insert Y_DEFER { Scheduler->Stop();} . But looking at the code, it seems like TMultiCellDescribeHandler does not need to hold Scheduler, so I made Scheduler parameter for Start method where it is needed.

### DIFF
--- a/cloud/blockstore/libs/cells/impl/describe_volume_ut.cpp
+++ b/cloud/blockstore/libs/cells/impl/describe_volume_ut.cpp
@@ -145,6 +145,7 @@ Y_UNIT_TEST_SUITE(TDescribeVolumeTest)
         bootstrap.Logging = CreateLoggingService("console");
         bootstrap.Scheduler = CreateScheduler();
         bootstrap.Scheduler->Start();
+        Y_DEFER { bootstrap.Scheduler->Stop(); };
 
         auto response = DescribeVolume(
             config,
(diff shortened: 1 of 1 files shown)


agalibin (dismissed):

cloud/blockstore/libs/cells/impl/describe_volume.cpp
```diff
@@ -89,7 +89,6 @@ class TDescribeResponseHandler
 struct TMultiCellDescribeHandler
     : public std::enable_shared_from_this<TMultiCellDescribeHandler>
 {
-    const ISchedulerPtr Scheduler;
```
SvartMetal, line 92: From my point of view, storing TScheduler as a member is a common practice and it would be better to modify the test's code instead

PR #5055, opened 2026-01-22
feat: enable ability to split workload between multiple runners

librarian:
We have 288 core hosts at our disposal, which split in two NUMA nodes 144 cores each. We use 96vcpu preset that can be allocated only on single NUMA node. So maximum we could allocate 2 VMs per node, leaving rest of it on the table. 

Meanwhile we have a set of tests that can be run in parallel, and if we run them on 48vcpu nodes we could fit 6 vms per node with nothing left.

Additionally it _can_ speed up builds by launching multiple vms at once. It would have been especially beneficial if we used pooled vms.

Technically I could've changed only pr*yaml files, without adding proxy workflow, but it looked ugly and was prone to mistakes from copy-paste code.
Additionally to future-proof code and be able to test it independently shell script to calculate which workflows to launch was replaced by python script `.github/scripts/pr_build_and_test_matrix_plan.py` that calculates which type of workflow to launch with what targets.
Then it calculates, in proxy workflow, it calculates how to split targets in `.github/scripts/set_build_and_test_targets.py` 

Nightly builds are not splitted (up for a debate) but instead are built on separate preset with more cores (128vcpu).

Comments on the changes itself (not related to feature itself):
1. I started uploading pytest results to s3 instead of using it as an GH artifact, since it is more comfortable to use.
2. `.temporary` dir added to gitignore to prevent scripts from workflow to be added to the repo.
3. Uploaded workflow test results to s3 and made it leave a comment. Maybe in the future to change it so it will put all test results in single comment and flood the pipeline.


.github/actions/nebius_calculate_pr_matrix/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+      shell: bash
+      run: |
+        pip install -r .github/scripts/requirements.txt
+    - id: plan
+      shell: bash
+      run: |
+        export PYTHONPATH=$PYTHONPATH:$GITHUB_WORKSPACE/.github
+        python3 -m scripts.pr_build_and_test_matrix_plan
+      env:
+          CONTAINS_BLOCKSTORE: ${{ contains(github.event.pull_request.labels.*.name, 'blockstore') && 'true' || 'false' }}
```
SealOfTime, line 27: It seems like something, that should be done via diff analysis, rather than arbitrary placed labels
As otherwise, it's easy to forget to mention one of the modules

PR #5059, opened 2026-01-22
issue-2674: RenameNodeInDestination - src/dst node type checking

qkrorlqr:
### Notes
Implemented the following check in `RenameNodeInDestination` - if dst is a directory, then src should also be a directory. This requires us to perform src+dst `GetNodeAttr` calls after the first `RenameNodeInDestination` tx run - if upon the first run we find out that the `NodeRef` = `NewParentId`/`NewName` exists and the `F_EXCHANGE` flag is not used.

Added missing `TryLockNodeRef`/`UnlockNodeRef` calls into `RenameNodeInDestination` as well.

Added a unit test for the scenario when src == directory and dst == non-empty directory - the error-checking part of it is currently commented-out, I will implement the check in the tablet logic in a separate PR because it will be a non-trivial change. 

// https://github.com/ydb-platform/nbs/issues/2674#issuecomment-2578785398 - overall cross-shard `RenameNode` algorithm

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_destination.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_destination.cpp
@@ -24,8 +24,8 @@ NProto::TError ValidateRequest(
 
 ////////////////////////////////////////////////////////////////////////////////
 
-class TGetNodeAttrActor final
-    : public TActorBootstrapped<TGetNodeAttrActor>
+class TGetExtraSourceAndDestinationInfoActor final
+    : public TActorBootstrapped<TGetExtraSourceAndDestinationInfoActor>
 {
 private:
     const TString LogTag;
@@ -40,7 +40,7 @@ private:
     ui32 ResultCount = 0;
 
 public:
-    TGetNodeAttrActor(
+    TGetExtraSourceAndDestinationInfoActor(
         TString logTag,
         TRequestInfoPtr requestInfo,
         const TActorId& parentId,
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_destination.cpp
```diff
@@ -21,6 +22,193 @@ NProto::TError ValidateRequest(
     return {};
 }
 
+////////////////////////////////////////////////////////////////////////////////
+
+class TGetNodeAttrActor final
```
ruslan-gar: I'd add a comment stating that the actor actually gets attributes of
TDoRenameNodeInDestination::SourceNodeAttr
TDoRenameNodeInDestination::DestinationNodeAttr
We discussed renaming of TGetNodeAttrActor but I think it's not worth doing, 
since this actor exists in the context of this file.

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_destination.cpp
```diff
@@ -288,6 +541,24 @@ void TIndexTabletActor::CompleteTx_RenameNodeInDestination(
     const TActorContext& ctx,
     TTxIndexTablet::TRenameNodeInDestination& args)
 {
+    if (args.SecondPassRequired) {
+        // TODO: verify args.NewChildRef
+
+        auto actor = std::make_unique<TGetNodeAttrActor>(
```
ruslan-gar: I’d add a comment explaining the two-stage processing, which involves creating TGetNodeAttrActor and then handling TEvDoRenameNodeInDestination

qkrorlqr: I'm planning to do something better - I'm planning to write a doc which describes the RenameNode operation in details for sharded filesystems. I'll refer to this doc in the comments.

Issue #5070, opened 2026-01-26
[NBS] use ibv_create_qp instead of rdma_create_qp


tpashkin:
In some cases QPs provided by `rdma_cm` show drastic performance degradation. Replacing `rdma_create_qp` with `ibv_create_qp` seems to help without any noticeable drawbacks

PR #5074, opened 2026-01-26
add an option to use plain verbs qp #5070

tpashkin:
rdma_cm modifies QP into RTS itself
1. INIT – https://github.com/linux-rdma/rdma-core/blob/6e9643ea97d1f312223801b07c19c0a16615eeb4/librdmacm/cma.c#L1283 
2. RTR – https://github.com/linux-rdma/rdma-core/blob/6e9643ea97d1f312223801b07c19c0a16615eeb4/librdmacm/cma.c#L1292 
3. RTS – https://github.com/linux-rdma/rdma-core/blob/6e9643ea97d1f312223801b07c19c0a16615eeb4/librdmacm/cma.c#L1320

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/client.cpp
+++ b/cloud/blockstore/libs/rdma/impl/client.cpp
@@ -595,10 +595,7 @@ TClientEndpoint::TClientEndpoint(
 TClientEndpoint::~TClientEndpoint()
 {
     // release any leftover resources if endpoint hasn't been properly stopped
-    if (CompletionQueue) {
-        DestroyQP();
-    }
-
+    DestroyQP();
     RDMA_INFO("stop endpoint");
 }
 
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -725,7 +729,11 @@ void TClientEndpoint::CreateQP()
 
 void TClientEndpoint::DestroyQP() noexcept
 {
-    Verbs->DestroyQP(Connection.get());
+    if (Config.VerbsQP) {
+        Verbs->DestroyQP(Connection->qp);
```
budevg: Need to set `Connection->qp` to NULL same  way it's done in `rdma_destroy_qp`
https://github.com/linux-rdma/rdma-core/blob/6e9643ea97d1f312223801b07c19c0a16615eeb4/librdmacm/cma.c#L1715

Otherwise it can lead to corruption when some internal rdma functions try to use it or if we call DestroyQp twice

cloud/blockstore/libs/rdma/impl/server.cpp
```diff
@@ -469,7 +473,11 @@ TServerSession::~TServerSession()
     RDMA_INFO("stop session");
 
     if (Connection->qp) {
-        Verbs->DestroyQP(Connection.get());
+        if (Config->VerbsQP) {
+            Verbs->DestroyQP(Connection->qp);
```
budevg: Need to set `Connection->qp` to NULL same  way it's done in `rdma_destroy_qp`
https://github.com/linux-rdma/rdma-core/blob/6e9643ea97d1f312223801b07c19c0a16615eeb4/librdmacm/cma.c#L1715

Otherwise it can lead to corruption when some internal rdma functions try to use it or if we call DestroyQp twice

cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -731,6 +731,7 @@ void TClientEndpoint::DestroyQP() noexcept
 {
     if (Config.VerbsQP) {
         Verbs->DestroyQP(Connection->qp);
```
budevg:
```
if (Connection->qp) {
   Verbs->DestroyQP(Connection->qp);
   Connection->qp = nullptr;
}
```

cloud/blockstore/libs/rdma/impl/client.cpp
tpashkin: There is a check in the calling code

cloud/blockstore/libs/rdma/impl/server.cpp
tpashkin: There is a check already

cloud/blockstore/libs/rdma/impl/client.cpp
tpashkin: There was a check in the calling code already. I've cleaned up the code to be a bit more readable, but it wasn't strictly necessary

PR #5098, opened 2026-01-29
[Blockstore/Disk Agent] Local NVMe service basic implementation

sharpeye:
PR implements the basic infrastructure for the Local NVMe service. Service runs in the embedded Disk Agent.

Introduced `ILocalNVMeDeviceProvider` interface for obtaining NVMe device lists from different sources and its implementation for loading NVMe devices from a file.

Added `TLocalNVMeConfig` with `DevicesSourceUri` field to specify device list source (e.g., `file:///etc/nbs/devices.txt`).
Added `--local-nvme-file` CLI option for config loading.

#5250

### DIFF
--- a/cloud/blockstore/libs/local_nvme/service_linux.cpp
+++ b/cloud/blockstore/libs/local_nvme/service_linux.cpp
@@ -1,0 +1,280 @@
+#include "service.h"
+
+#include "device_provider.h"
+
+#include <cloud/storage/core/libs/coroutine/executor.h>
+#include <cloud/storage/core/libs/diagnostics/logging.h>
+
+#include <library/cpp/threading/future/future.h>
+
+#include <util/generic/algorithm.h>
+#include <util/generic/hash_set.h>
+#include <util/stream/str.h>
+#include <util/string/builder.h>
+
+namespace NCloud::NBlockStore {
+
+using namespace NThreading;
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+const NProto::TError ServiceDestroyedError =
+    MakeError(E_REJECTED, "Local NVMe service is destroyed");
+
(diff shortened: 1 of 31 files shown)


komarevtsev-d (approved):

cloud/blockstore/libs/local_nvme/service_linux.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+const NProto::TError ServiceDestroyedError =
+    MakeError(E_REJECTED, "Local NVMe service is destroyed");
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <typename R>
+TString Join(TStringBuf delim, const R& range)
```
dvrazumov, line 29: Why not use `JoinSeq`?

cloud/blockstore/libs/local_nvme/service_linux_ut.cpp
```diff
@@ -0,0 +1,263 @@
...
+        {
+            auto [_, error] = ListNVMeDevices();
+            UNIT_ASSERT_VALUES_EQUAL_C(
+                E_FAIL,
+                error.GetCode(),
+                FormatError(error));
+        }
+
+        {
+            auto [_, error] = ListNVMeDevices();
```
dvrazumov, line 254: Why is it tested twice?

sharpeye: To make sure that the same error is returned

cloud/blockstore/libs/local_nvme/service_linux.cpp
sharpeye: It isn't part of util/string/join.h - it's just a helper function in one of the cpp files.

PR #5101, opened 2026-01-29
fix: make log files to be identified as text files in s3

librarian:


.github/actions/s3cmd/action.yaml
```diff
@@ -39,7 +39,7 @@ runs:
   steps:
     - name: install s3cmd and awscli
       if: inputs.install == 'true'
-      shell: bash
+      shell: bash
```
SvartMetal: ' ' at the end?

Issue #5103, opened 2026-01-30
[Filestore] Implement strict nodes count limit enforcement


ruslan-gar:
Currently when a node is created, the limit is checked the following way:
```
    if (GetUsedNodesCount() >= GetNodesCount()) {
        args.Error = ErrorNoSpaceLeft();
        return true;
    }
```
GetUsedNodesCount() and  GetNodesCount() are both local to a shard or main filesystem.
In case of strict filesystem GetNodesCount() is calculated using the total filesystem size and actually the limit is multiplied by shards count.
In case of non-strict fs, it is possible to hit the limit in a shard while the total file system limit is not reached.
The behavior of non-strict systems is not going to be changed but strict filesystems should enforce the limit correctly.
So, instead of GetUsedNodesCount, Metrics.AggregateUsedNodesCount should  be used.

PR #5104, opened 2026-01-30
issue-5103: [Filestore] Strictly enforce nodes count limit in filesystems with strict size enforcement

ruslan-gar:
The detail are in the issue  #5103

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
@@ -7045,9 +7045,9 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
         service.SendCreateNodeRequest(
             headers,
             TCreateNodeArgs::File(RootNodeId, "too_many_files"));
-        const ui32 status =
-            STATUS_FROM_CODE(service.RecvCreateNodeResponse()->GetStatus());
-        UNIT_ASSERT_EQUAL(static_cast<ui32>(NProto::E_FS_NOSPC), status);
+        UNIT_ASSERT_EQUAL(
+            E_FS_NOSPC,
+            service.RecvCreateNodeResponse()->GetStatus());
     }
 
     SERVICE_TEST_SID_SELECT_IN_LEADER_ONLY(
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1401,7 +1401,7 @@ bool TIndexTabletActor::HasNodesLeft() const
     // shards.
 
     if (Metrics.AggregateUsedNodesCount < 0) {
-        ReportAggregateUsedNodesCountIsNegative(
+        ReportCounterIsNegative(
             TStringBuilder() << "FileSystem: " << GetFileSystemId()
                              << ". TMetrics::AggregateUsedNodesCount should "
                                 "always be positive.");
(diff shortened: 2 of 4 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1391,6 +1391,33 @@ bool TIndexTabletActor::HasSpaceLeft(ui64 prevSize, ui64 newSize) const
 
 ////////////////////////////////////////////////////////////////////////////////
 
+bool TIndexTabletActor::HasNodesLeft() const
+{
+    if (!GetFileSystem().GetStrictFileSystemSizeEnforcementEnabled()) {
+        if (GetUsedNodesCount() >= GetNodesCount()) {
+            return false;
```
debnatkh: return GetUsedNodesCount() < GetNodesCount(), then else on 1400 wont be used

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1391,6 +1391,33 @@ bool TIndexTabletActor::HasSpaceLeft(ui64 prevSize, ui64 newSize) const
...
+        // shards.
+        TABLET_VERIFY(Metrics.AggregateUsedNodesCount >= 0);
+        TABLET_VERIFY(Metrics.TotalNodesCount >= 0);
+        // It makes sense for shardless filesystems, as it eliminates 15s delay
+        // in AggregateUsedNodesCount calculation.
+        const ui64 usedNodes = Max<ui64>(
+            static_cast<ui64>(Metrics.AggregateUsedNodesCount.load()),
+            GetUsedNodesCount());
+
+        if (usedNodes >= static_cast<ui64>(Metrics.TotalNodesCount.load())) {
```
debnatkh: Looking up Metrics for AggregateUsedNodesCount is fine since it is a regularly updated value, but not TotalNodesCount – each tablet will have a correctly set NodesCount()

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1391,6 +1391,26 @@ bool TIndexTabletActor::HasSpaceLeft(ui64 prevSize, ui64 newSize) const
...
 
+bool TIndexTabletActor::HasNodesLeft() const
+{
+    if (!GetFileSystem().GetStrictFileSystemSizeEnforcementEnabled()) {
+        return GetUsedNodesCount() < GetNodesCount();
+    }
+
+    // A new way to count available nodes uses the count aggregated by all
+    // shards.
+    TABLET_VERIFY(Metrics.AggregateUsedNodesCount >= 0);
```
qkrorlqr: We don't have to crash if the value of some counter is invalid. We can report a crit event and return false.

cloud/filestore/libs/storage/service/service_ut_sharding.cpp
```diff
@@ -6976,5 +6976,83 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
...
+            NActors::TDispatchOptions options;
+            options.FinalEvents.emplace_back(
+                TDispatchOptions::TFinalEventCondition(
+                    TEvIndexTabletPrivate::EvUpdateCounters,
+                    shardCount + 2));
+            env.GetRuntime().DispatchEvents(options);
+        }
+
+        // Creation of one more file should fail.
+        service.AssertCreateNodeFailed(
```
qkrorlqr: better to check that the error code is correct, not that we get "some" error

// that's one of the reasons why AssertXXXFailed method is bad overall

PR #5109, opened 2026-01-30
removed cells config parameter and changed the logic of checking describe volume requests for cells

yegorskii:
Previously describe volume request checking was implemented at both sides (both sender and receiver). Sender checked that response contains cell identifier (if cells feature was on and config parameter for checking was on). Receiver checked cell in incoming request if cells feature was on and incoming request contained cell identifier. 

Decided to change logic here. Only describe volume request will be checked for equivalence of incoming cell id and configured cell id. This is enough to avoid situations we care about.

Removing config parameter is ok? since it was not used in ay clusters

### DIFF
--- a/cloud/blockstore/libs/cells/impl/describe_volume.cpp
+++ b/cloud/blockstore/libs/cells/impl/describe_volume.cpp
@@ -311,12 +311,12 @@ void TDescribeResponseHandler::HandleResponse(const auto& future)
                 "id tag";
             *response.MutableError() = MakeError(E_NOT_FOUND, msg);
         } else {
-            response.SetCellId(Cell.CellId);
-            owner->Reply(std::move(response));
             STORAGE_DEBUG(
                 TStringBuilder() << "DescribeVolume: got success for disk "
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/cells/impl/describe_volume.cpp
```diff
@@ -301,38 +298,26 @@ void TDescribeResponseHandler::HandleResponse(const auto& future)
...
-                msg,
-                {{"expected", Cell.CellId}, {"actual", response.GetCellId()}});
-
-            *response.MutableError() = MakeError(E_REJECTED, msg);
+            response.SetCellId(Cell.CellId);
+            owner->Reply(std::move(response));
+            STORAGE_DEBUG(
+                TStringBuilder() << "DescribeVolume: got success for disk "
+                                 << Request.GetDiskId().Quote() << " from "
+                                 << HostInfo.Fqdn);
```
SvartMetal, line 317: Better to perform logging before replying

Issue #4853, opened 2025-12-16
[Filestore] TStorageServiceActor overhead optimization


qkrorlqr:
There are at least 2 big issues:
* there's a bug because of which `TStorageServiceActor` is spammed with `EvPingSessionRequest`s - we see that for high iops tests (high iodepth and/or numjobs and small request size) `TStorageServiceActor` is 100% busy and 20% of that time is spent processing `EvPingSessionRequest`s
* another 30% is spent in `TStorageServiceActor::CompleteRequest` which mostly does statistics accounting - this thing can be moved to a separate actor which wouldn't be on the critical path for IO processing
* and we also spend considerable amount of time in `TStorageServiceActor::HandleUpdateStats` and in request handlers (performing `TInFlightRequest` registration)

If we don't hit any other bottlenecks, freeing half of the time that `TStorageServiceActor` currently spends processing pings and stats to let it do IO instead may let us increase our max iops by x2.

A simple fio config which I used to trigger the issue:
```
[global]
filesize=2G
time_based=1
startdelay=5
exitall_on_error=1
create_serialize=0
filename_format=$filenum/$jobnum
group_reporting=1
clocksource=gettimeofday
ioengine=libaio
disk_util=0
direct=1

[init]
blocksize=1Mi
rw=write
size=100%
numjobs=64
time_based=0
description='pre-create files'

[read-4k]
stonewall
description='Read iops workload'
iodepth=8
bs=4k
rw=randread
numjobs=64
runtime=120
```

Another interesting thing which is worth testing here is how we scale if we increase guest queue count and max_background.

PR #5110, opened 2026-01-31
issue-4853: logging all requests to profile log, not dropping those where HasData() == false

qkrorlqr:
### Notes
There's a weird check that disables logging to profile log for some of the requests - we can drop important log records because of that check. Before removing that check I added counters to make sure that the amount of log records dropped by that check is not huge: https://github.com/ydb-platform/nbs/commit/c6faf93b5efcda8af4586d8678ff464d269c268a - turns out that it's rather small and we won't inflate our log size so it's safe to remove that check - removing it in this PR. 

### Issue
The commit where I added counters for requests w error / w logdata / w/o logdata was linked to https://github.com/ydb-platform/nbs/issues/4853 so linking this one to the same issue as well

### DIFF
--- a/cloud/filestore/libs/storage/service/service_state.cpp
+++ b/cloud/filestore/libs/storage/service/service_state.cpp
@@ -32,12 +32,10 @@ void TInFlightRequest::Complete(
         currentTs.MicroSeconds() - ProfileLogRequest.GetTimestampMcs());
     ProfileLogRequest.SetErrorCode(error.GetCode());
 
-    if (HasLogData()) {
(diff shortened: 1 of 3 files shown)


SvartMetal (approved):

neihar (approved):

komarevtsev-d (approved):

qkrorlqr:
> Do we need all of these counters after that or some can be removed as they add no value anymore?

they might be useful, they still tell us how many requests have data, how many requests have an error, etc

PR #5112, opened 2026-01-31
rename interface IIncompleteRequestProcessor  to IStatsHandler

neihar:
I am planing to add new lightweight stats without dependency to requests stats and want to reuse some code related to update stats mechanic.

### DIFF
--- a/cloud/blockstore/libs/diagnostics/public.h
+++ b/cloud/blockstore/libs/diagnostics/public.h
@@ -60,6 +60,9 @@ using IMetricConsumerPtr = std::shared_ptr<NMonitoring::IMetricConsumer>;
 struct IIncompleteRequestProvider;
 using IIncompleteRequestProviderPtr = std::shared_ptr<IIncompleteRequestProvider>;
 
+using IIncompleteRequestProcessor = NCloud::IStatsHandler;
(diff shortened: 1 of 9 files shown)


cloud/blockstore/libs/diagnostics/public.h
```diff
@@ -60,6 +60,9 @@ using IMetricConsumerPtr = std::shared_ptr<NMonitoring::IMetricConsumer>;
 struct IIncompleteRequestProvider;
 using IIncompleteRequestProviderPtr = std::shared_ptr<IIncompleteRequestProvider>;
 
+using IIncompleteRequestProcessor = NCloud::IStatsHandler;
```
drbasic, line 63: Are you planning to replace the IIncompleteRequestProcessor in the entire codebase with NCloud::IStatsHandler ?

neihar: No, I do not plan to touch IncompleteRequestProcessor, just Interface with that name (which was leaked from impl), renamed it to more appropriate name which i want to reuse together with StatsUpdater in another place that has 0 connection with IncompleteRequests.

drbasic (approved):

PR #5113, opened 2026-02-01
[Filestore] cleanup recipes: write to the PID files immediately after the service starts; correctly use logger; log the PID when the process is shutting down; remove unneeded clusters/local files

SvartMetal:


cloud/filestore/tests/clusters/local/cluster.yaml
SvartMetal, line 1: https://github.com/ydb-platform/nbs/tree/main/cloud/filestore/clusters/local

Issue #4604, opened 2025-11-04
[Filestore] Add metrics for vhost directory handles


neihar:
https://github.com/ydb-platform/nbs/issues/2390

Before enabling directory handles cache on vhost, we want to collect some metrics regarding cache size.

PR #5120, opened 2026-02-02
Filestore: Add ModuleStats and Directoryhandles stats

neihar:
- Introduced new ModuleStatsRegistry that manage simple stats not related to requests
- Introduced directory handles stats, that can provide information regarding direcotry handles cache expected load
https://github.com/ydb-platform/nbs/issues/4604

### DIFF
--- a/cloud/filestore/libs/diagnostics/request_stats_ut.cpp
+++ b/cloud/filestore/libs/diagnostics/request_stats_ut.cpp
@@ -932,8 +932,11 @@ Y_UNIT_TEST_SUITE(TRequestStatRegistryTest)
 
         // emulating assigning new cloud and folder (after session is created)
         // counters should migrate to new cloud and folder subgroup
-        stats = bootstrap.Registry
-                    ->GetFileSystemStats(FS, CLIENT, newCloud, newFolder);
+        bootstrap.FsCountersProvider->UpdateCloudAndFolder(
+            FS,
+            CLIENT,
+            newCloud,
+            newFolder);
 
         {
             auto fsCounters = rootCounters->FindSubgroup("filesystem", FS);
--- a/cloud/filestore/libs/server/server_ut.cpp
+++ b/cloud/filestore/libs/server/server_ut.cpp
@@ -6,6 +6,7 @@
 #include <cloud/filestore/libs/client/config.h>
 #include <cloud/filestore/libs/diagnostics/config.h>
 #include <cloud/filestore/libs/diagnostics/critical_events.h>
+#include <cloud/filestore/libs/diagnostics/filesystem_counters.h>
 #include <cloud/filestore/libs/diagnostics/profile_log.h>
 #include <cloud/filestore/libs/diagnostics/request_stats.h>
 #include <cloud/filestore/libs/service/context.h>
@@ -304,7 +305,8 @@ public:
             std::make_shared<TDiagnosticsConfig>(),
             Counters,
             CreateWallClockTimer(),
-            NCloud::NStorage::NUserStats::CreateUserCounterSupplierStub());
+            NCloud::NStorage::NUserStats::CreateUserCounterSupplierStub(),
+            CreateFsCountersProviderStub());
 
         ServerConfig = CreateConfig(serverConfig);
 


cloud/filestore/libs/daemon/vhost/bootstrap.h
```diff
@@ -24,6 +25,7 @@ struct TVhostModuleFactories
         ITimerPtr timer,
         ISchedulerPtr scheduler,
         IRequestStatsRegistryPtr requestStats,
+        IModuleStatsRegistryPtr moduleStats,
```
neihar, line 28: We use it currently only for vhost, if needed it can be uplifted to common/bootstrap in the future

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -205,7 +215,11 @@ void TFileSystem::ReadDir(
 
     if (!offset) {
         // directory contents need to be refreshed on rewinddir()
+        DirectoryHandlesStats->DecreaseCacheSize(handle->GetSerializedSize());
```
neihar, line 218: so, probably some wrapper class around all directoryhandles entities should be introduced, will think about that and probably do in separate PR

cloud/filestore/libs/vfs_fuse/directory_handles_stats.h
```diff
@@ -0,0 +1,92 @@
...
+        TStringBuf counterName)
+        : MaxCalc(
+              std::make_unique<TMaxCalculator<BucketCount>>(std::move(timer)))
+        , MaxCounter(counters->GetCounter(TString{counterName}, false))
+    {}
+
+    void Change(i64 delta)
+    {
+        i64 newVal = AtomicAdd(Value, delta);
+        Y_DEBUG_ABORT_UNLESS(newVal >= 0);
```
neihar, line 46: For the max metric, we don’t care if a race between the two atomics occurs, it will work either way.

cloud/filestore/libs/diagnostics/module_stats.cpp
```diff
@@ -0,0 +1,151 @@
...
+    void Unregister(
+        const TString& fileSystemId,
+        const TString& clientId) override
+    {
+        TWriteGuard guard(Lock);
+
+        auto key = std::make_pair(fileSystemId, clientId);
+        StatsMap.erase(key);
+
+        FsCounters->GetSubgroup("filesystem", fileSystemId)
```
qkrorlqr: this will delete all `client_fs` counters for this filesystem and not just module counters

doesn't look right

cloud/filestore/libs/vfs_fuse/directory_handles_stats.cpp
```diff
@@ -0,0 +1,73 @@
...
+    const TString& clientId,
+    const TString& cloudId,
+    const TString& folderId)
+{
+    auto moduleCounters = registry->GetFileSystemModuleCounters(
+        fileSystemId,
+        clientId,
+        cloudId,
+        folderId,
+        "DirectoryHandles");
```
qkrorlqr: nit: usually we don't use CamelCase in counter labels (e.g. we use ssd_nonrepl and not SSDNonrepl, ssd_mirror3 and not SSDMirror3, etc), let's call it "directory_handles"?

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
```diff
@@ -2,8 +2,31 @@
...
+// Fixed-size serialization overhead for TDirectoryHandleChunk
+constexpr size_t BaseSerializedSize = sizeof(ui64) +   // Index
+                                      sizeof(ui64) +   // UpdateVersion
+                                      sizeof(ui32);    // Cookie length
+
+}   // namespace
+
 ////////////////////////////////////////////////////////////////////////////////
 
+size_t TDirectoryHandleChunk::GetSerializedSize() const
```
qkrorlqr, line 18: this looks very fragile
let's at least add unit tests which would check that the actual serialized size is indeed the same as the return value of `GetSerializedSize()`
and let's consider switching to protobuf or flatbuf here instead of our custom format

cloud/filestore/libs/vfs_fuse/directory_handles_stats.cpp
neihar: during discussion we established that stats in CamelCase already in the project and decided to leave module stats as is

cloud/filestore/libs/diagnostics/module_stats.cpp
neihar: made dependency explicit

yegorskii (approved):

PR #5125, opened 2026-02-02
fix wrong function call in logging + fix potential memory leak

yegorskii:
1. Call to GetBlocksCount, but GetBlockSize should be
2. Fix memory leak during exception

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/verbs.h
+++ b/cloud/blockstore/libs/rdma/impl/verbs.h
@@ -55,7 +55,6 @@ struct ICompletionHandler
     virtual void HandleCompletionEvent(ibv_wc* wc) = 0;
 };
 
-
 struct IVerbs
 {
     virtual ~IVerbs() = default;


cloud/blockstore/libs/rdma/impl/test_verbs.cpp
```diff
@@ -132,14 +132,14 @@ struct TTestVerbs
...
-        void* addr,
+        TMemoryRegionHolder addr,
         size_t length,
         int flags) override
     {
         Y_UNUSED(flags);
 
         return {
-            static_cast<ibv_mr*>(new TMemoryRegion(pd, addr, length)),
+            static_cast<ibv_mr*>(new TMemoryRegion(pd, addr.Get(), length)),
```
tpashkin: Shouldn't we release `addr` here, so it won't get freed?

yegorskii: да, мой косяк

Issue #5095, opened 2026-01-29
[NBS] Volume Traffic Shaping: add the ability to throttle responses to clients


komarevtsev-d:
When the volume performs at a faster rate than the guarantees that NBS provides to its clients, it creates an unnecessary load of the storage system. To address this, we can introduce a "shaping throttler" – a new type of throttling mechanism that will delay the response after a request has been processed for a certain period of time.

The existing disk throttler limits the performance by IOPS and throughput. However, with a low parallel load, it does not work. In such case, a shaping throttler can delay responses and thus reduce the load.

PR #5127, opened 2026-02-02
Introduce TResponseHeaders to the blockstore; deprecate old TraceInfo and ThrottlerDelay proto fields

komarevtsev-d:
#5095
Added `NProto::TResponseHeaders`, which is similar to the version in the filestore. The plan is to add a delay of the "minimum throttler" to `NProto::TThrottlerInfo` in future PRs.  
Both `NProto::TTraceInfo` and `ThrottlerDelay` were deprecated visually in the code, but are still used and populated due to compatibility reasons. The newly added response headers are also populated.  After a sufficient period of time, the deprecated fields can be safely removed.

No change of the behavior is expected from this PR.

### DIFF
--- a/cloud/blockstore/libs/storage/service/volume_client_actor.cpp
+++ b/cloud/blockstore/libs/storage/service/volume_client_actor.cpp
@@ -362,7 +362,9 @@ void TVolumeClientActor::HandleResponse(
 
     if (it->second.CallContext->LWOrbit.HasShuttles()) {
         TraceSerializer->HandleTraceInfo(
-            msg->Record.GetDeprecatedTrace(),
+            msg->Record.GetHeaders().HasTrace()
+                ? msg->Record.GetHeaders().GetTrace()
+                : msg->Record.GetDeprecatedTrace(),
             it->second.CallContext->LWOrbit,
             it->second.SendTime,
             GetCycleCount());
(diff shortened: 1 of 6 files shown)


cloud/blockstore/libs/storage/volume_proxy/volume_proxy.cpp
```diff
@@ -809,19 +811,24 @@ void TVolumeProxyActor::HandleResponse(
...
-        msg->Record.ClearTrace();
+        msg->Record.ClearDeprecatedTrace();
+        msg->Record.MutableHeaders()->ClearTrace();
     }
 
     using TProtoType = decltype(TMethod::TResponse::Record);
-    if constexpr (THasGetThrottlerDelay<TProtoType>::value) {
+    static_assert(
+        THasGetDeprecatedThrottlerDelay<TProtoType>::value ==
+        RequiresThrottling<TMethod>);
```
komarevtsev-d, line 827: Ассерт явно показывает что поведение не поменялось на данный момент. Кажется, стало правильнее – наличие (или отсутствие) полей в прото спеке не должно менять поведение кода. 
А `RequiresThrottling` – это конкретный список методов в специально отведенном месте.

cloud/blockstore/libs/storage/service/volume_client_actor.cpp
```diff
@@ -1,9 +1,11 @@
 #include "volume_client_actor.h"
 
+#include "cloud/blockstore/libs/storage/core/forward_helpers.h"
```
drbasic: <>

vladstepanyuk (approved):

PR #5133, opened 2026-02-02
[Blockstore] store Local NVMe Service state in the cache; check if a device already acquired; move protos to local_nvme.proto

sharpeye:
Add acquire/release semantics to prevent concurrent access to NVMe devices;
Cache service state to a file to restore configuration after reboot;

#5250


cloud/blockstore/libs/storage/protos/local_nvme.proto
komarevtsev-d, line 1: коментиков бы к полям тут

dvrazumov (approved):

komarevtsev-d (approved):

Issue #4690, opened 2025-11-18
[Filestore] Optimize the amplification rate between requested bytes and actually read bytes upon ListNodes


debnatkh:
On a production sharded filesystem with 2,000,000+ files in a given directory, listing takes seconds:
```
ListNodes	5.276928s	S_OK	{node_id=146359, size=5796}
ListNodes	5.430266s	S_OK	{node_id=146359, size=5796}
ListNodes	5.258368s	S_OK	{node_id=146359, size=5796}
ListNodes	5.399740s	S_OK	{node_id=146359, size=5796}
ListNodes	5.378499s	S_OK	{node_id=146359, size=5796}
ListNodes	5.483648s	S_OK	{node_id=146359, size=5796}
ListNodes	5.414462s	S_OK	{node_id=146359, size=5796}
ListNodes	5.474946s	S_OK	{node_id=146359, size=5796}
```

Among other symptoms, we see MAX_TX_ListNodes_ReadBytes_ around 1.5GiB:
<img width="532" height="421" alt="Image" src="https://github.com/user-attachments/assets/a0e5f8e7-c3b1-41a2-914c-c3ca11b96e5c" />

This infers 3-8 MiB per request:

<img width="1054" height="593" alt="Image" src="https://github.com/user-attachments/assets/4cca2d1b-ccfa-4857-8c2b-843870d6e6d7" />

However, it should be closer to `150 (bytes per row) * 5796 < 1MiB`

PR #5143, opened 2026-02-03
issue-4690 [Filestore] Add PrepareAttempts metric to ListNodes metrics

debnatkh:
Slightly references #4690

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_ut_cache.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_ut_cache.cpp
@@ -132,8 +132,8 @@ TTxStats GetTxStats(TTestEnv& env, TIndexTabletClient& tablet)
 
 struct TListNodesTxStats
 {
-    ui64 BytesPrecharge;
-    ui64 PrepareAttempts;
+    ui64 BytesPrecharge = 0;
+    ui64 PrepareAttempts = 0;
 };
 
 TListNodesTxStats GetListNodesTxStats(


cloud/filestore/libs/storage/tablet/tablet_ut_cache.cpp
```diff
@@ -130,11 +130,17 @@ TTxStats GetTxStats(TTestEnv& env, TIndexTabletClient& tablet)
     return stats;
 }
 
-ui64 GetListNodesRequestedBytesPrecharge(
+struct TListNodesTxStats
+{
+    ui64 BytesPrecharge;
+    ui64 PrepareAttempts;
```
SvartMetal: Let’s explicitly initialize these to zero

cloud/filestore/libs/storage/tablet/tablet_tx.h
```diff
@@ -1339,6 +1340,7 @@ struct TTxIndexTablet
 
         void OnRestart() override
         {
+            ++PrepareAttempts;
```
SvartMetal, line 1343: `RestartsCount` would be a clearer name to me, but I don’t have a strong opinion here.

debnatkh: It would have been RestartsCount if it were initialized to zero, yet it is initialized to one

I believe it would be easier from the metrics calculations point of view to keep it to zero, meaning that the avg number of prefetches could be estimated as PrepareAttempts / Count as opposed to PrepareAttempts / Count + 1

SvartMetal: > yet it is initialized to one

It is not hard to initialize it to one :)

SvartMetal (approved):

Issue #4875, opened 2025-12-18
Write Fresh Blocks Requests should go past the Partition tablet


vladstepanyuk:
Currently, for YDB disks, all read/write requests go through the Partition tablet.

At the same time, the partition tablet performs synchronous read transactions from the local database during reads and asynchronous operations (Flush, Compression), during which it is blocked. That is, while the Execute transaction is running, the entire tablet is blocked and cannot process writes, even those that go to fresh and do not require writing to the index.

These transactions can take 5-10 ms to complete, as they require reads from the blobstorage.

Max Execution time of transactions on hosts and percentiles per all hosts:
<img width="1782" height="463" alt="Image" src="https://github.com/user-attachments/assets/4f2fcb93-b241-4624-9ea5-76de81492996" />

Because of this, when Compaction begins, the latency doubles and it becomes as unstable as possible.

It is supposed to write a Proxy actor in front of the Partition tablet, which will write fresh blocks in Blob storage, but in order to read, write something to the index in the local database, it will go to the Partition tablet, started nearby in the external boot.

PR #5167, opened 2026-02-05
issue-4875: mv write blob logic to separate companion

vladstepanyuk:
#4875 moving part_actor_writeblob.cpp  to separate companion

### DIFF
--- a/cloud/blockstore/libs/storage/partition/model/resource_metrics_updates_queue.h
+++ b/cloud/blockstore/libs/storage/partition/model/resource_metrics_updates_queue.h
@@ -100,7 +100,7 @@ public:
     TVector<TResourceMetricsUpdate> PopAll()
     {
         TVector<TResourceMetricsUpdate> resCollection;
-        ResourceMetricsUpdates.DequeueAll(&resCollection);
+        ResourceMetricsUpdates.DequeueAllSingleConsumer(&resCollection);
         return resCollection;
     }
 };


cloud/blockstore/libs/storage/partition_common/events_private.h
```diff
@@ -229,6 +233,70 @@ struct TEvPartitionCommonPrivate
...
+        TDuration RequestTime;
+        ui64 BSGroupOperationId = 0;
+
+        TWriteBlobCompleted() = default;
+
+        TWriteBlobCompleted(
+                const TPartialBlobId& blobId,
+                NKikimr::TStorageStatusFlags storageStatusFlags,
+                double approximateFreeSpaceShare,
+                TDuration requestTime)
```
komarevtsev-d: А поч тут BSGroupOperationId не заполняется?

komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition2/part2_actor.cpp
```diff
@@ -759,6 +759,15 @@ TDuration TPartitionActor::GetBlobStorageAsyncRequestTimeout() const
                : Config->GetBlobStorageAsyncRequestTimeoutHDD();
 }
 
+void TPartitionActor::HandleWriteBlob(
+    const TEvPartitionCommonPrivate::TEvWriteBlobRequest::TPtr& ev,
+    const NActors::TActorContext& ctx)
+{
+    Y_UNUSED(ev);
+    Y_UNUSED(ctx);
+    Y_ABORT("Unimplemented");
```
qkrorlqr, line 768: ?

cloud/blockstore/libs/storage/partition_common/writeblob_companion.h
```diff
@@ -0,0 +1,112 @@
...
+#include <cloud/storage/core/libs/actors/mortal_actor.h>
+
+#include <contrib/ydb/core/base/blobstorage.h>
+#include <contrib/ydb/library/actors/core/actor.h>
+
+namespace NCloud::NBlockStore::NStorage {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class IWriteBlobCompanionClient: public IMortalActor
```
qkrorlqr: Why do we need this client? Most of its methods look like `TPartitionChannelsState` methods.

And some methods don't belong to __Write__BlobAnythingActor at all - e.g. `ProcessIOQueue()` - it's related to `ReadBlob` logic as well

Overall I don't like these "companion clients" and think that we should extract most of these methods into `TPartitionChannelsState` (if we really want to extract them somewhere) and, probably, extract some of the others into other entities common to all ops, these methods are not confined to any specific operation

vladstepanyuk: renamed TWriteBlobCompanion to TIOCompaion and also moved read patch blob logic, moved ProcessIoQueue to TIOCompanion

i think we should  keep other functions in client
```
    virtual void UpdateWriteThroughput(
        const TInstant& now,
        const NKikimr::NMetrics::TChannel& channel,
        const NKikimr::NMetrics::TGroupId& group,
        ui64 value) = 0;
    virtual void UpdateReadThroughput(
        const TInstant& now,
        const NKikimr::NMetrics::TChannel& channel,
        const NKikimr::NMetrics::TGroupId& group,
        ui64 value,
        bool isOverlayDisk) = 0;
    virtual void UpdateNetworkStat(const TInstant& now, ui64 value) = 0;

    virtual void ScheduleYellowStateUpdate(
        const NActors::TActorContext& ctx) = 0;
    virtual void UpdateYellowState(const NActors::TActorContext& ctx) = 0;
    virtual void ReassignChannelsIfNeeded(
        const NActors::TActorContext& ctx) = 0;
    virtual void UpdateChannelPermissions(
        const NActors::TActorContext& ctx,
        ui32 channel,
        EChannelPermissions permissions) = 0;
```

I want to reuse this companion in my TFreshBlocksWriterActor, which should be separate actor with a separate from TPartitionActor mailbox. So some functions need to work in different ways.

`UpdateWriteThroughput`, `UpdateReadThroughput` and `UpdateNetworkStat` updates tablet executor resource metrics, TFreshBlocksWriterActor doesn't have direct access to the Executor, so he should send these updates to TPartitionActor. Therefore, different clients will have different logic.

Same with `ScheduleYellowStateUpdate`, `UpdateYellowState`, `ReassignChannelsIfNeeded`, `UpdateChannelPermissions`. These functions probably need different behavior because there are two separate actors and mailboxs also.

cloud/blockstore/libs/storage/partition2/part2_actor.cpp
vladstepanyuk: There are separate WriteBlob and PatchBlob events for part 2. For part 1, I moved these events to the partition_common so that they can be reused in my TFreshBlocksWriterActor. Later, if we need them, we can remove the separate WriteBlob/PatchBlob events from the part 2 tablet too, but for now it's unnecessary

cloud/blockstore/libs/storage/partition/model/group_downtimes.cpp
```diff
@@ -0,0 +1,23 @@
+#include "group_downtimes.h"
+
+namespace NCloud::NBlockStore::NStorage::NPartition {
+
```
komarevtsev-d, line 4: /// ?

cloud/blockstore/libs/storage/partition/model/group_downtimes.h
```diff
@@ -0,0 +1,29 @@
...
+#include <util/generic/hash.h>
+#include <util/system/mutex.h>
+#include <util/system/types.h>
+
+namespace NCloud::NBlockStore::NStorage::NPartition {
+
+// Thread safe
+class TGroupDowntimes {
+private:
+    TMutex Mutex;
```
komarevtsev-d: Поч не TAdaptiveLock?

cloud/blockstore/libs/storage/partition/io_companion_client.cpp
```diff
@@ -0,0 +1,40 @@
+#include "io_companion_client.h"
+
+#include "part_actor.h"
+
```
komarevtsev-d, line 4: ///

cloud/blockstore/libs/storage/partition_common/io_companion_patchblob.cpp
```diff
@@ -10,18 +12,15 @@
...
+namespace NCloud::NBlockStore::NStorage {
 
 using namespace NActors;
 
 using namespace NKikimr;
 
 LWTRACE_USING(BLOCKSTORE_STORAGE_PROVIDER);
 
-// TODO: get rid of code duplication with part_actor_writeblob.cpp
+// TODO: get rid of code duplication with writeblob_companion.cpp.
```
komarevtsev-d: io_companion

cloud/blockstore/libs/storage/core/config.cpp
```diff
@@ -665,6 +665,7 @@ NProto::TLinkedDiskFillBandwidth GetBandwidth(
     xxx(VolumeProxyPipeInactivityTimeout,     TDuration,   Minutes(1)         )\
     xxx(FreshChannelZeroRequestsEnabled,      bool,        false              )\
     xxx(AttachDetachPathRequestTimeout,       TDuration,   Seconds(5)         )\
+    xxx(ResourceMetricsUpdateInterval,        TDuration,   MSeconds(100)      )
```
qkrorlqr: )\

cloud/blockstore/libs/storage/partition/model/resource_metrics_updates_queue.h
```diff
@@ -0,0 +1,111 @@
...
+    TUpdateNetworkStat,
+    TUpdateStorageStat,
+    TUpdateCPUUsageStat>;
+
+// Thread safe
+class TResourceMetricsQueue
+{
+private:
+    TAdaptiveLock Lock;
+    TVector<TResourceMetricsUpdate> ResourceMetricsUpdates;
```
qkrorlqr: there's a very simple way to implement this API without any locks - just use `TLockFreeStack`

`PopAll()` -> `LFStack.DequeueAllSingleConsumer()`

cloud/blockstore/libs/storage/core/config.cpp
qkrorlqr: we don't need to do it this often, `Seconds(1)` will be enough

cloud/blockstore/libs/storage/partition/part_actor.cpp
```diff
@@ -1201,4 +1267,25 @@ NProto::TError VerifyBlockChecksum(
...
 
+////////////////////////////////////////////////////////////////////////////////
+
+void TPartitionActor::HandleUpdateResourceMetrics(
+    const TEvPartitionPrivate::TEvUpdateResourceMetrics::TPtr& ev,
+    const TActorContext& ctx)
+{
+    Y_UNUSED(ev);
+    auto updates = ResourceMetricsQueue->PopAll();
+    for (auto& update : updates) {
```
qkrorlqr: does clang-format autoformat this way? it should be `for (auto& update: updates) {`

cloud/blockstore/libs/storage/partition2/part2_actor_compaction.cpp
```diff
@@ -335,15 +337,18 @@ void TCompactionActor::WriteBlobs(const TActorContext& ctx)
...
     for (auto& req: WriteRequests) {
         if (!IsDeletionMarker(req.BlobId)) {
-            auto request = std::make_unique<TEvPartitionPrivate::TEvWriteBlobRequest>(
+            auto request = std::make_unique<
+                TEvPartitionCommonPrivate::TEvWriteBlobRequest>(
                 req.BlobId,
-                BuildBlobContent(req));
+                BuildBlobContent(req),
+                0,        // blockSizeForChecksums
+                false);   // async
```
qkrorlqr: why not async? reads/writes issued by compaction should be async

cloud/blockstore/libs/storage/partition2/part2_actor.cpp
qkrorlqr: why later, why not now?

Issue #5174, opened 2026-02-05
[Filestore] TCompletionQueue with single TMutex creates contention when using filestore with multi-queue in large number of threads


budevg:
<img width="3044" height="2580" alt="Image" src="https://github.com/user-attachments/assets/5d8822b0-7c02-4f0f-b9bc-6914eebc83c2" />

budevg:
Adding flame graph with shortcut code that immideatly returns on read/write io
```diff
diff --git a/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp b/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
index 666341b928..1708048ce5 100644
--- a/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
@@ -249,6 +249,9 @@ void TFileSystem::ReadLocal(
     off_t offset,
     fuse_file_info* fi)
 {
+
+    ReplyBuf(*callContext, {}, req, nullptr, size);
+    return;
     callContext->Unaligned = !IsAligned(offset, Config->GetBlockSize())
         || !IsAligned(size, Config->GetBlockSize());
 
@@ -558,6 +561,8 @@ void TFileSystem::WriteBufLocal(
         << " offset:" << offset
         << " size:" << size);
 
+    ReplyWrite(*callContext, {}, req, size);
+    return;
     auto request = StartRequest<NProto::TWriteDataLocalRequest>(ino);
     request->SetHandle(fi->fh);
     request->SetOffset(offset);
diff --git a/cloud/filestore/libs/vfs_fuse/loop.cpp b/cloud/filestore/libs/vfs_fuse/loop.cpp
index ab8e82b060..8503c8214f 100644
--- a/cloud/filestore/libs/vfs_fuse/loop.cpp
+++ b/cloud/filestore/libs/vfs_fuse/loop.cpp
@@ -661,12 +661,13 @@ public:
             ISessionPtr session)
         : Config(std::move(config))
         , Logging(std::move(logging))
-        , StatsRegistry(std::move(statsRegistry))
+        , StatsRegistry(std::move(CreateRequestStatsRegistryStub()))
         , Scheduler(std::move(scheduler))
         , Timer(std::move(timer))
         , ProfileLog(std::move(profileLog))
         , Session(std::move(session))
     {
+        Y_UNUSED(statsRegistry);
[...]

PR #5175, opened 2026-02-05
[Filestore] issue-5174: reduce completion queue contention

budevg:
#5174

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/loop.cpp
+++ b/cloud/filestore/libs/vfs_fuse/loop.cpp
@@ -106,9 +106,9 @@ class TCompletionQueue final
     , public IIncompleteRequestProvider
 {
 private:
-    // Shard of in-flight requests to reduce contention and track per shard
-    // counters.
-    struct TShard
+    // Bucket of in-flight requests to reduce contention and track per request
+    // bucket counters.
+    struct Y_CACHE_ALIGNED TRequestBucket
     {
         TAdaptiveLock Lock;
         THashMap<fuse_req_t, TCallContextPtr> Requests;
@@ -120,8 +120,8 @@ private:
     const IRequestStatsPtr RequestStats;
     TLog Log;
     const NProto::EStorageMediaKind RequestMediaKind;
-    const ui32 ShardCount;
-    TVector<TShard> Shards;
+    const ui32 RequestBucketCount;
+    TVector<TRequestBucket> RequestBuckets;
 
     enum fuse_cancelation_code CancelCode{};
     NAtomic::TBool ShouldStop = false;
@@ -134,15 +134,15 @@ public:
             IRequestStatsPtr stats,
             TLog& log,
             NProto::EStorageMediaKind requestMediaKind,
-            ui32 shardCount)
+            ui32 requestBucketCount)
         : FileSystemId(std::move(fileSystemId))
         , RequestStats(std::move(stats))
         , Log(log)
         , RequestMediaKind(requestMediaKind)
-        , ShardCount(shardCount)
-        , Shards(shardCount)
+        , RequestBucketCount(requestBucketCount)
+        , RequestBuckets(requestBucketCount)
     {
-        Y_ABORT_UNLESS(IsPowerOf2(ShardCount));
+        Y_ABORT_UNLESS(IsPowerOf2(RequestBucketCount));
     }
 
 public:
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -254,6 +299,19 @@ static ui32 CalcFrontendVhostQueueCount(const TVFSConfig& vfsConfig)
...
 
+static ui32 CalcCompletionQueueShardCount(const TVFSConfig& vfsConfig)
+{
+    constexpr ui32 MinShards = 16;
+    constexpr ui32 MaxShards = 256;
+
+    const ui32 queues = CalcFrontendVhostQueueCount(vfsConfig);
+
+    // Use ~2x queues and round to power of two for cheap shard indexing,
+    // then clamp to keep shard count in a reasonable range.
```
neihar: I would add that round to power of 2 it is mandatory for correct distribution

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -196,25 +194,28 @@ class TCompletionQueue final
...
+                for (auto& request: shard.Requests) {
+                    request.second->CancellationCode = CancelCode;
+                    request.second->Cancelled = true;
+                }
+                requestsSize += shard.Requests.size();
             }
         }
 
+        for (const auto& shard: Shards) {
+            completingCount += shard.Completing.Val();
```
neihar: is it safe to count without mutex? ( I see that before it was under mutex)

neihar: this is safe.

neihar (dismissed):

cloud/filestore/libs/vfs_fuse/loop.cpp
budevg: yes `Completing` is Atomic

neihar (dismissed):

qkrorlqr (dismissed):

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -105,17 +106,25 @@ class TCompletionQueue final
...
+        i64 Inflight = 0;
+        TAtomicCounter Completing = {0};
+    };
+
     const TString FileSystemId;
     const IRequestStatsPtr RequestStats;
     TLog Log;
     const NProto::EStorageMediaKind RequestMediaKind;
+    const ui32 ShardCount;
+    TVector<TShard> Shards;
```
debnatkh: Isn't there supposed to be some kind of struct padding to avoid false sharing?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -105,17 +106,25 @@ class TCompletionQueue final
     , public IIncompleteRequestProvider
 {
 private:
+    // Shard of in-flight requests to reduce contention and track per shard
+    // counters.
+    struct TShard
```
SvartMetal: Let’s not use the term “shard” here to avoid confusion with Filestore shards?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -124,68 +133,57 @@ class TCompletionQueue final
...
+        TGuard g{shard.Lock};
 
+        // Re-check after acquiring the shard lock to avoid enqueueing after
+        // stop is requested while we were waiting.
         if (ShouldStop) {
             return CancelCode;
         }
 
-        Requests[req] = std::move(context);
+        shard.Requests[req] = std::move(context);
```
SvartMetal: Can we extract all the related logic into a reusable primitive and add proper multi-threaded tests?

cloud/filestore/libs/vfs_fuse/loop.cpp
SvartMetal: Maybe TBucket?

cloud/filestore/libs/vfs_fuse/loop.cpp
budevg: In theory you are right, but I didn't see any performance gain from using cache line alignment here. But I will add it anyway since it's pretty cheap.

cloud/filestore/libs/vfs_fuse/loop.cpp
budevg: I'd prefer to keep this inline for now: it's a small lock scoped hot path, and extracting it into a separate primitive would mostly be code movement without clear correctness/readability gain. For multi threaded coverage, this flow is already exercised by our end to end tests (qemu-kikimr-mq-test, qemu-local-noserver-mq-test)

SvartMetal (approved):

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -294,15 +286,15 @@ void TFileSystem::ReplyAttr(
...
-        req);
-
     // notifying CompletionQueue about request completion to decrement inflight
     // request counter and unblock the stopping procedure
-    CompletionQueue->Complete(req, [&] (fuse_req_t) { return 0; });
+    CompletionQueue->Complete(
+        req,
+        [&](fuse_req_t request)
+        {
+            NFuse::CancelRequest(Log, *RequestStats, *callContext, request);
```
drbasic, line 295: а вызов этой лямбды произодет в какой момент и на каком потоке? паременные, захваченные по & живые будут еще?

budevg: вызов лямбды синхронно вызывается `int ret = cb(req);`
```c++
    int Complete(fuse_req_t req, TCompletionCallback cb) noexcept override
    {
        auto& requestBucket = RequestBuckets[GetRequestBucketIndex(req)];
        with_lock (requestBucket.Lock) {
            if (!requestBucket.Requests.erase(req)) {
                return 0;
            }
            --requestBucket.Inflight;
            requestBucket.Completing.Inc();
        }

        int ret = cb(req);
```

PR #5182, opened 2026-02-06
issue 1751: [Filestore] WriteBackCache refactoring - extract PersistentStorage

e673:
Extract the logic for storing WriteData requests to separate files/classes.

Part of https://github.com/ydb-platform/nbs/pull/5064

Changes:
* Stats are reported now from inside the wrapper.
* WriteBackCache is now unaware of the internal representation of the storage.
* An abstraction is used for better test isolation (simulation of storage full state).

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
@@ -1,22 +1,35 @@
 #pragma once
 
 #include <cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h>
-#include <cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_stats.h>
+
+#include <util/generic/hash.h>
 
 namespace NCloud::NFileStore::NFuse::NWriteBackCache {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct ITestPersistentStorage: public IPersistentStorage
+class TTestStorage: public IPersistentStorage
 {
-    virtual void SetCapacity(size_t capacity) = 0;
+private:
+    const IPersistentStorageStatsPtr Stats;
+    THashMap<const void*, std::unique_ptr<TString>> Data;
+    size_t Capacity = 0;
+
+public:
+    explicit TTestStorage(IPersistentStorageStatsPtr stats);
+
+    bool Empty() const override;
+    void Visit(const TVisitor& visitor) override;
+    ui64 GetMaxSupportedAllocationByteCount() const override;
+    TResultOrError<char*> Alloc(size_t size) override;
+    bool Commit() override;
+    void Free(const void* ptr) override;
+    TPersistentStorageStats GetStats() const override;
+
+    void SetCapacity(size_t capacity);
+
+private:
+    void UpdateStats();
 };
-
-using ITestPersistentStoragePtr = std::shared_ptr<ITestPersistentStorage>;
-
-////////////////////////////////////////////////////////////////////////////////
-
-ITestPersistentStoragePtr CreateTestPersistentStorage(
-    IWriteBackCacheStatsPtr stats);
 
 }   // namespace NCloud::NFileStore::NFuse::NWriteBackCache
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.cpp
@@ -1,113 +1,79 @@
 #include "test_persistent_storage.h"
 
-#include <util/generic/hash.h>
-#include <util/generic/intrlist.h>
-
 namespace NCloud::NFileStore::NFuse::NWriteBackCache {
 
-namespace {
-
 ////////////////////////////////////////////////////////////////////////////////
 
-class TTestStorage: public ITestPersistentStorage
+TTestStorage::TTestStorage(IPersistentStorageStatsPtr stats)
+    : Stats(std::move(stats))
+{}
+
+bool TTestStorage::Empty() const
 {
-private:
-    struct TChunk: TIntrusiveListItem<TChunk>
-    {
-        TString Data;
-    };
+    return Data.empty();
+}
 
-    const IWriteBackCacheStatsPtr Stats;
-    THashMap<const void*, std::unique_ptr<TChunk>> ChunkMap;
-    TIntrusiveList<TChunk> ChunkList;
-    size_t Capacity = 0;
-
-public:
-    explicit TTestStorage(IWriteBackCacheStatsPtr stats)
-        : Stats(std::move(stats))
-    {}
-
-    bool Empty() const override
-    {
-        return ChunkMap.empty();
-    }
-
-    void Visit(const TVisitor& visitor) override
-    {
-        for (const auto& it: ChunkList) {
-            visitor(it.Data);
-        }
-    }
-
-    ui64 GetMaxSupportedAllocationByteCount() const override
-    {
-        return Max<ui64>();
-    }
-
-    TResultOrError<char*> Alloc(size_t size) override
-    {
-        if (Capacity > 0 && ChunkMap.size() >= Capacity) {
-            return nullptr;
-        }
-
-        auto chunk = std::make_unique<TChunk>();
-        chunk->Data = TString::Uninitialized(size);
-
-        char* res = chunk->Data.begin();
-
-        ChunkList.PushBack(chunk.get());
-        ChunkMap[res] = std::move(chunk);
-
-        UpdateStats();
-
-        return res;
-    }
-
-    bool Commit() override
-    {
-        return true;
-    }
-
-    void Free(const void* ptr) override
-    {
-        auto it = ChunkMap.find(ptr);
-        Y_ENSURE(it != ChunkMap.end(), "Double free detected");
-
-        ChunkList.Remove(it->second.get());
-        ChunkMap.erase(it);
-
-        UpdateStats();
-    }
-
-    void SetCapacity(size_t capacity) override
-    {
-        Capacity = capacity;
-        UpdateStats();
-    }
-
-    TPersistentStorageStats GetStats() const override
-    {
-        return {
-            .RawCapacityByteCount = Capacity,
-            .RawUsedByteCount = ChunkMap.size(),
-            .EntryCount = ChunkMap.size(),
-            .IsCorrupted = false};
-    }
-
-    void UpdateStats()
-    {
-        Stats->UpdatePersistentStorageStats(GetStats());
-    }
-};
-
-}   // namespace
-
-////////////////////////////////////////////////////////////////////////////////
-
-ITestPersistentStoragePtr CreateTestPersistentStorage(
-    IWriteBackCacheStatsPtr stats)
+void TTestStorage::Visit(const TVisitor& visitor)
 {
-    return std::make_shared<TTestStorage>(std::move(stats));
+    for (const auto& it: Data) {
+        visitor(*it.second);
+    }
+}
+
+ui64 TTestStorage::GetMaxSupportedAllocationByteCount() const
+{
+    return Max<ui64>();
+}
+
+TResultOrError<char*> TTestStorage::Alloc(size_t size)
+{
+    if (Capacity > 0 && Data.size() >= Capacity) {
+        return nullptr;
+    }
+
+    auto item = std::make_unique<TString>(size, 0);
+    char* res = item->begin();
+
+    Data[res] = std::move(item);
+
+    UpdateStats();
+
+    return res;
+}
+
+bool TTestStorage::Commit()
+{
+    return true;
+}
+
+void TTestStorage::Free(const void* ptr)
+{
+    auto it = Data.find(ptr);
+    Y_ENSURE(it != Data.end(), "Double free detected");
+
+    Data.erase(it);
+
+    UpdateStats();
+}
+
+void TTestStorage::SetCapacity(size_t capacity)
+{
+    Capacity = capacity;
+    UpdateStats();
+}
+
+TPersistentStorageStats TTestStorage::GetStats() const
+{
+    return {
+        .RawCapacityByteCount = Capacity,
+        .RawUsedByteCount = Data.size(),
+        .EntryCount = Data.size(),
+        .IsCorrupted = false};
+}
+
+void TTestStorage::UpdateStats()
+{
+    Stats->UpdatePersistentStorageStats(GetStats());
 }
 
 }   // namespace NCloud::NFileStore::NFuse::NWriteBackCache


cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Allocates a buffer of the given size.
+     *
+     * On successful allocation, calls the writer, which should fill the buffer,
+     * and returns a pointer to the buffer in persistent storage.
+     * The returned pointer may differ from the pointer passed to the writer.
+     *
+     * Returns nullptr if there is not enough free space in the storage.
+     * Throws an exception if the allocation size exceeds
```
neihar: why do we need in some cases TResultOrError and in some Exception? Probably it worth to consider to make this class exception free.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
```diff
@@ -0,0 +1,165 @@
...
+namespace NCloud::NFileStore::NFuse::NWriteBackCache {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TFileRingBufferStorage: public IPersistentStorage
+{
+private:
+    IWriteBackCacheStatsPtr Stats;
```
neihar: the file name is 'persistent_storage', but actually it is just implementation of PersistentStorage interface via filering buffer and at the same time it depends explicitly on "WriteBackCacheStats". I strongly suggest either rename file and (or) class to reflect that this storage writebackcache specific or make it independent from writebackcache entities via inheritance or composition

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
```diff
@@ -0,0 +1,165 @@
...
+                }
+            });
+
+        UpdateStats();
+    }
+
+    ui64 GetMaxSupportedAllocationByteCount() const override
+    {
+        // Full capacity minus entry header
+        return Storage.GetRawCapacity() - 8;
```
neihar: looks weird to me, probably we need to work around api in fileringbuffer, for example in the method GetMaxAllocationBytesCount  we do similar calculations, but it is hidden in the fileringbuffer internals. I am against this -8 here.

cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
SvartMetal, line 1: test_persistent_storage.h?

by analogy with ITestPersistentStorage

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
```diff
@@ -0,0 +1,165 @@
...
+                    << Storage.GetMaxAllocationBytesCount());
+
+            return nullptr;
+        }
+
+        // TFileRingBuffer does not support in-place allocation yet
+        auto buffer = TString::Uninitialized(size);
+        writer(buffer.begin(), size);
+
+        Y_ABORT_UNLESS(Storage.PushBack(buffer));
```
neihar: In case if underlying storage has some kind of errors, for example corrupted, am I correct that after restart we will not able to init this class anyway?

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
neihar: Or, if we go exception way, I don't see a problem to throw in ctor

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
```diff
@@ -0,0 +1,72 @@
...
+using IPersistentStoragePtr = std::shared_ptr<IPersistentStorage>;
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TPersistentStorageConfig
+{
+    TString FilePath;
+    ui64 DataCapacity = 0;
+    ui64 MetadataCapacity = 0;
+    bool EnableChecksumCalculation = false;
```
neihar: From my perspective this is unneeded variable, we should calculate always, but validation should be optional

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
e673: Inability to store WriteData request into the storage because it exceeds the storage capacity will result in complete stuck. Probably it is better to return an enum.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
e673: Yes, the internal structure should be checked at startup.

cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
e673: From my point of view, it is harder to navigate when the files in the directory `test` have the same repeated prefix `test_`. I've renamed the files - you may check if it is better now.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
e673: I've removed the check - the size is already validated in `TWriteBackCache::WriteData`.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
e673: I've made it independent from `IWriteBackCacheStatsPtr`.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
e673: I've put this code to `TFileRingBuffer`.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
e673: I removed it for now because it is not used anyways.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+        if (Storage.GetAvailableByteCount() < size) {
+            return nullptr;
+        }
+
+        // TFileRingBuffer does not support in-place allocation yet
+        auto buffer = TString::Uninitialized(size);
+        writer(buffer.begin(), size);
+
+        Y_ABORT_UNLESS(Storage.PushBack(buffer));
```
qkrorlqr: let's not do things like `Y_ABORT_UNLESS(SomeImportantCall())` - someone might replace it with `Y_DEBUG_ABORT_UNLESS(SomeImporantCall())` and `SomeImportantCall()` will not be called in builds where `NDEBUG` is defined

and also - let's avoid `Y_ABORT_UNLESS` usage wherever we can because it gives zero info about the exact entity id (fs id, disk id, tablet id, etc) which caused the crash

```
const bool pushed = Storage.PushBack(buffer);
STORAGE_VERIFY(pushed, ... context ...);
```

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
```diff
@@ -0,0 +1,89 @@
...
+     * Gets the maximum possible buffer size that can be allocated.
+     * Alloc is guaranteed to succeed for any size <=
+     * MaxSupportedAllocationByteCount when the storage is empty.
+     */
+    virtual ui64 GetMaxSupportedAllocationByteCount() const = 0;
+
+    /**
+     * Allocates a buffer of the given size.
+     *
+     * On successful allocation, calls the writer, which should fill the buffer,
```
qkrorlqr: Why do we need this writer? Why can't the logic that calls `Alloc()` just write to the allocated memory by itself, after Alloc() returns a non-zero pointer? There're no atomicity guarantees here anyway. The `Storage.PushBack(buffer)` call is not guaranteed to write everything - we can crash in the middle of this operation.

e673: > Why can't the logic that calls Alloc() just write to the allocated memory by itself, after Alloc() returns a non-zero pointer?

This is due to checksum calculation. `TFileRingBuffer` is designed to store immutable data fragments:
1. The data is written.
2. The checksum is calculated.
3. The metadata is updated.
4. The data pointer is updated.

If we crash before step 4, then no corruption will happen.
And the step 4 is just an update of aligned ui64 value - it is likely to be atomic.

Also, `TFileRingBuffer` does not support in-place allocation yet but I plan to implement this in the similar way.

neihar: I don't see connection between Writer and Checksum. I can't disagree with point that alloc/free looks strange and more suitable solution would be preferable. But probably not in that PR

cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h>
+#include <cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_stats.h>
+
+namespace NCloud::NFileStore::NFuse::NWriteBackCache {
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct ITestPersistentStorage: public IPersistentStorage
```
neihar: why do we need this?

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
e673: Done.
I've added the filename (it is enough to investigate the case).

Upd: after implementing in-place allocation, this check is no more needed.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
e673: Ok, I've implemented in-place allocation in `TFileRingBuffer`.

cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
e673: I use it for `SetCapacity` method to test `TWriteBackCacheState` in the future PR.
I do not want to depend on `TFileRingBuffer` internal implementation and want the tests to be more explicit.

https://github.com/ydb-platform/nbs/blob/08bc76c5e45edeb5a41695bbdfe8295d89e98f80/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state_ut.cpp#L214-L286

e673: This is another reason why I use an abstraction for a storage.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -532,18 +553,36 @@ class TFileRingBuffer::TImpl
...
-        Header()->WritePos = writePos + sz;
-        Header()->LastEntrySize = sz;
+        char* ptr = Data.GetEntryData(CurrentAllocation);
+        Y_ABORT_UNLESS(ptr != nullptr);
+        return ptr;
+    }
+
+    void Commit()
+    {
+        Y_ENSURE(CurrentAllocation != nullptr, "Nothing to commit");
```
qkrorlqr: Let's try to avoid `Y_ENSURE` and exceptions - especially in common libs. Let's prefer error codes - with error codes it's much harder to forget to process the error. We use exceptions locally in some parts of the codebase - e.g. blockstore/libs/storage/disk_registry - but generally we avoid them and even in those places where we have them they are not exposed in the public interface - they are thrown and caught internally in the component where they're used.

cloud/storage/core/libs/common/file_ring_buffer.cpp
```diff
@@ -532,18 +553,36 @@ class TFileRingBuffer::TImpl
...
+
+        char* ptr = Data.GetEntryData(CurrentAllocation);
+        Y_ABORT_UNLESS(ptr != nullptr);
+
+        CurrentAllocation->Checksum = Crc32c(ptr, CurrentAllocation->DataSize);
```
qkrorlqr, line 583: stores to `Checksum`, `WritePos` and `LastEntrySize` might be reordered - will it be a problem for recovery upon restart?

on x86_64 a compiler barrier should be enough but on arm these stores might be reordered in runtime

probably we need `std::atomic_thread_fence`s here

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I changed the result of `Alloc` to `TResultOrError<char*>`.
It returns an error with an explicit verbal message that the caller may use.

cloud/storage/core/libs/common/file_ring_buffer.cpp
e673: I've added the barrier.

Also, there is no way to ensure that `WritePos` and `LastEntrySize` are updated atomically - this may lead to a corruption state after crash even if the buffer is not corrupted.

So I've removed validation for `LastEntrySize` - it will be restored at buffer validation.

cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
neihar: why we can't use just TestPersistentStorage without additional test interface?

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
```diff
@@ -0,0 +1,151 @@
...
+        return Storage.Empty();
+    }
+
+    void Visit(const TVisitor& visitor) override
+    {
+        Storage.Visit(
+            [this, &visitor](ui32 checksum, TStringBuf entry)
+            {
+                Y_UNUSED(checksum);
+                if (!DeletedEntries.contains(entry.data())) {
```
neihar, line 67: do we call it only on start? can we have deleted entries here?

neihar: As I understand it is just for "generic" api, that theoretically can be used.

cloud/filestore/libs/vfs_fuse/write_back_cache/test/test_persistent_storage.h
e673: Removed interface

PR #5187, opened 2026-02-07
issue-2674: RenameNode - support for the case when NewChildRef points to a directory but both ChildRef and NewNodeRef are managed by the same shard

qkrorlqr:
### Notes
Here https://github.com/ydb-platform/nbs/pull/5059 and here https://github.com/ydb-platform/nbs/pull/5102 I implemented node type checking and dst emptiness checking for cross-shard `RenameNode` ops - i.e. the ones where src and dst `NodeRefs` are managed by different shards. But the same logic needs to be present in local `RenameNode` logic when `DirectoryCreationInShardsEnabled` is true because even if both src and dst `NodeRefs` are local Node attrs are still not local. Implementing the same checks for local `RenameNode` ops in this PR.

Added rename op calls with directories as dst nodes to the `cloud/filestore/tests/client_sharded_dir/test.py` test.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode.cpp
@@ -430,7 +430,7 @@ bool TIndexTabletActor::PrepareTx_RenameNode(
                     args.SourceNodeAttr.GetType() == NProto::E_DIRECTORY_NODE;
             } else {
                 args.SecondPassRequired =
-                    Config->GetDirectoryCreationInShardsEnabled();
+                    GetFileSystem().GetDirectoryCreationInShardsEnabled();
             }
         } else {
             childIsDir =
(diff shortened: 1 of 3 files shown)


agalibin (dismissed):

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode.cpp
```diff
@@ -358,31 +419,79 @@ bool TIndexTabletActor::PrepareTx_RenameNode(
...
+        bool newChildIsEmpty = false;
+        ui64 newChildNodeId = InvalidNodeId;
+
+        if (args.ChildRef->IsExternal()) {
+            if (args.IsSecondPass) {
+                childIsDir =
+                    args.SourceNodeAttr.GetType() == NProto::E_DIRECTORY_NODE;
+            } else {
+                args.SecondPassRequired =
+                    Config->GetDirectoryCreationInShardsEnabled();
```
debnatkh: Are you sure we should access Config here? DirectoryCreationInShardsEnabled is a filesystem-level option. Isn't it better to use `message TFileSystem` field?

qkrorlqr: fixed

agalibin (approved):

PR #5188, opened 2026-02-08
[NBS] Local NVMe service: reset NVMe to single namespace during release operation

sharpeye:
Reset NVMe device to single namespace during ReleaseNVMeDevice operation

Detach and delete all existing namespaces, then create a single namespace spanning the entire device capacity.

#5250

### DIFF
--- a/cloud/blockstore/libs/nvme/nvme_linux.cpp
+++ b/cloud/blockstore/libs/nvme/nvme_linux.cpp
@@ -3,6 +3,7 @@
 #include <cloud/storage/core/libs/common/task_queue.h>
 #include <cloud/storage/core/libs/common/thread_pool.h>
 
+#include <util/generic/vector.h>
 #include <util/generic/yexception.h>
 #include <util/string/builder.h>
 #include <util/system/file.h>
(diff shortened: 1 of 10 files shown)


komarevtsev-d (approved):

cloud/blockstore/libs/nvme/nvme_linux.cpp
```diff
@@ -23,6 +25,159 @@ namespace {
...
+        << " cdw10: " << cmd.cdw10 << "): " << strerror(err);
+}
+
+auto ListAllocatedNamespaces(TFileHandle& file) -> TVector<ui32>
+{
+    ui32 namespaceIDs[1024]{};
+
+    nvme_admin_cmd cmd = {
+        .opcode = NVME_OPC_IDENTIFY,
+        .addr = std::bit_cast<ui64>(&namespaceIDs[0]),
```
dvrazumov, line 47:
```
std::bit_cast<ui64>(&namespaceIDs[0]),
```
->
```
reinterpret_cast<std::uintptr_t>(namespaceIDs),
```
?

Issue #5189, opened 2026-02-08
[Filestore] struct fuse_session with single mutex creates contention when using filestore with multi-queue in large number of threads


budevg:
<img width="1440" height="1220" alt="Image" src="https://github.com/user-attachments/assets/decb4233-bdd9-4c67-8913-02b6092b7a41" />

PR #5190, opened 2026-02-08
[Filestore] issue-5189: reduce fuse_session contention

budevg:
#5189

### DIFF
--- a/cloud/contrib/virtiofsd/fuse_lowlevel.c
+++ b/cloud/contrib/virtiofsd/fuse_lowlevel.c
@@ -124,6 +124,13 @@ static int req_queue_index_from_chan(struct fuse_session *se,
     int queue_index = virtio_queue_index((struct fuse_chan *)ch);
 
     if (queue_index < 0 || queue_index >= se->num_backend_queues) {
+        fuse_log(
+            FUSE_LOG_ERR,
+            "fuse: req queue index out of bounds in %s: index=%d, queues=%d, "
+            "using 0\n",
+            __func__,
+            queue_index,
+            se->num_backend_queues);
         return 0;
     }
 
@@ -143,6 +150,13 @@ void fuse_free_req(fuse_req_t req)
     list_kind = req->list_kind;
     queue_index = req->req_queue_index;
     if (queue_index < 0 || queue_index >= se->num_backend_queues) {
+        fuse_log(
+            FUSE_LOG_ERR,
+            "fuse: req queue index out of bounds in %s: index=%d, queues=%d, "
+            "using 0\n",
+            __func__,
+            queue_index,
+            se->num_backend_queues);
         queue_index = 0;
     }
     req->ch = NULL;
@@ -1702,6 +1716,13 @@ static int find_interrupted(struct fuse_session *se, struct fuse_req *req)
     int i;
 
     if (queue_index < 0 || queue_index >= se->num_backend_queues) {
+        fuse_log(
+            FUSE_LOG_ERR,
+            "fuse: req queue index out of bounds in %s: index=%d, queues=%d, "
+            "using 0\n",
+            __func__,
+            queue_index,
+            se->num_backend_queues);
         queue_index = 0;
     }
 


cloud/contrib/virtiofsd/fuse_lowlevel.c
```diff
@@ -118,18 +118,60 @@ static void destroy_req(fuse_req_t req)
     free(req);
 }
 
+static int req_queue_index_from_chan(struct fuse_session *se,
+                                     const struct fuse_chan *ch)
+{
+    int queue_index = virtio_queue_index((struct fuse_chan *)ch);
+
+    if (queue_index < 0 || queue_index >= se->num_backend_queues) {
```
qkrorlqr, line 126: why can `queue_index` be out of bounds?

cloud/contrib/virtiofsd/fuse_lowlevel.c
```diff
@@ -118,18 +118,60 @@ static void destroy_req(fuse_req_t req)
...
+        list_del_req(req);
+        req->list_kind = FUSE_REQ_LIST_NONE;
+        if (se->interrupts.next == &se->interrupts) {
+            atomic_store_explicit(&se->has_pending_interrupts, false,
+                                  memory_order_release);
+        }
+        ctr = --req->ctr;
+        pthread_mutex_unlock(&se->interrupts_lock);
+        break;
+    default:
```
qkrorlqr: what's this `default` for? `FUSE_REQ_LIST_NONE`? if so, in which cases can `list_kind` be `FUSE_REQ_LIST_NONE`?

and let's use exhaustive switches by the way - without default sections (to let the compiler warn us if we forget to process one of the enum values)

cloud/contrib/virtiofsd/fuse_lowlevel.c
```diff
@@ -1609,45 +1653,81 @@ static void do_setlkw(fuse_req_t req, fuse_ino_t nodeid,
...
             void *data;
+            int ctr;
 
             curr->ctr++;
-            pthread_mutex_unlock(&se->lock);
+            pthread_mutex_unlock(&se->req_queue_locks[queue_index]);
 
-            /* Ugh, ugly locking */
             pthread_mutex_lock(&curr->lock);
-            pthread_mutex_lock(&se->lock);
```
qkrorlqr, line 1626: why do we not need queue lock here anymore?

cloud/contrib/virtiofsd/fuse_lowlevel.c
```diff
@@ -1670,35 +1750,60 @@ static void do_interrupt(fuse_req_t req, fuse_ino_t nodeid,
...
+    pthread_mutex_lock(&se->interrupts_lock);
     for (curr = se->interrupts.next; curr != &se->interrupts;
          curr = curr->next) {
         if (curr->u.i.unique == req->unique) {
+            pthread_mutex_lock(&req->lock);
             req->interrupted = 1;
+            pthread_mutex_unlock(&req->lock);
             list_del_req(curr);
-            free(curr);
+            curr->list_kind = FUSE_REQ_LIST_NONE;
```
qkrorlqr: why do we need this? `curr` is going to be destroyed on line 1788

cloud/contrib/virtiofsd/fuse_lowlevel.c
```diff
@@ -2294,24 +2399,25 @@ const struct fuse_ctx *fuse_req_ctx(fuse_req_t req)
...
+    int interrupted;
+
     pthread_mutex_lock(&req->lock);
-    pthread_mutex_lock(&req->se->lock);
     req->u.ni.func = func;
     req->u.ni.data = data;
-    pthread_mutex_unlock(&req->se->lock);
-    if (req->interrupted && func) {
+    interrupted = req->interrupted;
+    pthread_mutex_unlock(&req->lock);
```
qkrorlqr: are you sure that it's enough to keep only `req->interrupted` in the scope of the lock and that `func()` call doesn't rely on the fact that `req->lock` is held?

cloud/contrib/virtiofsd/fuse_lowlevel.c
```diff
@@ -2294,24 +2399,25 @@ const struct fuse_ctx *fuse_req_ctx(fuse_req_t req)
 void fuse_req_interrupt_func(fuse_req_t req, fuse_interrupt_func_t func,
                              void *data)
 {
+    int interrupted;
+
     pthread_mutex_lock(&req->lock);
-    pthread_mutex_lock(&req->se->lock);
```
qkrorlqr, line 2298: why do we not need queue lock here anymore?

cloud/contrib/virtiofsd/fuse_lowlevel.c
budevg: it's just defensive guard. Negative value is not possible since `virtio_queue_index` returns it only in case `ch, ch->virtio_queue` not set and we do set it during init. Upper bound is also not possible since we initialize the index for each queue ... do you think it's better to put assert here ? I prefer guard which will not lead to abort in production

cloud/contrib/virtiofsd/fuse_lowlevel.c
budevg: yes it's the FUSE_REQ_LIST_NONE (I've changed to use it explicitly).
It can be used in case of error case in the `fuse_session_process_buf_int`. For example in the
```c
    if (se->deny_others && in->uid != se->owner && in->uid != 0 &&
        in->opcode != FUSE_INIT && in->opcode != FUSE_READ &&
        in->opcode != FUSE_WRITE && in->opcode != FUSE_FSYNC &&
        in->opcode != FUSE_RELEASE && in->opcode != FUSE_READDIR &&
        in->opcode != FUSE_FSYNCDIR && in->opcode != FUSE_RELEASEDIR &&
        in->opcode != FUSE_NOTIFY_REPLY && in->opcode != FUSE_READDIRPLUS) {
        goto reply_err;
    }
```
fuse_reply_none will be invoked and it will call to `fuse_free_req(req);` with kind set to `FUSE_REQ_LIST_NONE`

cloud/contrib/virtiofsd/fuse_lowlevel.c
budevg: we increased  `curr->ctr++;` under queue lock 
```c
        if (curr->unique == req->u.i.unique) {
            fuse_interrupt_func_t func;
            void *data;
            int ctr;

            curr->ctr++;
```
so the curr request can't be dropped anymore. Then we can invoke it's callback and release it if needed

cloud/contrib/virtiofsd/fuse_lowlevel.c
budevg: Good catch! it's actually regression from previous behavior for some reason I thought it would be safe but it's not. I've reverted to previous behavior with lock held during `func()`

cloud/contrib/virtiofsd/fuse_lowlevel.c
budevg: previously
```
req->u.ni.func = func;
req->u.ni.data = data;
```
was protected by session lock but now we protect it with `req->lock` across the code

cloud/contrib/virtiofsd/fuse_lowlevel.c
qkrorlqr: let's at least put some logging here in case it's out of bounds

qkrorlqr (approved):

yegorskii (approved):

PR #5196, opened 2026-02-09
[Filestore] always use GetNodeAttrBatch while performing listing in a sharded filesystem

debnatkh:


cloud/filestore/libs/storage/service/service_ut_sharding.cpp
```diff
@@ -1530,11 +1530,8 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
                             auto* msg =
                                 event
                                     ->Get<TEvService::TEvGetNodeAttrResponse>();
-                            if (msg->Record.GetError().GetCode() == E_FS_NOENT)
```
qkrorlqr, line 1533: why was this changed?

debnatkh: I made a bug in this test in the past: https://github.com/ydb-platform/nbs/commit/b7123d3b819e439764f29a46f0e3fc787f87c175

I wanted to simulate a behaviour when we entered the StateCheck and were observing E_REJECTED from the leader. But accidentally returned and propagated E_REJECTED from the shard to StateWork

After getting rid of individual GetNodeAttr requests, this filter stopped matching any message at all, and this test started failing

PR #5207, opened 2026-02-10
List nodes max count

zveznicht:
Add ability to set max elements count in ListNodes request


cloud/filestore/libs/storage/tablet/tablet_database.cpp
```diff
@@ -2243,12 +2254,16 @@ bool TIndexTabletDatabaseProxy::ReadNodeRefs(
...
     bool noAutoPrecharge,
-    NProto::EListNodesSizeMode sizeMode)
+    NProto::EListNodesSizeMode sizeMode,
+    ui32 maxRows)
 {
     ui32 skipped = 0;
     if (!skippedRefs) {
         skippedRefs = &skipped;
     }
+    if (maxRows == 0) {
```
qkrorlqr: you check `maxRows == 0` in listnodes.cpp - seems to be enough, we can just assume that maxRows is non-zero in the database implementations

cloud/filestore/libs/storage/tablet/tablet_ut_nodes.cpp
```diff
@@ -685,6 +685,19 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Nodes)
 
             UNIT_ASSERT(response->Record.GetCookie().empty());
         }
+
+        // Limit by elements count.
+        {
+            auto response = tablet.ListNodes(RootNodeId, 0, TString{}, 1);
+            const auto& names = response->Record.GetNames();
+            UNIT_ASSERT(names.size() == 1);
```
qkrorlqr: it's better to use `UNIT_ASSERT_VALUES_EQUAL(expected, actual)` - it prints the operands and their diff upon failure

Issue #4649, opened 2025-11-11
[Filestore] shared memory api for passing data in read/write grpc requests to filestore-server


debnatkh:
Currently, all `ReadData`/`WriteData` operations transfer data through gRPC protobufs, which might be CPU-intensive.

In #3361, we allowed for passing iovecs in write requests. We would like to allow the use of this feature also via gRPC API. 

The proposed solution will include a gRPC-server-level API that will allow for mmap-ing of shared files, which any local client will be able to read from/write to, allowing for sharing address space with the gRPC server. Proposed API:

* `Mmap(relativeFilepath, size) → id (any string)`
* `Munmap(id)`
* `PingMmap(id)`
* `ListMmaps`

This will also allow us to pre-process every incoming `WriteData`/`ReadData` gRPC request by verifying that their addresses are within mmaped ranges and adjusting their addresses.

---

_Note:_ this issue is loosely linked to https://github.com/ydb-platform/nbs/issues/3699, where we try to provide our gRPC clients with a leaner and faster way to access filestore data index

PR #5208, opened 2026-02-10
issue-4649 [Filestore]: add server-level sensors for the shared memory logic

debnatkh:
References #4649

### DIFF
--- a/cloud/filestore/libs/server/server.cpp
+++ b/cloud/filestore/libs/server/server.cpp
@@ -1550,7 +1550,6 @@ private:
     std::unique_ptr<NStorage::NServer::TEndpointPoller> EndpointPoller;
 
     ISchedulerPtr Scheduler;
-    static constexpr TDuration RefreshInterval = TDuration::Seconds(15);
 
 public:
     template <typename T>
@@ -1591,9 +1590,11 @@ private:
     void Refresh()
     {
         if constexpr (requires { AppCtx.State; }) {
-            if (AppCtx.State && AppCtx.ServerStats) {
-                AppCtx.ServerStats->Update(AppCtx.State->GetStateStats());
-            }
+            Y_ABORT_UNLESS(
+                AppCtx.State && AppCtx.Stats,
+                "State and Stats should be initialized upon TServer::Refresh "
+                "call");
+            AppCtx.ServerStats->Update(AppCtx.State->GetStateStats());
         }
 
         ScheduleRefresh();
@@ -1607,8 +1608,9 @@ private:
 
         auto weak_ptr = this->weak_from_this();
         Scheduler->Schedule(
-            TInstant::Now() + RefreshInterval,
-            [weak_ptr]() {
+            TInstant::Now() + UpdateCountersInterval,
+            [weak_ptr]()
+            {
                 if (auto self = weak_ptr.lock()) {
                     self->Refresh();
                 }


cloud/filestore/libs/server/server.cpp
```diff
@@ -1546,32 +1549,70 @@ class TServer final
 
     std::unique_ptr<NStorage::NServer::TEndpointPoller> EndpointPoller;
 
+    ISchedulerPtr Scheduler;
+    static constexpr TDuration RefreshInterval = TDuration::Seconds(15);
```
SvartMetal: Maybe use this symbol? https://github.com/ydb-platform/nbs/blob/fec7c419f44d95d91995742cd8b155b04b62a76b/cloud/storage/core/libs/diagnostics/public.h#L31

cloud/filestore/libs/server/server.cpp
```diff
@@ -1546,32 +1549,70 @@ class TServer final
...
             if (Config->GetSharedMemoryTransportEnabled()) {
                 ctx.State = std::make_shared<TServerState>(
                     Config->GetSharedMemoryBasePath());
+
+                ctx.ServerStats = std::make_shared<TServerStats>(counters);
+            }
+        }
+    }
+
+    void Refresh()
```
SvartMetal, line 1590: UpdateCounters?

cloud/filestore/libs/server/server.cpp
```diff
@@ -1546,32 +1549,70 @@ class TServer final
...
+    {
+        if constexpr (requires { AppCtx.State; }) {
+            if (AppCtx.State && AppCtx.ServerStats) {
```
SvartMetal: `Y_ABORT_UNLESS(AppCtx.ServerStats);`?

SvartMetal: or `Y_DEBUG_ABORT_UNLESS`

cloud/filestore/libs/server/server.cpp
debnatkh: I am going to use this entrypoint both for region invalidation and counters update, so I propose to keep `Refresh`

cloud/filestore/libs/server/server.cpp
debnatkh: ServerStats can be uninitialized if GetSharedMemoryTransportEnabled is false

SvartMetal: Here you check the pointer https://github.com/ydb-platform/nbs/pull/5208/changes#diff-d6f2cf095000365f24c0741d0eda6d4bb4f502e92ae2438aaecd569c42b8b5c9R1722. 
Refresh is called only after `ScheduleRefresh`, so it should not be possible to reach this point with `ServerStats == nullptr`.

cloud/filestore/libs/server/server_memory_state.cpp
```diff
@@ -161,4 +161,16 @@ NProto::TError TServerState::PingMmapRegion(ui64 mmapId)
     return NProto::TError{};
 }
 
+TServerStateStats TServerState::GetStateStats() const
+{
+    TLightReadGuard guard(StateLock);
+
+    TServerStateStats info;
+    info.MmapRegionCount = MmapRegions.size();
+    for (const auto& [id, region] : MmapRegions) {
```
agalibin, line 170: maybe rename 'id' -> '_' as not used

debnatkh: Will fix in another PR. In order not to rerun pipelines

PR #5213, opened 2026-02-11
[NBS] add logging to NVMe manager; minor fixes

### LINKED ISSUE #5250: [NBS] Add VFIO-PCI passthrough support for local NVMe devices in QEMU VMs
Currently, NBS can expose local NVMe disks to QEMU VMs via the vhost-user protocol. However, vhost-user has significant drawbacks: it introduces performance overhead that prevents full utilization of disk capabilities, and it doesn't support NVMe-specific features like NVIDIA GPUDirect Storage (GDS).
To address these limitations, NBS should support exposing local NVMe disks via VFIO-PCI passthrough. This would eliminate the virtualization overhead, enable maximum I/O performance, and provide full access to advanced NVMe features required by GPU-accelerated workloads.

sharpeye:
#5250

### DIFF
--- a/cloud/blockstore/libs/nvme/nvme_linux.cpp
+++ b/cloud/blockstore/libs/nvme/nvme_linux.cpp
@@ -1,8 +1,9 @@
 #include "nvme.h"
 
+#include <cloud/storage/core/libs/common/format.h>
 #include <cloud/storage/core/libs/common/task_queue.h>
 #include <cloud/storage/core/libs/common/thread_pool.h>
-
+#include <cloud/storage/core/libs/diagnostics/logging.h>
 #include <util/generic/vector.h>
 #include <util/generic/yexception.h>
 #include <util/string/builder.h>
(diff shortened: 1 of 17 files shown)


cloud/blockstore/libs/nvme/nvme_linux.cpp
```diff
@@ -591,15 +593,48 @@ class TNvmeManager final
...
+
+        std::span formats(ns.lbaf, ns.nlbaf + 1);
+
+        // Pick the most performant format without metadata
+        auto cmp = [](const auto& lhs, const auto& rhs)
+        {
+            return std::tie(lhs.ms, lhs.rp) < std::tie(rhs.ms, rhs.rp);
+        };
+
+        auto it = std::ranges::min_element(formats, cmp);
```
dvrazumov, line 624: Why `min_element` is most performant?  `.rp` is "relative performance", so here the least performant format is choosed, isn't it?

dvrazumov: Ouch, turns out that `pr=0` is best performance

Issue #5216, opened 2026-02-11
[Filestore] Shards balancer crashes immediately after the filesystem resizes and increases shard count


ruslan-gar:
1. TEvAggregateStatsCompleted was sent
2. The filesystem was resized,  TIndexTabletActor::ExecuteTx_ConfigureShards was called, 
ShardBalancer was recreated with new number of shards.
3. TEvAggregateStatsCompleted was handled, UpdateShardBalancer with old number of shards was called,
which caused   `Y_ABORT_UNLESS(stats.size() == Metas.size());`

The solution is to ignore TEvAggregateStatsCompleted in TIndexTabletActor::StateZombie

PR #5221, opened 2026-02-11
issue-5216: [Filestore] Fix crash in shard balancer

ruslan-gar:
The details are in issue #5216


cloud/filestore/libs/storage/service/service_ut_sharding.cpp
```diff
@@ -7056,5 +7056,66 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
...
+            TInstant());
+
+        env.GetRuntime().Send(
+            new IEventHandle(
+                fsInfo.MainTabletActorId,   // recipient
+                TActorId(),                 // sender
+                completion.release(),
+                0,   // flags
+                0),
+            0);
```
debnatkh, line 7118: Shouldn't you wait here for the tablet to accept the released completion?

PR #5229, opened 2026-02-12
[NBS] moved partition client from part_ut.cpp to separate lib

vladstepanyuk:
will be used for testing #4875

### DIFF
--- a/cloud/blockstore/libs/storage/partition/ut/ya.make
+++ b/cloud/blockstore/libs/storage/partition/ut/ya.make
@@ -10,7 +10,6 @@ SRCS(
 
 PEERDIR(
     cloud/blockstore/libs/storage/testlib
-    cloud/blockstore/libs/storage/partition/testlib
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/partition/testlib/part_client.cpp
```diff
@@ -0,0 +1,552 @@
+#include "part_client.h"
+
+
```
qkrorlqr: nit: extra blank line

cloud/blockstore/libs/storage/partition/ut/ya.make
```diff
@@ -10,6 +10,7 @@ SRCS(
 
 PEERDIR(
     cloud/blockstore/libs/storage/testlib
```
qkrorlqr, line 12: let's put this client into this existing lib instead of creating a new lib

PR #5234, opened 2026-02-12
issue-4649 [Filestore] Add shared memory region timeout invalidation logic

debnatkh:
References #4649

### DIFF
--- a/cloud/filestore/libs/server/server_memory_state_ut.cpp
+++ b/cloud/filestore/libs/server/server_memory_state_ut.cpp
@@ -266,7 +266,8 @@ Y_UNIT_TEST_SUITE(TServerStateTest)
 
         Sleep(TDuration::Seconds(2));
 
-        state.InvalidateTimedOutRegions();
+        auto error = state.InvalidateTimedOutRegions();
+        UNIT_ASSERT_C(!HasError(error), FormatError(error));
         UNIT_ASSERT_VALUES_EQUAL(0u, state.ListMmapRegions().size());
     }
 


cloud/filestore/libs/server/server.cpp
```diff
@@ -1595,6 +1596,11 @@ class TServer final
                 "State and Stats should be initialized upon TServer::Refresh "
                 "call");
             AppCtx.ServerStats->Update(AppCtx.State->GetStateStats());
+            const auto error = AppCtx.State->InvalidateTimedOutRegions();
+            if (HasError(error)) {
+                ReportInvalidateTimedOutRegionsError(
```
SvartMetal, line 1601: Maybe call this inside `InvalidateTimedOutRegions`? 
Since all relevant information and context are available there, this would simplify the caller.
This is preferable because `server.cpp` is already large

cloud/filestore/libs/server/server_memory_state.cpp
```diff
@@ -173,4 +176,30 @@ TServerStateStats TServerState::GetStateStats() const
...
+NProto::TError TServerState::InvalidateTimedOutRegions()
+{
+    auto now = TInstant::Now();
+
+    TLightWriteGuard guard(StateLock);
+
+    for (auto it = MmapRegions.begin(); it != MmapRegions.end(); ) {
+        if (now - it->second.GetLatestActivityTimestamp() > RegionTimeout) {
+            auto metadata = it->second.ToMetadata();
+            if (munmap(metadata.Address, metadata.Size) != 0) {
```
SvartMetal, line 188: Should we log the `munmap` error code?

cloud/filestore/libs/server/server_memory_state_ut.cpp
```diff
@@ -252,10 +252,28 @@ Y_UNIT_TEST_SUITE(TServerStateTest)
...
+        TString path1 = env.CreateTestFile("file1.dat", 4096);
+
+        auto result = state.CreateMmapRegion(path1, 4096);
+        UNIT_ASSERT_C(!HasError(result), FormatError(result.GetError()));
+
+        UNIT_ASSERT_VALUES_EQUAL(1u, state.ListMmapRegions().size());
+
+        Sleep(TDuration::Seconds(2));
+
+        state.InvalidateTimedOutRegions();
```
SvartMetal: Should we check for an error?

cloud/filestore/libs/server/server.cpp
debnatkh: I consider it reasonable to keep state class monitoring-level agnostic. Its callee should be responsible for processing errors.

> This is preferable because server.cpp is already large

Probably because both endpoint-level and controlplane-level server logic is kept in the same source file. Would you like me to create a separate refactoring PR/issue?

cloud/filestore/libs/server/server_memory_state.cpp
debnatkh: LastSystemErrorText is passed in an error, same as here: https://github.com/ydb-platform/nbs/blob/f9a42a8ba3b43b798d0a26e3250a0a68d3c3c6e4/cloud/filestore/libs/server/server_memory_state.cpp#L114-L117

PR #5236, opened 2026-02-12
Some fixes to run ya test on arm

proller:
USE_SYSTEM_VIRTIOFSD : Use  virtiofs-server from system
USE_BUNDLED_YDBD : Use ydbd from source instead sandbox
FORCE_BUILD_FLAKE8_LINTER : Build dummy tool for avoid errors (we dont need to run linters to arm)


build/scripts/fetch_from_sandbox.py
SvartMetal, line 1: Why?

PR #5249, opened 2026-02-14
issue-2674: introduced ResponseLog to implement RenameNodeInDestination idempotency

qkrorlqr:
### Notes
Implemented `RenameNodeInDestination` request idempotency via an explicit "ResponseLog". We can't use `DupCache` to solve this problem for several reasons:
* `DupCache` is stored per-session and `RenameNodeInDestination` request/response idempotency shouldn't be linked to any session
* `DupCache` is limited in size
* `DupCache` uses client-generated request_id as its key - it's not 100% reliable and for this task we need to be 100% confident that there'll be no collisions

So I simply store responses in a separate `ResponseLog` table, indexed by a <clientTabletId, requestId> pair where the client tablet just uses its `CommitId` as requestId.

`ResponseLog` table cleanup will be implemented in a separate PR - this one is already big enough. I left the following TODO item:
```
// TODO(#2674): send ResponseLogEntry deletion request
```  
In order to have some balance between code complexity and reliability, I'm planning to implement this cleanup in the following way:
* after `CommitRenameNodeInSource` succeeds, we'll send a `DeleteResponseLogEntryRequest` to the shard in charge of the destination dir - this will work in the majority of the cases but if the source (client) tablet dies before sending this request it won't be sent
* to clean up the responses left in `ResponseLog` because of the occasional `DeleteResponseLogEntryRequest` failures, we'll have a TTL for `ResponseLog` entries but it will be big enough to be confident that no one's going to retry a `RenameNodeInDestination` request for which we'll delete the response - e.g. 1 hour or so 

### Issue


cloud/filestore/libs/storage/api/tablet.h
```diff
@@ -173,6 +176,20 @@ struct TEvIndexTablet
...
+        EvDeleteResponseLogEntryResponse,
+
+        EvGetResponseLogEntryRequest = EvBegin + 73,
+        EvGetResponseLogEntryResponse,
+
+        EvWriteResponseLogEntryRequest = EvBegin + 75,
+        EvWriteResponseLogEntryResponse,
+
+        // After the TABLET sub-namespace we have TABLET_WORKER and TABLET_PROXY
+        // sub-namespaces which don't have any non-local events so we if we run
```
agalibin, line 189: "so we if we" - "so if we"?

cloud/filestore/libs/storage/tablet/tablet_actor_responselog.cpp
```diff
@@ -0,0 +1,217 @@
...
+    TABLET_VERIFY(!HasError(args.Error));
+    LOG_DEBUG(ctx, TFileStoreComponents::TABLET,
+        "%s WriteResponseLogEntry completed: %s",
+        LogTag.c_str(),
+        args.Entry.ShortUtf8DebugString().Quote().c_str());
+
+    CommitResponseLogEntry(std::move(args.Entry));
+
+    using TResponse = TEvIndexTablet::TEvWriteResponseLogEntryResponse;
+
```
agalibin, line 212: extra empty line

cloud/filestore/libs/storage/api/tablet.h
qkrorlqr: will fix in the next PR

cloud/filestore/libs/storage/tablet/tablet_actor_responselog.cpp
qkrorlqr: will fix in the next PR

debnatkh (approved):

PR #5251, opened 2026-02-16
[NBS] Added hard limit for fresh zero requests

vladstepanyuk:


komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition/part_actor_writefreshblocks.cpp
```diff
@@ -846,6 +846,28 @@ void TPartitionActor::ZeroFreshBlocks(
...
+                flags));
+
+        LWTRACK(
+            ResponseSent_Partition,
+            requestInfo->CallContext->LWOrbit,
+            "ZeroBlocks",
+            requestInfo->CallContext->RequestId);
+
+        NCloud::Reply(ctx, *requestInfo, std::move(response));
+        return;
```
budevg, line 868: `TPartitionActor::HandleZeroBlocks` updates in flight bookkeeping before delegating to `ZeroFreshBlocks`:

```
  ++WriteAndZeroRequestsInProgress;
  ...
  State->AccessCommitQueue().AcquireBarrier(commitId);
```

In the hard limit rejection path, `ZeroFreshBlocks` replies and returns early without running the corresponding completion cleanup. As a result, both the in flight counter and commit barrier are leaked for that request.

cloud/blockstore/libs/storage/partition/part_actor_writefreshblocks.cpp
```diff
@@ -846,6 +846,28 @@ void TPartitionActor::ZeroFreshBlocks(
     TBlockRange32 writeRange,
     ui64 commitId)
 {
+    if (State->GetUnflushedFreshBlobByteCount() >=
```
budevg, line 849: We need to add some unit test for this

budevg (changes requested):

PR #5255, opened 2026-02-17
Send TEvNavigateKeySet to SchemeCache with undelivery flag

PeshkovMikhail:
When manually preempted volumes are present and `UseSchemeCache` is true, the `DescribeVolume` request might be sent before `SchemeCache` is activated. Under such conditions, `DescribeVolume` will hit a timeout.

If the `NavigateKeySet` request isn’t delivered, the request will be rejected instead of waiting.

### DIFF
--- a/cloud/blockstore/libs/storage/ss_proxy/ss_proxy_ut.cpp
+++ b/cloud/blockstore/libs/storage/ss_proxy/ss_proxy_ut.cpp
@@ -1827,26 +1827,16 @@ Y_UNIT_TEST_SUITE(TSSProxyTest)
 
         TActorId sender;
 
-        TEvSSProxy::TEvDescribeSchemeResponse* response = nullptr;
-
         CreateVolume(runtime, "vol0", 4096);
 
         runtime.SetEventFilter(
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/ss_proxy/ss_proxy_ut.cpp
```diff
@@ -1811,6 +1811,72 @@ Y_UNIT_TEST_SUITE(TSSProxyTest)
...
+            [&](auto&, TAutoPtr<IEventHandle>& event)
+            {
+                if (event->GetTypeRewrite() ==
+                    TEvTxProxySchemeCache::EvNavigateKeySet)
+                {
+                    sender = event->Sender;
+                    return true;
+                }
+                return false;
+            });
```
komarevtsev-d, line 1842: это clang-fornat?

PeshkovMikhail: да

komarevtsev-d (approved):

SvartMetal (approved):

PR #5258, opened 2026-02-17
Do not mix postponed and backoff time in server stats

komarevtsev-d:
https://github.com/ydb-platform/nbs/issues/5095
1) `postponedTime` in `TRequestCounters` is splitted into 3: posponed (throttled time), backoff (durable client backoff), shaping (for shaping throttler that will be introduced later).
2) ThrottlerDelay histogram fixed. Previously it was reporting the "wait time" (backoff + posponed)
3) MaxTime calculation will be more accurate, since it no longer subtracts backoff time
4) Profile log will now also contain posponed time and not "wait time"


qkrorlqr (dismissed):

qkrorlqr:
Let's add a separate sensor for WaitTime?

drbasic (approved):

PR #5269, opened 2026-02-18
issue-2674: sending DeleteResponseLogEntryRequest upon CommitRenameNodeInSource completion

qkrorlqr:
### Notes
Follow-up for https://github.com/ydb-platform/nbs/pull/5249
`DeleteResponseLogEntryRequest` is now sent upon `CommitRenameNodeInSource` completion (except for completion with a retriable error)

Remaining TODO (separate PR): clean up the responses left in `ResponseLog` because of the occasional `DeleteResponseLogEntryRequest` failures - we'll have a TTL for `ResponseLog` entries but it will be big enough to be confident that no one's going to retry a `RenameNodeInDestination` request for which we'll delete the response - e.g. 1 hour or so 

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
@@ -96,9 +96,7 @@ void TRenameNodeInDestinationActor::SendRequest(const TActorContext& ctx)
         Request.GetNewParentId(),
         Request.GetNewName().c_str());
 
-    ctx.Send(
-        MakeIndexTabletProxyServiceId(),
-        request.release());
+    ctx.Send(MakeIndexTabletProxyServiceId(), request.release());
 }
 
 void TRenameNodeInDestinationActor::HandleRenameNodeInDestinationResponse(
@@ -177,6 +175,8 @@ void TRenameNodeInDestinationActor::ReplyAndDie(
         Request.GetHeaders().GetSessionId(),
         RequestId,
         OpLogEntryId,
+        Request.GetFileSystemId(),
+        Request.GetHeaders().GetRequestId(),
         Request.GetOriginalRequest(), // TODO: move
         std::move(ProfileLogRequest),
         std::move(response)));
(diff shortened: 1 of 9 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -202,6 +202,191 @@ STFUNC(TRenameNodeInDestinationActor::StateWork)
...
+
+        const auto message = Sprintf(
+            "DeleteResponseLogEntry failed for %s, %lu, %lu with error %s"
+            ", will not retry",
+            ShardFileSystemId.c_str(),
+            ClientTabletId,
+            TabletRequestId,
+            FormatError(msg->GetError()).Quote().c_str());
+
+        LOG_ERROR(
```
agalibin, line 313: same log in the ReplyAndDie, are you sure that is need here?

cloud/filestore/libs/storage/tablet/tablet_tx.h
```diff
@@ -1267,6 +1267,8 @@ struct TTxIndexTablet
         const NProto::TRenameNodeRequest Request;
         const NProtoPrivate::TRenameNodeInDestinationResponse Response;
         const ui64 OpLogEntryId;
+        /* const */ TString ShardFileSystemId;
```
agalibin, line 1270: /* const */?

qkrorlqr: I wanted to highlight that this field should actually be const but since we want to move from it upon tx completion, we have to make it non-const

cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
qkrorlqr: good catch, will delete it (in the current PR, if there are other comments, or in the next one - if there're no more comments for this PR)

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -1461,6 +1680,7 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
         env.GetRuntime().SetEventFilter(rebootTracker.GetEventFilter());
 
         const ui64 tabletId = env.BootIndexTablet(nodeIdx);
+        //OverrideDescribeFileStore(env.GetRuntime(), nodeIdx, tabletId);
```
debnatkh, line 1683: //

qkrorlqr: will remove, thx

PR #5279, opened 2026-02-19
issue-2674: deleting stale ResponseLogEntries in background

qkrorlqr:
### Notes
Follow-up for https://github.com/ydb-platform/nbs/pull/5249 and https://github.com/ydb-platform/nbs/pull/5269

`ResponseLogEntry` deletion via explicit `TEvDeleteResponseLogEntryRequest`s is implemented as "best-effort" (for simplicity) so we need to have some background op that deletes really old entries from time to time. Introduced `TEvRunRegularTasks` event - planning to add similar rarely firing tablet background tasks there in the future.

The number of `ResponseLogEntr`ies should be tiny so iterating over all of them once a minute should not be a problem at all. I added `ResponseLogEntryCount` sensor to be able to notice its growth (if we introduce a bug which causes the number of these entries to grow). 

Fixes filestore tablet ut build after https://github.com/ydb-platform/nbs/pull/5269 vs https://github.com/ydb-platform/nbs/pull/5272 conflict as well

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
@@ -302,21 +302,6 @@ void TDeleteResponseLogEntryActor::HandleDeleteResponseLogEntryResponse(
             return;
         }
 
-        const auto message = Sprintf(
-            "DeleteResponseLogEntry failed for %s, %lu, %lu with error %s"
-            ", will not retry",
-            ShardFileSystemId.c_str(),
-            ClientTabletId,
-            TabletRequestId,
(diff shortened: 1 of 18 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -302,21 +302,6 @@ void TDeleteResponseLogEntryActor::HandleDeleteResponseLogEntryResponse(
             return;
         }
 
-        const auto message = Sprintf(
-            "DeleteResponseLogEntry failed for %s, %lu, %lu with error %s"
```
qkrorlqr, line 306: this message is printed in `ReplyAndDie()`

Issue #5094, opened 2026-01-28
[Filestore] Add Filestore scrubbing


jkuradobery:
We would like to know about all the nodes missing from shards within the fs. For now we can only know about it during directory listing.

- [x] implement filesystem traversal
- [x] add flag to report nodes missing from shards but present in the index tablet.
- [x] Abstract nodes listing within the filesystem traversal. 
- [x] Add scrubbing task
- [x] Use unsafe flag in scrubbing task
- [x] Add scrubbing task to admin

PR #5291, opened 2026-02-23
issue-5094: [Disk Manager] Add maxBytes parameter to ListNodes()

jkuradobery:
#5094


cloud/filestore/public/sdk/go/client/client.go
```diff
@@ -309,6 +310,9 @@ func (client *Client) ListNodes(
 			SessionId:    []byte(session.SessionID),
 		},
 	}
+	if maxBytes > 0 {
+		req.MaxBytes = maxBytes
```
debnatkh: is this if needed? In proto3 zero is the dafault

PR #5336, opened 2026-02-27
[Logging] Introduce contextLogger to enhance logging with context support

jkuradobery:
We have removed passing the context in #3139. Bring it back.

### DIFF
--- a/cloud/tasks/logging/logger.go
+++ b/cloud/tasks/logging/logger.go
@@ -48,6 +48,7 @@ func getLogger(ctx context.Context) Logger {
 
 	return logger
 }
+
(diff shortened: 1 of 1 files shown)


cloud/tasks/logging/logger.go
```diff
@@ -48,6 +48,17 @@ func GetLogger(ctx context.Context) Logger {
 
 	return logger
 }
```
SvartMetal, line 50: \n

jkuradobery: fixed

BarkovBG (approved):

Issue #5340, opened 2026-02-28
[Filestore] Decrease FUSE loop thread CPU consumption


qkrorlqr:
We can get rid of per-request `TRequestActor`s - actor registration is very expensive and can take around 25% of FUSE loop thread time:

<img width="1193" height="486" alt="Image" src="https://github.com/user-attachments/assets/5d28f10b-5737-4ba4-925d-9b5a464c56ba" />

qkrorlqr:
the switch to `THandlerActor`s gives approximately 20-40% iops improvement (depending on the configuration and the test scenario) according to the results from our production clusters, closing the issue

PR #5341, opened 2026-02-28
issue-5340: TKikimrFileStore - implemented no-TRequestActor mode

qkrorlqr:
### Notes
`TRequestActor` creation and registration for each request is expensive and shows up in flamegraphs. It's especially bad for the case when we have just 1 FUSE loop thread. Implemented a mode where we reuse the same actor for all requests and thus do `Send()` instead of `Register()` in FUSE loop threads per each request:
* `Send()` is cheaper than `Register()`
* and we do this same `Send()` anyway in `TRequestActor` so the end result is -1 `Register()` call per each request and the number of `Send()` calls stays the same

### Issue
https://github.com/ydb-platform/nbs/issues/5340

### DIFF
--- a/cloud/blockstore/libs/rdma/fake/client_ut.cpp
+++ b/cloud/blockstore/libs/rdma/fake/client_ut.cpp
@@ -104,6 +104,13 @@ struct TTestActorSystem
         return true;
     }
 
+    bool Send(IEventHandlePtr event) override
+    {
+        Runtime.SendAsync(event.release());
+
+        return true;
+    }
+
     TProgramShouldContinue& GetProgramShouldContinue() override
     {
         return ProgramShouldContinue;
--- a/cloud/blockstore/libs/service_kikimr/ut/kikimr_test_env.cpp
+++ b/cloud/blockstore/libs/service_kikimr/ut/kikimr_test_env.cpp
@@ -91,6 +91,12 @@ bool TTestActorSystem::Send(const TActorId& recipient, IEventBasePtr event)
     return true;
 }
 
+bool TTestActorSystem::Send(IEventHandlePtr event)
+{
+    Runtime->Send(event.release());
+    return true;
+}
+
 TProgramShouldContinue& TTestActorSystem::GetProgramShouldContinue()
 {
     return ProgramShouldContinue;
--- a/cloud/blockstore/libs/service_kikimr/ut/kikimr_test_env.h
+++ b/cloud/blockstore/libs/service_kikimr/ut/kikimr_test_env.h
@@ -60,6 +60,8 @@ public:
         const NActors::TActorId& recipient,
         NActors::IEventBasePtr event) override;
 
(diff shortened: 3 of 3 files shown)


cloud/filestore/libs/service_kikimr/service.cpp
```diff
@@ -143,6 +143,152 @@ class TRequestActor final
...
+    using TResponseEvent = typename TMethod::TResponseEvent;
+
+private:
+    THashMap<ui64, TPromise<TResponse>> Responses;
+    ui64 Cookie = 0;
+    TAdaptiveLock Lock;
+
+public:
+    void SendRequest(
+        IActorSystem& ass,
```
drbasic: ass ?)

cloud/filestore/libs/service_kikimr/service.cpp
```diff
@@ -143,6 +143,152 @@ class TRequestActor final
...
+        ass.Send(std::move(event));
+    }
+
+    void HandleResponse(
+        const typename TResponseEvent::TPtr& ev,
+        const TActorContext& ctx)
+    {
+        auto* msg = ev->Get();
+
+        LOG_TRACE_S(ctx, TFileStoreComponents::SERVICE_PROXY,
```
drbasic, line 193: clang-format?

cloud/filestore/libs/service_kikimr/service.cpp
```diff
@@ -143,6 +143,152 @@ class TRequestActor final
...
+////////////////////////////////////////////////////////////////////////////////
+
+struct THandler
+{
+    TActorId SelfId;
+
+public:
+#define FILESTORE_SEND_REQUEST(name, ...)                                      \
+    TMethodHandler<T##name##Method> name##Handler;                             \
+    auto SendRequest(                                                          \
```
drbasic: void

cloud/filestore/libs/service_kikimr/service.cpp
qkrorlqr: I'm maintaining the style of the other code in this file

cloud/filestore/libs/service_kikimr/service.cpp
```diff
@@ -334,6 +490,15 @@ class TKikimrFileStore final
         std::shared_ptr<typename T::TRequest> request,
         TPromise<typename T::TResponse> response)
     {
+        if (UsePermanentActor) {
+            Handler->SendRequest(
+                *ActorSystem,
+                std::move(callContext),
+                std::move(*request),
```
e673, line 497: Note: this corrupts the request for other holders of `std::shared_ptr<typename T::TRequest>`.

qkrorlqr: fix: https://github.com/ydb-platform/nbs/pull/5452

Issue #5349, opened 2026-03-02
[Filestore] add possibility to add AlignedData to index out of order


neihar:
part of #2293
Currently we have possibility to only apply index updates in commit order. Deletion markers and block lifetimes are built assuming newer commits always arrive later, which breaks replay and rebuild flows where older writes are inserted after newer state is already present.

What was done in this commit:
- Added out-of-order insertion path for TDeletionMarkers (AddOutOfOrder) with interval handling that keeps newer commits authoritative.
- Added TMixedBlocks::AddDeletionMarkerOutOfOrder and used it from tablet state when marker commit is not the current commit.
- Added TIndexTabletState::WriteMixedBlocksCommitOrderAware for mixed block writes from older commits.
- Updated fresh block deletion condition to handle out-of-order deletes (MinCommitId <= commitId < MaxCommitId).
- Added deterministic and randomized unit tests for order-independent behavior in deletion markers, fresh blocks, and mixed blocks.

PR #5350, opened 2026-03-02
[Filestore] out of order support for index #5349

neihar:
Currently we have possibility to only apply index updates in commit order. Deletion markers and block lifetimes are built assuming newer commits always arrive later, which breaks replay and rebuild flows where older writes are inserted after newer state is already present.

What was done in this commit:
- Added out-of-order insertion path for TDeletionMarkers (AddOutOfOrder) with interval handling that keeps newer commits authoritative.
- Added TMixedBlocks::AddDeletionMarkerOutOfOrder and used it from tablet state when marker commit is not the current commit.
- Added TIndexTabletState::WriteMixedBlocksCommitOrderAware for mixed block writes from older commits.
- Updated fresh block deletion condition to handle out-of-order deletes (MinCommitId <= commitId < MaxCommitId).
- Added deterministic and randomized unit tests for order-independent behavior in deletion markers, fresh blocks, and mixed blocks.

### DIFF
--- a/cloud/filestore/libs/storage/tablet/model/deletion_markers.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/deletion_markers.cpp
@@ -73,6 +73,8 @@ private:
 
     IAllocator* Alloc;
 
+    void AddOutOfOrder(TDeletionMarker deletionMarker, TDisjointRangeMap& map);
+
 public:
     TImpl(IAllocator* alloc)
         : DisjointRangeMapByNodeId{alloc}
@@ -102,7 +104,10 @@ void TDeletionMarkers::TImpl::Add(
     TDisjointRangeMapByNodeId::insert_ctx ctx;
     auto it = DisjointRangeMapByNodeId.find(deletionMarker.NodeId, ctx);
     if (it == DisjointRangeMapByNodeId.end()) {
-        it = DisjointRangeMapByNodeId.emplace_direct(ctx, deletionMarker.NodeId, Alloc);
+        it = DisjointRangeMapByNodeId.emplace_direct(
+            ctx,
+            deletionMarker.NodeId,
+            Alloc);
     }
 
     auto& map = it->second;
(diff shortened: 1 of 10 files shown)


cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -854,9 +884,17 @@ void TIndexTabletState::MarkMixedBlocksDeleted(
     IncrementDeletionMarkersCount(db, blocksCount);
 
     if (Impl->MixedBlocks.IsLoaded(rangeId)) {
-        Impl->MixedBlocks.AddDeletionMarker(
-            rangeId, {nodeId, commitId, blockIndex, blocksCount}
-        );
+        if (commitId == GetCurrentCommitId()) {
```
qkrorlqr, line 887: let's add a `TABLET_VERIFY` for the `commitId > GetCurrentCommitId()` case

and probably a counter for the number of out of order additions would also be useful

cloud/filestore/libs/storage/tablet/model/mixed_blocks_ut.cpp
```diff
@@ -190,6 +190,177 @@ Y_UNIT_TEST_SUITE(TMixedBlocksTest)
...
+                UNIT_ASSERT_VALUES_EQUAL(blocks[i].BlockIndex, blockIndex + i);
+                UNIT_ASSERT_VALUES_EQUAL(blocks[i].MinCommitId, 20);
+                UNIT_ASSERT_VALUES_EQUAL(
+                    blocks[i].MaxCommitId,
+                    InvalidCommitId);
+            }
+        }
+    }
+
+    Y_UNIT_TEST(ShouldKeepNonOverlappedPartVisibleForOutOfOrderInsertion)
```
qkrorlqr, line 283: nit: Overlapped -> Overlapping (according to ChatGPT)

cloud/filestore/libs/storage/tablet/model/mixed_blocks_ut.cpp
```diff
@@ -190,6 +190,177 @@ Y_UNIT_TEST_SUITE(TMixedBlocksTest)
...
+            seen[idx] = true;
+            minCommitId[idx] = block.MinCommitId;
+            maxCommitId[idx] = block.MaxCommitId;
+        }
+
+        for (ui32 i = 0; i < blocksCount; ++i) {
+            UNIT_ASSERT(seen[i]);
+        }
+
+        UNIT_ASSERT_VALUES_EQUAL(minCommitId[0], 10);
```
qkrorlqr: we usually do `UNIT_ASSERT_VALUES_EQUAL(expected, actual);`

cloud/filestore/libs/storage/tablet/model/mixed_blocks_ut.cpp
```diff
@@ -190,6 +190,177 @@ Y_UNIT_TEST_SUITE(TMixedBlocksTest)
...
+        auto blocks = visitor.Finish();
+        UNIT_ASSERT_VALUES_EQUAL(blocks.size(), blocksCount);
+
+        bool seen[blocksCount] = {};
+        ui64 minCommitId[blocksCount] = {};
+        ui64 maxCommitId[blocksCount] = {};
+
+        for (const auto& block: blocks) {
+            UNIT_ASSERT_VALUES_EQUAL(block.NodeId, nodeId);
+            UNIT_ASSERT(block.BlockIndex >= blockIndex);
```
qkrorlqr: `UNIT_ASSERT_LE/GE/LT/GT` (these macros print the operands in case of failure)

cloud/filestore/libs/storage/tablet/model/fresh_blocks.cpp
```diff
@@ -107,9 +89,7 @@ TVector<TFreshBlocks::TVersionedBlock> TFreshBlocks::MarkBlocksDeleted(
     for (auto it = start; it != end; ++it) {
         auto& block = const_cast<TBlock&>(it->first);
 
-        if (block.MinCommitId <= commitId &&
-            block.MaxCommitId == InvalidCommitId)
-        {
+        if (block.MinCommitId <= commitId && commitId < block.MaxCommitId) {
```
qkrorlqr, line 92: why?

cloud/filestore/libs/storage/tablet/model/deletion_markers.cpp
```diff
@@ -107,25 +109,110 @@ void TDeletionMarkers::TImpl::Add(TDeletionMarker deletionMarker)
 
     const ui32 start = deletionMarker.BlockIndex;
     const ui32 end = deletionMarker.BlockIndex + deletionMarker.BlockCount;
+    Y_ABORT_UNLESS(start < end);
```
qkrorlqr, line 117: `TABLET_VERIFY` or `STORAGE_VERIFY(fileSystemId or tabletId)`

cloud/filestore/libs/storage/tablet/model/deletion_markers.cpp
```diff
@@ -107,25 +109,110 @@ void TDeletionMarkers::TImpl::Add(TDeletionMarker deletionMarker)
...
 
-    auto lo = map.upper_bound(start);
```
qkrorlqr, line 111: let's structure the code a bit differently to decrease the number of changes:
```
if (!isLatestCommitId) {
    AddOutOfOrder(deletionMarker);
    return;
}

... old code ...
```

qkrorlqr (commented):
let's add a big comment referencing the related issue and saying that this out-of-order logic is not fully implemented yet and is not expected to be used outside of the code relevant to the aforementioned issue

cloud/filestore/libs/storage/tablet/model/fresh_blocks.cpp
qkrorlqr: I mean - it looks suspicious

because for example we may have already rebased the old `MaxCommitId` to some checkpoint's `CommitId` and the `commitId` that we're applying now might actually be greater than the rebased one

qkrorlqr: > than the rebased one

than the one that we had __before__ rebase

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
neihar: Added verify, metrics for out of order insertions should be added in the higher layer when actual usage of this will be declared. I will add this clause to the task where it is planned to be used to avoid missing this. Task: #5353

cloud/filestore/libs/storage/tablet/model/fresh_blocks.cpp
neihar: Should it be a problem? here we will make hole starting from CommitId of outOfOrder commit and ending at min of commitsId from the future. And this hole will be filled with later insertion. If commitId is greater we will not apply it. Probably I missing something about checkpoints

cloud/filestore/libs/storage/tablet/model/deletion_markers.cpp
neihar: it is not available here. We can add to model dependens on tablet in separate PR

Issue #3497, opened 2025-05-13
[NBS] data race on handling multiple disconnects when using poll waiting mode


tpashkin:
HandlePollEvent doesn't handle multiple disconnects correctly
<details>
<summary>tsan output</summary>

```
==================
WARNING: ThreadSanitizer: data race (pid=45681)
  Write of size 8 at 0x7b6000000d90 by thread T2:
    #0 std::__y1::unique_ptr<ibv_cq, int (*)(ibv_cq*)>::reset[abi:v15000](ibv_cq*) /work/nbs3/contrib/libs/cxxsupp/libcxx/include/__memory/unique_ptr.h:304:20 (cloud-blockstore-libs-rdma-impl-ut+0x4b41ae9) (BuildId: dd2692c2bf7a3021f6cdbe33d0193cd76d43783f)
    #1 NCloud::NBlockStore::NRdma::(anonymous namespace)::TClientEndpoint::DestroyQP() /work/nbs3/cloud/blockstore/libs/rdma/impl/client.cpp:761:21 (cloud-blockstore-libs-rdma-impl-ut+0x4b41e2c) (BuildId: dd2692c2bf7a3021f6cdbe33d0193cd76d43783f)
    #2 NCloud::NBlockStore::NRdma::(anonymous namespace)::TClient::Reconnect(NCloud::NBlockStore::NRdma::(anonymous namespace)::TClientEndpoint*) /work/nbs3/cloud/blockstore/libs/rdma/impl/client.cpp:2054:23 (cloud-blockstore-libs-rdma-impl-ut+0x4b189a9) (BuildId: dd2692c2bf7a3021f6cdbe33d0193cd76d43783f)
    #3 non-virtual thunk to NCloud::NBlockStore::NRdma::(anonymous namespace)::TClient::Reconnect(NCloud::NBlockStore::NRdma::(anonymous namespace)::TClientEndpoint*) /work/nbs3/cloud/blockstore/libs/rdma/impl/client.cpp (cloud-blockstore-libs-rdma-impl-ut+0x4b18b39) (BuildId: dd2692c2bf7a3021f6cdbe33d0193cd76d43783f)
[...]

PR #5361, opened 2026-03-03
fix disconnect race; issue-3497

tpashkin:
#3497

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/client.cpp
+++ b/cloud/blockstore/libs/rdma/impl/client.cpp
@@ -445,7 +445,7 @@ private:
     TPromise<IClientEndpointPtr> StartResult = NewPromise<IClientEndpointPtr>();
     TPromise<void> StopResult = NewPromise<void>();
 
-    ui64 FlushStartCycles = 0;
+    std::atomic<ui64> FlushStartCycles = 0;
 
     TBufferPool SendBuffers;
     TBufferPool RecvBuffers;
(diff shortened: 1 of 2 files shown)


cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -1287,24 +1307,14 @@ void TClientEndpoint::Disconnect() noexcept
...
-            ChangeState(
-                EEndpointState::Connected,
-                EEndpointState::Disconnecting);
-
-            FlushQueues();
-
-            if (WaitMode == EWaitMode::Poll) {
-                DisconnectEvent.Set();
-            }
+            DisconnectEvent.Set();
```
budevg, line 1315: In the old code we were switching to Disconnecting state from the CQ thread synchronously. Now we just signal event and there could be race that the CQ keeps processing events using HandleInputRequests, HandleQueuedRequests even though the qp is broken and the CM thread didn't acknowledged the event yet.

tpashkin: > In the old code we were switching to Disconnecting state from the CQ thread synchronously

In the old code `Disconnect` could have been called by both `CQ` and `CM`, so there still was a race, which this PR is all about

> there could be race that the CQ keeps processing events using HandleInputRequests, HandleQueuedRequests even though the qp is broken and the CM thread didn't acknowledged the event yet

True, but `QP` being broken is a wrong premise. `Disconnect` is called in 3 cases:
- Upon receiving `RDMA_CM_EVENT_DISCONNECTED` by `CM` thread. In this case `QP` will be transitioned to the `Error`. Calls like `ibv_post_send` which would be called as a result of `CQ` processing new requests, would fail and ultimately end up calling another `Disconnect`
- Upon receiving completion with status `!= IBV_WC_SUCCESS` by `CQ`. In this case `QP` is transitioned either to `ERR` or `CQE` and the same reasoning applies
- On calling `Stop` when `QP` is still in `RTS` state. In this case we can safely continue to process incoming requests

Issue #5362, opened 2026-03-03
[CSI] Support multiple nfs-vhost sockets


antonmyagkov:
Multiple nfs vhost process can be run on the same node. Each nfs-vhost process has unix socket nfs{id}.sock. csi driver should choose one of the nfs-vhost sockets to start endpoint for filesystem.

PR #5363, opened 2026-03-03
[CSI] Support multiple nfs-vhost

antonmyagkov:
issue: #5362 

Add support for multiple NFS host processes in the CSI driver.
When handling NodeStageVolume, the CSI driver selects one of the available NFS vhost sockets based on a hash of the Pod ID (or instance ID).

### DIFF
--- a/cloud/blockstore/tools/csi_driver/internal/driver/helper.go
+++ b/cloud/blockstore/tools/csi_driver/internal/driver/helper.go
@@ -1,6 +1,9 @@
 package driver
 
 import (
+	"hash/fnv"
+	"strconv"
+
 	storagecoreapi "github.com/ydb-platform/nbs/cloud/storage/core/protos"
 )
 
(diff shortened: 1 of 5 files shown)


cloud/blockstore/tools/csi_driver/internal/driver/helper.go
```diff
@@ -45,3 +48,19 @@ func isDiskRegistryMediaKind(mediaKind storagecoreapi.EStorageMediaKind) bool {
...
 }
+
+func getNfsSocket(socketKey string, nfsVhostReplicaCount uint) string {
+	if nfsVhostReplicaCount == 0 {
+		return defaultNfsSocketName
+	}
+
+	h := fnv.New64a()
+	h.Write([]byte(socketKey))
+	index := h.Sum64() % uint64(nfsVhostReplicaCount)
```
tpashkin, line 59: If `nfsVhostReplicaCount` ever changes, we will try to stop endpoint on a different replica 🤔

tpashkin: Or start another one in case of e.g. kubelet restart

tpashkin: Ok, it gets saved to stageData

tpashkin: And yet we can overwrite stageData with the new value in case of a kubelet restart. So the point still stands

Issue #5293, opened 2026-02-23
[Filestore] Don't allow node attr caching in guest if node attr might have changed after the request


qkrorlqr:
Node attr caching in guest can lead to stale node size being cached if size-modifying requests (append via WriteData, size extension via SetNodeAttr, AllocateData) overlap with attr-fetching requests (GetNodeAttr, ListNodes, CreateHandle). If a size-modifying request completes while an attr-fetching request is in flight, we should not let the guest cache the results of this attr-fetching request.

We seem to have this problem for example when we run compilation+linking with -j > 1 on top of filestore. Detailed description of one of the cases follows:

The problem seems to be caused by the fact that sometimes after appending to a file the guest ends up having old (pre-append) inode size cached. An example scenario:

1. guest sends an append request for file X in dir Y
2. guest sends a listing request for dir Y
3. listing partially completes in the backend - prev size (N1) for file X is retrieved
4. append op completes in the backend - file size increased to N2 - guest receives the response
5. [not sure whether this step is necessary] the inode for file X is evicted from the guest cache
6. listing request completes, returns a response which contains size N1 for file X and contains entry_timeout=15s, attr_timeout=15s for this file
7. guest caches size=N1 for file X for 15 seconds
8. fstat call is performed by an application within those 15 seconds - size=N1 is returned, the call doesn’t reach the backend
9. read call with offset>=N1 is performed by an application within those 15 seconds - zeroes are returned, the call doesn’t reach the backend

Example logs:

file created:
```
2026-02-02T23:14:33.080367Z→proller_test_nfs_postgresql-2-4-0→→→CreateHandle→→→→0.005894s→→→S_OK→→→→{parent_node_id=450, node_name=varbit.o, flags=38, mode=420, node_id=72057594038579030, handle=72325632859339675, size=0}→→→loop_thread_id=3758812→→client_id=local-qemu-32202
```

listing request:
```
2026-02-02T23:14:33.086454Z→proller_test_nfs_postgresql-2-4-0→→→ListNodes→→→0.007108s→→→S_OK→→→→{node_id=450, size=641}→loop_thread_id=3758812→→client_id=local-qemu-32202
```

append and flush - both start AFTER the listing starts and complete BEFORE the listing completes:
```
[...]

PR #5369, opened 2026-03-03
issue-5293: ResetAttrTimeout - allowing incomplete direntries (w/o name or w incomplete name)

qkrorlqr:
### Notes
* using the same code that's used in libvirtiofsd to calculate serialized `fuse_direntplus` size (even though the old code returned the same value, it's still safer to use exactly the same code that's used in libvirtiofsd)
* allowing missing or cropped `name` field - the code that limits the size of the returned directory content chunk is not format-aware and can cut the buffer at a random spot
* added a unittest for large (relatively large) dir listing
* added some comments describing these caveats

### Issue
https://github.com/ydb-platform/nbs/issues/5293

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_directory_content_format.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_directory_content_format.cpp
@@ -97,8 +97,7 @@ NProto::TError ResetAttrTimeout(
             }
         }
 
-        const ui64 fullSize =
-            FUSE_DIRENT_ALIGN(FUSE_NAME_OFFSET_DIRENTPLUS + de->dirent.namelen);
+        const ui64 fullSize = FUSE_DIRENTPLUS_SIZE(de);
 
         if (remainingLen < fullSize) {
             //
--- a/cloud/filestore/libs/vfs_fuse/fs_directory_content_format_ut.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_directory_content_format_ut.cpp
@@ -245,6 +245,81 @@ Y_UNIT_TEST_SUITE(TDirectoryContentFormatTest)
             FormatError(error));
     }
 
+    Y_UNIT_TEST(ShouldHandleIncompleteEntry)
+    {
+        const ui64 size = 4_KB;
+        const ui64 attrTimeout = 15;
+        const ui64 entryTimeout = 10;
+        const ui32 preferredBlockSize = 4_KB;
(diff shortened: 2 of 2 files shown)


cloud/filestore/libs/vfs_fuse/fs_directory_content_format.cpp
```diff
@@ -99,13 +97,17 @@ NProto::TError ResetAttrTimeout(
             }
         }
 
-        const ui64 fullSize = sizeof(fuse_direntplus)
-            + AlignUp<ui64>(de->dirent.namelen, sizeof(ui64));
+        const ui64 fullSize =
+            FUSE_DIRENT_ALIGN(FUSE_NAME_OFFSET_DIRENTPLUS + de->dirent.namelen);
```
budevg: you can use macro `FUSE_DIRENTPLUS_SIZE`

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -4058,6 +4058,97 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
             bootstrap.SetSize(12_KB);
         });
     }
+
+    Y_UNIT_TEST(ShouldListWithDirectoryContentChunkAtNameBoundary)
+    {
```
budevg, line 4063: It does not need to end at the name boundary. The final entry may be truncated, and the missing bytes can be anywhere within `fuse_direntplus`. Maybe it worth to add unit test for this case as well.

qkrorlqr: It shouldn't matter, the logic is still the same - if remainingLen < align(sizeof(fuse_direntplus) + namelen), then return

qkrorlqr: I can add a testcase into fs_directory_content_format_ut.cpp

cloud/filestore/libs/vfs_fuse/fs_directory_content_format.cpp
qkrorlqr: done

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
qkrorlqr: done

budevg (approved):

Issue #5356, opened 2026-03-02
[Filestore] filestore-vhost hangs while stopping upon lease expiration


SvartMetal:
when NodeBroker lease is expired (due to network problems for example) https://github.com/ydb-platform/nbs/blob/5cf58a77d9adcc346425923bb922d6a2c8a4f0b5/contrib/ydb/core/mind/lease_holder.cpp#L233, `EndpointManager::Drain` is called https://github.com/ydb-platform/nbs/blob/5cf58a77d9adcc346425923bb922d6a2c8a4f0b5/cloud/filestore/libs/daemon/vhost/bootstrap.cpp#L524 which waits for in-flight requests to complete and hangs because Interconnect is not working https://github.com/ydb-platform/nbs/blob/5cf58a77d9adcc346425923bb922d6a2c8a4f0b5/cloud/filestore/libs/endpoint/endpoint_manager.cpp#L184

PR #5372, opened 2026-03-03
issue-5356: [Filestore] should not wait for requests completion when NodeBroker lease is expired

SvartMetal:
### Notes
When NodeBroker lease is expired (due to network problems for example) https://github.com/ydb-platform/nbs/blob/5cf58a77d9adcc346425923bb922d6a2c8a4f0b5/contrib/ydb/core/mind/lease_holder.cpp#L233, `EndpointManager::Drain` is called https://github.com/ydb-platform/nbs/blob/5cf58a77d9adcc346425923bb922d6a2c8a4f0b5/cloud/filestore/libs/daemon/vhost/bootstrap.cpp#L524 which waits for in-flight requests to complete and hangs because Interconnect is not working https://github.com/ydb-platform/nbs/blob/5cf58a77d9adcc346425923bb922d6a2c8a4f0b5/cloud/filestore/libs/endpoint/endpoint_manager.cpp#L184

So the safe approach is to call _exit, which skips process shutdown

### Issue
#5356

### DIFF
--- a/cloud/filestore/libs/daemon/vhost/bootstrap.cpp
+++ b/cloud/filestore/libs/daemon/vhost/bootstrap.cpp
@@ -40,6 +40,7 @@
 #include <cloud/storage/core/libs/common/thread_pool.h>
 #include <cloud/storage/core/libs/common/timer.h>
 #include <cloud/storage/core/libs/daemon/mlock.h>
+#include <cloud/storage/core/libs/daemon/public.h>
 #include <cloud/storage/core/libs/diagnostics/logging.h>
 #include <cloud/storage/core/libs/diagnostics/monitoring.h>
 #include <cloud/storage/core/libs/diagnostics/stats_updater.h>
@@ -528,6 +529,16 @@ void TBootstrapVhost::StopComponents()
 
 void TBootstrapVhost::Drain()
 {
+    if (Configs->Options->Service == NDaemon::EServiceKind::Kikimr &&
+        GetShouldContinue().PollState() != TProgramShouldContinue::Continue)
+    {
+        auto code = GetShouldContinue().GetReturnCode();
+        if (code == NodeLeaseExpirationExitCode) {
+            ythrow TAppShouldExitWithoutShutdownException()
+                << "Node lease is expired";
+        }
+    }
+
     if (EndpointManager) {
         EndpointManager->Drain();
     }
--- a/cloud/filestore/libs/daemon/vhost/bootstrap.cpp
+++ b/cloud/filestore/libs/daemon/vhost/bootstrap.cpp
@@ -532,7 +532,7 @@ void TBootstrapVhost::Drain()
     if (Configs->Options->Service == NDaemon::EServiceKind::Kikimr &&
         GetShouldContinue().PollState() != TProgramShouldContinue::Continue)
     {
-        auto code = GetShouldContinue().GetReturnCode();
+        const int code = GetShouldContinue().GetReturnCode();
         if (code == NodeLeaseExpirationExitCode) {
             ythrow TAppShouldExitWithoutShutdownException()
                 << "Node lease is expired";
--- a/cloud/filestore/libs/daemon/vhost/bootstrap.h
+++ b/cloud/filestore/libs/daemon/vhost/bootstrap.h
@@ -18,6 +18,11 @@ namespace NCloud::NFileStore::NDaemon {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// NodeBroker lease expiration exit code
+constexpr int NodeLeaseExpirationExitCode = 2;
+
+////////////////////////////////////////////////////////////////////////////////
+
 struct TVhostModuleFactories
 {
     std::function<NVFS::IFileSystemLoopFactoryPtr(
--- a/cloud/filestore/libs/daemon/vhost/bootstrap_ut.cpp
+++ b/cloud/filestore/libs/daemon/vhost/bootstrap_ut.cpp
@@ -192,14 +192,14 @@ auto MakeSslServerCredentials()
 ////////////////////////////////////////////////////////////////////////////////
 
 // Needed to mock NodeBroker to allow our dynamic node to register
-class FakeYdbDiscoveryService final
+class TFakeYdbDiscoveryService final
     : public Ydb::Discovery::V1::DiscoveryService::Service
 {
 private:
     ui64 DynamicNodeIcPort;
 
 public:
-    FakeYdbDiscoveryService(ui64 dynamicNodeIcPort)
+    explicit TFakeYdbDiscoveryService(ui64 dynamicNodeIcPort)
         : DynamicNodeIcPort(dynamicNodeIcPort)
     {}
 
@@ -232,12 +232,12 @@ public:
     }
 };
 
-// Grpc server for FakeYdbDiscoveryService.
+// Grpc server for TFakeYdbDiscoveryService.
 // Needed to mock NodeBroker to allow our dynamic node to register
 class TestYdbDiscoveryServer
 {
 private:
-    FakeYdbDiscoveryService Service;
+    TFakeYdbDiscoveryService Service;
     std::unique_ptr<grpc::Server> Server;
 
     TPortManager PortManager;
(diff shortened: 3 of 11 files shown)


contrib/ydb/core/mind/node_broker_impl.h
```diff
@@ -75,7 +75,7 @@ class TNodeBroker : public TActor<TNodeBroker>
 private:
     using TActorBase = TActor<TNodeBroker>;
```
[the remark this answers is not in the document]
SvartMetal: This setting applies only to YDB built in our repo and is used ONLY for tests. 
It reduces the minimum possible lease duration, but does not change the default lease duration, which remains 2 hours. 
That is why I believe it is safe to keep this patch in our codebase, along with other patches from https://github.com/ydb-platform/nbs/issues/5654

cloud/filestore/libs/daemon/vhost/bootstrap.cpp
```diff
@@ -528,6 +529,16 @@ void TBootstrapVhost::StopComponents()
 
 void TBootstrapVhost::Drain()
 {
+    if (Configs->Options->Service == NDaemon::EServiceKind::Kikimr &&
+        GetShouldContinue().PollState() != TProgramShouldContinue::Continue)
+    {
+        auto code = GetShouldContinue().GetReturnCode();
```
qkrorlqr: nit: let's not use `auto` for primitive types

cloud/filestore/libs/daemon/vhost/bootstrap.h
```diff
@@ -18,6 +18,11 @@ namespace NCloud::NFileStore::NDaemon {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// NodeBroker lease expiration exit code
+constexpr int NodeLeaseExpirationExitCode = 2;
```
qkrorlqr, line 22: is it possible to include some ydb header where this code is defined? in order not to maintain a copy of this `2`

cloud/filestore/libs/daemon/vhost/bootstrap_ut.cpp
```diff
@@ -0,0 +1,436 @@
...
+namespace NCloud::NFileStore::NDaemon {
+
+using namespace NActors;
+using namespace NKikimr;
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+const TString SysConfigStr = R"(
```
qkrorlqr, line 37: nit: probably it will look better if we put these configs into separate file / files and use them via `RESOURCE()`

cloud/filestore/libs/daemon/vhost/bootstrap_ut.cpp
```diff
@@ -0,0 +1,436 @@
...
+    keyCert.private_key = ReadFile(certKeyFile);
+    opts.pem_key_cert_pairs.push_back(keyCert);
+
+    return grpc::SslServerCredentials(opts);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+// Needed to mock NodeBroker to allow our dynamic node to register
+class FakeYdbDiscoveryService final
```
qkrorlqr: nit: T

cloud/filestore/libs/daemon/vhost/bootstrap_ut.cpp
```diff
@@ -0,0 +1,436 @@
...
+    : public Ydb::Discovery::V1::DiscoveryService::Service
+{
+private:
+    ui64 DynamicNodeIcPort;
+
+public:
+    FakeYdbDiscoveryService(ui64 dynamicNodeIcPort)
```
qkrorlqr: explicit

cloud/filestore/tests/lease_expiration/test.py
```diff
@@ -0,0 +1,394 @@
...
+    def stop(self):
+        self.qemu.stop()
+
+
+@retry(stop_max_attempt_number=RETRY_COUNT, wait_fixed=WAIT_TIMEOUT_MS)
+def update_current_epoch_end(
+    ydb_cluster,
+    current_epoch_end_us,
+):
+    current_epoch_end_key = 5
```
qkrorlqr, line 128: is it possible to import this code (`5`) from some ydb .py module?

cloud/filestore/tests/lease_expiration/test.py
```diff
@@ -0,0 +1,394 @@
...
+    # is a special type of "message bus" call
+    assert YdbMessageBusStatus.MSTATUS_OK == response.Status
+
+
+@retry(stop_max_attempt_number=RETRY_COUNT, wait_fixed=WAIT_TIMEOUT_MS)
+def update_next_epoch_end(
+    ydb_cluster,
+    next_epoch_end_us,
+):
+    next_epoch_end_key = 6
```
qkrorlqr, line 153: and this one as well

cloud/storage/core/libs/daemon/app.h
```diff
@@ -45,6 +45,10 @@ int DoMain(TBootstrap& bootstrap, int argc, char** argv)
 
         try {
             bootstrap.Stop();
+        } catch (const TAppShouldExitWithoutShutdownException& ex) {
```
qkrorlqr, line 48: a hack =/

probably this code would look better w/o exceptions - if we just return exit codes from all the funcs that are called here - including bootstrap.Stop()

// can be cleaned up in a separate PR, all the `catch (...)` blocks that already exist in this code also look hacky

cloud/filestore/libs/daemon/vhost/bootstrap.h
SvartMetal: Unfortunately, no. The main source of truth is NNodeBroker::TLeaseHolder, where this exit code is hardcoded https://github.com/ydb-platform/nbs/blob/cf1a73afcd02efbb32ce4fb0a43d3db317e60785/contrib/ydb/core/mind/lease_holder.cpp#L233

But I have added a unit test for this exit code to reduce maintenance cost https://github.com/ydb-platform/nbs/pull/5372/changes#diff-e2f43f6be6e5ade693b77ac83dbc378984a45c13a608e0810b87ece0af8e48adR424

cloud/filestore/libs/daemon/vhost/bootstrap_ut.cpp
SvartMetal: Maybe, but I prefer to keep them in one place, especially template ones, as they are used directly in this file.

But if you insist, I will move them.

cloud/filestore/tests/lease_expiration/test.py
SvartMetal: Unfortunately, no. These are used only in C++ code https://github.com/ydb-platform/nbs/blob/cf1a73afcd02efbb32ce4fb0a43d3db317e60785/contrib/ydb/core/mind/node_broker_impl.h#L89

Good news: I do not expect these parameters to change, as they are stored in the DB and changing them would require a migration

cloud/filestore/tests/lease_expiration/test.py
SvartMetal: Answered here https://github.com/ydb-platform/nbs/pull/5372/changes#r3184283114

cloud/storage/core/libs/daemon/app.h
SvartMetal: Yes, I do not like this hack either =/
Alternative solutions:
* Call _exit directly in TBootstrapVhost::Drain - I do not like this because it is impossible to write a unit test for such code.
* Use exit codes - then we would need to define a special exit code for calling _exit, because reusing the lease expiration exit code also looks hacky. Returning exit codes from Stop also looks uglier because they conflict with TProgramShouldContinue return codes, which are the main source of truth now.
* A better approach, in my opinion, is to use an enum class with stop actions, something like this:
```
enum class EStopAction {
    ExitWithShutdown,
    ExitWithoutShutdown
};
```

SvartMetal: But maybe the best approach is to avoid using _exit at all and instead implement a proper way to cancel FUSE requests

qkrorlqr (dismissed):

SvartMetal:
image

Not relevant to this PR

qkrorlqr (approved):

yegorskii (approved):

Issue #3783, opened 2025-06-27
[NBS, Filestore] Support for cell-based architecture in Blockstore/Filestore,Disk Manager


yegorskii:
Currently, Blockstore and Filestore, when working with virtual machines, can only service those entities (disk devices, file systems) that were created by themselves. This forces clusters to be scaled by increasing the number of hosts. This approach works well until the cluster becomes too large and bottlenecks appear in components such as NodeBroker, DiskRegistry, Hive, ScheShard, etc. Furthermore, such clusters are less resilient to failures, for example, power outages. It is proposed to build smaller clusters (hereafter referred to as cells), but this creates a new issue: for example, disks attached to a virtual machine may end up being located in different cells. As part of this feature, it is proposed to solve this problem by using RDMA/GRPC for communication between cells.

PR #5373, opened 2026-03-03
fix get_cluster_capacity service private action

yegorskii:
#3783 
Changes for getclustercapacity private api call used in cells:
- added pipe disconnection handling
- switched to another api, because old one triggered heavy transaction in BSC
- changed the way we account YDB based resources

### DIFF
--- a/cloud/blockstore/libs/storage/service/service_actor_actions_get_cluster_capacity.cpp
+++ b/cloud/blockstore/libs/storage/service/service_actor_actions_get_cluster_capacity.cpp
@@ -46,9 +46,6 @@ class TGetClusterCapacityActor final
     : public TActorBootstrapped<TGetClusterCapacityActor>
 {
 private:
-    using TPoolToKindMapping =
-        THashMap<TString, NProto::EStorageMediaKind>;
-
     struct TStoragePoolStats
     {
         ui64 TotalBytes = 0;
@@ -99,17 +96,6 @@ private:
     void HandleEmptyClusterCapacity(
         const TActorContext& ctx,
         const TString& component);
-
-    const TPoolToKindMapping& GetYdbPoolToStorageMediaKind() const
-    {
-        static const TPoolToKindMapping PoolToKind
-        {
-            {"Type:ROT", NProto::STORAGE_MEDIA_HDD},
-            {"Type:SSD", NProto::STORAGE_MEDIA_SSD},
-        };
-
-        return PoolToKind;
-    }
 
 private:
     STFUNC(StateGetDiskRegistryBasedCapacity);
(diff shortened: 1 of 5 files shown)


BarkovBG (approved):

cloud/blockstore/libs/storage/core/volume_model.cpp
```diff
@@ -977,4 +900,71 @@ TVolumeParams ComputeVolumeParams(
     return volumeParams;
 }
 
+TPoolKinds GetPoolKinds(
```
qkrorlqr, line 903: I don't see any reason to move the code of this function. It creates an unneeded diff.

cloud/blockstore/libs/storage/service/service_actor_actions_get_cluster_capacity.cpp
```diff
@@ -26,12 +27,11 @@ LWTRACE_USING(BLOCKSTORE_STORAGE_PROVIDER)
 
 namespace {
 
-////////////////////////////////////////////////////////////////////////////////
-
-inline bool IsDynamicGroup(ui32 groupId)
+THashMap<TString, NProto::EStorageMediaKind> YdbPoolToStorageMediaKind
```
qkrorlqr: please don't use globals

especially non-const globals

it introduces issues with initialization/destruction order across the application

cloud/blockstore/libs/storage/service/service_actor_actions_get_cluster_capacity.cpp
```diff
@@ -52,6 +52,17 @@ class TGetClusterCapacityActor final
...
 private:
+    struct TStoragePoolStats
+    {
+        ui64 TotalBytes = 0;
+        ui64 FreeBytes = 0;
+    };
+
+    THashSet<TString> UsedPools;
+    THashMap<NProto::EStorageMediaKind, TStoragePoolStats> Stats;
+    THashMap<std::pair<ui64, ui64>, NProto::EStorageMediaKind> StorageIdToKind;
```
qkrorlqr: let's not use things like `std::pair<ui64, ui64>` - it's not clear at all what it is

cloud/blockstore/libs/storage/service/service_actor_actions_get_cluster_capacity.cpp
```diff
@@ -96,7 +125,25 @@ TGetClusterCapacityActor::TGetClusterCapacityActor(
         TStorageConfigPtr config)
     : RequestInfo(std::move(requestInfo))
     , Config(config)
-{}
+{
+    NProto::EStorageMediaKind kinds[] {
```
qkrorlqr: static const or constexpr?

cloud/blockstore/libs/storage/core/volume_model.cpp
yegorskii: Initially function was defined in anon namespace. Since now it is public I've just moved its body out of that namespace.

cloud/blockstore/libs/storage/service/service_actor_actions_get_cluster_capacity.cpp
yegorskii: Forgot const, agree. But did not understand statement about "issues with initialization/destruction", it is defined in anon namespace, so in our case should be initialized just before main() is called.

PR #5379, opened 2026-03-04
[Filestore] correct block in write check druing compaction load

neihar:


cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
```diff
@@ -80,9 +80,7 @@ void TIndexTabletActor::HandleWriteData(
                 b < range.FirstBlock() + range.BlockCount();
                 ++b)
         {
-            const auto rangeId = GetMixedRangeIndex(
-                msg->Record.GetNodeId(),
-                range.FirstBlock());
+            const auto rangeId = GetMixedRangeIndex(msg->Record.GetNodeId(), b);
```
qkrorlqr, line 83: Let's modify this test https://github.com/ydb-platform/nbs/blob/a9f3fd157fe874a417b5edf3ede8a5c1b06c6abe/cloud/filestore/libs/storage/tablet/tablet_ut_data.cpp#L3708 to use larger write requests to catch this bug

PR #5425, opened 2026-03-07
issue-5094: [Disk Manager] Add metrics for nfs client

jkuradobery:
Extracted nbs client's metrics to a separate metrics package, added interface and mocks for clientMetrics.
Used new metrics package in nfs metrics.
Added interface for nfs-go-sdk client Client, added mock for nfs client.
#5094

### DIFF
--- a/cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
+++ b/cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
@@ -764,6 +764,7 @@ func TestClientDestroyCheckpointWrappedError(t *testing.T) {
 	require.Error(t, err)
 	var clientErr *nfs_client.ClientError
 	require.ErrorAs(t, err, &clientErr)
+	require.ErrorIs(t, err, retriableError)
 
 	nfsMock.AssertExpectations(t)
 	metricsMock.AssertExpectations(t)


cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
```diff
@@ -0,0 +1,1285 @@
...
+
+func TestClientCreateFileStoreWrappedError(t *testing.T) {
+	ctx := context.Background()
+
+	metricsMock := client_metrics_mocks.NewMetricsMock()
+	metricsMock.On("StatRequest", "CreateFileStore").Return(func(err *error) {
+		require.Error(t, *err)
+		var clientErr *nfs_client.ClientError
+		require.ErrorAs(t, *err, &clientErr)
+		require.ErrorIs(t, *err, retriableError)
```
gy2411, line 632: Do we check exactly the same thing in lines 629-632 and in lines 646-649 (and similarly in other tests below)? Why do we need such double check?

cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
```diff
@@ -0,0 +1,1285 @@
...
+	nfsMock := nfs_client_mocks.NewClientInterfaceMock()
+	nfsMock.On("DestroyCheckpoint", mock.Anything, mock.Anything, mock.Anything).
+		Return(testNfsClientError)
+
+	c := newTestClientWithMetricsMock(nfsMock, metricsMock)
+
+	err := c.DestroyCheckpoint(ctx, "fs-1", "cp-1")
+	require.Error(t, err)
+	var clientErr *nfs_client.ClientError
+	require.ErrorAs(t, err, &clientErr)
```
gy2411, line 766: Why we don't check
`require.ErrorIs(t, *err, retriableError)`
here?

cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
```diff
@@ -0,0 +1,1285 @@
...
+
+func TestSessionCreateNodeNonRetriableError(t *testing.T) {
+	ctx := context.Background()
+
+	metricsMock := client_metrics_mocks.NewMetricsMock()
+	metricsMock.On("StatRequest", "CreateNode").Return(func(err *error) {
+		require.Error(t, *err)
+		var clientErr *nfs_client.ClientError
+		require.ErrorAs(t, *err, &clientErr)
+		require.NotErrorIs(t, *err, retriableError)
```
gy2411, line 1240: Why we don't check
`require.ErrorIs(t, err, testNfsClientNonRetriableError)`
here?

cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
jkuradobery: The return check is a sanity check. We need to check that the metrics receive the same error as the error returned.

cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
jkuradobery: Nice catch, TY

cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
jkuradobery: We check matching with clientErr, that is enough.

PR #5428, opened 2026-03-09
issue-5094: [Disk Manager] Fetch metrics from all  instances in tests

jkuradobery:
Some tests were unstable because they ran on several instances of disk-manager. Collect metrics from all instances.

#5094

### DIFF
--- a/cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_metrics_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_metrics_test.go
@@ -83,7 +83,6 @@ func TestNbsClientReportsMetrics(t *testing.T) {
 		"count",
 		map[string]string{"component": "nbs_client", "request": "Create"},
 	)
-	require.Len(t, createRequestCounters, 1)
 	require.Greater(t, createRequestCounters[0], float64(0))
 
 	require.Greater(t, testcommon.GetCountersDataplane(


cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_metrics_test.go
debnatkh: GetCountersControlplane?

cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_metrics_test.go
```diff
@@ -82,19 +82,19 @@ func TestNbsClientReportsMetrics(t *testing.T) {
 		t,
 		"count",
 		map[string]string{"component": "nbs_client", "request": "Create"},
-	), float64(0))
+	)[0], float64(0))
```
debnatkh: do you need assert that array size is 1?

cloud/disk_manager/test/recipe/__main__.py
```diff
@@ -374,10 +375,15 @@ def start(argv):
         append_recipe_err_files(
             ERR_LOG_FILE_NAMES_FILE, disk_manager.disk_manager.stderr_file_name
         )
+        controlplane_mon_ports.append(disk_manager.monitoring_port)
 
-    set_env("DISK_MANAGER_RECIPE_DISK_MANAGER_MON_PORT", str(disk_managers[0].monitoring_port))
+    set_env(
+        "DISK_MANAGER_RECIPE_DISK_MANAGER_MON_PORT",
+        ",".join([str(port) for port in controlplane_mon_ports]),
```
debnatkh: no need in [ ] ; map(str, controlplane_mon_ports)

cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_metrics_test.go
jkuradobery: No, we do not. The test is ran within a chunk, and it is expected that some tests might come before this test and create disks in another instances.

Issue #2293, opened 2024-10-16
[Filestore] should add blobs to index in parallel while writing them


SvartMetal:
The idea is similar to the one from #122

neihar:
Main task is done. Possible improvements for different scenarios created as separate issues.

PR #5459, opened 2026-03-11
[Filestore] add new requests and adjust old API

neihar:
this needs for later usage in filestore feature #2293


cloud/filestore/libs/storage/tablet/tablet_actor_adddata.cpp
```diff
@@ -592,4 +592,14 @@ void TIndexTabletActor::HandleAddDataCompleted(
     EnqueueBlobIndexOpIfNeeded(ctx);
 }
 
+void TIndexTabletActor::HandleConfirmAddData(
+    const TEvIndexTablet::TEvConfirmAddDataRequest::TPtr&,
+    const TActorContext&)
+{}
```
qkrorlqr, line 598: let's reply with E_NOT_IMPLEMENTED

PR #5487, opened 2026-03-14
issue-2674: E_FS_ISDIR/E_FS_NOTEMPTY/E_FS_NOENT/E_FS_EXIST errors received from shard upon RenameNodeInDestination shouldn't trigger crit events

qkrorlqr:
### Notes
These errors may happen during normal operation - the client can issue a command which would trigger them. So we shouldn't raise crit events for these errors.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
@@ -123,8 +123,9 @@ void TRenameNodeInDestinationActor::HandleRenameNodeInDestinationResponse(
         }
 
         const auto message = Sprintf(
-            "RenameNodeInDestination failed for %s, %lu, %s with error %s"
+            "%s RenameNodeInDestination failed for %s, %lu, %s with error %s"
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_renamenode_source.cpp
```diff
@@ -130,14 +130,35 @@ void TRenameNodeInDestinationActor::HandleRenameNodeInDestinationResponse(
...
+            // Other non-retriable errors are not expected.
+            //
 
-        ReportReceivedNodeOpErrorFromShard(message);
+            LOG_ERROR(
+                ctx,
+                TFileStoreComponents::TABLET_WORKER,
+                "%s %s",
+                LogTag.c_str(),
+                message.c_str());
```
SvartMetal, line 155: We can include LogTag in the message - it will be printed when reporting a critical event

https://github.com/ydb-platform/nbs/blob/16214c8f3497712d9bf07ba248ffdea02c06aceb/cloud/storage/core/libs/diagnostics/critical_events.cpp#L85

PR #5489, opened 2026-03-15
issue-5094: [Disk Manager] Implement regular fs scrubbing

jkuradobery:
Implement task to scrubb a subset of filesystems continuously (the filesystem MUST always be scrubbed).
The following invariants must be maintained: 
1) filesystem is scrubbed by a single scrubbing task all the time.
2) Crash of a single scrubbing task must not affect other filesystems scrubbing.
3) Filesystem scrubbing must be always restarted even in case of a failure. 
#5094

### DIFF
--- a/cloud/disk_manager/internal/pkg/facade/testcommon/common.go
+++ b/cloud/disk_manager/internal/pkg/facade/testcommon/common.go
@@ -760,6 +760,7 @@ func GetTaskMetadata(
 	ctx context.Context,
 	taskID string,
 ) proto.Message {
+
 	scheduler, err := newScheduler(ctx)
 	require.NoError(t, err)
 


jkuradobery:
On the second thought, I think that adding re-scheduling capability to the scrubbing task itself, would make more simplicity-wise.

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
```diff
@@ -749,6 +755,19 @@ func WaitOperationEnded(
 	require.NoError(t, err)
 }
 
+func GetTaskMetadata(
+	t *testing.T,
+	ctx context.Context,
+	taskID string,
+) proto.Message {
```
gy2411, line 762: Blank line

cloud/disk_manager/internal/pkg/dataplane/filesystem/scrubbing/regular_scrub_filesystems_task.go
```diff
@@ -0,0 +1,146 @@
...
+	err = execCtx.SaveState(ctx)
+	if err != nil {
+		return err
+	}
+
+	if egErr != nil {
+		return egErr
+	}
+
+	return errors.NewInterruptExecutionError()
```
gy2411, line 124: If some of ScrubFilesystem tasks fails, how we know about that? Is it entirely up to monitoring?

cloud/disk_manager/internal/pkg/dataplane/filesystem/scrubbing/regular_scrub_filesystems_task.go
```diff
@@ -0,0 +1,146 @@
...
+	t.state = &scrubbing_protos.RegularScrubFilesystemsTaskState{}
+	return proto.Unmarshal(state, t.state)
+}
+
+func (t *regularScrubFilesystemsTask) idempotencyKey(
+	filesystemID string,
+	generation uint64,
+) string {
+
+	return fmt.Sprintf("%v_%v", filesystemID, generation)
```
gy2411, line 41: Should we also use task id in idempotancy key?

Now, if RegularScrubFilesystemsTask fails, the next RegularScrubFilesystemsTask will check already ended 
ScrubFilesystem tasks until its generations grow big enough. Is it ok for us?

On the other hand, if we use task id in idempotancy key, we might end up creating uncontrollably more and more ScrubFilesystem tasks since they are not cancelled on RegularScrubFilesystemsTask cancellation...

cloud/disk_manager/internal/pkg/dataplane/filesystem/scrubbing/regular_scrub_filesystems_task.go
jkuradobery: From monitoring. We want to monitor several file systems, and we do not want the monitoring to be interrupted if one of the scrubbing's fail.

cloud/disk_manager/internal/pkg/dataplane/filesystem/scrubbing/regular_scrub_filesystems_task.go
jkuradobery: If something goes wrong with regular scrubbing, e.g. YDB flap or sth else, we would like to avoid instantiating new tasks per filesystem.

PR #5493, opened 2026-03-16
issue-1751: [Filestore] Introduce barriers to WriteBackCache

e673:
### Notes

There are operations that should not interfere with write-back cache, for example:
- SetAttr with node size specified;
- ReadData/WriteData for handles with O_DIRECT;
- WriteData with O_SYNC, O_DSYNC flags.

Need to implement a way to empty cache before executing these operations and prevent new writes to be flushed.

Proposed implementation:
```
NThreading::TFuture<TResultOrError<ui64>> AcquireBarrier(ui64 nodeId);
void ReleaseBarrier(ui64 nodeId, ui64 barrierId);
```

Acquiring a barrier ensures that all prior WriteData requests are flushed and evicted from cache and no new WriteData requests will be flushed until the barrier is released.

A error may be returned if flush fails - in this case, a barrier will not be acquired.

In other word, barrier BarrierId guarantees operation sequencing:
- All cached writes with SequenceId < BarrierId are synchronized with filestore.
- No cached writes with SequenceId > BarrierId will be sent to filestore.

Barriers should not prevent from concurrent execution of non-cached operations, it only affects interleaving between cached and non-cached operations.

### Issue

### Linux behavior

1. All read/write requests are sequenced, newest requests should observe the effects of completed previous requests no matter of whether O_DIRECT flag is set or not.
2. Flag O_DIRECT only instructs that the data cannot be read from or written to a cache:
* Direct read: the cached data is flushed then the request is processed:
https://github.com/torvalds/linux/blob/30f09200cc4aefbd8385b01e41bde2e4565a6f0e/mm/filemap.c#L2907
* Direct write: the cache is invalidated then the request is processed:
https://github.com/torvalds/linux/blob/30f09200cc4aefbd8385b01e41bde2e4565a6f0e/mm/filemap.c#L4181
3. For FUSE:
* Direct reads trigger flushing write-back cache:
https://github.com/torvalds/linux/blob/c875a6c3246713a018d8b7b143deb3ebaeecaf1c/fs/fuse/file.c#L1677

Note: O_DIRECT flag is a non-POSIX feature so we should follow Linux behavior.

### Current behavior in filestore

* Direct read: ignore O_DIRECT flag and read from cache. The read data is consistent but this breaks rule 2.

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
@@ -167,27 +167,6 @@ public:
 
     // Used to adjust node size according to cached data
     ui64 GetMaxWrittenOffset(ui64 nodeId) const;
-
-    // Used to clear max written offset in SetNodeAttr handler
-    // Note: a barrier should be acquired via AcquireBarrier
-    void ResetMaxWrittenOffset(ui64 nodeId);
-
-    /* Acquiring a barrier ensures that all prior WriteData requests are flushed
-     * and evicted from cache and no new WriteData requests will be flushed
-     * until the barrier is released.
-     *
-     * Used to execute requests that should not interfere with cache:
-     * SetNodeAttr, ReadData/WriteData with O_DIRECT/O_SYNC/O_DSYNC.
-     *
-     * On success, returns barrierId - it should be released by ReleaseBarrier.
-     *
-     * An error may be returned if flush fails - in this case, a barrier will
-     * not be acquired.
-     */
-    NThreading::TFuture<TResultOrError<ui64>> AcquireBarrier(ui64 nodeId);
-
-    // Releases a barrier previously acquired by AcquireBarrier
-    void ReleaseBarrier(ui64 nodeId, ui64 barrierId);
 };
 
 }   // namespace NCloud::NFileStore::NFuse
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
@@ -353,21 +353,6 @@ public:
         return State.GetMaxWrittenOffset(nodeId);
     }
 
-    void ResetMaxWrittenOffset(ui64 nodeId)
-    {
-        State.ResetMaxWrittenOffset(nodeId);
-    }
-
-    NThreading::TFuture<TResultOrError<ui64>> AcquireBarrier(ui64 nodeId)
-    {
-        return State.AcquireBarrier(nodeId);
-    }
-
-    void ReleaseBarrier(ui64 nodeId, ui64 barrierId)
-    {
-        State.ReleaseBarrier(nodeId, barrierId);
-    }
-
 private:
     // Implementation of IQueuedOperationsProcessor
     void ScheduleFlushNode(ui64 nodeId) override
@@ -626,22 +611,6 @@ void TWriteBackCache::ReleaseNodeStateRef(ui64 refId)
 ui64 TWriteBackCache::GetMaxWrittenOffset(ui64 nodeId) const
 {
     return Impl->GetMaxWrittenOffset(nodeId);
-}
-
-void TWriteBackCache::ResetMaxWrittenOffset(ui64 nodeId)
-{
-    Impl->ResetMaxWrittenOffset(nodeId);
-}
-
-NThreading::TFuture<TResultOrError<ui64>> TWriteBackCache::AcquireBarrier(
-    ui64 nodeId)
-{
-    return Impl->AcquireBarrier(nodeId);
-}
-
-void TWriteBackCache::ReleaseBarrier(ui64 nodeId, ui64 barrierId)
-{
-    Impl->ReleaseBarrier(nodeId, barrierId);
 }
 
 }   // namespace NCloud::NFileStore::NFuse
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state_ut.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state_ut.cpp
@@ -133,6 +133,31 @@ struct TBootstrap
             });
         return out;
     }
+
+    ui64 AcquireBarrier(ui64 nodeId) const
+    {
+        auto future = State->AcquireBarrier(nodeId);
+        while (!future.HasValue()) {
+            State->FlushSucceeded(nodeId, 1);
+        }
+        const auto& result = future.GetValue();
+        UNIT_ASSERT(!HasError(result.GetError()));
+        return result.GetResult();
+    }
+
+    void ReleaseBarrier(ui64 nodeId, ui64 barrierId) const
+    {
+        State->ReleaseBarrier(nodeId, barrierId);
+    }
+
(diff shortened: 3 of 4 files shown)


cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
```diff
@@ -115,8 +125,17 @@ struct TNodeState
         if (!Cache.HasUnflushedRequests()) {
             return ENodeFlushStatus::NothingToFlush;
         }
+
+        const ui64 minUnflushedSequenceId = Cache.GetMinUnflushedSequenceId();
+
+        if (!Barriers.empty()) {
+            return minUnflushedSequenceId < Barriers.cbegin()->first
+                       ? ENodeFlushStatus::FlushRequested
```
SvartMetal, line 138: Let's add a comment on why should we change the previous behavior for data which is to the left from the barrier

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -431,17 +507,53 @@ void TWriteBackCacheState::EvictUnpinnedFlushedEntries(
...
+
+void TWriteBackCacheState::TriggerBarrierAcquisitions(TNodeState& nodeState)
+{
+    if (nodeState.Barriers.empty()) {
+        return;
     }
 
-    if (!entriesDeleted) {
+    auto it = nodeState.Barriers.begin();
+    if (it->second.IsAcquired) {
```
SvartMetal, line 537: Let's add a comment about expectations this if is based on

cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.cpp
```diff
@@ -118,6 +119,34 @@ bool TNodeCache::Empty() const
...
+    if (!FlushedRequests.empty()) {
+        return FlushedRequests.front()->GetSequenceId();
+    }
+    if (!UnflushedRequests.empty()) {
+        return UnflushedRequests.front()->GetSequenceId();
+    }
+    if (!PendingRequests.empty()) {
+        return PendingRequests.front()->GetSequenceId();
+    }
+    Y_ABORT("Cache is empty");
```
SvartMetal: It is better not to crash here

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
```diff
@@ -122,6 +126,15 @@ class TWriteBackCacheState
...
         const NCloud::NProto::TError& error);
 
+    // Barriers enforce sequencing and allow execution of operations without
+    // interfering with cache (such as SetNodeAttr or ReadData/WriteData with
+    // O_DIRECT/O_SYNC/O_DSYNC)
+    // Successfully acquired barrier ensures that:
+    // - all prior WriteData requests are flushed and evicted;
+    // - no flush will take place until the barrier is released.
+    NThreading::TFuture<TResultOrError<ui64>> AcquireBarrier(ui64 nodeId);
+    void ReleaseBarrier(ui64 nodeId, ui64 barrierId);
```
SvartMetal, line 139: Let's add a comment that barrierId should be valid and previously acquired

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -563,48 +618,71 @@ void TFileSystem::DoWrite(
...
+                    auto callbackWithReleaseBarrier =
+                        [ptr = std::move(ptr),
+                         ino = request->GetNodeId(),
+                         callback = std::move(callback),
+                         barrierId = result.GetResult()](const auto& future)
+                    {
+                        callback(future);
+
+                        if (auto self = ptr.lock()) {
+                            self->WriteBackCache.ResetMaxWrittenOffset(ino);
```
SvartMetal: Shouldn't we move this call inside of ReleaseBarrier?

Caller code would be simpler, IMO

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -484,7 +478,68 @@ void TFileSystem::Read(
...
+        case EServerWriteBackCacheState::Draining: {
+            // Read data directly from the underlying storage.
+            // We need to acquire a barrier that ensures:
+            // - all cached writes are flushed;
+            // - no new cached writes will be executed until the barrier is
+            // released
+            auto acquireBarrierFuture =
+                WriteBackCache.AcquireBarrier(request->GetNodeId());
+
+            acquireBarrierFuture.Subscribe(
```
qkrorlqr: I would suggest to log acquire/release barrier events as separate events into `vhost-profile-log.bin`

can be done in a separate PR of course

qkrorlqr: And let's add a doc about these barriers and O_SYNC / O_DIRECT behaviour

cloud/filestore/libs/vfs_fuse/write_back_cache/node_state.h
e673: Comment added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Comment added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.h
e673: Added

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: I've massively refactored and deduplicated the code and moved the logic to WriteBackCache.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
```diff
@@ -136,14 +136,58 @@ class TWriteBackCache final
...
     void ReleaseNodeStateRef(ui64 refId);
 
-    ui64 GetCachedNodeSize(ui64 nodeId) const;
-    void SetCachedNodeSize(ui64 nodeId, ui64 size);
+    // Used to adjust node size according to cached data
+    ui64 GetMaxWrittenOffset(ui64 nodeId) const;
+
+    // Used to clear max written offset in SetNodeAttr handler
+    // Note: a barrier should be acquired via AcquireBarrier
+    void ResetMaxWrittenOffset(ui64 nodeId);
```
SvartMetal: Should it be public?

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: Added ToDo to documentation PR
https://github.com/ydb-platform/nbs/pull/5168#issuecomment-4090777313

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.h
e673: I've removed methods that are no more used in public and moved corresponding tests.

PR #5505, opened 2026-03-17
Added TVolumeShapingThrottler that throttles responses

komarevtsev-d:
### Notes

Introduce `TVolumeShapingThrottler` along with its documentation.
TL;DR:
Allow to throttle the response from the volume to the client. Throttler parameters are set in the storage config. The shaper takes the real `executionTime` of the completed request and calculates the `cost` of a request based on config parameters. If `cost > executionTime`  then some part of the difference (from 0% to 100%) becomes shaping throttler delay.

### Issue
#5095

### DIFF
--- a/cloud/blockstore/libs/storage/partition_common/model/ya.make
+++ b/cloud/blockstore/libs/storage/partition_common/model/ya.make
@@ -18,7 +18,3 @@ PEERDIR(
 )
 
 END()
-
-RECURSE_FOR_TESTS(
-    ut
-)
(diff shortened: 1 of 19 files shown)


cloud/blockstore/libs/storage/partition_common/model/ya.make
```diff
@@ -18,7 +18,3 @@ PEERDIR(
 )
 
 END()
-
```
drbasic, line 21: зачем?

komarevtsev-d: ya make ругается из-за того что ut не существует

Issue #5507, opened 2026-03-17
[Filestore] Directory structure viewer on tablet monpage


qkrorlqr:
After https://github.com/ydb-platform/nbs/issues/2674 the Node and NodeRef distribution across the shards got even more complex - it would be nice to have a visualization of this structure from which it would be clear which Nodes are managed by which shards.

PR #5509, opened 2026-03-17
issue-5507: filestore directory structure viewer

qkrorlqr:
### Notes
Screenshot from 2026-03-21 17-10-56

Will add some automatic smoke tests in a separate PR.

Made a small modification in the tablet's `ListNodes` handler - not returning an error if there's no session.

`nextCookie` looks like this:
Screenshot from 2026-03-21 17-16-05

### Issue
https://github.com/ydb-platform/nbs/issues/5507

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring_directory_viewer.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring_directory_viewer.cpp
@@ -150,6 +150,12 @@ TString JsonError(const NProto::TError& e)
 
 ////////////////////////////////////////////////////////////////////////////////
 
+/*
+ * This code is similar to the TListNodesActor code in storage service but it's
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/storage/tablet/resources/js/dir-viewer.js
```diff
@@ -0,0 +1,275 @@
+function render() {
+    // Palette for shard coloring - up to 10 distinct shards
```
neihar, line 2: up to 10 distinct shards colors

qkrorlqr: ?

neihar: it reads like we can support only 10 distinct shards, while we support 10 distinct colors for them

PR #5398, opened 2026-03-05
[CSI] Support multiple nfs-vhost clients

antonmyagkov:
issue: https://github.com/ydb-platform/nbs/issues/5362

Add support for multiple NFS host processes in the CSI driver.
When handling NodeStageVolume, the CSI driver selects one of the available NFS vhost sockets based on a hash of the instance ID. Selected nfs vhost index is saved on the disk in the stage data and can be read again to stop endpoint or to start endpoint after kubelet restart.

The CSI driver initializes NfsVhostReplicaCount NFS clients. However, only the first NfsVhostReplicaCountOnNode clients are used to create new endpoints.
To scale down the number of nfs-vhost instances on a node, reduce the value of NfsVhostReplicaCountOnNode. After updating this value, restart all virtual machines that have endpoints attached to nfs-vhost instances with indexes greater than NfsVhostReplicaCountOnNode so that their endpoints are recreated on the remaining instances.

PR #5516, opened 2026-03-18
[CSI] Support multiple blockstore servers in VM mode

antonmyagkov:
issue: https://github.com/ydb-platform/nbs/issues/5527

Add support for multiple blockstore-server processes in the CSI driver.
When handling NodeStageVolume, the CSI driver selects one of the available blockstore-server sockets based on a hash of the instance ID. Selected blockstore-server index is saved on the disk in the stage data and can be read again to stop endpoint or to start endpoint after kubelet restart.

The CSI driver initializes NbsServerReplicaCountTotal NBS clients. However, only the first NbsServerReplicaCountUsed clients are used to create new endpoints.
To scale down the number of blockstore-server instances on a node, reduce the value of NbsServerReplicaCountTotal. After updating this value, restart all virtual machines that have endpoints attached to blockstore-server instances with indexes greater than NbsServerReplicaCountUsed so that their endpoints are recreated on the remaining instances.

nbd mode and legacy VM mode are not supported multiple blockstore servers

### DIFF
--- a/cloud/blockstore/tools/csi_driver/internal/driver/node.go
+++ b/cloud/blockstore/tools/csi_driver/internal/driver/node.go
@@ -889,7 +889,9 @@ func (s *nodeService) nodeStageDiskAsFilesystem(
 	err = nil
 	defer func() {
 		if err != nil {
-			s.cleanupNbsEndpoint(ctx, instanceId, diskId, 0)
+			// Use index 0, as multiple blockstore servers are supported only in VM mode.
+			clientIndex := uint(0)
+			s.cleanupNbsEndpoint(ctx, instanceId, diskId, clientIndex)
 		}
 	}()
 
@@ -964,7 +966,9 @@ func (s *nodeService) nodeStageDiskAsBlockDevice(
 	err = s.mountBlockDevice(diskId, resp.NbdDeviceFile, devicePath, false)
 
 	if err != nil {
-		s.cleanupNbsEndpoint(ctx, instanceId, diskId, 0)
+		// Use index 0, as multiple blockstore servers are supported only in VM mode.
+		clientIndex := uint(0)
+		s.cleanupNbsEndpoint(ctx, instanceId, diskId, clientIndex)
 	}
 	return err
 }
@@ -1860,8 +1864,7 @@ func (s *nodeService) NodeGetVolumeStats(
 		return nil, fmt.Errorf("NodeGetVolumeStats is not supported in vmMode")
 	}
 
-	nbsClient := s.getNbsClient(0)
-	if nbsClient == nil {
+	if len(s.nbsClients) == 0 {
 		return nil, fmt.Errorf("NBS client is not available")
 	}
 


cloud/blockstore/tools/csi_driver/internal/driver/driver.go
```diff
@@ -139,19 +154,18 @@ func createClients(cfg Config) (*driverClients, error) {
...
-				nfsEndpointClient, err := nfsclient.NewGrpcEndpointClient(
-					&nfsclient.GrpcClientOpts{
-						Endpoint: getEndpoint(nfsVhostSocket, cfg.NfsVhostHost, cfg.NfsVhostPort),
-					}, nfsclient.NewStderrLog(nfsclient.LOG_DEBUG),
-				)
-				if err != nil {
-					return nil, err
-				}
-				nfsEndpointClients = append(nfsEndpointClients, nfsEndpointClient)
+		for i := 0; i < int(cfg.NfsVhostReplicaCountTotal); i++ {
```
antonmyagkov, line 157: I forgot to pass correct nfs vhost port in https://github.com/ydb-platform/nbs/pull/5398
Fixed here as the same logic is applied for blockstore client too

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -306,16 +310,18 @@ func (s *nodeService) NodeStageVolume(
...
 					}
 					stageData = &StageData{
-						Backend:        backend,
-						InstanceId:     instanceId,
-						RealStagePath:  s.getEndpointDir(instanceId, nbsId),
-						NfsClientIndex: nfsClientIndex,
+						Backend:       backend,
+						InstanceId:    instanceId,
+						RealStagePath: s.getEndpointDir(instanceId, nbsId),
+						ClientIndex:   clientIndex,
```
antonmyagkov, line 324: generalize StageData for nfs/nbs backend and rename NfsClientIndex to ClientIndex. 
It's possible to do it as we don't use this feature yet.

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -625,7 +632,9 @@ func (s *nodeService) nodePublishDiskAsVhostSocket(
 			HostType: &hostType,
 		},
 	}
-	_, err := s.nbsClient.StartEndpoint(ctx,
+
+	nbsClient := s.getNbsClient(0)
```
antonmyagkov, line 637: We don't support multiple clients in nbd(infrakuber)
 mode, so I always pass 0 index.

cloud/blockstore/tools/csi_driver/internal/driver/node.go
tpashkin: > It's possible to do it as we don't use this feature yet

Can you expand on that? If this version of csi driver is deployed, it would have already saved stage data with this field, right?

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -1016,15 +1028,20 @@ func (s *nodeService) startNbsEndpointForNBD(
 			HostType: &hostType,
 		},
 	}
-	resp, err := s.nbsClient.StartEndpoint(ctx,
-		s.resolveEndpoint(ctx, startEndpointRequest))
+
+	nbsClient := s.getNbsClient(0)
```
tpashkin, line 1038: Comment explaining that multiple servers aren't supported in this mode would be welcome

tpashkin: Why don't we support it though? Doesn't seem to be all that different from vhost

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -329,15 +335,16 @@ func (s *nodeService) NodeStageVolume(
...
 				} else {
 					err = s.nodeStageDiskAsVhostSocket(
 						ctx,
 						instanceId,
 						nbsId,
 						req.VolumeContext,
 						req.VolumeCapability.GetMount(),
-						vhostSettings)
+						vhostSettings,
+						stageData.ClientIndex)
```
tpashkin, line 347: If it's an old disk/filesystem, wouldn't this field be missing from stage data?

cloud/blockstore/tools/csi_driver/internal/driver/node.go
tpashkin: Comment explaining this would be welcome. Why we don't support this though?

cloud/blockstore/tools/csi_driver/internal/driver/node.go
antonmyagkov: default value is 0 which is a correct index for already mounted disks/filesystems

cloud/blockstore/tools/csi_driver/internal/driver/node.go
antonmyagkov: We didn’t deploy this version, but even if it were deployed, the parser would ignore excess fields.

cloud/blockstore/tools/csi_driver/internal/driver/node.go
antonmyagkov: I didn't implement it as we store StageData metadata only in vm mode.

antonmyagkov: I've added comments

cloud/blockstore/tools/csi_driver/internal/driver/node.go
antonmyagkov: done. No metadata(StageData) in nbd mode and it's not required yet.

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -876,7 +889,7 @@ func (s *nodeService) nodeStageDiskAsFilesystem(
 	err = nil
 	defer func() {
 		if err != nil {
-			s.cleanupNbsEndpoint(ctx, instanceId, diskId)
+			s.cleanupNbsEndpoint(ctx, instanceId, diskId, 0)
```
SvartMetal: Let’s add a comment explaining why we pass zero here

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -951,7 +964,7 @@ func (s *nodeService) nodeStageDiskAsBlockDevice(
 	err = s.mountBlockDevice(diskId, resp.NbdDeviceFile, devicePath, false)
 
 	if err != nil {
-		s.cleanupNbsEndpoint(ctx, instanceId, diskId)
+		s.cleanupNbsEndpoint(ctx, instanceId, diskId, 0)
```
SvartMetal: Let’s add a comment explaining why we pass zero here

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -1826,7 +1860,8 @@ func (s *nodeService) NodeGetVolumeStats(
 		return nil, fmt.Errorf("NodeGetVolumeStats is not supported in vmMode")
 	}
 
-	if s.nbsClient == nil {
+	nbsClient := s.getNbsClient(0)
```
SvartMetal: Let’s add a comment explaining why we pass zero here

antonmyagkov: I've changed condition here to check that at least one nbs client is available

PR #5529, opened 2026-03-19
fix rdma server error handling; issue-5433; issue-3375

tpashkin:
This PR is based on https://github.com/ydb-platform/nbs/pull/5522

- Move completion error handling to ValidateRequest
- Properly adjust counters and inflight in case of an error
- Increment error counters in most of the catch blocks
- Set up server error handling test

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/server.cpp
+++ b/cloud/blockstore/libs/rdma/impl/server.cpp
@@ -121,6 +121,7 @@ struct TEndpointCounters
     TDynamicCounters::TCounterPtr QueuedRequests;
     TDynamicCounters::TCounterPtr ActiveRequests;
     TDynamicCounters::TCounterPtr CompletedRequests;
+    TDynamicCounters::TCounterPtr AbortedRequests;
     TDynamicCounters::TCounterPtr ThrottledRequests;
 
     TDynamicCounters::TCounterPtr ActiveSend;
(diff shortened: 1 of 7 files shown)


cloud/blockstore/libs/rdma/impl/server.cpp
```diff
@@ -849,8 +889,18 @@ void TServerSession::ReadRequestData(TRequestPtr req, TSendWr* send)
     wr.wr.rdma.rkey = req->In.Key;
     wr.wr.rdma.remote_addr = req->In.Address;
 
-    RDMA_TRACE("READ " << TWorkRequestId(wr.wr_id));
-    Verbs->PostSend(Connection->qp, &wr);
+    try {
+        Verbs->PostSend(Connection->qp, &wr);
+        RDMA_TRACE("READ " << TWorkRequestId(wr.wr_id) << " posted");
+
```
budevg, line 895: remove extra line

tpashkin: are you sure? extra line is consistent with all other try-catch blocks in this file

tpashkin: i'd very much not use it, but our code style requires catch/else to be on the same line as closing curly brace. so there is no other way to visually separate try and catch blocks

budevg: Right it's the only place we use this convention. Other places in libs don't put extra newline before catch. Anyway I don't think it's important to fix this in this PR .

PR #5537, opened 2026-03-19
Allow to use ydbd built from current source ( -DUSE_BUNDLED_YDBD=1 )

proller:
https://github.com/ydb-platform/nbs/issues/6645

Change in contrib/ydb/apps/ydbd/ya.make can be rewritten by next ydb code sync


contrib/ydb/apps/ydbd/ya.make
```diff
@@ -19,6 +19,10 @@ IF (OS_WINDOWS)
     )
 ENDIF()
 
+SRCDIR(
+     contrib/ydb/apps/ydbd
+)
```
SvartMetal, line 24: We try to avoid patching YDB code directly. Have you considered submitting a PR to YDB?

cloud/storage/core/tools/testing/ydb/bin/ya.make
```diff
@@ -1,10 +1,14 @@
-PACKAGE()
+IF(USE_BUNDLED_YDBD)
+    INCLUDE(${ARCADIA_ROOT}/contrib/ydb/apps/ydbd/ya.make)
```
SvartMetal, line 4: Why do we need this? 

Maybe it would be better for the package to depend directly on YDB sources instead of using `cloud/storage/core/tools/testing/ydb/bin` with `USE_BUNDLED_YDBD=1`?

SvartMetal: OK, we need it on ARM in order to make already existing targets that depend on `cloud/storage/core/tools/testing/ydb/bin` to build ydbd from sources instead of using prebuilt binary

SvartMetal (dismissed):

cloud/storage/core/tools/testing/ydb/bin/ya.make
```diff
@@ -18,4 +23,5 @@ ELSE()
     MESSAGE(FATAL_ERROR "Unsupported platform")
 ENDIF()
 
-END()
+    END()
```
SvartMetal: ?

PR #5539, opened 2026-03-20
[Disk Manager] Fix leak of base disk slots

gy2411:
In https://github.com/ydb-platform/nbs/pull/4339 there was a bug. This bug leads to a leaking base disk slots: the overlay disk is deleted, but its slot remains in the database.

Scenario:
- Overlay disk is creating
- Concurrenlty, this overlay disk is deleting
- Disk is deleted successfully
- CreateOverlayDiskTask successfully acquires base disk slot
- CreateOverlayDiskTask sees that disk is deleted and starts to cancel
- On cancel, CreateOverlayDiskTask sees that the disk is deleted and does not proceed with releasing base disk

Leaking slots result in hanging RetireBaseDisk task: the base disk with leaking slot will never retire.

---

In this pr we make the following changes:
- Fix race condition in acquire/release slot. On ReleaseBaseDisk request, we add tombstone slot to the database if there. On AcquireBaseDisk request, we don't acquire a new slot if there is a tombstone.
- Improve the test on concurrent creation/deletion using CheckBaseDiskSlotReleased.
- Increase DeletedDiskExpirationTimeout and ReleasedSlotExpirationTimeout parameters in test config. Otherwise, the deleted disk / released slots gets cleaned up from the database too quickly and the test does not work.

Without the fix, the improve test fails with error:
```
common.go:711: 
                Error Trace:    /-S/cloud/disk_manager/internal/pkg/facade/testcommon/common.go:711
                                                        /-S/cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_test.go:1273
                Error:          Received unexpected error:
                                Non retriable error, Silent=false: internal inconsistency: slot &{overlayDiskID:TestDiskServiceDeleteDiskWhenCreationIsInFlightAAAA overlayDiskKind:0 overlayDiskSize:134217728 baseDiskID:base-6c1b5f58-9f2f-44fb-8f27-6856e31de178 imageID:TestDiskServiceDeleteDiskWhenCreationIsInFlightAAAA zoneID:zone-a allottedSlots:1 allottedUnits:1 releasedAt:{wall:0 ext:62135596800 loc:0x1fe8200} targetZoneID: targetBaseDiskID: targetAllottedSlots:0 targetAllottedUnits:0 generation:0 status:0} should be released
```

### DIFF
--- a/cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
+++ b/cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
@@ -1170,13 +1170,13 @@ func (s *storageYDB) releaseBaseDiskSlot(
 			releasedAt:    time.Now(),
 		}
 
-		err = s.updateSlots(
+		err = s.updateSlot(
 			ctx,
 			tx,
-			[]slotTransition{slotTransition{
+			slotTransition{
 				oldState: nil,
 				state:    &slotTombstone,
-			}},
+			},
 		)
 		if err != nil {
 			return BaseDisk{}, err
@@ -1184,8 +1184,9 @@ func (s *storageYDB) releaseBaseDiskSlot(
 
 		logging.Info(
 			ctx,
-			"created slot tombstone %+v",
+			"Created tombstone %+v while releasing base disk slot in zone %v",
 			slotTombstone,
+			overlayDisk.ZoneId,
 		)
 
 		return BaseDisk{}, tx.Commit(ctx)
--- a/cloud/tasks/storage/storage_ydb_impl.go
+++ b/cloud/tasks/storage/storage_ydb_impl.go
@@ -2177,10 +2177,10 @@ func (s *storageYDB) isTaskEnded(
 	taskID string,
 ) (bool, error) {
 
-	taskState, err := s.getTask(ctx, session, taskID)
+	state, err := s.getTask(ctx, session, taskID)
 	if err != nil {
 		return false, err
 	}
 
-	return IsEnded(taskState.Status), nil
+	return IsEnded(state.Status), nil
 }


cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_test.go
```diff
@@ -239,9 +239,6 @@ func TestDiskServiceDeleteDiskWhenCreationIsInFlight(t *testing.T) {
 		true, // pooled
 	)
 
-	// Need to add some variance for better testing.
-	common.WaitForRandomDuration(time.Millisecond, 2*time.Second)
```
gy2411, line 243: Moved this wait a ltltle bit lower. It is strange what this wait was here.

cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -187,7 +187,7 @@
     ScheduleCollectClusterCapacityTask: false
 >
 DisksConfig: <
-    DeletedDiskExpirationTimeout: "1s"
+    DeletedDiskExpirationTimeout: "30s"
```
gy2411: There are no tests that need small value of this parameter, aren't there?

cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_test.go
```diff
@@ -260,12 +257,19 @@ func TestDiskServiceDeleteDiskWhenCreationIsInFlight(t *testing.T) {
...
 
 	testcommon.DeleteDisk(t, ctx, client, diskID)
 
-	_ = internal_client.WaitOperation(ctx, client, createOp.Id)
+	err = internal_client.WaitOperation(ctx, client, createOp.Id)
+	require.NoError(t, err)
+
+	// Should wait here because base disk slot might be released on overlay disk
+	// creation operation cancel (and exact time of this event is unknown).
+	time.Sleep(time.Second * 10)
```
gy2411: It would be very convinient (here and in some other test) if we new either the task completed its cancellation.

Maybe it's ok to add this information to the Operation message in dm api? https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/api/operation.proto#L11.

SvartMetal: I think we can look at task state directly using `tasks.Storage` similar to this code https://github.com/ydb-platform/nbs/blob/c409d0c4d3191085b49d156cb19f9f4d0154254a/cloud/disk_manager/internal/pkg/facade/placement_group_service_test/placement_group_service_test.go#L75

cloud/disk_manager/internal/pkg/services/disks/create_overlay_disk_task.go
```diff
@@ -219,6 +217,43 @@ func (t *createOverlayDiskTask) GetResponse() proto.Message {
...
+		headers.SetIncomingIdempotencyKey(ctx, selfTaskID+"_release"),
+		&pools_protos.ReleaseBaseDiskRequest{
+			OverlayDisk: overlayDisk,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	_, err = t.scheduler.WaitTask(ctx, execCtx, taskID)
```
SvartMetal: `return t.scheduler.WaitTask(...`

cloud/disk_manager/internal/pkg/services/disks/create_overlay_disk_task.go
```diff
@@ -170,35 +167,36 @@ func (t *createOverlayDiskTask) Cancel(
 	}
 
 	if diskMeta == nil {
```
SvartMetal, line 172: If diskMeta is nil, it means we could not start acquiring the base disk, so releasing it is not necessary

cloud/disk_manager/internal/pkg/services/disks/create_overlay_disk_task.go
```diff
@@ -219,6 +217,43 @@ func (t *createOverlayDiskTask) GetResponse() proto.Message {
...
+	return &types.Disk{
+		DiskId: t.request.Params.Disk.DiskId,
+		ZoneId: t.state.SelectedCellId,
+	}, nil
+}
+
+func (t *createOverlayDiskTask) releaseBaseDisk(
+	ctx context.Context,
+	execCtx tasks.ExecutionContext,
+	overlayDisk *types.Disk,
```
SvartMetal: Maybe use `getOverlayDisk` instead?

SvartMetal: But maybe there is no need to extract releaseBaseDisk as a separate method

cloud/disk_manager/test/recipe/disk_manager_launcher.py
SvartMetal: If you change this value it is better to add a comment on why it is needed

gy2411: There is a multi-line string """, we can not add a comment inside it.

Seems that the best we could do here is to use string concatination here like this
```
long_string = (
    "First part of the string\n" # This is a comment
    "Second part of the string\n" # This is another comment
    "Third part"                 # Comment on the last line
)
```

cloud/disk_manager/internal/pkg/services/disks/create_overlay_disk_task.go
gy2411: Unfortunately, is't not true -- this is just the reason why we need a fix. The disk might have been deleted concurrently by DeleteDisk task -- see the scenario in pr description

SvartMetal: > The disk might have been deleted concurrently by DeleteDisk task 

In this case base disk should be released by DeleteDisk task

cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_test.go
gy2411: This is a good  idea. Added new method to tasks storage and used it in the test.

cloud/disk_manager/test/recipe/disk_manager_launcher.py
SvartMetal: You can pass this value as a parameter and add a comment somewhere in the code

cloud/disk_manager/internal/pkg/services/disks/create_overlay_disk_task.go
gy2411: The release by DeleteDisk task could happen earliar then the acquire by CreateOverayDisk. If so, nothing happens on release, then the slot is acquired and never released.

The is an alternative way to fix this bug -- we could change the behaviour of ReleaseBaseDisk in case then there is no slot for this overlay disk. Now it is noop. But we could insert a tombstone or something like that, so the AcquireDisk will see this tombstone and will not acquire a slot. But such approach seems too difficult.

gy2411: One more possible approach is to check the disk state in one transaction with acquiring base disk slot. Then acquire will see that the disk is already deleting/deleted and will not acquire a slot.

But here we need to query the disks table https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/resources/disks.go#L227 from the pool storage https://github.com/ydb-platform/nbs/tree/main/cloud/disk_manager/internal/pkg/services/pools/storage. Is it ok for us?

SvartMetal: > The is an alternative way to fix this bug -- we could change the behaviour of ReleaseBaseDisk in case then there is no slot for this overlay disk. Now it is noop. But we could insert a tombstone or something like that, so the AcquireDisk will see this tombstone and will not acquire a slot

This is the correct approach. The CreateOverlayDisk and DeleteDisk task code was designed with this behavior in mind

When releasing a base disk slot, a tombstone should be used, similar to the one used when deleting a disk https://github.com/ydb-platform/nbs/blob/8d7cd75892f5a875ef0c5871ffbf8a23fcf87623/cloud/disk_manager/internal/pkg/resources/disks.go#L539

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -1159,7 +1159,36 @@ func (s *storageYDB) releaseBaseDiskSlot(
...
+		// operations. The tombstone prevents a new slot with the same
+		// overlay disk ID from being acquired.
+		slotTombstone := slot{
+			overlayDiskID: overlayDiskID,
+			baseDiskID:    "",
+			status:        slotStatusReleased,
+			releasedAt:    time.Now(),
+		}
+
+		err = s.updateBaseDisksAndSlots(
```
SvartMetal: It would be simpler to use `s.updateSlot(...)` here

cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -577,6 +577,10 @@ def __init__(
         filesystem_dataplane_enabled=False,
         list_nodes_max_bytes=0,
         scrubbing_config_content="",
+        # 30s is long enough in tests with concurrent resource creation and deletion to prevent
+        # creating an already deleted resourse.
```
SvartMetal: Let's add an issue (or current PR) number here for reference

cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -577,6 +577,10 @@ def __init__(
         filesystem_dataplane_enabled=False,
         list_nodes_max_bytes=0,
         scrubbing_config_content="",
+        # 30s is long enough in tests with concurrent resource creation and deletion to prevent
```
SvartMetal: I would suggest increase the timeout to 100s

cloud/tasks/storage/compound_storage.go
```diff
@@ -451,6 +451,14 @@ func (s *compoundStorage) GetNode(
 	return s.storage.GetNode(ctx, host)
 }
 
+func (s *compoundStorage) IsTaskEnded(
+	ctx context.Context,
+	taskID string,
+) (isTerminal bool, err error) {
```
SvartMetal: You don't need named return args here

cloud/tasks/storage/storage_ydb.go
```diff
@@ -507,3 +507,21 @@ func (s *storageYDB) ResumeTask(ctx context.Context, taskID string) error {
 		},
 	)
 }
+
+func (s *storageYDB) IsTaskEnded(
+	ctx context.Context,
+	taskID string,
+) (bool, error) {
+
+	var isEnded bool
```
SvartMetal: We prefer shorter variable names, `ended`

cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_test.go
```diff
@@ -260,12 +257,20 @@ func TestDiskServiceDeleteDiskWhenCreationIsInFlight(t *testing.T) {
...
 	testcommon.DeleteDisk(t, ctx, client, diskID)
 
-	_ = internal_client.WaitOperation(ctx, client, createOp.Id)
+	tasksStorage, err := testcommon.NewTaskStorage(ctx)
+	require.NoError(t, err)
+
+	require.Eventually(t, func() bool {
+		ended, err := tasksStorage.IsTaskEnded(ctx, createOp.Id)
+		return err == nil && ended
+	}, 30*time.Second, 100*time.Millisecond)
```
SvartMetal: 30 second timeout might be not enough for such a heavy test, let's have the timeout at least 200 seconds

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_test.go
```diff
@@ -167,6 +167,14 @@ func TestStorageYDBReleaseNonExistent(t *testing.T) {
 	require.Empty(t, baseDisk)
 	require.False(t, baseDiskShouldBeDeletedSoon(t, ctx, storage, baseDisk))
 
+	baseDisk, err = storage.AcquireBaseDiskSlot(ctx, "image", slot)
+	require.Error(t, err)
+	require.True(t, errors.Is(err, errors.NewEmptyNonRetriableError()))
+	require.True(t, errors.IsSilent(err))
+	require.ErrorContains(t, err, "already released")
+	require.Empty(t, baseDisk)
+	require.False(t, baseDiskShouldBeDeletedSoon(t, ctx, storage, baseDisk))
```
SvartMetal, line 178: Let's add a clarifying comment that the tombstone is expected to prevent acquisition

cloud/tasks/storage/storage_ydb_test.go
```diff
@@ -5200,6 +5200,78 @@ func TestStorageYDBNonCancellableTask(t *testing.T) {
...
+	nonExistentTaskID := "non-existent-task-id"
+	_, err = storage.IsTaskEnded(ctx, nonExistentTaskID)
+	require.Error(t, err)
+	require.True(t, errors.Is(err, errors.NewEmptyNonRetriableError()))
+	require.True(t, errors.Is(err, errors.NewNotFoundErrorWithTaskID(
+		nonExistentTaskID,
+	)))
+}
+
+func TestStorageYDBIsTaskEnded(t *testing.T) {
```
SvartMetal: Looks like the test names were swapped

Below is TestStorageYDBNonCancellableTask

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -1159,7 +1159,35 @@ func (s *storageYDB) releaseBaseDiskSlot(
...
+		// operations. The tombstone prevents a new slot with the same
+		// overlay disk ID from being acquired.
+		slotTombstone := slot{
+			overlayDiskID: overlayDiskID,
+			baseDiskID:    "",
+			status:        slotStatusReleased,
+			releasedAt:    time.Now(),
+		}
+
+		err = s.updateSlots(
```
SvartMetal: You can you use `updateSlot` method, it is slightly less verbose

https://github.com/ydb-platform/nbs/blob/1ac31d13524fafa1c365e3b6e42d8e2118d05ef8/cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go#L559

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -1159,7 +1159,35 @@ func (s *storageYDB) releaseBaseDiskSlot(
...
+				state:    &slotTombstone,
+			}},
+		)
+		if err != nil {
+			return BaseDisk{}, err
+		}
+
+		logging.Info(
+			ctx,
+			"created slot tombstone %+v",
```
SvartMetal:
```
....
"Created tombstone %+v while releasing base disk slot in zone %v",
overlayDisk.ZoneId,
```

cloud/tasks/storage/storage_ydb_impl.go
```diff
@@ -2170,3 +2170,17 @@ func (s *storageYDB) resumeTask(
...
 	return tx.Commit(ctx)
 }
+
+func (s *storageYDB) isTaskEnded(
+	ctx context.Context,
+	session *persistence.Session,
+	taskID string,
+) (bool, error) {
+
+	taskState, err := s.getTask(ctx, session, taskID)
```
SvartMetal: taskState -> state

It is usually named this way in this file

PR #5554, opened 2026-03-22
be a little bit more verbose for rdma endpoint events

yegorskii:
### Notes
When reading logs related to rdma endpoints sometimes it is not clear what connection the message is related to. Just added prefixes (host:port) to every message related to endpoints

### Issue
no

### DIFF
--- a/cloud/blockstore/libs/rdma/impl/client.cpp
+++ b/cloud/blockstore/libs/rdma/impl/client.cpp
@@ -49,6 +49,23 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+#define RDMA_ENDPOINT_PREFIX(ep) "[" << (ep)->Host << ":" << (ep)->Port << "]: "
+
+#define RDMA_ENDPOINT_LOG2(priority, stream) \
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -588,7 +605,7 @@ TClientEndpoint::TClientEndpoint(
         return TStringBuilder() << "[" << Id << "] " << msg;
     });
 
-    RDMA_INFO(
+    RDMA_ENDPOINT_INFO(
```
tpashkin, line 608: There is no need for a new macro. You can simply add this info to the Formatter here – https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/impl/client.cpp#L588

However, I'm not convinced that it's necessary – we already log host on connects and disconnects
- connect – https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/impl/client.cpp#L2175
- disconnect – https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/impl/client.cpp#L2094

PR #5557, opened 2026-03-23
Fix unregistration total used quota metric if one of VMs stopped

PeshkovMikhail:
### Notes
If `TThrottlerMetrics::UnregisterAll` is called on one of the VMs, the total used quota metrics will disappear on the entire host. Now, one instance of TThrottlerMetrics is shared between all client throttlers

### DIFF
--- a/cloud/blockstore/libs/client/throttling.cpp
+++ b/cloud/blockstore/libs/client/throttling.cpp
@@ -382,6 +382,7 @@ private:
     const NMonitoring::TDynamicCountersPtr RootGroup;
     const IRequestStatsPtr RequestStats;
     const IVolumeStatsPtr VolumeStats;
+    const IThrottlerMetricsPtr ThrottlerMetrics;
 
     TMutex ThrottlerLock;
     THashMap<TString, TThrottlerInfo> Throttlers;
@@ -402,6 +403,7 @@ public:
         , RootGroup(std::move(rootGroup))
         , RequestStats(std::move(requestStats))
         , VolumeStats(std::move(volumeStats))
+        , ThrottlerMetrics(CreateThrottlerMetrics(Timer, RootGroup, "server"))
     {}
 
     IThrottlerPtr GetThrottler(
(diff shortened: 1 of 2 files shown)


cloud/blockstore/libs/throttling/throttler_metrics_ut.cpp
```diff
@@ -192,6 +192,43 @@ Y_UNIT_TEST_SUITE(TThrottlerBaseTests)
...
+        UNIT_ASSERT_VALUES_EQUAL(0, quotaCounterTotal->Val());
+        UNIT_ASSERT_VALUES_EQUAL(0, maxQuotaCounter1->Val());
+        UNIT_ASSERT_VALUES_EQUAL(0, maxQuotaCounter2->Val());
+        UNIT_ASSERT_VALUES_EQUAL(0, maxQuotaCounterTotal->Val());
+
+        SecondVMThrottler->Unregister("Disk-2", "Client-2");
+
+        UNIT_ASSERT(TotalGroup->FindCounter("UsedQuota"));
+        UNIT_ASSERT(TotalGroup->FindCounter("MaxUsedQuota"));
+
```
SvartMetal: excessive \n

cloud/blockstore/libs/throttling/throttler_metrics.cpp
```diff
@@ -239,7 +239,6 @@ void TThrottlerMetrics::UpdateMaxUsedQuota()
 
 void TThrottlerMetrics::UnregisterAll()
 {
-    Total.Metrics.Unregister();
```
SvartMetal, line 242: Help me understand the root cause of the bug.
As far as I can see, UnregisterAll is called when the last disk metric is unregistered. This means the metrics are unregistered when the last VM stops, not the first one

PeshkovMikhail: There can be multiple disks on a single VM. UnreginsterAll is called if all disks on the same VM have been unmounted. The bug is that if there are several VMs on the host, then each vm creates its own instance of the trottler, and if UnreginsterAll is called on one of the VMs, then this metric will disappear on all VMs.

SvartMetal: > The bug is that if there are several VMs on the host, then each vm creates its own instance of the trottler,

But why is this happening? It looks like the TThrottlerMetrics interface was designed to support multiple disk-to-instance pairs: https://github.com/ydb-platform/nbs/blob/76157bc26be42a0fde5a6c03fd24e6ded6abe886/cloud/blockstore/libs/throttling/throttler_metrics.cpp#L136

Maybe it would be more reasonable to fix how TThrottlerMetrics is used and make it per host rather than per VM?

PR #5558, opened 2026-03-23
feat: add ability to match files to mark as plain by regex

librarian:


.github/scripts/index.py
```diff
@@ -11,7 +11,10 @@
 from jinja2 import Environment, FileSystemLoader
 from datetime import datetime, timedelta, timezone
 
-FORCE_TEXT_PLAIN_FILES = {"stdout", "stderr"}
+FORCE_TEXT_PLAIN_PATTERNS = (
+    r"stdout.*",
```
SvartMetal: Let's add .* prefix also?

It will match files like `disk_manager_stderrXXX` then

Issue #5561, opened 2026-03-23
[Filestore] Don't return schemeshard errors from public API handlers - translate them into NBS errors


qkrorlqr:
For example, here https://github.com/ydb-platform/nbs/blob/14393dbe9ea0ef1d3add0ec6793c15e5ac87078b/cloud/filestore/libs/storage/service/service_actor_getfsinfo.cpp#L86 we can return schemeshard's `StatusPathDoesNotExist` error - we should return our own `E_NOT_FOUND` instead. I'm planning to change only filestore code here - to make the PR simpler and smaller and not to risk breaking blockstore's backward compatibility (there are fewer dependencies on filestore error codes in the project - because there're no snapshots, no images, no base disk pools for filestore, etc). 

Probably, the best way to fix error codes would be to change the behaviour of `cloud/filestore/libs/storage/ss_proxy` and a couple places in the code where schemeshard error codes are explicitly used, e.g. this: https://github.com/ydb-platform/nbs/blob/14393dbe9ea0ef1d3add0ec6793c15e5ac87078b/cloud/filestore/libs/endpoint/endpoint_manager.cpp#L271

PR #5563, opened 2026-03-23
issue-5561: filestore ss_proxy - translating schemeshard's StatusPathDoesNotExist code to E_NOT_FOUND

qkrorlqr:
### Notes
The goal is to avoid showing schemeshard's error codes in our public API and use our own error codes for all our public methods.

I grepped the code in `cloud/` and I don't see any dependencies on `StatusPathDoesNotExist` outside of some local interactions within either `filestore-vhost` or `filestore-server` - i.e. I don't see any cross-node communication where one side can send `StatusPathDoesNotExist` and the other side needs to process this code in some special way. So looks like we can easily start translating `StatusPathDoesNotExist` to `E_NOT_FOUND` without any backwards-compat-related logic.

### Issue
https://github.com/ydb-platform/nbs/issues/5561

### DIFF
--- a/cloud/filestore/tests/client_sharded/canondata/test.test_explicit_shard_count_addition/results.txt
+++ b/cloud/filestore/tests/client_sharded/canondata/test.test_explicit_shard_count_addition/results.txt
@@ -1,7 +1,7 @@
 {"Sessions":[{"SessionId":"session0","ClientId":"client0"}]}
 {"Sessions":[{"SessionId":"session0","ClientId":"client0"}]}
-SEVERITY_ERROR | FACILITY_SCHEMESHARD | 2 Path not found, path="/Root/nfs/_385/fs0_s2"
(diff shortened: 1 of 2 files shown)


ruslan-gar (approved):

cloud/filestore/libs/storage/ss_proxy/ss_proxy_actor_destroyfs.cpp
```diff
@@ -108,7 +108,9 @@ void TDestroyFileStoreActor::HandleModifySchemeResponse(
     const TActorContext& ctx)
 {
     NProto::TError error = ev->Get()->GetError();
-    if (error.GetCode() == MAKE_SCHEMESHARD_ERROR(NKikimrScheme::EStatus::StatusPathDoesNotExist)) {
+    const ui32 pathDoesNotExist =
```
agalibin, line 111: это мне кажется лишним, но ок

qkrorlqr: ну, это ж мелочи

Issue #3727, opened 2025-06-21
[NBS] Send notification for broken NRD when volume has broken devices


komarevtsev-d:
As of now, only DiskRegistry can notify when a has been broken. However, sometimes the disk may break from the perspective of the volume, e.g., due to network issues, and this should also trigger a notification of a broken disk.

PR #5565, opened 2026-03-24
[NBS] Add IDeviceStatObserver to notify VolumeActor on device status changes

zavik001:
https://github.com/ydb-platform/nbs/issues/3727

Introduced IDeviceStatObserver interface to TDeviceStat so that
TNonreplicatedPartitionActor gets notified when a device becomes
broken or recovers, and forwards these events to VolumeActor.

### DIFF
--- a/cloud/blockstore/libs/storage/volume/volume_actor_broken_devices.cpp
+++ b/cloud/blockstore/libs/storage/volume/volume_actor_broken_devices.cpp
@@ -25,6 +25,8 @@ void TVolumeActor::HandleBrokenDeviceNotification(
     // TODO: save the broken device to the database
 }
 
+////////////////////////////////////////////////////////////////////////////////
+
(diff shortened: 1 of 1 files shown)


cloud/blockstore/libs/storage/partition_nonrepl/part_nonrepl_actor.cpp
```diff
@@ -23,6 +23,25 @@ LWTRACE_USING(BLOCKSTORE_STORAGE_PROVIDER);
 
 ////////////////////////////////////////////////////////////////////////////////
```
qkrorlqr: nit: we usually put "///..." right before class or function declaration, not before namespace declaration

zavik001: > nit: we usually put "///..." right before class or function declaration, not before namespace declaration

done

PR #5567, opened 2026-03-24
issue-4875: Fix data loss on fresh blobs data loading with Fresh Blocks Writer Feature

vladstepanyuk:
### Notes
We persist the current TrimFreshLogToCommitId for every AddBlobs transaction in the local database. On partition start, we load all the blobs starting from that persisted TrimFreshLogCommitId. So, if at the time of AddBlob transaction execution, there are some inflight requests with a commit id less than the TrimFreshLog to CommitId in the partition and a reboot occurs after the transaction completes, we will not load fresh blobs with a commit id less than the TrimFreshLog from the FBW , though response was already sent.

In this PR i persist Last TrimFreshLogToCommitId, which takes into account commits from FBW

### Issue
#4875

### DIFF
--- a/cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
+++ b/cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
@@ -1010,6 +1010,75 @@ Y_UNIT_TEST_SUITE(TFreshBlocksWriterTest)
             response->GetStatus(),
             response->GetErrorReason());
     }
+
+    Y_UNIT_TEST(ShouldNotLooseInFlightWritesOnReboot)
+    {
+        auto testEnv = PrepareTestActorRuntime();
+        auto& runtime = *testEnv.Runtime;
+
+        TPartitionClient partition(runtime);
+        partition.WaitReady();
+
+        TFreshBlocksWriterClient fbwClient(
+            runtime,
+            testEnv.FreshBlocksWriterActorId);
+        fbwClient.WaitReady();
+
+        std::unique_ptr<IEventHandle> stollenPutRequest;
+        bool steelPutRequest = true;
+
+        runtime.SetObserverFunc(
+            [&](TAutoPtr<IEventHandle>& event)
+            {
+                if (event->GetTypeRewrite() == TEvBlobStorage::EvPut &&
+                    !stollenPutRequest && steelPutRequest)
+                {
+                    Cerr << "Stollen put request" << Endl;
+                    stollenPutRequest.reset(event.Release());
+                    return TTestActorRuntime::EEventAction::DROP;
+                }
+
+                if (event->GetTypeRewrite() == TEvPartitionCommonPrivate::EvTrimFreshLogRequest)
+                {
+                    return TTestActorRuntime::EEventAction::DROP;
+                }
+
+                return TTestActorRuntime::DefaultObserverFunc(event);
+            });
+
+        fbwClient.SendWriteBlocksRequest(0, '1');
+        runtime.DispatchEvents({}, 10ms);
+
+        fbwClient.WriteBlocks(1, '2');
+
+        partition.Flush();
+
+        fbwClient.WriteBlocks(2, '3');
+
+        partition.Flush();
+
+        steelPutRequest = false;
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
```diff
@@ -1010,6 +1010,75 @@ Y_UNIT_TEST_SUITE(TFreshBlocksWriterTest)
             response->GetStatus(),
             response->GetErrorReason());
     }
+
+    Y_UNIT_TEST(ShouldNotLooseInFlightWritesOnReboot)
```
qkrorlqr, line 1014: Loose -> Lose

cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
```diff
@@ -1010,6 +1010,75 @@ Y_UNIT_TEST_SUITE(TFreshBlocksWriterTest)
...
+        TPartitionClient partition(runtime);
+        partition.WaitReady();
+
+        TFreshBlocksWriterClient fbwClient(
+            runtime,
+            testEnv.FreshBlocksWriterActorId);
+        fbwClient.WaitReady();
+
+        std::unique_ptr<IEventHandle> stollenPutRequest;
+        bool steelPutRequest = true;
```
qkrorlqr, line 1028: steel -> steal

cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
```diff
@@ -1010,6 +1010,75 @@ Y_UNIT_TEST_SUITE(TFreshBlocksWriterTest)
...
+
+        TPartitionClient partition(runtime);
+        partition.WaitReady();
+
+        TFreshBlocksWriterClient fbwClient(
+            runtime,
+            testEnv.FreshBlocksWriterActorId);
+        fbwClient.WaitReady();
+
+        std::unique_ptr<IEventHandle> stollenPutRequest;
```
qkrorlqr, line 1027: stollen -> stolen

cloud/blockstore/libs/storage/partition/part_state.cpp
```diff
@@ -97,6 +97,7 @@ TPartitionState::TPartitionState(
     , CleanupScoreHistory(cleanupScoreHistorySize)
 {
     InitChannels();
+    SetLastTrimFreshLogToCommitId(Meta.GetTrimFreshLogToCommitId());
```
qkrorlqr, line 100: why do we need this? if we don't set last-trim-fresh-log-to-commit-id we won't drop any data - we will just run one trim op which might be useless (might not delete any data) but it is not a problem

cloud/blockstore/libs/storage/partition/part_state.h
```diff
@@ -950,7 +950,7 @@ class TPartitionState
 
     void UpdateTrimFreshLogToCommitIdInMeta()
     {
-        Meta.SetTrimFreshLogToCommitId(GetTrimFreshLogToCommitId());
```
qkrorlqr, line 953: `GetTrimFreshLogToCommitId()` is supposed to return a commit-id up to which it is safe to drop all fresh blobs: https://github.com/ydb-platform/nbs/blob/247ad52d669ce0d9528c0d2d48c69737871cc439/cloud/blockstore/libs/storage/partition_common/part_fresh_blocks_state.h#L155

if this is not true, then this is the thing that must be fixed, not something else like using last-trim-fresh-log-to-commit-id instead of trim-fresh-log-to-commit-id

last-trim-fresh-log-to-commit-id is supposed to be used only for the purpose of not running the trim-fresh-log-op when there's nothing to do

this fix doesn't look correct

qkrorlqr: We should at least try to make an implementation that uses shared barrier queues + locks. There will be almost no contention on those locks. But the code will probably be much simpler than managing 2 independent incomplete states and then somehow implementing the coordination.

PR #5576, opened 2026-03-25
[NBS] Add BrokenDevices table to volume schema and database

zavik001:
https://github.com/ydb-platform/nbs/issues/3727

TNonreplicatedPartitionActor detects silent broken devices and notifies TVolumeActor. This information is persisted in the local tablet database to survive actor restarts.

### DIFF
--- a/cloud/blockstore/libs/storage/volume/volume_database.cpp
+++ b/cloud/blockstore/libs/storage/volume/volume_database.cpp
@@ -865,8 +865,6 @@ bool TVolumeDatabase::ReadBrokenDevices(TVector<TBrokenDeviceInfo>& devices)
 {
     using TTable = TVolumeSchema::BrokenDevices;
 
-    devices.clear();
-
     auto it = Table<TTable>().Range().Select<TTable::TColumns>();
 
     if (!it.IsReady()) {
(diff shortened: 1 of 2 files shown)


drbasic (dismissed):

cloud/blockstore/libs/storage/volume/volume_database.cpp
```diff
@@ -841,4 +841,50 @@ bool TVolumeDatabase::ReadLeaders(TLeaderDisks& leaders)
...
+    using TTable = TVolumeSchema::BrokenDevices;
+
+    Table<TTable>().Key(deviceUUID).Delete();
+}
+
+bool TVolumeDatabase::ReadBrokenDevices(TVector<TBrokenDeviceInfo>& devices)
+{
+    using TTable = TVolumeSchema::BrokenDevices;
+
+    devices.clear();
```
qkrorlqr: usually we don't rely on ReadXXX funcs clearing the argument - we usually clear such things in the tx args' `Clear()` method

I see that the other similar methods in `TVolumeDatabase` clear the arguments but in the other tablets (partition tablet, filestore index tablet) we don't do this

qkrorlqr (dismissed):

cloud/blockstore/libs/storage/volume/volume_database.cpp
zavik001: > usually we don't rely on ReadXXX funcs clearing the argument - we usually clear such things in the tx args' `Clear()` method
> 
> I see that the other similar methods in `TVolumeDatabase` clear the arguments but in the other tablets (partition tablet, filestore index tablet) we don't do this

done

Issue #5527, opened 2026-03-19
[CSI] Support multiple blockstore-server sockets


antonmyagkov:
Multiple blockstore-server process can be run on the same node. Each blockstore-server process has unix socket grpc{id}.sock. csi driver should choose one of the blockstore-server sockets to start endpoint for disk.

PR #5577, opened 2026-03-25
[CSI] Select the backend instance with the fewest endpoints

antonmyagkov:
issue :#5527
issue: #5362

Select the backend instance with the fewest endpoints, rather than choosing one based on a hash.
For now, a simple algorithm is used: each time, the CSI driver lists the endpoints for each instance and selects the one with the fewest endpoints.

### DIFF
--- a/cloud/blockstore/tools/csi_driver/internal/driver/node.go
+++ b/cloud/blockstore/tools/csi_driver/internal/driver/node.go
@@ -435,10 +435,6 @@ func (s *nodeService) getNbsClientIndex(ctx context.Context) uint {
 			continue
 		}
 
-		if listEndpointsResp.EndpointsWereRestored == false {
-			continue
-		}
-
 		if len(listEndpointsResp.Endpoints) < endpointsCount {
 			endpointsCount = len(listEndpointsResp.Endpoints)
 			nbsClientIndex = uint(index)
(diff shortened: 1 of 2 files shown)


cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -422,6 +422,54 @@ func (s *nodeService) NodeUnstageVolume(
...
+	endpointsCount := math.MaxInt32
+	nbsClientIndex := uint(0)
+	for index := uint(0); index < s.nbsServerReplicaCount; index++ {
+		listEndpointsResp, err := s.nbsClients[index].ListEndpoints(ctx,
+			&nbsapi.TListEndpointsRequest{})
+		if err != nil {
+			continue
+		}
+
+		if listEndpointsResp.EndpointsWereRestored == false {
```
tpashkin: Does it actually matter if they were restored or not?

tpashkin: This may introduce a slight imbalance when the volume is staged during restart. It's unlikely to ever be a problem, but what is the logic here?

antonmyagkov: imbalance can be introduced without this check too as the list of endpoints in request can be empty but endpoints are not yet restored.
 
 I prefer to skip instance in this case.

antonmyagkov: agreed with @SvartMetal to remove usage of EndpointsWereRestored

PR #5580, opened 2026-03-25
Add shaping throttler config; allow to persist shaping throttler state in volume's DB

komarevtsev-d:
#5095

This change introduces `TShapingThrottlerConfig` / `TShapingThrottlerQuota` in blockstore storage config. Volume throttler state is now persisted when shaping throttler is enabled for that kind.

### DIFF
--- a/cloud/storage/core/protos/diagnostics.proto
+++ b/cloud/storage/core/protos/diagnostics.proto
@@ -12,3 +12,12 @@ enum EStatsFetcherType
     CGROUP = 0;
     TASKSTATS = 1;
 };
+
+////////////////////////////////////////////////////////////////////////////////
+// Performance profile
+
+message TPerformanceProfile
+{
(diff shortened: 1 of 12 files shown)


cloud/storage/core/protos/diagnostics.proto
```diff
@@ -12,3 +12,12 @@ enum EStatsFetcherType
     CGROUP = 0;
     TASKSTATS = 1;
 };
+
+////////////////////////////////////////////////////////////////////////////////
+// Performance profile
+
+message TPerformanceProfile
```
qkrorlqr, line 19: let's rename it, otherwise there's a name collision with other `TPerformanceProfile`s

qkrorlqr: ok, seems that there're actually no collisions

PR #5582, opened 2026-03-25
issue-5103: [Filestore] Don't enforce node count limit if BehaveAsShard

ruslan-gar:
### Notes
If create node message in shard is initiated by the main fs, we should not enforce node count limit.
If we do, it may lead to dangling NodeRefs in the main fs, as it may happen that AggregateUsedNodesCount 
is updated in a shard earlier than in the main fs. Thus, the limit is enforced in in the shard when NodeRef 
is already created in the main fs.

### Issue
#5103

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
@@ -7312,14 +7312,12 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
             options.FinalEvents.emplace_back(
                 TDispatchOptions::TFinalEventCondition(
                     TEvIndexTabletPrivate::EvUpdateCounters,
-                    shardCount + static_cast<ui64>(updateMainFsCounters)));
+                    shardCount + updateMainFsCounters));
             env.GetRuntime().DispatchEvents(options);
         }
 
-        const ui32 critEventCount =
-            strictFileSystemSizeEnforcementEnabled ? 0 : 1;
         const ui32 response =
-            (!strictFileSystemSizeEnforcementEnabled || updateMainFsCounters)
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -233,14 +233,21 @@ bool TIndexTabletActor::PrepareTx_CreateHandle(
...
 
-            // validate there are enough free inodes
-            if (!HasNodesLeft()) {
+            const bool behaveAsShard = BehaveAsShard(args.Request.GetHeaders());
+
+            // Validate there are enough free inodes.
+            // In the strict mode the restriction is not enforced if the
+            // request comes from the main FS to the shard.
+            if (!HasNodesLeft() &&
+                (!GetFileSystem().GetStrictFileSystemSizeEnforcementEnabled() ||
```
debnatkh: Do we need this? I mean, ignoring this limitation is reasonable regardless of strict filesystem enforcement feature

cloud/filestore/libs/storage/service/service_ut_sharding.cpp
```diff
@@ -7287,39 +7290,63 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
...
+            }
+            updateCounters(fsInfo.Shard1ActorId);
+            updateCounters(fsInfo.Shard2ActorId);
+
             NActors::TDispatchOptions options;
             options.FinalEvents.emplace_back(
                 TDispatchOptions::TFinalEventCondition(
                     TEvIndexTabletPrivate::EvUpdateCounters,
-                    shardCount + 2));
+                    shardCount + static_cast<ui64>(updateMainFsCounters)));
```
debnatkh: Is static cast necessary here? I thought that int + bool does not require explicit cast

PR #5586, opened 2026-03-25
[Filestore] Implement backpressure for write based on FreshBytesItemCount

e673:
### Notes
Out of memory / transaction limit hit may happen if there are too much FreshBytes entries.
Previously, a `FreshBytesItemCountThreshold` was introduced to start flushing data in order to solve this problem.
But it is still possible that a client may write data faster that we can process it — in this case, OOM is still likely to happen.

Currently, there is a backpressure threshold based on the amount of FreshBytes but not the number of entries.
This PR adds `FlushBytesItemCountThresholdForBackpressure`.

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring.cpp
@@ -1172,6 +1172,7 @@ void TIndexTabletActor::HandleHttpInfo_Default(
 
             DUMP_BACKPRESSURE_FIELD(Flush);
             DUMP_BACKPRESSURE_FIELD(FlushBytes);
+            DUMP_BACKPRESSURE_FIELD(FlushBytesItemCount);
             DUMP_BACKPRESSURE_FIELD(CompactionScore);
             DUMP_BACKPRESSURE_FIELD(CleanupScore);
 


cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -295,6 +295,12 @@ bool TIndexTabletState::IsWriteAllowed(
         return false;
     }
 
+    if (values.FlushBytesItemCount >= thresholds.FlushBytesItemCount) {
```
debnatkh, line 298: Could you, please, support Flush items count also here: https://github.com/ydb-platform/nbs/blob/users/nasonov/fresh-bytes-count-backpressure/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring.cpp#L1176

So we can see the backpressure reason on the monitoring page of a filesystem

e673: Added

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -942,22 +931,11 @@ FILESTORE_DUPCACHE_REQUESTS(FILESTORE_DECLARE_DUPCACHE)
...
-            , FlushBytes(flushBytes)
-            , CompactionScore(compactionScore)
-            , CleanupScore(cleanupScore)
-        {
-        }
+        const ui64 Flush;
+        const ui64 FlushBytes;
+        const ui64 FlushBytesItemCount;
+        const ui64 CompactionScore;
+        const ui64 CleanupScore;
```
SvartMetal, line 938: Let’s avoid leaving values uninitialized

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -181,20 +181,9 @@ struct TBackgroundOpsBackpressureStatus
...
-            EBackgroundOpBackpressureStatus flush,
-            EBackgroundOpBackpressureStatus flushBytes,
-            EBackgroundOpBackpressureStatus compaction,
-            EBackgroundOpBackpressureStatus cleanup)
-        : Flush(flush)
-        , FlushBytes(flushBytes)
-        , Compaction(compaction)
-        , Cleanup(cleanup)
-    {
-    }
```
SvartMetal, line 197: Same

cloud/filestore/libs/storage/tablet/tablet_state.h
e673: Compiler will throw an error if any of these values is left uninitialized (because the values are const).

SvartMetal (approved):

Issue #5588, opened 2026-03-26
[Disk Manager] cleanup tasks (such as DeleteDisk, DeleteFilesystem, etc.) should only be allowed to complete successfully or fail with a NonCancellableError


SvartMetal:
Needed to mitigate the impact of bugs like #5511.

If a DeleteDisk task encounters a non-retriable error, it persists that error in its task state. 
As a result, the corresponding Compute API operation becomes stuck, because it keeps retrying the same Disk Manager task and gets the same error every time.
Cleanup tasks should not be allowed to fail with retriable or non-retriable errors. 
The only exception is a non-cancellable error, which means the task itself must not run and therefore must not be cancelled, to avoid further damaging the system.
In other words, cleanup tasks should always finish successfully, regardless of fatal errors returned by underlying services. 
Such errors should still be reflected in metrics, but they should not cause the task to fail.
Whenever a cleanup task finishes unsuccessfully, this indicates an abnormal situation that likely requires manual intervention from the on-duty engineer

PR #5511, opened 2026-03-17
[Blockstore] DestroyVolume should be idempotent when the disk is not found during graceful shutdown of a non-replicated (DiskRegistry-based) disk

SvartMetal:
### Notes
Fixes a race condition where a disk found by StatVolume could be deleted before GracefulShutdownRequest, causing `disks.DeleteDisk` Disk Manager task to fail with a non-retriable error and the Compute API operation to get stuck

```
2026-03-10T22:55:35.812626Z :BLOCKSTORE_SERVER INFO: cloud/blockstore/libs/diagnostics/server_stats.cpp:317: [d:xxx] [c:69b0a162-354d-081b-eed0-a5d9-d7037488] DestroyVolume #1636853003130062538 REQUEST { Headers { ClientId: "69b0a162-354d-081b-eed0-a5d9-d7037488" Timestamp: 1773183335812549 RequestId: 1636853003130062538 RequestTimeout: 20000 } DiskId: "xxx" }
...
2026-03-10T22:55:35.995844Z :BLOCKSTORE_SERVER INFO: cloud/blockstore/libs/diagnostics/server_stats.cpp:565: [d:xxx] [c:69b0a162-354d-081b-eed0-a5d9-d7037488] DestroyVolume #1636853003130062538 RESPONSE request completed (execution: 183.179ms, postponed: 0, predicted: 0, backoff: 0, size: 0 B, unaligned: 0, error: SEVERITY_ERROR | FACILITY_SCHEMESHARD | 2 Path not found, path="/xxx/NBS/xxx/xxx-copy" f@silent)
...
2026-03-10T22:55:35.812686Z :BLOCKSTORE_SERVICE INFO: Deleting volume: diskId = "xxx", destroyIfBroken = 0, sync = 0, fillGeneration = 0
...
2026-03-10T22:55:35.988903Z :BLOCKSTORE_VOLUME_PROXY WARN: [vp:72075186244428604 d:xxx pg:1 t:20.195s] Connection to volume failed
...
2026-03-10T22:55:35.995716Z :BLOCKSTORE_VOLUME_PROXY ERROR: [vp:72075186244428604 d:xxx pg:1 t:20.202s] Could not resolve path for volume. Error: SEVERITY_ERROR | FACILITY_SCHEMESHARD | 2 Path not found, path="/xxx/NBS/_1E9/xxx-copy" f@silent
...
[...]

PR #5589, opened 2026-03-26
issue-5588: [Tasks] support non-cancellable tasks that can only either complete successfully or fail with a NonCancellableError

SvartMetal:
### Notes
Needed to mitigate the impact of bugs like #5511.

If a DeleteDisk task encounters a non-retriable error, it persists that error in its task state.
As a result, the corresponding Compute API operation becomes stuck, because it keeps retrying the same Disk Manager task and gets the same error every time.
Cleanup tasks should not be allowed to fail with retriable or non-retriable errors.
The only exception is a non-cancellable error, which means the task itself must not run and therefore must not be cancelled, to avoid further damaging the system.
In other words, cleanup tasks should always finish successfully, regardless of fatal errors returned by underlying services.
Such errors should still be reflected in metrics, but they should not cause the task to fail.
Whenever a cleanup task finishes unsuccessfully, this indicates an abnormal situation that likely requires manual intervention from the on-duty engineer

### Issue
#5588

### DIFF
--- a/cloud/tasks/execution_context.go
+++ b/cloud/tasks/execution_context.go
@@ -27,6 +27,9 @@ type ExecutionContext interface {
 
 	GetTaskID() string
 
+	// NOTE: non-cancellable tasks can only fail with a NonCancellableError, so
+	// they ignore NonRetriableError and do not enforce a retry limit for
+	// RetriableError
 	IsNonCancellable() bool
 
 	// Dependencies are automatically added by Scheduler.WaitTask.
@@ -300,6 +303,8 @@ func (c *executionContext) incrementPanicCount(
 	})
 }
 
+// NOTE: returns an error for non-cancellable tasks. Such tasks can only fail
+// with a NonCancellableError
 func (c *executionContext) setError(ctx context.Context, e error) error {
 	state, err := c.task.Save()
 	if err != nil {
--- a/cloud/tasks/storage/storage_ydb_impl.go
+++ b/cloud/tasks/storage/storage_ydb_impl.go
@@ -1619,17 +1619,18 @@ func (s *storageYDB) updateTaskTx(
 
 	lastState := states[0]
 
-	// A non-cancellable task may transition to the cancelled state, as this
-	// simply skips the actual cancellation process
-	if lastState.NonCancellable && state.Status != TaskStatusCancelled {
+	if lastState.NonCancellable && IsCancellationRequested(state.Status) {
+		err = tx.Commit(ctx)
+		if err != nil {
+			return TaskState{}, err
+		}
+
 		// It is forbidden to start the cancellation process for a
 		// non-cancellable task
-		if IsCancellationRequested(state.Status) {
-			return TaskState{}, errors.NewNonRetriableErrorf(
-				"unexpected status for a non-cancellable task, got %v",
-				TaskStatusToString(state.Status),
-			)
-		}
+		return TaskState{}, errors.NewNonRetriableErrorf(
+			"unexpected status for a non-cancellable task, got %v",
+			TaskStatusToString(state.Status),
+		)
 	}
 
 	if lastState.GenerationID != state.GenerationID {


cloud/tasks/runner.go
SvartMetal, line 1: More tests

cloud/tasks/scheduler.go
```diff
@@ -45,6 +45,29 @@ type Scheduler interface {
...
+	//
+	// A task can end in only two ways:
+	// * If it encounters a NonCancellableError, it finishes with a "cancelled"
+	//   status without invoking any cancel function.
+	// * Otherwise, it completes successfully.
+	//
+	// This behaviour is useful for cleanup tasks
+	// (e.g., DeleteDisk, DeleteFilesystem)
+	//
+	// If zoneID == nil, the task is not zonal
```
jkuradobery: zoneID can't be nil, it is string.

cloud/tasks/runner.go
```diff
@@ -174,7 +175,7 @@ func (r *runnerForRun) executeTask(
 		taskType,
 		taskID,
 	)
-	r.metrics.OnExecutionError(err)
+	r.onExecutionError(execCtx, err)
```
jkuradobery, line 178: Why do we attempt task cancellation if the task is non cancellable? 
If I understand correctly,  we will receive `unexpected status for a non-cancellable task, got %v`
on the attempt to setError. Is it expected behaviour?

SvartMetal: The task does not allow transitioning to the cancellation state. It’s fine for the runner to receive an error - we want it to be visible in the logs.

All I need from the runner is to not suppress the task error escalation

cloud/tasks/storage/storage_ydb_impl.go
```diff
@@ -1608,6 +1619,19 @@ func (s *storageYDB) updateTaskTx(
 
 	lastState := states[0]
 
+	// A non-cancellable task may transition to the cancelled state, as this
+	// simply skips the actual cancellation process
```
gy2411: `IsCancellationRequested` checks that status in not `Cancelled` -- so why not just
```
if lastState.NonCancellable && IsCancellationRequested(state.Status) {
    return ...
}
```

cloud/tasks/storage/storage_ydb_impl.go
```diff
@@ -1608,6 +1619,19 @@ func (s *storageYDB) updateTaskTx(
...
+	if lastState.NonCancellable && state.Status != TaskStatusCancelled {
+		// It is forbidden to start the cancellation process for a
+		// non-cancellable task
+		if IsCancellationRequested(state.Status) {
+			return TaskState{}, errors.NewNonRetriableErrorf(
```
gy2411: Seems that will we account one and the same error twice for non-cancellable tasks.

E.g. if the task got a non-retriable error, we will account the error inself in `onExecutionError` https://github.com/ydb-platform/nbs/pull/5589/changes#diff-e17825d18ff2e356fdaf163881058e0cc95b82194123ea7d232049b8e80c2737R178 and also account this error here https://github.com/ydb-platform/nbs/blob/main/cloud/tasks/runner.go#L236. Is it ok?

Maybe we should return InterruptExecutionError here?

cloud/tasks/execution_context.go
```diff
@@ -27,6 +27,8 @@ type ExecutionContext interface {
 
 	GetTaskID() string
 
+	IsNonCancellable() bool
```
gy2411, line 33: Maybe write a comment here https://github.com/ydb-platform/nbs/blob/main/cloud/tasks/execution_context.go#L295 that setError returns an error for non-cancellable tasks? Is doesn't seem obvious and requires diving deep into the storage code to understand it.

cloud/tasks/storage/storage_ydb_impl.go
SvartMetal: I think this is acceptable - more alerts in such situations are better.

cloud/tasks/execution_context.go
SvartMetal: Added a comment

cloud/tasks/storage/storage_ydb_impl.go
SvartMetal: Thanks, fixed

Issue #5590, opened 2026-03-26
[NBS] Get rid of reading block masks in compaction transactions


vladstepanyuk:
Reading block masks takes A LOT of CPU time during Compaction transaction 
<img width="4804" height="916" alt="Image" src="https://github.com/user-attachments/assets/10635421-745f-4e57-8896-8410e2e57d24" />

Block masks are used to track blobs that have been overwritten with other blobs. When the block mask is full, we add a blob to the cleanup queue.

But if blob lies entirely in one compaction range, it should be overwritten with new blob during compaction. So we can get rid of reading block masks for this type of blobs.

 Took some tests of this optimization and there is some results:

I brought the two disks to the same compaction score and started fio:
`fio --rw=randwrite --name=randwrite --filename=/dev/disk/by-id/virtio-blockmask --direct=1 --bs=4k --ioengine=libaio --iodepth=32 --group_reporting --size=512GB --rate_iops=4500`

Compaction score:

<img width="2933" height="487" alt="Image" src="https://github.com/user-attachments/assets/7395fcf4-0249-49b4-a9b9-f5914a50ac40" />

Compaction Time Percentiles:

50:

<img width="2933" height="487" alt="Image" src="https://github.com/user-attachments/assets/78e0741b-d9bc-4cf7-8479-034f2de31edd" />

90:

<img width="2933" height="487" alt="Image" src="https://github.com/user-attachments/assets/cea05ab4-3778-4f22-b0d4-770c9479ae8f" />

99:

<img width="2933" height="487" alt="Image" src="https://github.com/user-attachments/assets/ea7a2bf8-63c9-47d6-acba-348263565bc4" />

ExecutionCpuTime of Compaction transaction:

<img width="2933" height="487" alt="Image" src="https://github.com/user-attachments/assets/3c57204c-da74-48a4-9a3b-2ad7c7da348f" />

We can see strong perfomance boost  in compaction performance

PR #5591, opened 2026-03-26
[NBS] issue-5590: don't read block mask for merged blobs during compaction if blob lies entirely in one compaction range

vladstepanyuk:
### Issue
#5590

### DIFF
--- a/cloud/blockstore/config/storage.proto
+++ b/cloud/blockstore/config/storage.proto
@@ -1529,7 +1529,7 @@ message TStorageServiceConfig
     // Shaping throttler configuration.
     optional TShapingThrottlerConfig ShapingThrottlerConfig = 490;
 
-    // Don't read block masks on compaction if blob entirely contains
+    // Don't read block masks on compaction if blob is entirely contained
     // in one compaction range.
     optional bool ReadBlockMaskOnCompactionOptimizationEnabled = 491;
 }
--- a/cloud/blockstore/libs/storage/partition/model/cleanup_queue.cpp
+++ b/cloud/blockstore/libs/storage/partition/model/cleanup_queue.cpp
@@ -26,7 +26,8 @@ struct TCleanupQueue::TImpl
     {
         bool result = BlobIds.insert(item.BlobId).second;
         if (result) {
-            Items.insert(item);
+            auto inserted = Items.insert(item).second;
+            Y_DEBUG_ABORT_UNLESS(inserted);
         }
 
         return result;
@@ -36,6 +37,7 @@ struct TCleanupQueue::TImpl
     {
         auto itBlob = BlobIds.find(item.BlobId);
         if (itBlob == BlobIds.end()) {
+            Y_DEBUG_ABORT_UNLESS(!Items.contains(item));
             return false;
         }
 
--- a/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
@@ -584,7 +584,6 @@ private:
                     TWellKnownEntityTypes::TABLET,
                     TabletId,
                     "Cleanup queue: blob already in cleanup queue");
-                Y_DEBUG_ABORT_UNLESS(inserted);
                 if (inserted) {
                     db.WriteCleanupQueue(kv.first, DeletionCommitId);
                 }
--- a/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
@@ -1740,10 +1740,12 @@ void PrepareRangeCompaction(
             !readBlockMaskOnCompactionOptimizationEnabled)
         {
             if (db.ReadBlockMask(kv.first, kv.second.BlockMask)) {
-                Y_ABORT_UNLESS(
+                STORAGE_VERIFY_C(
                     kv.second.BlockMask.Defined(),
-                    "Could not read block mask for blob: %s",
-                    ToString(MakeBlobId(tabletId, kv.first)).data());
+                    TWellKnownEntityTypes::TABLET,
+                    tabletId,
+                    TStringBuilder() << "Could not read block mask for blob: "
+                                     << MakeBlobId(tabletId, kv.first));
             } else {
                 ready = false;
             }
@@ -1755,10 +1757,12 @@ void PrepareRangeCompaction(
 
         if (args.ChecksumsEnabled) {
             if (db.ReadBlobMeta(kv.first, kv.second.BlobMeta)) {
-                Y_ABORT_UNLESS(
+                STORAGE_VERIFY_C(
                     kv.second.BlobMeta.Defined(),
-                    "Could not read blob meta for blob: %s",
-                    ToString(MakeBlobId(tabletId, kv.first)).data());
+                    TWellKnownEntityTypes::TABLET,
+                    tabletId,
+                    TStringBuilder() << "Could not read blob meta for blob: "
+                                     << MakeBlobId(tabletId, kv.first));
             } else {
                 ready = false;
             }
--- a/cloud/blockstore/libs/storage/partition/part_events_private.h
+++ b/cloud/blockstore/libs/storage/partition/part_events_private.h
@@ -118,7 +118,7 @@ struct TWriteFreshBlocksRequest
 
 struct TAffectedBlob
 {
-    // All blocks in the blob contained in BlockRangeHint. Actual blob range can
+    // All blocks in the blob contained in BlobRangeHint. Actual blob range can
     // be smaller.
     TBlockRange32 BlobRangeHint = TBlockRange32::Max();
     TVector<ui16> Offsets;


cloud/blockstore/libs/storage/partition/part_actor_cleanup.cpp
```diff
@@ -312,6 +310,11 @@ void TPartitionActor::CompleteCleanup(
         LogTitle.GetWithTime().c_str(),
         args.CommitId);
 
+    for (const auto& item: args.CleanupQueue) {
+        State->RemoveCleanupQueueItem(item);
+    }
+
+
```
vladstepanyuk: лишнее

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -684,17 +684,23 @@ void TCompactionActor::AddBlobs(const TActorContext& ctx)
...
-            // could already be full
-            if (IsBlockMaskFull(blockMask, MaxBlocksInBlob)) {
-                continue;
+                // mask overwritten blocks
+                for (ui16 blobOffset: blob.Offsets) {
+                    blockMask->Set(blobOffset);
+                }
+            } else {
+                TBlockMask mask;
+                mask.Set(0, MaxBlocksInBlob);
```
qkrorlqr: which blocks in this affected blob are actually overwritten then?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1701,12 +1717,19 @@ void PrepareRangeCompaction(
...
     for (auto& kv: args.AffectedBlobs) {
-        if (db.ReadBlockMask(kv.first, kv.second.BlockMask)) {
-            Y_ABORT_UNLESS(kv.second.BlockMask.Defined(),
-                "Could not read block mask for blob: %s",
-                ToString(MakeBlobId(tabletId, kv.first)).data());
-        } else {
-            ready = false;
+        auto& blobRange = kv.second.BlobRange;
+        bool compactRangeContainsBlob =
+            blobRange && args.BlockRange.Contains(*blobRange);
```
qkrorlqr: how can `blobRange` be equivalent to `false`?

IMO the code would be clearer if we always initialized this range

cloud/blockstore/libs/storage/partition/part_database.cpp
```diff
@@ -365,6 +365,7 @@ void TPartitionDatabase::DeleteMergedBlocks(
 
 bool TPartitionDatabase::FindMergedBlocks(
     IBlocksIndexVisitor& visitor,
+    IBlobsVisitor* blobsVisitor,
```
qkrorlqr: let's just use no-op visitor if we don't want to process anything

instead of having a possibly null visitor ptr and then having to check whether it's null

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
qkrorlqr: did you actually mean to write `mask.Set(1, MaxBlocksInBlob)`?

cloud/blockstore/libs/storage/partition/part_actor_cleanup.cpp
```diff
@@ -312,6 +310,10 @@ void TPartitionActor::CompleteCleanup(
         LogTitle.GetWithTime().c_str(),
         args.CommitId);
 
+    for (const auto& item: args.CleanupQueue) {
+        State->RemoveCleanupQueueItem(item);
```
qkrorlqr: why do we need to move this call to the `Complete` stage?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
vladstepanyuk: We overwrite all blocks in blob
```
// Fills the specified [start, end) bit range by the 1. Other bits are kept unchanged
    TThis& Set(size_t start, size_t end)
```

cloud/blockstore/config/storage.proto
```diff
@@ -1474,4 +1474,7 @@ message TStorageServiceConfig
     // Maximum number of inflight attach/detach path requests.
     // VolumeBalancerMaxInProgress == 0 means no limit
     optional uint64 VolumeBalancerMaxInProgress = 488;
+
+    // Enables block mask optimization.
+    optional bool BlockMaskOptimizationEnabled = 489;
```
komarevtsev-d: Может быть более конкретными? 
Типа ReadBlockMaskOnCompactionOptimazation?
Ну или хотя бы одно из слов Read/Compaction

cloud/blockstore/config/storage.proto
```diff
@@ -1528,4 +1528,8 @@ message TStorageServiceConfig
 
     // Shaping throttler configuration.
     optional TShapingThrottlerConfig ShapingThrottlerConfig = 490;
+
+    // Don't read block masks on compaction if blob entirely contains
```
qkrorlqr: nit: "entirely contains" -> "is entirely contained" or "lies entirely"

cloud/blockstore/libs/storage/partition/model/cleanup_queue.cpp
```diff
@@ -20,20 +20,38 @@ struct TCleanupQueue::TImpl
...
 
     bool Remove(const TCleanupQueueItem& item)
     {
-        auto it = Items.find(item);
-        if (it != Items.end()) {
-            Items.erase(it);
-            return true;
+        auto itBlob = BlobIds.find(item.BlobId);
+        if (itBlob == BlobIds.end()) {
+            return false;
```
qkrorlqr, line 41: let's also verify that `Items` doesn't contain `item` - in debug builds probably?

cloud/blockstore/libs/storage/partition/model/cleanup_queue.cpp
```diff
@@ -20,20 +20,38 @@ struct TCleanupQueue::TImpl
...
 
     TSet<TCleanupQueueItem, TLess> Items;
+    THashSet<TPartialBlobId, TPartialBlobIdHash> BlobIds;
 
     bool Add(const TCleanupQueueItem& item)
     {
-        return Items.insert(item).second;
+        bool result = BlobIds.insert(item.BlobId).second;
+        if (result) {
+            Items.insert(item);
```
qkrorlqr: seems that we need to verify that this insertion succeeds

cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -576,8 +576,18 @@ class TAddBlobsExecutor
...
-                State.GetCleanupQueue().Add({ kv.first, DeletionCommitId });
+                bool inserted =
+                    State.GetCleanupQueue().Add({kv.first, DeletionCommitId});
+
+                STORAGE_VERIFY_DEBUG_C(
+                    inserted,
+                    TWellKnownEntityTypes::TABLET,
+                    TabletId,
+                    "Cleanup queue: blob already in cleanup queue");
+                Y_DEBUG_ABORT_UNLESS(inserted);
```
qkrorlqr: you have already verified it in the line above this line

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1701,17 +1724,30 @@ void PrepareRangeCompaction(
...
+        {
+            if (db.ReadBlockMask(kv.first, kv.second.BlockMask)) {
+                Y_ABORT_UNLESS(
+                    kv.second.BlockMask.Defined(),
+                    "Could not read block mask for blob: %s",
+                    ToString(MakeBlobId(tabletId, kv.first)).data());
+            } else {
+                ready = false;
+            }
+        } else if (state.GetCleanupQueue().HasBlob(kv.first)) {
```
qkrorlqr, line 1752: how is this situation possible? if the blob is already in the cleanup queue, then it should not be read from the index - it's not visible, it's already been overwritten by other blobs

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1701,17 +1724,30 @@ void PrepareRangeCompaction(
...
+            // If the blob is in the cleanup queue, we should not try to add
+            // this blob to cleanup queue again.
+            kv.second.BlockMask = GetFullBlockMask();
         }
 
         if (args.ChecksumsEnabled) {
             if (db.ReadBlobMeta(kv.first, kv.second.BlobMeta)) {
-                Y_ABORT_UNLESS(kv.second.BlobMeta.Defined(),
+                Y_ABORT_UNLESS(
```
qkrorlqr: STORAGE_VERIFY_C

cloud/blockstore/libs/storage/partition/part_events_private.h
```diff
@@ -118,6 +118,9 @@ struct TWriteFreshBlocksRequest
 
 struct TAffectedBlob
 {
+    // All blocks in the blob contained in BlockRangeHint. Actual blob range can
```
qkrorlqr: BlockRangeHint -> BlobRangeHint

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
vladstepanyuk: The blob should be invisible only after the execute cleanup transaction is completed, until then we can read it from the index. If we start two compactions in the same 4MB region and there is no cleanup between them, the first compaction will put the blob in the cleanup queue, and the second will read it from index. However, we should not attempt to add it back to the cleanup queue. Additionally, we already have a deduplication logic for cleanup queue items in TCompactionActor. https://github.com/ydb-platform/nbs/blob/9a18fb05d603e31939b14810068c1bb4a7a7e97b/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp#L697-L700

komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -576,8 +576,17 @@ class TAddBlobsExecutor
             db.WriteBlockMask(kv.first, blockMask);
 
             if (IsBlockMaskFull(blockMask, MaxBlocksInBlob)) {
-                db.WriteCleanupQueue(kv.first, DeletionCommitId);
-                State.GetCleanupQueue().Add({ kv.first, DeletionCommitId });
+                bool inserted =
+                    State.GetCleanupQueue().Add({kv.first, DeletionCommitId});
+
+                STORAGE_VERIFY_DEBUG_C(
+                    inserted,
```
qkrorlqr, line 583: probably it's even better to change this to "impossible event"

can be done in a separate PR

PR #5596, opened 2026-03-26
issue-5507: implemented file name hashing for tablet dir viewer

qkrorlqr:
### Notes
Enabled via the `HideFileNamesInTabletDirectoryViewer` storage config flag.

| HideFileNamesInTabletDirectoryViewer=false | HideFileNamesInTabletDirectoryViewer=true |
| --- | ----------- |
| Screenshot from 2026-03-26 19-25-43 | Screenshot from 2026-03-26 19-25-26 |

### Issue
https://github.com/ydb-platform/nbs/issues/5507

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring_directory_viewer.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_monitoring_directory_viewer.cpp
@@ -446,6 +446,10 @@ void TDirViewerActor::ReplyAndDie(const TActorContext& ctx)
         for (auto& name: Names) {
             TStringBuf sbuf(name);
             size_t pos = sbuf.rfind('.');
+            const ui32 maxExtensionLength = 4;
+            if (pos == 0 || pos + 1 + maxExtensionLength < sbuf.size()) {
+                pos = TString::npos;
+            }
             TStringBuilder maskedName;
             maskedName << MD5::Calc(sbuf.substr(0, pos));
             if (pos != TString::npos) {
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_monitoring_directory_viewer.cpp
```diff
@@ -437,6 +442,19 @@ void TDirViewerActor::HandlePoisonPill(
 
 void TDirViewerActor::ReplyAndDie(const TActorContext& ctx)
 {
+    if (HideFileNames) {
+        for (auto& name: Names) {
+            TStringBuf sbuf(name);
+            size_t pos = sbuf.rfind('.');
+            TStringBuilder maskedName;
+            maskedName << MD5::Calc(sbuf.substr(0, pos));
```
debnatkh, line 454: Is it reasonable to reuse `Transform` here?

https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/tools/analytics/profile_tool/lib/mask.cpp#L153

qkrorlqr: I don't think that it's necessary to extract these couple lines into a separate common func. The least common peerdir would be somewhere in `cloud/storage/core/libs/common` then. Or we'll need to create `cloud/filestore/libs/common` specifically for this function.

qkrorlqr: but I'll add extension len and pos check into my func

qkrorlqr: done

PR #5598, opened 2026-03-27
issue-5590: dont read block masks for mixed blobs if its completely in one range

vladstepanyuk:
### Notes

Added `CompactionRangeCount` field to mixed index: number of compaction ranges blob overlaps with.
if `CompactionRangeCount` for blob equals to 1, block masks are not readed

In a local db, it is stored in a single column with the BlobOffset value.
### Issue
#5590

### DIFF
--- a/cloud/blockstore/libs/storage/partition/model/mixed_index_cache.h
+++ b/cloud/blockstore/libs/storage/partition/model/mixed_index_cache.h
@@ -18,38 +18,38 @@ struct TMixedBlock
     ui64 CommitId;
     ui32 BlockIndex;
     ui16 BlobOffset;
-    ui8 CompactionRangeCountOverlapped;
+    ui8 CompactionRangeCount;
 
     TMixedBlock(
             TPartialBlobId blobId,
             ui64 commitId,
             ui32 blockIndex,
             ui16 blobOffset,
-            ui8 compactionRangeCountOverlapped)
+            ui8 compactionRangeCount)
         : BlobId(blobId)
         , CommitId(commitId)
         , BlockIndex(blockIndex)
         , BlobOffset(blobOffset)
-        , CompactionRangeCountOverlapped(compactionRangeCountOverlapped)
+        , CompactionRangeCount(compactionRangeCount)
     {}
 
     bool operator==(const TMixedBlock& other) const
     {
         return BlockIndex == other.BlockIndex && CommitId == other.CommitId &&
                BlobId == other.BlobId && BlobOffset == other.BlobOffset &&
-               CompactionRangeCountOverlapped ==
-                   other.CompactionRangeCountOverlapped;
+               CompactionRangeCount ==
+                   other.CompactionRangeCount;
     }
 };
 
 static_assert(sizeof(TMixedBlock) == 32);
 
-// BlobOffset and CompactionRangeCountOverlapped are stored in a single 32-bit
+// BlobOffset and CompactionRangeCount are stored in a single 32-bit
 // integer in the local database.
 static_assert(sizeof(TMixedBlock::BlobOffset) == 2);
-static_assert(sizeof(TMixedBlock::CompactionRangeCountOverlapped) == 1);
+static_assert(sizeof(TMixedBlock::CompactionRangeCount) == 1);
 static_assert(
-    sizeof(TMixedBlock::CompactionRangeCountOverlapped) +
+    sizeof(TMixedBlock::CompactionRangeCount) +
         sizeof(TMixedBlock::BlobOffset) <=
     4);
 
--- a/cloud/blockstore/libs/storage/partition/model/mixed_index_cache.h
+++ b/cloud/blockstore/libs/storage/partition/model/mixed_index_cache.h
@@ -18,16 +18,19 @@ struct TMixedBlock
     ui64 CommitId;
     ui32 BlockIndex;
     ui16 BlobOffset;
+    ui32 BlobAlignment;
 
     TMixedBlock(
             TPartialBlobId blobId,
             ui64 commitId,
             ui32 blockIndex,
-            ui16 blobOffset)
-       : BlobId(blobId)
-       , CommitId(commitId)
-       , BlockIndex(blockIndex)
-       , BlobOffset(blobOffset)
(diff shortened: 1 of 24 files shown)


cloud/blockstore/libs/storage/partition/model/mixed_index_cache.h
komarevtsev-d: тут не надо добавить BlobAlignment? Иначе зачем static_assert?

komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1693,12 +1725,19 @@ void PrepareRangeCompaction(
     args.ChecksumsEnabled = args.BlockRange.Start < checksumBoundary;
 
     for (auto& kv: args.AffectedBlobs) {
+        state.IncrementBlobsProcessedDuringCompaction();
+
         const bool compactRangeContainsBlob =
             args.BlockRange.Contains(kv.second.BlobRangeHint);
+        const bool blobOnlyInOneCompactRange =
+            state.GetCompactionMap().GetRangeSize() ==
+            kv.second.EnclosingCompactionRangeSize;
```
qkrorlqr: why do we need to store range size then? why not just store the number of compaction ranges that overlap with our blob?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1693,12 +1725,19 @@ void PrepareRangeCompaction(
...
 
-        if (!compactRangeContainsBlob ||
+        if ((!compactRangeContainsBlob && !blobOnlyInOneCompactRange) ||
```
qkrorlqr: I don't understand this condition. According to it there can be a situation when the blob is not fully contained in the range that we're processing (which is a compaction range) BUT at the same time it's contained in a single compaction range? Which range is it then?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1693,12 +1725,19 @@ void PrepareRangeCompaction(
...
             !readBlockMaskOnCompactionOptimizationEnabled)
         {
+            state.IncrementBlockMaskReadedDuringCompaction();
```
qkrorlqr: Readed -> Read

// the forms of "to read" are "read / read / read"

cloud/blockstore/libs/storage/partition/part_schema.h
```diff
@@ -111,14 +111,19 @@ struct TPartitionSchema
         {
         };
 
+        struct EnclosingCompactionRangeSize
+            : public Column<6, NKikimr::NScheme::NTypeIds::Uint32>
```
qkrorlqr: keep in mind that extra columns are not free

you might want to encode this extra info in one of the other columns

vladstepanyuk: I changed the EnclosingCompactionRangeSize to CompactionRangeCountOverlaped, which is of type ui8, so I added only one byte for each row, it's not that expensive.

Encoding this information in BlobOffset or BlobId columns can be difficult to understand and may lead to a bug.

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
vladstepanyuk: https://github.com/ydb-platform/nbs/blob/4c2d3457892d0ea781adbd4bf322e078d18e0e52/cloud/blockstore/libs/storage/partition/part_events_private.h#L127-L128

BlobRangeHint is filled only for merged blobs, for mixed blobs it is equal to `TBlockRange32::Max()` 

I can fill it based on `CompactionRangeCountOverlaped`, but it's more complicated than just looking at the `CompactionRangeCountOverlaped` field by itself.

cloud/blockstore/libs/storage/partition/part_schema.h
qkrorlqr: > so I added only one byte for each row, it's not that expensive.

there's constant overhead per each column and AFAIK it's pretty big - around 8 bytes or so (if not more)

cloud/blockstore/libs/storage/partition/part_actor_flush.cpp
```diff
@@ -510,21 +520,42 @@ class TFlushBlocksVisitor final
...
         }
     }
 
 private:
+    void FlushBlob(
+        TBlockBuffer blobContent,
+        TVector<TBlock> blocks,
+        TVector<ui32> checksums)
+    {
+        const auto compactionRangeCount =
```
qkrorlqr: auto -> ui32?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1079,8 +1084,35 @@ class TCompactionBlockVisitor final
...
+        const TPartialBlobId& blobId,
+        ui16 blobOffset,
+        ui8 compactionRangeCount) override
+    {
+        if (commitId > MaxCommitId) {
+            return true;
+        }
+
+        auto& ab = Args.AffectedBlobs[blobId];
+        ab.CompactionRangeCount = compactionRangeCount;
```
qkrorlqr, line 1102: as I can see, this should in theory be forward-compatible - if you deploy this code and then roll it back then the old code will probably work with the new BlobOffset values because it seems to just cast them to ui16 without any checks

but I would prefer to see a feature flag w/o which CompactionRangeCount wouldn't be calculated at all

vladstepanyuk: https://github.com/ydb-platform/nbs/blob/66c9a4c3ff8c0f4e58f8e6df0feb19a713869a44/cloud/blockstore/libs/storage/partition/part_actor_flush.cpp#L551-L556
Filling CompactionRangeCount only if Block Mask Optimization enabled

qkrorlqr (approved):

komarevtsev-d (approved):

PR #5604, opened 2026-03-28
[Filestore] Allow module stats to be registered multiple times for the same filesystem/client pair.

e673:
### Notes

During migration, a new session with the same FileSystem/ClientId pair is created. If migration happens within the same node, a module may be registered twice.

Previous behavior: associate module metrics with FileSystem/ClientId pair.

New behavior: associate module metrics with SessionId, allow registering module metrics without restrictions.

Resolve failed test `cloud/filestore/tests/fio_index_migration/qemu-intrahost-migration-kikimr-nemesis-test`

```
VERIFY failed (2026-03-28T01:08:13.811715Z): Module DirectoryHandles is already registered for (fsId: nfs_test, clientId: )
  cloud/filestore/libs/diagnostics/module_stats.cpp:121
  Register(): requirement !fileSystemStatsEntry.ModuleStatsMap.contains(moduleName) failed
```

### DIFF
--- a/cloud/filestore/libs/diagnostics/module_stats.cpp
+++ b/cloud/filestore/libs/diagnostics/module_stats.cpp
@@ -19,36 +19,18 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TFileSystemKey
-{
-    const TString FileSystemId;
-    const TString ClientId;
-
-    auto AsTuple() const
-    {
-        return std::tie(FileSystemId, ClientId);
-    }
-
-    bool operator==(const TFileSystemKey& other) const
-    {
-        return AsTuple() == other.AsTuple();
-    }
-
-    // Used as hash function in THashMap
-    explicit operator size_t() const
-    {
-        return ComputeHash(AsTuple());
-    }
-};
+// (TFileSystemId, TClientId)
+using TFileSystemKey = std::pair<TString, TString>;
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TFileSystemEntry
+struct TFileSystemMetricsRegistryEntry
 {
     const IMainMetricsRegistryPtr FileSystemMetricsRegistry;
     ui64 RefCount = 0;
 
-    explicit TFileSystemEntry(NMonitoring::TDynamicCountersPtr fsCounters)
+    explicit TFileSystemMetricsRegistryEntry(
+        NMonitoring::TDynamicCountersPtr fsCounters)
         : FileSystemMetricsRegistry(
               CreateMetricsRegistry({}, std::move(fsCounters)))
     {}
@@ -58,7 +40,8 @@ struct TFileSystemEntry
 
 struct TModuleStatsEntry
 {
-    TFileSystemKey FileSystemKey;
+    TString FileSystemId;
+    TString ClientId;
 
     // DynamicCounters are automatically deleted when all registries referencing
     // them are deleted
@@ -79,7 +62,8 @@ private:
     TAdaptiveLock Lock;
 
     // Map: Fs -> local metrics registry + ref count
-    THashMap<TFileSystemKey, TFileSystemEntry> FileSystemMetricsRegistries;
+    THashMap<TFileSystemKey, TFileSystemMetricsRegistryEntry>
+        FileSystemMetricsRegistries;
 
     // Map: SessionId -> registered module metrics per session
     THashMap<TString, TVector<TModuleStatsEntry>> StatsPerSession;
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/diagnostics/module_stats_ut.cpp
```diff
@@ -307,13 +337,70 @@ Y_UNIT_TEST_SUITE(TModuleStatsRegistryTest)
...
+            "TestModule",
+            "fs1",
+            "client1",
+            "session1");
+
+        auto stats2 = b.CreateAndRegisterStats(
+            "TestModule",
+            "fs1",
+            "client1",
+            "session1");
```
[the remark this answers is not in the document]
e673: Added a test case

cloud/filestore/libs/diagnostics/module_stats.cpp
```diff
@@ -19,36 +19,54 @@ namespace {
...
-    IMetricsRegistryPtr AggregatableMetricsRegistry;
-    IModuleStatsPtr ModuleStats;
+    const TString FileSystemId;
+    const TString ClientId;
+    const TString CloudId;
+    const TString FolderId;
+
+    auto AsTuple() const
+    {
+        return std::tie(FileSystemId, ClientId, CloudId, FolderId);
```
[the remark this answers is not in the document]
e673: I just didn't like std::pair in the previous code.
Removed CloudId and FolderId from the key.

cloud/filestore/libs/diagnostics/module_stats.cpp
```diff
@@ -132,26 +149,66 @@ class TModuleStatsRegistry final: public IModuleStatsRegistry
...
+        args.ModuleStats->RegisterCounters(
             *localMetricsRegistry,
             *aggregatableMetricsRegistry);
 
-        fileSystemStatsEntry.ModuleStatsMap[moduleName] = {
-            .LocalMetricsRegistry = std::move(localMetricsRegistry),
-            .AggregatableMetricsRegistry =
-                std::move(aggregatableMetricsRegistry),
-            .ModuleStats = std::move(stats)};
+        StatsPerSession[args.SessionId].push_back(
```
neihar, line 137: low: emplace_back?

cloud/filestore/libs/diagnostics/module_stats.cpp
```diff
@@ -61,7 +77,12 @@ class TModuleStatsRegistry final: public IModuleStatsRegistry
     IMainMetricsRegistryPtr TotalMetricsRegistry;
 
     TAdaptiveLock Lock;
-    THashMap<std::pair<TString, TString>, TFileSystemStatsEntry> StatsMap;
+
+    // Map: Fs -> local metrics registry + ref count
+    THashMap<TFileSystemKey, TFileSystemEntry> FileSystemMetricsRegistries;
+
+    // Map: SessionId -> registered module metrics per session
+    THashMap<TString, TVector<TModuleStatsEntry>> StatsPerSession;
```
neihar, line 69: while current solution is okay for me, I still think, that expressing it via seqno could be better to emphasize that this logic related only to local live migrations.

cloud/filestore/libs/diagnostics/module_stats.cpp
```diff
@@ -19,36 +19,52 @@ namespace {
...
+    explicit operator size_t() const
+    {
+        return ComputeHash(AsTuple());
+    }
 };
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TFileSystemStatsEntry
+struct TFileSystemEntry
```
neihar: probably better TFileSystemMetricsRegistryEntry

cloud/filestore/libs/diagnostics/module_stats.cpp
```diff
@@ -19,36 +19,52 @@ namespace {
...
+    bool operator==(const TFileSystemKey& other) const
+    {
+        return AsTuple() == other.AsTuple();
+    }
+
+    // Used as hash function in THashMap
+    explicit operator size_t() const
+    {
+        return ComputeHash(AsTuple());
+    }
```
neihar: using TFileSystemKey = std::pair<TString, TString>; ??

neihar (dismissed):

cloud/filestore/libs/diagnostics/module_stats.cpp
e673: I prefer explicitly named fields (especially when they have the same types) but emplace_back does not work with designated initializers.

cloud/filestore/libs/diagnostics/module_stats.cpp
e673: I don't like the idea of seqno because it will create an additional entity and increase the cognitive complexity.

`SessionId` is more than ok here: we may register any number of modules within the session - and they are unregistered at once at session destroy.

cloud/filestore/libs/diagnostics/module_stats.cpp
e673: Renamed

cloud/filestore/libs/diagnostics/module_stats.cpp
e673: There were 4 fields in the initial PR.
Reverted back to `std::pair`

PR #5613, opened 2026-03-30
[NBS] BlocksPerStripe in TVolume and use it in SplitRequestService

zavik001:
continue https://github.com/ydb-platform/nbs/pull/5282

### DIFF
--- a/cloud/blockstore/libs/service/split_request_service.cpp
+++ b/cloud/blockstore/libs/service/split_request_service.cpp
@@ -408,11 +408,8 @@ TSplitConfig TSplitConfig::Create(const NProto::TVolume& volume)
 
     if (IsBlobStorageMediaKind(volume.GetStorageMediaKind())) {
         if (volume.GetPartitionsCount() > 1) {
-            // Need to find a way to get the size of the stripe for
-            // multipartition YDB-based volume.
-            // Used default BytesPerStripe from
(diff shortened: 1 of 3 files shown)


cloud/blockstore/libs/service/split_request_service.cpp
```diff
@@ -408,11 +408,8 @@ TSplitConfig TSplitConfig::Create(const NProto::TVolume& volume)
 
     if (IsBlobStorageMediaKind(volume.GetStorageMediaKind())) {
         if (volume.GetPartitionsCount() > 1) {
-            // Need to find a way to get the size of the stripe for
-            // multipartition YDB-based volume.
-            // Used default BytesPerStripe from
-            // cloud/blockstore/libs/storage/core/config.cpp
-            stripeSize = 16_MB;
+            stripeSize =
+                static_cast<size_t>(volume.GetBlocksPerStripe()) * blockSize;
```
qkrorlqr, line 412: but this field is empty for the existing volumes, why will this change be backwards-compatible?

drbasic: Here we take this number and trust it.
https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/storage/volume/volume_actor_forward.cpp#L246

PR #5616, opened 2026-03-30
Disable local service if is_tight_service_memory_platform is set in infra throttling config file

efimov-ivan-yteam:
Support for `ClientConfig::ThrottlingConfig::InfraThrottlingConfigPath`s `is_tight_service_memory_platform` flag
This flag denotes that the host has more constrained less RAM capacity than usual
Supporting this flag allows NBS to prevent enabling Local service on such nodes to avoid undesired memory consumption by brought-in hive's tablets

### DIFF
--- a/cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
+++ b/cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
@@ -846,6 +846,10 @@ void TBootstrapYdb::InitKikimrService()
 
     STORAGE_INFO("Local NVMe service initialized");
 
+    const bool isHiveLocalServiceEnabled =
+        !Configs->StorageConfig->GetDisableLocalService() &&
+        !Configs->HostPerformanceProfile.IsTightServiceMemoryPlatform;
+
     NStorage::TServerActorSystemArgs args;
     args.ModuleFactories = ModuleFactories;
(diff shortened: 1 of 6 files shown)


cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
```diff
@@ -846,6 +846,10 @@ void TBootstrapYdb::InitKikimrService()
 
     STORAGE_INFO("Local NVMe service initialized");
 
+    auto isLocalEnabled =
```
komarevtsev-d: `const bool`
Лучше для базовых типов auto не пользоваться. Буквы всё равно не экономятся

cloud/blockstore/libs/storage/init/server/actorsystem.cpp
```diff
@@ -391,6 +391,7 @@ class TCustomLocalServiceInitializer final
     const NServer::IEndpointEventHandlerPtr EndpointEventHandler;
     const TPartitionBudgetManagerPtr PartitionBudgetManager;
     const bool IsDiskRegistrySpareNode;
+    const bool IsLocalEnabled;
```
SvartMetal: IsLocalServiceEnabled

or more explicitly IsHiveLocalServiceEnabled

PR #5620, opened 2026-03-30
[Disk Manager] Do not create bucket if it exists

jkuradobery:
It is useful to create S3 buckets externally and issue permissions per-bucket. 
Bucket creation permission can't be assigned per-bucket, since per-bucket permissions require bucket to exist.

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/schema/schema.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/schema/schema.go
@@ -122,7 +122,9 @@ func Create(
 		if err != nil {
 			return err
 		}
-
+		// It is useful to create S3 buckets externally and assign permissions
+		// per-bucket. Bucket creation permissions can't be assigned per-bucket,
+		// since per-bucket permissions require the bucket to exist.
 		if !exists {
 			err = s3.CreateBucket(ctx, config.GetS3Bucket())
 			if err != nil {


cloud/tasks/persistence/s3.go
```diff
@@ -158,6 +158,33 @@ func (c *S3Client) CreateBucket(
 
 ////////////////////////////////////////////////////////////////////////////////
 
+func (c *S3Client) BucketExists(
+	ctx context.Context,
+	bucket string,
+) (exists bool, err error) {
```
SvartMetal, line 164: \n

cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/schema/schema.go
```diff
@@ -118,12 +118,19 @@ func Create(
 	logging.Info(ctx, "Created chunk_map table")
 
 	if s3 != nil && len(config.GetS3Bucket()) != 0 {
-		err = s3.CreateBucket(ctx, config.GetS3Bucket())
+		ok, err := s3.BucketExists(ctx, config.GetS3Bucket())
```
SvartMetal: ok -> exists

cloud/disk_manager/internal/pkg/dataplane/snapshot/storage/schema/schema.go
```diff
@@ -118,12 +118,19 @@ func Create(
...
 
 	if s3 != nil && len(config.GetS3Bucket()) != 0 {
-		err = s3.CreateBucket(ctx, config.GetS3Bucket())
+		exists, err := s3.BucketExists(ctx, config.GetS3Bucket())
 		if err != nil {
 			return err
 		}
 
-		logging.Info(ctx, "Created s3 bucket")
+		if !exists {
```
SvartMetal, line 128: Let's add a comment from description here

Issue #4823, opened 2025-12-11
Safe way to enable Discard/ZeroBlocks feature.


gy2411:
There is a VhostDiscardEnabled feature in the service config https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/config/server.proto#L243. However, this feature is non-revertible: once a disk starts supporting discards, there is no way to turn that off. We need the safest possible way to enable discards.

Plan:

- Add a VhostDiscardEnabled flag to VolumeConfig in Schemeshard https://github.com/ydb-platform/nbs/blob/main/contrib/ydb/core/protos/blockstore_config.proto#L34. Note that once this flag has been set, it cannot be unset.
- Add a feature to the storage config that will set the VhostDiscardEnabled flag in VolumeConfig when creating new disks. This feature can be enabled for specific clouds/folders, as well as for a percentage of clouds/folders.
- Add a feature to the storage config that will set the VhostDiscardEnabled flag in VolumeConfig for already existing disks. This feature can also be enabled for specific clouds/folders, as well as for a percentage of clouds/folders.

This way, discards can be rolled out gradually.

Approximate order of the feature deployment:
1. Turn on VhostDiscardEnabled in SchemeShard for new volumes on test clouds/folders.
2. Turn on VhostDiscardEnabled in SchemeShard for small persentage of clouds/folders.
3. Turn on VhostDiscardEnabled in SchemeShard for all persentage of clouds/folders.
4. Like 1-3, but for existing volumes. 
5. Trun on VhostDiscardEnabled in server config.

PR #5625, opened 2026-03-31
[Blockstore] Make possible to turn on Discard and Write Zeroes features separately in server config

gy2411:
### Notes

Now, when VhostDiscardEnabled is enabled, two options are turned on: VHD_BDEV_F_DISCARD and VHD_BDEV_F_WRITE_ZEROES (https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/vhost/vhost.cpp#L160). We want the possibility to enable these options separately.

So we add theVhostDiscardOnlyEnabled and VhostWriteZeroesEnabled flags.

In future we will mark VhostDiscardEnabled feature as deprecated.

### Issue
#4823

### DIFF
--- a/cloud/blockstore/libs/endpoints_vhost/vhost_server.h
+++ b/cloud/blockstore/libs/endpoints_vhost/vhost_server.h
@@ -19,7 +19,6 @@ IEndpointListenerPtr CreateVhostEndpointListener(
     NVhost::IServerPtr server,
     const NProto::TChecksumFlags& checksumFlags,
     bool vhostDiscardEnabled,
-    bool vhostDiscardOnlyEnabled,
     bool vhostWriteZeroesEnabled,
     ui32 maxZeroBlocksSubRequestSize,
     ui32 optimalIoSize);


cloud/blockstore/libs/server/config.cpp
```diff
@@ -106,6 +106,7 @@ constexpr TDuration Seconds(int s)
     xxx(VhostServerTimeoutAfterParentExit, TDuration,       Seconds(60)       )\
     xxx(ChecksumFlags,               NProto::TChecksumFlags, {}               )\
     xxx(VhostDiscardEnabled,         bool,                   false            )\
+    xxx(VhostWriteZeroesEnabled,     bool,                   true             )\
```
antonmyagkov: @gy2411 @komarevtsev-d 
I understand the intention, but we should avoid introducing features that are enabled by default.
Instead, consider adding two new features that are disabled by default: VhostDiscardOnlyEnabled (feel free to suggest a better name) and VhostWriteZeroesEnabled. The existing VhostDiscardEnabled feature can then be marked as deprecated.


```
options.WriteZeroesEnabled =
            (VhostDiscardEnabled || VhostWriteZeroesEnabled) && !IsDiskRegistryMediaKind(volume.GetStorageMediaKind());
```

antonmyagkov (changes requested):

cloud/blockstore/libs/server/config.cpp
gy2411: Yes, this makes sence. Now the flags VhostDiscardOnlyEnabled and VhostWriteZeroesEnabled are added.

I'm not sure about a better version of the name `VhostDiscardOnlyEnabled`. But this name seems ok since there is a clarifying commeb in the proto file and this name is used almost nowhere else.

PR #5626, opened 2026-03-31
[Blockstore] Drop discard requests when needed

gy2411:
### Notes

Add possibility to drop discard requests instead of executing ZeroBlocks. 

Discards are dropped in 3 cases:
1. If the volume has corresponding tag
2. If the corresponding flag in server config is on
3. If the disk is Disk Registry based.

1-2 will allow to turn on discards more safely.
3 is useful in case when a disk changes kind from ydb-based to disk registry based.

### Issue
https://github.com/ydb-platform/nbs/issues/4823

### DIFF
--- a/cloud/blockstore/libs/common/constants.h
+++ b/cloud/blockstore/libs/common/constants.h
@@ -29,6 +29,6 @@ constexpr TStringBuf SourceDiskIdTagName = "source-disk-id";
 
 constexpr TStringBuf UseFastPathTagName = "use-fastpath";
 
-constexpr TStringBuf DropDiscardRequestsTagName = "drop-discards-requests";
+constexpr TStringBuf DropDiscardRequestsTagName = "drop-discard-requests";
 
 }   // namespace NCloud::NBlockStore


cloud/blockstore/libs/endpoints_vhost/vhost_server.cpp
```diff
@@ -109,17 +120,33 @@ bool ShouldEnableVhostDiscardForVolume(
...
+bool ShouldDropDiscardRequestsForVolume(
+    bool dropDiscardRequests,
+    const NProto::TVolume& volume)
+{
+    return dropDiscardRequests ||
+           volume.GetTags().contains(DropDiscardRequestsTagName) ||
+           IsDiskRegistryMediaKind(
+               volume.GetStorageMediaKind());   // There is no implementation of
+                                                // ZeroBlocks for disk registry
+                                                // based disks yet.
```
komarevtsev-d: В коменте не совсем правда. Реализация ZeroBlocks-то как раз есть, но мы не хотим её использовать для дискрадов

PR #5630, opened 2026-03-31
fix Disconnected transition; resolves #5602

### LINKED ISSUE #5602: [NBS] TRequest::~TRequest double free
case #1:
- stacktrace – https://gist.github.com/yegorskii/07e0358f6a44006429807537b4f83c4c
- log – https://gist.github.com/tpashkin/2d54e58f62207f5a36b6395cec3acded

case #2
- stacktrace – https://gist.github.com/komarevtsev-d/10f65a486298672a4b1ce33c0ffca9a1
- log – https://gist.github.com/komarevtsev-d/092ec8a290234e413122297a9177bac9

Endpoint can be mistakingly transitioned to the `Disconnected` state when `ActiveRequests` still isn't empty – https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/impl/client.cpp#L1742. Unfortunately, we've recently stopped aborting requests in `Disconnected` state – https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/impl/client.cpp#L969. This can ultimately lead to some requests surviving reconnect. But during reconnect, buffer pool gets recreated, so the memory these requests are pointing to will be freed. When these requests eventually time out, we will try to free the same buffers again

Interestingly we've recieved an error completion around the same time we were supposed to `AbortRequest` for such dangling request
```
SEND C6229471.7.3F7 [wr_id=C6229471.7.3F7 opcode=SEND status=IBV_WC_LOC_PROT_ERR vendor_err=4] failed with IBV_WC_LOC_PROT_ERR
```
I suppose this could have been triggered by writing to the same memory this WR has been pointing to

tpashkin:


cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -915,6 +932,14 @@ bool TClientEndpoint::HandleCancelRequests() noexcept
...
+        if (!CheckState(EEndpointState::Connected)) {
+            RDMA_ERROR(
+                __func__ << " has been called in " << State
+                         << " state, expected Connected");
+            ReportRdmaError("invalid state");
+            return false;
+        }
     }
 
     if (!CheckState(EEndpointState::Connected)) {
```
budevg: so this check in Busy/AdaptiveWait should be removed now as in other places ?

cloud/blockstore/libs/rdma/impl/client.cpp
```diff
@@ -963,12 +988,18 @@ void TClientEndpoint::AbortRequests() noexcept
...
+        if (!CheckState(EEndpointState::Disconnecting)) {
+            RDMA_ERROR(
+                __func__ << " has been called in " << State
+                         << " state, expected Disconnecting");
+            ReportRdmaError("invalid state");
+            return;
+        }
     }
 
+    CancelRequests.DequeueAll();
```
budevg: so now canceled requests will be aborted with E_RDMA_UNAVAILABLE instead of E_CANCELLED shouldn't we use the same logic as in HandleCancelRequests to abort canceled requests with the right code ?

cloud/blockstore/libs/rdma/impl/client_ut.cpp
```diff
@@ -409,6 +409,71 @@ Y_UNIT_TEST_SUITE(TRdmaClientTest)
...
+            std::make_unique<TNullContext>(),
+            4096,   // requestBytes
+            4096);  // responseBytes
+
+        endpoint->SendRequest(
+            request.ExtractResult(),
+            MakeIntrusive<TCallContext>());
+
+        Disconnect(testContext);
+        disconnected.WaitT(5s);
```
budevg: shouldn't we assert that the completion handler (HandleResponse) was actually called

cloud/blockstore/libs/rdma/impl/client.cpp
tpashkin: Forgot that one

cloud/blockstore/libs/rdma/impl/client_ut.cpp
tpashkin: We could, but it's not strictly necessary, because this test actually reproduces the original double free

cloud/blockstore/libs/rdma/impl/client.cpp
tpashkin: Not quite. There was a condition in `CancelRequests` that prevented it from running in `Disconnecting` state, so `CancelRequests` basically leaked. At worst it would survive reconnect and try to cancel request leading to another double free

Besides, connection went down before we got a chance to actually cancel request, I don't think it's necessarily a bad thing to return `E_RDMA_UNAVAILABLE` instead of `E_CANCELLED` in this case

komarevtsev-d: `E_RDMA_UNAVAILABLE` is okay i think. 
`E_CANCELLED` will be changed to `E_REJECTED` with `EF_INSTANT_RETRIABLE` flag which is not needed here: https://github.com/ydb-platform/nbs/blob/89f5bdd/cloud/blockstore/libs/storage/partition_nonrepl/rdma_device_request_handler.cpp#L17

tpashkin: I've actually decided to go along with this suggestions, because `E_CANCELLED` (unlike `E_RDMA_UNAVAILABLE`) is a non-retriable error

> E_CANCELLED will be changed to E_REJECTED with EF_INSTANT_RETRIABLE flag which is not needed here: https://github.com/ydb-platform/nbs/blob/89f5bdd/cloud/blockstore/libs/storage/partition_nonrepl/rdma_device_request_handler.cpp#L17

Even if right now there will be discernible difference

PR #5704, opened 2026-04-09
issue-5694: [Disk Manager] Recreate session on E_FS_INVALID_SESSION in nfs-client

jkuradobery:
### Issue
#5694

### DIFF
--- a/cloud/disk_manager/internal/pkg/clients/nfs/interface.go
+++ b/cloud/disk_manager/internal/pkg/clients/nfs/interface.go
@@ -81,6 +81,8 @@ type Session interface {
 
 	SetSession(nfsSession nfs_client.Session)
 
+	GetID() string
+
 	Close(ctx context.Context) error
 }
 
--- a/cloud/disk_manager/internal/pkg/clients/nfs/interface.go
+++ b/cloud/disk_manager/internal/pkg/clients/nfs/interface.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/types"
+	nfs_client "github.com/ydb-platform/nbs/cloud/filestore/public/sdk/go/client"
 )
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -77,6 +78,8 @@ type Session interface {
 		name string,
 		unlinkDirectory bool,
 	) error
+
+	SetSession(nfsSession nfs_client.Session)
 
 	Close(ctx context.Context) error
 }
--- a/cloud/disk_manager/internal/pkg/clients/nfs/interface.go
+++ b/cloud/disk_manager/internal/pkg/clients/nfs/interface.go
@@ -126,6 +126,14 @@ type Client interface {
 		checkpointID string,
 		readonly bool,
 	) (Session, error)
+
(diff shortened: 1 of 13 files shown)


jkuradobery:
The problem with E_FS_INVALID_SESSION was caused by GetNodeAttrBatch sent from main tablet to shard tablets, 
and shard tablet was restarted (tablet should not validate session for shard, but it validates because of an issue with tablet-in-directory feature.

cloud/disk_manager/internal/pkg/clients/nfs/session.go
```diff
@@ -183,3 +192,196 @@ func (s *session) UnlinkNode(
...
+		newSession, err := s.nfs.CreateSession(
+			ctx,
+			s.filesystemID,
+			s.clientID,
+			s.checkpointID,
+			s.readonly,
+		)
+		if err != nil {
+			var zero T
+			return zero, err
```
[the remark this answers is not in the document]
jkuradobery: We wrap only client calls, wrapping error twice would be unwise

cloud/disk_manager/internal/pkg/clients/nfs/session.go
```diff
@@ -183,3 +194,207 @@ func (s *session) UnlinkNode(
...
+	var err error
+
+	for i := 0; i < SessionReEstablishMaxRetries+1; i++ {
+		result, err = callback()
+		if !isSessionInvalidError(err) {
+			return result, err
+		}
+
+		if i == SessionReEstablishMaxRetries {
+			continue
```
debnatkh: break?

cloud/disk_manager/internal/pkg/clients/nfs/interface.go
```diff
@@ -78,6 +79,10 @@ type Session interface {
 		unlinkDirectory bool,
 	) error
 
+	SetSession(nfsSession nfs_client.Session)
```
debnatkh, line 82: should it be public? Isn't it called only from within the implementation?

Same question about the next method

jkuradobery: We need GetSession in tests, which are not in the same package, also I want to pass interface and not a concrete to the wrapper, this way we would be able to add more middlewares in the future.

debnatkh (approved):

cloud/disk_manager/internal/pkg/clients/nfs/session.go
```diff
@@ -183,3 +194,207 @@ func (s *session) UnlinkNode(
...
+	filesystemID string
+	clientID     string
+	checkpointID string
+	readonly     bool
+}
+
+func requestWithReEstablishSession[T any](
+	ctx context.Context,
+	s *sessionWithReEstablish,
+	callback func() (T, error),
```
gy2411, line 212: Why this parameter is called 'callback', not 'doRequest' or something like that?

gy2411 (approved):

cloud/disk_manager/internal/pkg/clients/nfs/session.go
jkuradobery: I agree with you since it brings consistency and callback is a poor naming. I will fix it in the following PR.

PR #5705, opened 2026-04-09
expose rdma buffer pool parameters in the config; closes #5652

tpashkin:


cloud/blockstore/config/rdma.proto
```diff
@@ -15,6 +15,20 @@ enum EWaitMode
     WAIT_MODE_ADAPTIVE_WAIT = 2;
 }
 
+message TBufferPool
```
komarevtsev-d, line 18: `TBufferPoolParams`/`TBufferPoolConfig` ?

cloud/blockstore/libs/rdma/impl/buffer.cpp
```diff
@@ -110,6 +104,7 @@ class TBufferPool::TImpl
     using TChunkList = TIntrusiveListWithAutoDelete<TChunk, TDelete>;
 
 private:
+    TBufferPoolConfig Config;
```
komarevtsev-d: nit: const

cloud/blockstore/config/rdma.proto
```diff
@@ -15,6 +15,20 @@ enum EWaitMode
...
+{
+    // Memory chunks of this size will be kept for reuse. Buffers smaller or
+    // equal to ChunkSize will be baked by one of the available chunks
```
budevg: baked -> backed

cloud/blockstore/libs/rdma/iface/buffer.h
```diff
@@ -0,0 +1,16 @@
...
+
+namespace NCloud::NBlockStore::NRdma {
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TBufferPoolConfig
+{
+    size_t ChunkSize = 4_MB;
+    size_t MaxChunkAlloc = ChunkSize / 4;
+    size_t MaxFreeChunks = 10;
```
budevg, line 15: Probably need to add some sanity check that config make sense. For example MaxChunkAlloc > ChunkSize is a bad config. If such config is provided we need to replace it with safe config

cloud/blockstore/tools/testing/rdma-test/bootstrap.cpp
```diff
@@ -85,6 +88,9 @@ void TBootstrap::Init()
         config->IpTypeOfService = Options->Tos;
         config->SourceInterface = Options->SourceInterface;
         config->VerbsQP = Options->VerbsQP;
+        config->BufferPoolConfig.ChunkSize = Options->ChunkSize;
```
budevg: Don't you want to allow changing config in blockstore/disk-agent and not just rdma-test utility ?
https://github.com/ydb-platform/nbs/blob/users/tpashkin/configure-rdma-buffer-pool-parameters/cloud/blockstore/libs/rdma/iface/server.cpp#L50

cloud/blockstore/config/rdma.proto
tpashkin: These are all configs, so I went with the common theme here and left the Config part out 🤷

budevg (changes requested):

cloud/blockstore/tools/testing/rdma-test/bootstrap.cpp
tpashkin: Welp, this was the whole point... I'll fix it

cloud/blockstore/libs/rdma/iface/buffer.h
tpashkin: I'd rather leave it as is. Smart configs tend to make debugging a nightmare. We can break things with a client config too

cloud/blockstore/config/rdma.proto
tpashkin: oops

cloud/blockstore/config/rdma.proto
```diff
@@ -15,6 +15,20 @@ enum EWaitMode
...
+    uint64 ChunkSize = 1;
+
+    // Maximum buffer size, that will be baked by a chunk. Anything bigger will
```
SvartMetal: baked -> backed

cloud/blockstore/config/rdma.proto
```diff
@@ -15,6 +15,20 @@ enum EWaitMode
...
+    // spawn a custom chunk that won't be saved for reuse
```
SvartMetal: that won't be reused

cloud/blockstore/libs/rdma/iface/buffer.h
SvartMetal: > I'd rather leave it as is. Smart configs tend to make debugging a nightmare. We can break things with a client config too

Config validation is a good practice - it's better to reject configurations that can cause crashes, as they "tend to make debugging a nightmare"

SvartMetal: If configuration validation is properly logged - debugging is easy

cloud/blockstore/tools/testing/rdma-test/options.h
```diff
@@ -74,6 +74,11 @@ struct TOptions
     TString TracePath;
     ui32 TraceRate = 1;
 
+    // buffer pool options
+    size_t ChunkSize = 4_MB;
+    size_t MaxChunkAlloc = ChunkSize / 4;
+    size_t MaxFreeChunks = 10;
```
SvartMetal: Default values are copy-pasted here and there https://github.com/ydb-platform/nbs/pull/5705/changes#diff-dedc0d86c3461cd462579947de89aa0c176346c39f1a4e168e852b165efc058dR11

Let's have a single source of truth - a default config, for example

cloud/blockstore/libs/rdma/iface/buffer.h
tpashkin: There is nothing to check. MaxChunkAlloc > ChunkSize is perfectly valid configuration that will allocate custom chunks that won't be pooled

komarevtsev-d (approved):

budevg (approved):

PR #5721, opened 2026-04-13
added include to fix vfs fuse arcadia build

vladstepanyuk:
### Notes
Fixing build error
```
$(SOURCE_ROOT)/cloud/filestore/libs/vfs_fuse/loop.cpp:1289:27: error: use of undeclared identifier 'FUSE_CAP_POSIX_ACL'
 1289 |             conn->want |= FUSE_CAP_POSIX_ACL;
      |                           ^
1 error generated.

```

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/loop.cpp
+++ b/cloud/filestore/libs/vfs_fuse/loop.cpp
@@ -28,6 +28,8 @@
 #include <cloud/storage/core/libs/common/thread.h>
 #include <cloud/storage/core/libs/diagnostics/logging.h>
 
+#include <cloud/contrib/virtiofsd/fuse_common.h>
+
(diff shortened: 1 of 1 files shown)


cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -28,6 +28,8 @@
 #include <cloud/storage/core/libs/common/thread.h>
 #include <cloud/storage/core/libs/diagnostics/logging.h>
 
+#include <cloud/contrib/virtiofsd/fuse_common.h>
```
qkrorlqr, line 31: this include doesn't look correct - libs/vfs_fuse doesn't always link with virtiofsd

qkrorlqr (changes requested):

PR #5725, opened 2026-04-13
Fix using lkey for remote memory in RDMA; Use GTest in RDMA tests

komarevtsev-d:
### Notes
This PR fixes the usage of `lkey` of registered memory region for remote reading and writing. This worked because RDMA implementations use the same number for both keys.

Added a unit test `ShouldSetLkeyAndRkeyToBuffers` to verify this behavior. I found it more convenient to use gtest, as it has a very handy gmock framework. 

### Issue
#5707


cloud/blockstore/libs/rdma/impl/buffer_ut.cpp
```diff
@@ -1,96 +1,292 @@
 #include "buffer.h"
 
-#include <library/cpp/testing/unittest/registar.h>
+#include <library/cpp/testing/gtest/gtest.h>
```
budevg, line 3: GTEST is indeed very usefully but we don't use it across NBS project ... How hard would it be to just add the rkey/lkey validation test without rewriting all the unit tests to gtest ?

budevg (changes requested):

cloud/blockstore/libs/rdma/impl/buffer_ut.cpp
komarevtsev-d: We have a couple of gtests: `cloud/blockstore/vhost-server/gtest`, `cloud/blockstore/libs/diagnostics/gtest`.

I think that gtest is strictly better than unittest in every possible dimension: more comparison macros, mocking, better error messages, tests with parameters, etc.

The only problem with it is that, as of now, tests using the actor system can't be gtest because of YDB's runtime dependency on unittest. I've been dreaming of getting rid of this dependecy for a long time. 

Opensource projects should strive to well-known technologies :)

>  How hard would it be to just add the rkey/lkey validation ... 

Not that hard. Just stupid and irritating

budevg (approved):

vladstepanyuk (approved):

cloud/blockstore/libs/rdma/impl/buffer_ut.cpp
```diff
@@ -1,96 +1,292 @@
...
-        TBufferPool pool;
-        pool.Init(nullptr, nullptr, 0);
+    delete mr;
+    return 0;
+}
 
-        auto buffer = pool.AcquireBuffer(1234);
-        UNIT_ASSERT(buffer.Length == NSystemInfo::GetPageSize());
-    }
+class TVerbsMock: public NVerbs::IVerbs
```
tpashkin, line 24: Это же по сути TTestVerbs. Я не имею ничего против gtest, но не обязательно было переколбашивать все тесты ради одной функции

tpashkin: Заведу-ка я тикет на это – https://github.com/ydb-platform/nbs/issues/5738

PR #5735, opened 2026-04-14
Fix flapping test cloud/filestore/tests/fio/qemu-local-noserver-mq-test

proller:
```
Error occured: (yexception) library/cpp/eventlog/logparser.cpp:201: Could not read frame payload: premature end of stream
```

### DIFF
--- a/cloud/filestore/tests/fio/qemu-local-noserver-mq-test/test.py
+++ b/cloud/filestore/tests/fio/qemu-local-noserver-mq-test/test.py
@@ -21,7 +21,10 @@ def test_fio(name):
 
     fio.run_test(file_name, TESTS[name], fail_on_errors=True)
 
-    time.sleep(15)  # Waiting for profile logs to be written
+    # Sleep for a while to ensure that the profile log is flushed
+    # before we start analyzing it
+    # The default value of ProfileLogTimeThreshold for tests is 100ms
+    time.sleep(2)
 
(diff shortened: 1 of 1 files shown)


cloud/filestore/tests/fio/qemu-local-noserver-mq-test/test.py
```diff
@@ -20,6 +21,8 @@ def test_fio(name):
 
     fio.run_test(file_name, TESTS[name], fail_on_errors=True)
 
+    time.sleep(15)  # Waiting for profile logs to be written
```
qkrorlqr: 1. 15 seconds is too much
2. we have multiple 2-second sleeps in different tests: https://github.com/search?q=repo%3Aydb-platform%2Fnbs%20path%3A%2F%5Ecloud%5C%2Ffilestore%5C%2Ftests%5C%2F%2F%20%22Sleep%20for%20a%20while%20to%20ensure%20that%20the%20profile%20log%20is%20flushed%22&type=code - better to extract this into a function called `flush_profile_log()` with a TODO item saying that we shouldn't sleep and should explicitly trigger flushing via some private API method

qkrorlqr (changes requested):

qkrorlqr (changes requested):
Let's discard this, I created a better one: https://github.com/ydb-platform/nbs/pull/5740

Issue #5747, opened 2026-04-16
[Filestore] add memory deallocation for dynamic persistent table


neihar:
Currently, we do not deallocate memory for the dynamic persistent table. This causes problems when the table load is uneven.

The proposed approach is as follows:
Add a lowOps counter, along with threshold and reserve percentages for deallocation. Once the number of operations at low utilization reaches the threshold, we trigger shrinking.
Add an external API to force an attempt to deallocate memory, since the client may have a better understanding of when shrinking can be applied.

PR #5748, opened 2026-04-16
[Filestore] add deallocation for dynamic persistent table #5747

neihar:
### Notes
- add simple mechanic that counts operations at low memory utilization and if we hit limit then trigger shrinking
- add public api to class for external usage when client class have information when shrinking can make sense
- use DataAreaSize for compaction threshold
- introduce MaxDataAreaStepSize which will limit max step for any memory increase related operation to avoid huge spikes
- add tests for all cases

### Issue
#5747

### DIFF
--- a/cloud/storage/core/libs/common/dynamic_persistent_table.h
+++ b/cloud/storage/core/libs/common/dynamic_persistent_table.h
@@ -9,6 +9,7 @@
 #include <util/generic/string.h>
 #include <util/generic/vector.h>
 #include <util/stream/mem.h>
+#include <util/system/align.h>
 #include <util/system/file.h>
 #include <util/system/filemap.h>
 #include <util/system/yassert.h>
@@ -17,6 +18,19 @@
 #include <cstring>
 
 namespace NCloud {
+
+////////////////////////////////////////////////////////////////////////////////
+
+namespace {
+
+inline ui64 CalculatePercentage(ui64 value, ui64 percent)
+{
+    Y_ABORT_UNLESS(percent <= 100, "Invalid percentage %lu", percent);
+
+    return value / 100 * percent + value % 100 * percent / 100;
+}
+
+}   // namespace
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -90,6 +104,15 @@ public:
         Stored,
     };
 
+    // When the deallocation condition is met, AnyOp will attempt to shrink the
+    // data area for all operations, while AllocOnlyOp will attempt to shrink it
+    // only during allocations.
+    enum class EShrinkMode
+    {
+        AnyOp = 0,
+        AllocOnlyOp,
+    };
+
     struct TRecordDescriptor
     {
         ui64 DataOffset = 0;
@@ -99,6 +122,12 @@ public:
         ERecordState State = ERecordState::Free;
         ui32 Crc32 = 0;
     };
+
+    static constexpr ui64 DefaultGapSpacePercentageCompactionThreshold = 30;
+    static constexpr ui64 DefaultShrinkLowMemoryOpThreshold = 100;
+    static constexpr ui64 DefaultShrinkTriggerPercent = 25;
+    static constexpr ui64 DefaultShrinkReservePercent = 25;
+    static constexpr EShrinkMode DefaultShrinkMode = EShrinkMode::AllocOnlyOp;
 
 private:
     enum class EDataRetrievalMode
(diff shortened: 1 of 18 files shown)


cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -20,6 +21,19 @@ namespace NCloud {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace {
+
```
budevg: putting anonymous namespace in header doesn't ring as a good idea. Instead you can extract this into some common place like `cloud/storage/core/libs/common/numeric.h` also maybe rename to `PercentOf`

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -100,6 +123,12 @@ class TDynamicPersistentTable
         ui32 Crc32 = 0;
     };
 
+    static constexpr ui64 DefaultGapSpacePercentageCompactionThreshold = 30;
+    static constexpr ui64 DefaultShrinkLowMemoryOpThreshold = 100;
+    static constexpr ui64 DefaultShrinkTriggerPercent = 25;
+    static constexpr ui64 DefaultShrinkReservePercent = 25;
+    static constexpr EShrinkMode DefaultShrinkMode = EShrinkMode::AllocOnlyOp;
```
budevg: These constants are only used by external callers and are not part of the table's internal  implementation. Please define  them in the modules that actually use them, rather than keeping caller specific policy defaults in this generic component.

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -226,16 +262,36 @@ class TDynamicPersistentTable
         const TString& fileName,
         ui64 maxRecords,
         ui64 initialDataAreaSize,
+        ui64 maxDataAreaStepSize,
         ui64 initialDataCompactionBufferSize,
-        ui64 gapSpacePercentageCompactionThreshold)
+        ui64 gapSpacePercentageCompactionThreshold,
```
budevg: too much parameters in this constructor. Use a hepler struct `TDynamicPersistentTableConfig` to pass parameters by name and avoid mistakes

budevg (changes requested):

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: agreed, fixed

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: fixed

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: agreed

cloud/storage/core/libs/common/numeric.cpp
```diff
@@ -0,0 +1 @@
+#include "numeric.h"
```
budevg, line 1: add numeric.cpp to ya.make

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -21,16 +23,29 @@ namespace NCloud {
...
+    AnyOp = 0,
+    AllocOnlyOp,
+};
 
-}   // namespace
+struct TDynamicPersistentTableConfig
+{
+    using EShrinkMode = EDynamicPersistentTableShrinkMode;
+
+    const ui64 MaxRecords;
```
budevg: no need to declare members as const if needed the struct can be declared as const instead
```
const TDynamicPersistentTableConfig cfg;
```
Otherwise this just disables struct assignment and move constructor

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -62,6 +77,8 @@ template <typename H>
 class TDynamicPersistentTable
 {
 public:
+    using EShrinkMode = TDynamicPersistentTableConfig::EShrinkMode;
```
budevg: better remove this and use EDynamicPersistentTableShrinkMode::AnyOp directly

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -21,16 +23,29 @@ namespace NCloud {
...
-    return value / 100 * percent + value % 100 * percent / 100;
-}
+    AnyOp = 0,
+    AllocOnlyOp,
+};
 
-}   // namespace
+struct TDynamicPersistentTableConfig
+{
+    using EShrinkMode = EDynamicPersistentTableShrinkMode;
```
budevg: better remove this and use  EDynamicPersistentTableShrinkMode::AnyOp directly

budevg (changes requested):

budevg (dismissed):

budevg (approved):

agalibin (approved):

PR #5773, opened 2026-04-17
[Blockstore, Filestore] add retries for FDNHostName function used in (blockstore|filestore)-client

SvartMetal:
### Notes
Sometimes the blockstore-client (or filestore-client) tool fails to retrieve the FQDN:
```
util/system/hostname.cpp:74: can not get FQDN (return code is -3, hostname is "XXX"
```
These errors can cause our release tool (krelease) to trigger a rollback, which is frustrating for DevOps engineers since the errors are flapping. 
To address this, I decided to add synchronous retries for this case.

This approach is preferable to adding retries in krelease, as the tool does not expect blockstore-client to fail and cannot reliably distinguish which errors are safe to retry

### Issue

### DIFF
--- a/cloud/storage/core/libs/common/hostname.cpp
+++ b/cloud/storage/core/libs/common/hostname.cpp
@@ -1,0 +1,36 @@
+#include "hostname.h"
+
+#include <library/cpp/retry/retry.h>
+
+#include <util/datetime/base.h>
+#include <util/system/hostname.h>
+
+namespace NCloud {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+const ui32 GetFqdnHostNameRetryCount = 10;
+// Duration between two consecutive retries
+const TDuration GetFqdnHostNameSleepDuration = TDuration::Seconds(2);
+
+}   // namespace
+
+////////////////////////////////////////////////////////////////////////////////
+
+TString GetFqdnHostNameWithRetries(
+    const std::function<void(const yexception&)>& onFail)
+{
+    const auto hostname = DoWithRetry<TString, yexception>(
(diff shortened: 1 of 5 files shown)


cloud/storage/core/libs/common/hostname.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace NCloud {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+const ui32 GetFqdnHostNameRetryCount = 10;
+// Duration between two consecutive retries
+const TDuration GetFqdnHostNameSleepDuration = TDuration::Seconds(2);
```
neihar, line 16: I would use increment or exponential multiplier, retry supports it

neihar (approved):

cloud/storage/core/libs/common/hostname.cpp
SvartMetal: I don't think it is needed here - we do not expect that FQDNHostName calls would overload the OS

neihar: The main idea was to reduce first retry step to very small time. I don't no the nature of failures, but if it fails once for some reason, we will delay for 2 sec. Whether it could be a problem for us, I don’t know.

SvartMetal: It should not be a problem, since disk downtime should not depend on blockstore-client execution time. However, the overall release time could increase if these failures are frequent (I hope that is not the case). 

We should also monitor the situation and make sure these errors are visible to the relevant duty engineers

drbasic (approved):

Issue #5778, opened 2026-04-19
[Blockstore] Throttler race on Postpone/Advance leads to "Request was not postponed" crash


budevg:
`cloud/blockstore/tests/loadtest/local-user-metrics/test.py.test_user_metrics` fails

```
VERIFY failed (2026-04-15T01:15:27.294031Z): Request was not postponed.
  cloud/storage/core/libs/common/context.cpp:97
  Advance(): requirement start != 0 failed
```

Looks like during throttling some request reached LogAdvancedRequest without passing in LogPostponedRequest

```
2026-04-15T01:15:27.252232Z :BS_PROXY_PUT INFO: [2ebe159fb222751f] bootstrap ActorId# [50000:7628788458419823170:73] Group# 2181038080 BlobCount# 1 BlobIDs# [[72075186224037889:4:12:0:0:373:0]] HandleClass# TabletLog Tactic# MinLatency RestartCounter# 0 Marker# BPP13
2026-04-15T01:15:27.252766Z :BS_PROXY_PUT INFO: [2ebe159fb222751f] SendReply putResult# TEvPutResult {Id# [72075186224037889:4:12:0:0:373:0] Status# OK StatusFlags# { Valid } ApproximateFreeSpaceShare# 0.999479} ResponsesSent# 0 PutImpl.Blobs.size# 1 Last# true Marker# BPP21
2026-04-15T01:15:27.252823Z :BLOCKSTORE_PARTITION DEBUG: [72075186224037889] Complete WriteBlocks (gen: 4, step: 12)
VERIFY failed (2026-04-15T01:15:27.294031Z): Request was not postponed.
  cloud/storage/core/libs/common/context.cpp:97
  Advance(): requirement start != 0 failed
2026-04-15T01:15:28.232945Z :BS_PROXY_PUT INFO: [855029f16899db43] bootstrap ActorId# [50000:7628788462714790469:81] Group# 2181038082 BlobCount# 1 BlobIDs# [[72075186224037890:4:3:1:24576:91:0]] HandleClass# TabletLog Tactic# MinLatency RestartCounter# 0 Marker# BPP13
[...]

PR #5779, opened 2026-04-19
[Blockstore] issue-5778: Fix throttler race on Postpone/Advance logging

budevg:
#5778

### DIFF
--- a/cloud/blockstore/libs/throttling/throttler_ut.cpp
+++ b/cloud/blockstore/libs/throttling/throttler_ut.cpp
@@ -15,6 +15,7 @@
 #include <library/cpp/testing/unittest/registar.h>
 #include <library/cpp/threading/future/subscription/wait_all.h>
 
+#include <util/datetime/cputimer.h>
 #include <util/generic/queue.h>
 #include <util/generic/scope.h>
 #include <util/generic/xrange.h>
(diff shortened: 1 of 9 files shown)


qkrorlqr (changes requested):
The code is deliberately structured this way to avoid calling these 2 methods while holding the lock

I don't see why we would want to call anything that's called inside those methods while holding the lock apart from maybe this `callContext.Postpone(nowCycles);`

qkrorlqr (approved):

komarevtsev-d (dismissed):

cloud/blockstore/libs/throttling/throttler_ut.cpp
```diff
@@ -1668,6 +1653,41 @@ Y_UNIT_TEST_SUITE(TThrottlerTest)
...
+        UNIT_ASSERT(!future.HasValue());
+        UNIT_ASSERT_VALUES_EQUAL(
+            TDuration::Zero(),
+            callContext->Time(EProcessingStage::Postponed));
+
+        env.Policy->SetPostponeDelay(TDuration::Zero());
+        env.Timer->AdvanceTime(TDuration::Seconds(1));
+
+        NWait::WaitAll(
+            TVector<TFuture<NProto::TWriteBlocksResponse>>{future}).Wait();
```
drbasic: а почему не просто  future.Wait() ?

drbasic (approved):

qkrorlqr (approved):

Issue #5519, opened 2026-03-18
[NBS] Static and dynamic configs parsing failures


nisqatsi:
**Terminology**

**Static configs** - configs, read form local files at startup.
**Dynamic configs** - configs, received form CONSOLE tablet at node registration (so called CMS configs) or via ConfigsDispatcher during process runtime.

**Srcs**

cloud/storage/core/libs/
  - kikimr/{config_initializer.h/cpp, options.h/cpp}

cloud/blockstore/libs/daemon:
  - common/{bootstrap.h/cpp, config_initializer.h/cpp, options.h/cpp}
  - ydb/{bootstrap.h/cpp, config_initializer.h/cpp, options.h/cpp}
  - local/{bootstrap.h/cpp, config_initializer.h/cpp, options.h/cpp}

cloud/blockstore/libs/
  - disk_agent/{bootstrap.h/cpp, config_initializer.h/cpp, options.h/cpp}

**Problem 1: Different methods and transformations are used for applying static and dynamic configs**

Reading and applying static configs is performed by the methods {TConfigInitializerYdbBase, TConfigInitializerYdb}::Init\*Config().
Reading and applying dynamic configs is performed by the methods {TConfigInitializerYdbBase, TConfigInitializerYdb}::Apply\*Config().
In both cases, some transformations are applied to the certain config params, but the transformations differ between static and dynamic config handling.

**Problem 2: Potential cluster failure due to incorrect / new versioned dynamic configs**

All configs, except LogConfig, are parsed from a protobuf string using the function ParseProtoTextFromString(), which throws an exception on parsing errors, such as incorrect configuration (protobuf) format or presence of unknown parameters. This will lead to termination and restart of the NBS process, since there's no exception handling in the call stack. Only LogConfig is parsed using ParseProtoTextFromStringRobust(), which ignores unknown parameters but still throws an exception if the protobuf string structure is invalid.

While such behavior is acceptable when reading static configs (as the process would not start with an erroneous config), receiving an incorrect or incompatible dynamic config (e.g., containing unknown parameters for the current NBS version) can bring down the entire cluster (causing cyclic restarts of NBS). This situation can also occur when rolling back binaries to a previous version while using a newer config.

**Problem 3: Inconsistent error handling for static configs and potential cluster failure during binary rollback**

[...]

PR #5786, opened 2026-04-20
[Blockstore] Ignore unknown params for all NBS and DA static and cms configs

nisqatsi:
### Notes
Previously unknown parameters were ignored for the most part of NBS static configs and only for LogConfig in dynamic (cms) configs. Now all Kikimr, NBS and DiskAgent unknown parameters are not treated as errors for all types of static and dynamic configs. This prevents potential cluster failures:
* due to incompatible cms configs
* due to binaries rollback to older versions with configuration for newer version (with new parameters) left.

### Issue
This PR is part of #5519

### DIFF
--- a/cloud/storage/core/libs/kikimr/config_initializer_ut.cpp
+++ b/cloud/storage/core/libs/kikimr/config_initializer_ut.cpp
@@ -39,39 +39,12 @@ TOptionsYdbBasePtr CreateOptions()
 struct TConfigInitializerYdb
     : public TConfigInitializerYdbBase
 {
-    using TBase = TConfigInitializerYdbBase;
-
     explicit TConfigInitializerYdb(TOptionsYdbBasePtr options)
         : TConfigInitializerYdbBase(std::move(options))
     {}
 
     void ApplyCustomCMSConfigs(const NKikimrConfig::TAppConfig&) override
     {}
-
-    void ApplyActorSystemConfig(const TString& text)
-    {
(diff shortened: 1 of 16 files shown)


cloud/storage/core/libs/kikimr/config_initializer_ut.cpp
```diff
@@ -39,12 +39,39 @@ TOptionsYdbBasePtr CreateOptions()
...
+
+    void ApplyLogConfig(const TString& text)
+    {
+        TBase::ApplyLogConfig(text);
+    }
+
+    void ApplyMonitoringConfig(const TString& text)
+    {
+        TBase::ApplyMonitoringConfig(text);
+    }
```
SvartMetal: Aren't these methods supposed to be inherited from the base class?

SvartMetal: Oh, they are protected there https://github.com/ydb-platform/nbs/blob/aa7da2dbc05c11051ca075e06c799ded09e04c8d/cloud/storage/core/libs/kikimr/config_initializer.h#L46

Let's just make them public maybe?

nisqatsi: Давай, будет проще пользовать.
В cloud/blockstore/libs/daemon/ydb/config_initializer.h они private. М.б. тоже в public вынести?

SvartMetal (approved):

agalibin (approved):

PR #5820, opened 2026-04-23
add MaxSkippedBlobsDuringCompactionHDD to split hdd and ssd parameters

agalibin:


neihar (approved):

cloud/blockstore/libs/storage/partition/part_ut.cpp
```diff
@@ -9947,16 +9977,17 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
         TStatsChecker statsChecker;
         auto config = DefaultConfig();
         config.SetBlobPatchingEnabled(true);
-        config.SetWriteBlobThreshold(1);   // disable FreshBlocks
+        config.SetWriteBlobThresholdSSD(1);   // disable FreshBlocks
```
gy2411, line 9980: Why do we need to change this test? Just for readability?

gy2411 (approved):

cloud/blockstore/libs/storage/partition/part_ut.cpp
agalibin: readability and an explicitly defined disk type

PR #5827, opened 2026-04-24
clean up rdma tests; issue-5804

### LINKED ISSUE #5804: Get rid of CallContext in RDMA tests
CallContext isn't used to calculate timeout duration, so it's pointless to set it

tpashkin:
- drop pointless callContext initialization
- return logic introduced in e3b20ba25b362ffb3ee293f06f21021761b0ba40 that was lost during one of the merges
- check AllocateRequest results

### DIFF
--- a/cloud/storage/core/libs/rdma/impl/client_ut.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client_ut.cpp
@@ -344,6 +344,8 @@ TEST(TRdmaClientTest, ShouldProcessRequests)
             }
         };
 
+        long timedOutRequests = 0;
+
         for (size_t i = 0; i < RequestCount; ++i) {
             TManualEvent ev;
             TResponse response;
@@ -361,7 +363,7 @@ TEST(TRdmaClientTest, ShouldProcessRequests)
 
             if (i != 0 && i != RDMA_MAX_REQID - 3) {
                 // complete request right away
-                 handleRequest(*testContext);
+                handleRequest(*testContext);
 
                 ev.WaitT(5s);
                 ASSERT_TRUE(response.Received);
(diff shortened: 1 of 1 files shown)


cloud/storage/core/libs/rdma/impl/client_ut.cpp
```diff
@@ -345,70 +345,45 @@ TEST(TRdmaClientTest, ShouldProcessRequests)
...
-                ASSERT_TRUE(response.Received);
-                ASSERT_EQ(
-                    static_cast<ui32>(RDMA_PROTO_FAIL),
-                    response.Status);
+                // request duration is measured against the wall clock, so it
+                // can time out if the process stalls for some reason
+                if (response.Status != RDMA_PROTO_OK) {
+                    NProto::TError error =
+                        ParseError(response.Buffer.Head(response.Bytes));
+                    ASSERT_EQ(E_TIMEOUT, error.GetCode());
```
budevg, line 376: Even though you intention was not to fail the test when timeout due to clock occurs. The test will fail since  in the end of the test
```
ASSERT_EQ(aborted->Val(), 1);
```
will be invoked and the counter will be 2

cloud/storage/core/libs/rdma/impl/client_ut.cpp
```diff
@@ -345,70 +345,45 @@ TEST(TRdmaClientTest, ShouldProcessRequests)
...
-                    GetCycleCount() - retryDelay);
-                ep->SendRequest(std::move(request), callContext);
+            ep->SendRequest(
+                request.ExtractResult(),
+                MakeIntrusive<TCallContextBase>(0u));
 
-                handleRequest(*testContext);
+            if (i != 0 && i != RDMA_MAX_REQID - 3) {
+                // complete request right away
+                 handleRequest(*testContext);
```
budevg: remove extra space

budevg (changes requested):

budevg (approved):

komarevtsev-d (approved):

Issue #5388, opened 2026-03-04
[Filestore] filestore-vhost should communicate with the tablet via a side channel over RoCEv2


SvartMetal:
One possible solution is to implement a custom pipe with a side channel over RDMA. This pipe could be used inside `TIndexTabletProxyActor`.

### Basic approach
1. This, extended, pipe first establishes a connection using a regular Interconnect pipe, which is used to determine the tablet’s host.
2. Once the tablet host is known, the side channel is established via RDMA.
3. If the RDMA connection fails, it is re-established via step 1.
4. If the tablet moves to another host, the connection is also re-established via step 1.

We could reuse our existing RDMA library: https://github.com/ydb-platform/nbs/tree/main/cloud/blockstore/libs/rdma
However, it should probably be moved to cloud/storage first.

### Improved approach (decoupling from slow Actor System)
We can further improve the design by making the happy path less dependent on the Actor System, which is relatively slow.
1. Detach the happy path from `TIndexTabletProxyActor` entirely, using it only for connection establishment and disconnection handling.
2. Requests from `TStorageServiceActor` are placed into a lock-free / wait-free queue.
3. A dedicated thread polls this queue and:
* Dispatches requests via RDMA
* Handles connection failures and initiates re-discover the tablet RDMA endpoint via Interconnect (which involves `TIndexTabletProxyActor`)
* Initiates re-discover and retry in case the RDMA request cannot reach the tablet because the tablet has moved to another host

### More notes
We must account for possible reordering scenarios, for example:
* The tablet moves A → B → A.
* The client sends request 1 and request 2.
* Request 1 is delayed on the way to host A.
* Request 2 reaches tablet on host B
* After the tablet moves back to host A, request 1 arrives there later.

It may also be useful to establish a back channel from the tablet to filestore-vhost, for example to notify about:
* Cache flush events.
* Addition of unconfirmed blobs to the index.

PR #5835, opened 2026-04-24
[Blockstore] Init RDMA without logging and monitoring usage

neihar:
### Notes
Provide logging and monitoring tags to RDMA for later initialization
Remove log and monitoring initialization during RDMA init

### Issue
#5388


yegorskii (approved):

yegorskii:
We discussed that this might not be the best solution and noted its drawbacks, but since it's an urgent fix, it’s acceptable for now

sharpeye (approved):

PR #5847, opened 2026-04-27
[Storage] make rdma log and monitoring dependency cleaner

neihar:
### Notes
Created RdmaObservabilityTagProvided
Suplied blockstore tags via this provider with lazy log and monitoring counters creation

### Issue
#5388

### DIFF
--- a/cloud/storage/core/libs/rdma/impl/observability.cpp
+++ b/cloud/storage/core/libs/rdma/impl/observability.cpp
@@ -1,0 +1,33 @@
+#include "observability.h"
+
+namespace NCloud::NStorage::NRdma {
+
+////////////////////////////////////////////////////////////////////////////////
+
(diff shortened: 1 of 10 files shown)


cloud/storage/core/libs/rdma/impl/observability.cpp
```diff
@@ -0,0 +1,33 @@
+#include "observability.h"
+
+namespace NCloud::NStorage::NRdma {
+
+////////////////////////////////////////////////////////////////////////////////
+
+TRdmaObservabilityProvider::TRdmaObservabilityProvider(
```
tpashkin: Let's remove `Rdma` prefix. We are already inside the `NRdma` namespace, there is no ambiguity

tpashkin (dismissed):

tpashkin (approved):

yegorskii (approved):

PR #5848, opened 2026-04-27
fmdtest: fixing stolen file validation race btw producer append and stealer stat

qkrorlqr:
### Notes
Fixing the following race:
1. producer opens a file in its dir with `O_CREAT`
2. stealer steals this file, calls stat, reads it, saves size & content
3. producer writes data to the file
4. in the end of the test we validate that stealer's saved stat & content are equal to the actual stat & content of the file -> this fails

Fixed this race by saving only the inode number upon stealing and then comparing the actual size & content of the file to the size & content saved by the producer 

Also removed an unneeded `NFs::Exists(filePath)` check for stealer dir validation - all final file paths of the stolen files must point to existing files

### Issue
none

### DIFF
--- a/cloud/filestore/tools/testing/fmdtest/bin/app.cpp
+++ b/cloud/filestore/tools/testing/fmdtest/bin/app.cpp
@@ -122,6 +122,7 @@ struct TFsPathHash
 
 struct TValidationContext
 {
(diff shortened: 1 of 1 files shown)


neihar (approved):

neihar:
```
Issue
none
```
Noted :)

yegorskii (approved):

PR #5851, opened 2026-04-28
issue-4875: start FreshBlocksWriter only for Partition ver1 tablet

vladstepanyuk:
### Notes
Fresh Blocks Writer is not supported for partitions v2, so we shouldn't try to start fbw for v2.

### Issue
#4875

### DIFF
--- a/cloud/blockstore/libs/storage/partition2/part2_actor.cpp
+++ b/cloud/blockstore/libs/storage/partition2/part2_actor.cpp
@@ -766,7 +766,7 @@ void TPartitionActor::HandlePatchBlob(
     Y_UNUSED(ctx);
 
     HandleUnexpectedEvent(
-        ev.Release(),
+        *ev,
         TBlockStoreComponents::PARTITION,
         __PRETTY_FUNCTION__);
 }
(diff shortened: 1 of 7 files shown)


komarevtsev-d (dismissed):

neihar:
PR description and Issue would be nice

vladstepanyuk:
> PR description and Issue would be nice

added description and issue

komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition2/part2_actor.cpp
```diff
@@ -763,9 +763,12 @@ void TPartitionActor::HandlePatchBlob(
     const TEvPartitionCommonPrivate::TEvPatchBlobRequest::TPtr& ev,
     const NActors::TActorContext& ctx)
 {
-    Y_UNUSED(ev);
     Y_UNUSED(ctx);
-    Y_ABORT("Unimplemented");
+
+    HandleUnexpectedEvent(
```
budevg, line 768: Don't you have double free here and in other places where you call ev.Get()
```
void HandleUnexpectedEvent(
    const TAutoPtr<NActors::IEventHandle>& ev,
    int component,
    const TString& location);
```
converts the raw pointer from ev.Get() to AutoPtr then releases it and the original `ev` will also release the same pointer

budevg (changes requested):

komarevtsev-d (dismissed):

budevg (approved):

komarevtsev-d (approved):

Issue #5830, opened 2026-04-24
[Filestore] realtime records increase for dynamic persistent table


neihar:
Currently we have static records size for dynamic persistent table. For some case scenarios it is not suitable and can grow beyond default limits. At the same time, we don't want to allocate huge amount of memory if it is not needed. So, the solution is to explan it in realtime if needed

PR #5855, opened 2026-04-28
[Storage] increase record table size if no space left

neihar:
### Notes
Renamed Compaction* entities to Move*
Added possibility to expand records size dynamically
Added tests for crash scenarios and correctness

### Issue
 #5830

### DIFF
--- a/cloud/storage/core/libs/common/dynamic_persistent_table.h
+++ b/cloud/storage/core/libs/common/dynamic_persistent_table.h
@@ -37,7 +37,7 @@ struct TDynamicPersistentTableConfig
     ui64 MaxRecords = 0;
     ui64 InitialDataAreaSize = 0;
     ui64 MaxDataAreaStepSize = 0;
-    ui64 InitialDataCompactionBufferSize = 0;
+    ui64 InitialDataMoveBufferSize = 0;
     ui64 GapSpacePercentageCompactionThreshold = 30;
     ui64 ShrinkLowMemoryOpThreshold = 100;
     ui64 ShrinkTriggerPercent = 25;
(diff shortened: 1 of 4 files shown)


cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -443,6 +452,138 @@ class TDynamicPersistentTable
...
+
+    void ExpandRecordArea()
+    {
+        if (GapSpaceSize != 0) {
+            CompactDataArea();
+        }
+
+        const ui64 targetMaxRecords = MaxRecords + RecordAreaExpansionStep;
+        const ui64 targetDataAreaOffset = CalcDataAreaOffset(targetMaxRecords);
+        const ui64 expansionBytes = targetDataAreaOffset - DataAreaOffset;
```
budevg, line 545: if the `targetDataAreaOffset` lands in the middle of big data record it will be moved to end. It's first half will be used for descriptors but the second half will be lost since the `GapSpaceSize` is not increase and so we can have leaked data size

```
before:   <spec space> | record A                    | record B     |
after:    <spec space> | <spec space> |stale record A| record B ... | record A| 
```

budevg (changes requested):

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: It will be reused after next compaction, but will be missed from coefficient calculations. In current situation, it should be pretty small to make impact, but will fix GapSize calculations for that case anyway.

budevg (approved):

yegorskii (approved):

Issue #5791, opened 2026-04-21
[NBS] Get rid of strict validation logic


tpashkin:
Currently it's not used anywhere

tpashkin:
Discussed internally – @komarevtsev-d has plans to use it during `QueueSize -> Send/RecvQueueSize` split rollout. So I'm repurposing this issue into fixing this logic

PR #5888, opened 2026-05-04
Second attempt to fix RDMA StrictValidation logic

komarevtsev-d:
### Notes
Fixing the logic of StrictValidation after notes in the https://github.com/ydb-platform/nbs/pull/5734

The first attempt was here: https://github.com/ydb-platform/nbs/pull/5754

Now, during the "connection" phase, the client and server negotiate a protocol version. The client sends its version in the `TConnectMessage`, the server validates it, and responds with either a reject message or an accept message with the supported protocol version. The client must know and accept the server's protocol version. In this way, if a protocol version needs to be changed in the future, the update process should be straightforward.

### Issue
https://github.com/ydb-platform/nbs/issues/5791


cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2236,17 +2270,29 @@ void TClient::HandleConnected(
...
+        return;
+    }
+
+    const int version = ParseMessageHeader(param->private_data);
+    if (version < RDMA_PROTO_PREV_VERSION || version > RDMA_PROTO_CURR_VERSION)
     {
-        RDMA_ERROR(endpoint->Log, "unable to parse accept message");
+        RDMA_ERROR(
+            endpoint->Log,
+            "unable to parse message version: " << version);
```
tpashkin: I'd say something like "unsupported message version: " would be more precise

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2254,17 +2270,29 @@ void TClient::HandleConnected(
 {
     const rdma_conn_param* param = &event->param.conn;
 
-    RDMA_DEBUG(endpoint->Log, "validate");
+    RDMA_DEBUG(endpoint->Log, "validate connect message");
```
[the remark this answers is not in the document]
tpashkin: It's not wrong I suppose

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2236,17 +2270,29 @@ void TClient::HandleConnected(
...
 
     if (param->private_data == nullptr ||
-        param->private_data_len < sizeof(TAcceptMessage) ||
-        ParseMessageHeader(param->private_data) != RDMA_PROTO_VERSION)
+        param->private_data_len < sizeof(TAcceptMessage))
+    {
+        RDMA_ERROR(
+            endpoint->Log,
+            "couldn't parse private data in accept message");
```
tpashkin: Sorry, this starts to feel like a editorial review. I think "unable to parse accept message" was just fine

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2268,31 +2314,81 @@ void TClient::HandleRejected(
...
     const rdma_conn_param* param = &event->param.conn;
 
+    RDMA_DEBUG(endpoint->Log, "validate reject message");
+
     if (param->private_data == nullptr ||
-        param->private_data_len < sizeof(TRejectMessage) ||
-        ParseMessageHeader(param->private_data) != RDMA_PROTO_VERSION)
+        param->private_data_len < sizeof(TRejectMessage))
     {
+        RDMA_ERROR(endpoint->Log, "no private data in reject message");
```
tpashkin: I think whatever this message is, it should be similar to the TAcceptMessage case

tpashkin:
Looks good, I'll do a second pass looking over the tests tomorrow

cloud/storage/core/libs/rdma/iface/protocol.h
```diff
@@ -27,7 +27,9 @@ constexpr size_t RDMA_MAX_DATA_LEN = Max<ui32>();
 enum {
     RDMA_PROTO_VERSION_0        = 0,
     RDMA_PROTO_VERSION_1        = 1,
-    RDMA_PROTO_VERSION          = RDMA_PROTO_VERSION_1,
+    RDMA_PROTO_VERSION_2        = 2,
+    RDMA_PROTO_PREV_VERSION     = RDMA_PROTO_VERSION_1,
+    RDMA_PROTO_CURR_VERSION     = RDMA_PROTO_VERSION_2,
```
tpashkin: Maybe drop the CURR suffix and continue to use RDMA_PROTO_VERSION?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1230,10 +1234,11 @@ void TClientEndpoint::RecvResponseCompleted(TRecvWr* recv) noexcept
     auto* msg = recv->Message();
 
     int version = ParseMessageHeader(msg);
-    if (version != RDMA_PROTO_VERSION) {
+    if (version != NegotiatedProtocolVersion) {
         RDMA_ERROR(
             recv << " incompatible protocol version " << version
-                 << ", expected " << static_cast<int>(RDMA_PROTO_VERSION));
+                 << ", expected "
+                 << static_cast<int>(NegotiatedProtocolVersion));
```
tpashkin: it's already int, so no need to cast

tpashkin (dismissed):

cloud/storage/core/libs/rdma/impl/server.cpp
tpashkin, line 1864: I think there is a bit of a discrepancy here – server accepts equally sized queues, but client adjusts it's queues to be half the size the server ones

cloud/storage/core/libs/rdma/iface/protocol.h
```diff
@@ -27,7 +27,9 @@ constexpr size_t RDMA_MAX_DATA_LEN = Max<ui32>();
 enum {
     RDMA_PROTO_VERSION_0        = 0,
     RDMA_PROTO_VERSION_1        = 1,
-    RDMA_PROTO_VERSION          = RDMA_PROTO_VERSION_1,
+    RDMA_PROTO_VERSION_2        = 2,
+    RDMA_PROTO_PREV_VERSION     = RDMA_PROTO_VERSION_1,
+    RDMA_PROTO_VERSION     = RDMA_PROTO_VERSION_2,
```
neihar: format?

cloud/storage/core/libs/rdma/impl/server.cpp
komarevtsev-d: I did it intentionally. Technically, the same queue sizes can work.

komarevtsev-d: Plus, not that long ago queue sizes weren't split. Equal values should be accepted until the original "QueueSize" field is not removed from configs.

cloud/storage/core/libs/rdma/iface/protocol.h
```diff
@@ -27,7 +27,9 @@ constexpr size_t RDMA_MAX_DATA_LEN = Max<ui32>();
 enum {
     RDMA_PROTO_VERSION_0        = 0,
     RDMA_PROTO_VERSION_1        = 1,
-    RDMA_PROTO_VERSION          = RDMA_PROTO_VERSION_1,
+    RDMA_PROTO_VERSION_2        = 2,
+    RDMA_PROTO_PREV_VERSION     = RDMA_PROTO_VERSION_1,
```
budevg, line 31: let's use explicit versions in code (RDMA_PROTO_VERSION_1 , RDMA_PROTO_VERSION_2) instead of PREV and CURRENT version. Later it easier to see what version is processed in handler and what versions are supported (for example maybe we need to support 3 versions though it's unlikely)

cloud/storage/core/libs/rdma/iface/protocol.h
```diff
@@ -142,6 +144,24 @@ static_assert(sizeof(TRejectMessage) == RDMA_PRIVATE_SIZE);
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct Y_PACKED TRejectMessage2
+{
+    union {
+        struct {
+            TMessageHeader Header;
+            ui32 Status : 16;
```
budevg, line 155: let's add explicit padding field to make it more clear
```
ui32 Padding: 16;
```
or even better we can change it to `ui16` why do we need the bit field for the new message

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2268,31 +2311,84 @@ void TClient::HandleRejected(
...
-
-    if (msg->Status == RDMA_PROTO_CONFIG_MISMATCH) {
-        if (endpoint->Config.QueueSize > msg->QueueSize) {
-            RDMA_INFO(endpoint->Log, "set QueueSize=" << msg->QueueSize
-                << " supported by " << endpoint->Host);
-
-            endpoint->Config.QueueSize = msg->QueueSize;
+    const int version = ParseMessageHeader(param->private_data);
+    switch (version) {
+        case RDMA_PROTO_PREV_VERSION: {
```
budevg, line 2326: if old server (VERSION1) rejects new client due to queue size or buffer size we will keep failing and will never connect

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1817,10 +1843,18 @@ void TServer::HandleConnectRequest(
...
-        ParseMessageHeader(connectParams->private_data) < RDMA_PROTO_VERSION)
+        connectParams->private_data_len < sizeof(TConnectMessage))
     {
-        return Reject(event->id, RDMA_PROTO_INVALID_REQUEST);
+        Reject(event->id, RDMA_PROTO_INVALID_REQUEST, RDMA_PROTO_VERSION);
+        return;
+    }
+
+    const int protocolVersion = ParseMessageHeader(connectParams->private_data);
+    if (protocolVersion < RDMA_PROTO_PREV_VERSION ||
```
budevg, line 1853: better explicitly reject anything which is not protocol 1,2

cloud/storage/core/libs/rdma/impl/server.cpp
budevg, line 1863: shouldn't we check here based on protocol version for old client the `connectMsg->RecvQueueSize` is not defined

budevg (changes requested):

cloud/storage/core/libs/rdma/impl/client.cpp
komarevtsev-d: The old server can't reject anyone. `StrictValidation` is always false, since it wasn't in the proto config :)

cloud/storage/core/libs/rdma/impl/server.cpp
komarevtsev-d: I think It's water under the bridge. And `StrictValidation` is always false anyway

cloud/storage/core/libs/rdma/iface/protocol.h
komarevtsev-d: Are you sure? 
It seems that current code handles adding +1 to `RDMA_PROTO_PREV_VERSION` and `RDMA_PROTO_VERSION ` well. 
In your case, we would introduce a lot of minor changes when bumping up a version, rather than making some local changes. E.g., to change one message, we would simply update the handler for it.

cloud/storage/core/libs/rdma/iface/protocol.h
tpashkin: Bit fields are used in other messages, let's keep it consistent. Explicit padding – good idea

cloud/storage/core/libs/rdma/iface/protocol.h
tpashkin: > maybe we need to support 3 versions though it's unlikely

Let's be honest – we don't really need to keep compatibility between versions except for the smooth transition. There are only 2 real users and we can coordinate our releases just fine

I'm strongly against supporting more than 2 consecutive versions. Intention is whenever version increases, code that handles `version - 2` goes

We can pin versions wherever specific message gets parsed. But let's keep prev and current aliases in generic checks like "validate accept message" and such

budevg: My main concern is that it's difficult to understand how each version is handled when you have switch case with all the version it's easy to see the code and reason about it. When we dealing with current and prev version then in the next switch it will be difficult to understand what code we need to keep and what to modified. Aditionally it will be easier to discard old version by just removing the switch case with the specific version
```
switch (version) {
case version_1:
  ///
case version_2:
 ///
case version_3:
///
}
```

cloud/storage/core/libs/rdma/impl/client.cpp
budevg: let's add a comment about this here that we don't handle rejects from old server since it was never enabled in prod

cloud/storage/core/libs/rdma/impl/server.cpp
budevg: ok but let's add comment

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2268,31 +2311,84 @@ void TClient::HandleRejected(
...
+            {
+                RDMA_WARN(
+                    endpoint->Log,
+                    "connection rejected, retry connect with previous protocol "
+                    "version");
+                endpoint->SetNegotiatedProtocolVersion(RDMA_PROTO_PREV_VERSION);
+            }
+            break;
         }
+        case RDMA_PROTO_VERSION_2: {
```
tpashkin: we use RDMA_PROTO_PREV_VERSION alias in another case and explicit version here. kind of incosistent

budevg (approved):

vladstepanyuk (approved):

tpashkin (approved):

PR #5910, opened 2026-05-06
filestore: not returning TFileStore from shards to main tablet upon session creation

qkrorlqr:
### Notes
Otherwise the total number of bytes that needs to be transferred is quadratic relative to the total number of shards which is a huge issue for filesystems with thousands of shards 

Tablets don't use the contents of `TCreateSessionResponse`s anyway - they're needed only in the storage service layer

### Issue
None - it's just a small tweak to tablet<->tablet `TCreateSessionResponse`s

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
@@ -418,16 +418,18 @@ void TIndexTabletActor::CompleteTx_CreateSession(
     auto response = std::make_unique<TResponse>(args.Error);
     response->Record.SetSessionId(std::move(args.SessionId));
     response->Record.SetSessionState(session->GetSessionState());
-    auto& fileStore = *response->Record.MutableFileStore();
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
```diff
@@ -506,6 +508,7 @@ void TIndexTabletActor::CreateSessionsInShards(
         logTag.c_str(),
         JoinSeq(",", shardIds).c_str());
 
+    request.SetNoFileStoreInfo(true);
```
SvartMetal, line 518: It would be helpful to add a comment explaining that TFileStore is not needed in the tablet code, and that the amount of received data can be huge for filesystems with many shards and clients

qkrorlqr: added comment

SvartMetal (approved):

yegorskii (approved):

PR #5921, opened 2026-05-07
Allow to skip volume performance profile for suffer calculation

komarevtsev-d:
### Notes
Small volumes have low performance profile values. However, "execution time" (without any waiting time) does not really differ much between small and large volumes. This PR allows to override the volume's throttling settings and rely only on the performance profile set in the diagnostic config to calculate the `Suffer`/`SmoothSuffer`/`CriticalSuffer`.

### DIFF
--- a/cloud/blockstore/libs/diagnostics/config.h
+++ b/cloud/blockstore/libs/diagnostics/config.h
@@ -15,56 +15,72 @@ namespace NCloud::NBlockStore {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TVolumePerfSettings:
-    public TAtomicRefCount<TVolumePerfSettings>
+struct TVolumePerfSettings: public TAtomicRefCount<TVolumePerfSettings>
 {
-    ui32 ReadIops = 0;
-    ui64 ReadBandwidth = 0;
+    struct TPerformanceProfile {
+        ui32 Iops = 0;
+        ui64 Bandwidth = 0;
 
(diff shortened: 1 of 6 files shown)


vladstepanyuk (dismissed):

cloud/blockstore/libs/diagnostics/config.h
```diff
@@ -15,56 +15,72 @@ namespace NCloud::NBlockStore {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TVolumePerfSettings:
-    public TAtomicRefCount<TVolumePerfSettings>
+struct TVolumePerfSettings: public TAtomicRefCount<TVolumePerfSettings>
 {
-    ui32 ReadIops = 0;
-    ui64 ReadBandwidth = 0;
+    struct TPerformanceProfile {
```
qkrorlqr: nit: { to the next line plz

cloud/blockstore/libs/diagnostics/volume_perf.cpp
```diff
@@ -14,10 +14,27 @@
...
 
+#include <limits>
+#include <type_traits>
+
 namespace NCloud::NBlockStore {
 
 using namespace NMonitoring;
 
+namespace {
+
```
qkrorlqr, line 25: ////////////////

qkrorlqr (approved):

vladstepanyuk (approved):

PR #5925, opened 2026-05-08
[NBS] Local NVMe service: resolve vfio-dev when acquiring NVMe device

sharpeye:
Resolving vfio-dev for an NVMe device has been moved into `AcquireNVMeDevice`, since vfioN only becomes available after the device has been rebound to the vfio-pci driver.

### DIFF
--- a/cloud/blockstore/libs/local_nvme/service_linux.cpp
+++ b/cloud/blockstore/libs/local_nvme/service_linux.cpp
@@ -32,6 +32,8 @@ const TErrorResponse ServiceDestroyedError =
     MakeError(E_REJECTED, "Local NVMe service is destroyed");
 
 const TDuration SanitizeStatusProbeTimeout = TDuration::MilliSeconds(100);
+const TDuration VfioDeviceRetryDelay = TDuration::MilliSeconds(100);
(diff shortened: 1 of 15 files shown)


yegorskii (approved):

cloud/blockstore/libs/local_nvme/service_linux.cpp
```diff
@@ -434,7 +447,62 @@ auto TLocalNVMeService::AcquireDevice(TCont* c, const TString& serialNumber)
...
+
+    if (!IsVfioDevSupported) {
+        return r;
+    }
+
+    auto [vfioDev, error] = GetVfioDevName(c, r.GetPCIAddress());
+    if (HasError(error)) {
+        STORAGE_ERROR(
+            "Failed to get vfio device for " << r << ": "
+                                             << FormatError(error));
```
dvrazumov, line 464: Why the error is not returned here?

sharpeye: If vfio-dev is unavailable, we log the error but don't propagate it, since we can still fall back to legacy mode. The new method will be enforced in the future.

dvrazumov (approved):

Issue #5934, opened 2026-05-10
[Filestore]: false positive CreateNodeRequestResponseMismatchInShard crit events


debnatkh:
`ProcessNodeAttr` fires a `CreateNodeRequestResponseMismatchInShard` critical event when it gets `EEXIST` from a shard, does `GetNodeAttr`, and finds the node's size doesn't match the original request attrs AND the node's mtime looks old relative to the request time:

https://github.com/ydb-platform/nbs/blob/7bde8dbead29a91c4f9b1a87486f1cb67f15c9aa/cloud/filestore/libs/storage/tablet/tablet_actor_createnode.cpp#L223-L226

The mtime condition was supposed to suppress false positives when a node was modified after creation. But clients can set arbitrary timestamps via `SetNodeAttr` — for example, a timestamp-preserving copy sets mtime to the source file's original modification time, which can be arbitrarily old. This causes the critical event to fire even though the node is perfectly valid.

PR #5935, opened 2026-05-10
issue-5934 [Filestore]: add timestamp attributes to the profile log's SetNodeAttr requests

debnatkh:
### Notes
Right now, it is impossible to know if the node's time attributes were set by the SetNodeAttr from the profile log 

### Issue
#5934

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,3 +170,6 @@ cloud/storage/core/tools/testing/qemu/bin/usr/**
 cloud/storage/core/tools/testing/qemu/bin/qemu-bin.tar.gz
 
 cloud/filestore/bin/pids.txt
+
+# nbs_internal directory
+nbs_internal/**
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -170,3 +170,6 @@ cloud/storage/core/tools/testing/qemu/bin/usr/**
 cloud/storage/core/tools/testing/qemu/bin/qemu-bin.tar.gz
 
 cloud/filestore/bin/pids.txt
+
+# nbs_internal directory
+nbs_internal/**
```
debnatkh: this is intentional

qkrorlqr: this is not related to the github repo at all, let's not put it here

e673 (approved):

yegorskii (approved):

PR #5963, opened 2026-05-13
[Filestore] Forbid modification of TFileRingBuffer when IsCorrupted flag is set

e673:
### Notes
Corruption in `TFileRingBuffer` is no more a recoverable state. Previously, restoration of read and write position happened and alloc/free was allowed. Now the strategy is changed from 'try to restore whatever is possible' to the strategy 'halt and wait for a manual resolution'.

### DIFF
--- a/cloud/storage/core/libs/common/file_ring_buffer.cpp
+++ b/cloud/storage/core/libs/common/file_ring_buffer.cpp
@@ -383,6 +383,7 @@ private:
 
         if (front.IsInvalid() || front.ActualPos != Header()->ReadPos) {
             SetCorrupted();
+            return;
         }
 
         while (cur.HasValue()) {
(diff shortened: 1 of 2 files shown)


cloud/storage/core/libs/common/file_ring_buffer.cpp
SvartMetal: Do we need it?

SvartMetal, line 712: Do we need to handle 'corrupted' state here?

e673: removed unused variables

e673: After discussion, we decided to implement this as critevents (instead of Y_DEBUG_ABORT).
Will do this separately in a corresponding PR.

SvartMetal (approved):

yegorskii (approved):

PR #5971, opened 2026-05-14
[Blockstore] optimize compaction map: do not recalculate score/garbage of the group when not needed

gy2411:
### Notes

- Optimize the performance of the compaction map in some scenarios
- Improve the compaction map benchmark

---

Compaction map optimization:

`RecalculateGroupScore` and `RecalculateGroupGarbageBlockCount` are quite expensive because they iterate over all ranges in the group. However, we only need this recalculation when the score/garbage value in the top range of the group decreases. Before this fix, we also performed this recalculation when the top range changed and the score/garbage value did not change.

For example, if we fill the entire compaction map with ranges that have `0` garbage, we call `RecalculateGroupGarbageBlockCount` for each range. This is inefficient.

---

Benchmark modifications:

1. Create a new compaction map object for each iteration. If we do not do this, all iterations after the first one are much faster (they just don't dive into this `if` https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/storage/core/compaction_map.cpp#L207). This leads to strange results like this -- avarage for 18 iteration is much better then expected:

```
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
Update_1_TB              29805000 ns     29802429 ns           18
Update_5_TB            2015158198 ns   2015002645 ns            1
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
Update_1_TB             679578910 ns    679552834 ns            1
Update_5_TB            3473159005 ns   3472999281 ns            1
```

2. Added new benchmarks that try to simulate real load better. Old benchmarks represent a special case when garbage in all ranges is 0.

---

Benchmark results:

Without the fix:

```
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
Update_1_TB              401368067 ns    401338871 ns            2
Update_5_TB             2063699893 ns   2063571520 ns            1
Update_10_TB            4178160549 ns   4177690753 ns            1
Update_50_TB            7114521451 ns   7114163624 ns            1
Update_100_TB           7650246420 ns   7649679902 ns            1
Update_500_TB           1.1903e+10 ns   1.1902e+10 ns            1
UpdateRandomized_1_TB     17984053 ns     17982492 ns           39
UpdateRandomized_5_TB    131754655 ns    131745974 ns            6
UpdateRandomized_10_TB   285124737 ns    285108213 ns            3
UpdateRandomized_50_TB  1301530570 ns   1301425944 ns            1
UpdateRandomized_100_TB 2547280139 ns   2547114097 ns            1
UpdateRandomized_500_TB 1.2384e+10 ns   1.2383e+10 ns            1
```

With the fix:

```
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
Update_1_TB              139999779 ns    139998215 ns            5
Update_5_TB              763097292 ns    763067559 ns            1
Update_10_TB            1600163868 ns   1600088125 ns            1
Update_50_TB            2943275623 ns   2943165508 ns            1
Update_100_TB           3470560086 ns   3470379386 ns            1
Update_500_TB           7601042094 ns   7600382811 ns            1
UpdateRandomized_1_TB     18083756 ns     18082438 ns           40
UpdateRandomized_5_TB    126521303 ns    126512374 ns            6
UpdateRandomized_10_TB   268469964 ns    268464845 ns            3
UpdateRandomized_50_TB  1283381287 ns   1283216432 ns            1
UpdateRandomized_100_TB 2541680311 ns   2541274731 ns            1
UpdateRandomized_500_TB 1.2210e+10 ns   1.2210e+10 ns            1
```

Old benchmarks performance improved significantly.

### DIFF
--- a/cloud/blockstore/libs/storage/core/compaction_map.cpp
+++ b/cloud/blockstore/libs/storage/core/compaction_map.cpp
@@ -234,15 +234,12 @@ struct TCompactionMap::TImpl
                 : Policy->CalculateScore(group->Stats[index]);
             group->Stats[index].CompactionScore = newScore;
 
-            if (prev.CompactionScore.Score < newScore.Score) {
+            if (prev.CompactionScore.Score <= newScore.Score) {
                 if (group->Score < newScore.Score) {
                     group->Score = newScore.Score;
                     group->Range = index;
                 }
-            } else if (
-                index == group->Range &&
-                newScore.Score < prev.CompactionScore.Score)
-            {
+            } else if (index == group->Range) {
                 // Score in the top range has decreased, need to recalculate the
                 // maximal score.
                 RecalculateGroupScore(group);
@@ -251,14 +248,11 @@ struct TCompactionMap::TImpl
             const auto newGarbageBlockCount =
                 compacted ? 0 : group->Stats[index].GarbageBlockCount();
 
-            if (prev.GarbageBlockCount() < newGarbageBlockCount) {
+            if (prev.GarbageBlockCount() <= newGarbageBlockCount) {
                 if (GetGarbageBlockCount(*group) < newGarbageBlockCount) {
                     group->GarbageBlockCount = newGarbageBlockCount;
                 }
-            } else if (
-                prev.GarbageBlockCount() == group->GarbageBlockCount &&
-                newGarbageBlockCount < prev.GarbageBlockCount())
-            {
+            } else if (prev.GarbageBlockCount() == GetGarbageBlockCount(*group)) {
                 // Garbage block count in the top range has decreased, need to
                 // recalculate the maximal garbage block count.
                 RecalculateGroupGarbageBlockCount(group);
@@ -293,7 +287,7 @@ struct TCompactionMap::TImpl
             auto newScore = Policy->CalculateScore(group->Stats[index]);
             group->Stats[index].CompactionScore = newScore;
 
-            if (prev.CompactionScore.Score < newScore.Score) {
+            if (prev.CompactionScore.Score <= newScore.Score) {
                 if (group->Score < newScore.Score) {
                     group->Score = newScore.Score;
                     group->Range = index;
--- a/cloud/blockstore/libs/storage/core/compaction_map.cpp
+++ b/cloud/blockstore/libs/storage/core/compaction_map.cpp
@@ -239,19 +239,28 @@ struct TCompactionMap::TImpl
                     group->Score = newScore.Score;
                     group->Range = index;
                 }
-            } else if (index == group->Range) {
+            } else if (
+                index == group->Range &&
+                newScore.Score < prev.CompactionScore.Score)
+            {
+                // Score in the top range has decreased, need to recalculate the
+                // maximal score.
                 RecalculateGroupScore(group);
             }
 
-            const auto newGarbageBlockCount = compacted
-                ? 0
-                : group->Stats[index].GarbageBlockCount();
+            const auto newGarbageBlockCount =
+                compacted ? 0 : group->Stats[index].GarbageBlockCount();
 
             if (prev.GarbageBlockCount() < newGarbageBlockCount) {
                 if (GetGarbageBlockCount(*group) < newGarbageBlockCount) {
                     group->GarbageBlockCount = newGarbageBlockCount;
                 }
-            } else if (prev.GarbageBlockCount() == group->GarbageBlockCount) {
+            } else if (
+                prev.GarbageBlockCount() == group->GarbageBlockCount &&
+                newGarbageBlockCount < prev.GarbageBlockCount())
+            {
+                // Garbage block count in the top range has decreased, need to
+                // recalculate the maximal garbage block count.
                 RecalculateGroupGarbageBlockCount(group);
             }
         }
--- a/cloud/blockstore/libs/storage/core/compaction_map.cpp
+++ b/cloud/blockstore/libs/storage/core/compaction_map.cpp
@@ -293,6 +293,8 @@ struct TCompactionMap::TImpl
                     group->Range = index;
                 }
             } else if (index == group->Range) {
+                // Score in the top range has decreased, need to recalculate the
+                // maximal score.
                 RecalculateGroupScore(group);
             }
         }
--- a/cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
+++ b/cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
@@ -3,6 +3,7 @@
 #include <library/cpp/testing/gbenchmark/benchmark.h>
 
 #include <util/generic/size_literals.h>
+#include <util/random/random.h>
 
 namespace NCloud::NBlockStore::NStorage {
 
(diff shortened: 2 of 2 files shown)


vladstepanyuk (approved):

cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
```diff
@@ -48,6 +38,81 @@ void DoUpdate(ui64 diskSize, benchmark::State& state)
...
+    const ui64 rangeCount = blockCount / RangeSize;
+
+    TCompressedBitmap usedBlocks(rangeCount * RangeSize);
+
+    TVector<TCompactionCounter> counters(Reserve(rangeCount));
+    for (size_t i = 0; i < rangeCount; ++i) {
+        const ui16 blobCount =
+            static_cast<ui16>(RandomNumber<ui32>(5) + 1);
+        const ui16 statBlockCount =
+            static_cast<ui16>(RandomNumber<ui32>(901) + 100);
```
vladstepanyuk: мб вынести куда-то эти захардкоженные значения, и поч они такие странные
1-5 блобов
100-1000 блоков

vladstepanyuk (changes requested):

cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
```diff
@@ -48,6 +38,81 @@ void DoUpdate(ui64 diskSize, benchmark::State& state)
...
+ui16 RandomUsedBlockCount(ui16 blockCount)
+{
+    const ui32 bucket = RandomNumber<ui32>(4);
+    if (bucket == 0) {
+        return blockCount;
+    }
+    if (bucket == 1) {
+        return 0;
+    }
+    return static_cast<ui16>(RandomNumber<ui32>(blockCount - 1) + 1);
```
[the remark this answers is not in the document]
gy2411: Handled `blockCount <= 1` case properly.

cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
gy2411: Changed range for blockCount: [0, 1024] instead of [100, 1000]. Here 1024 is the number of blocks in the range.

I don't think that other constanst are really worth giving them a name, they are not so important.

cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
```diff
@@ -48,6 +38,89 @@ void DoUpdate(ui64 diskSize, benchmark::State& state)
...
+
+    if (blockCount <= 1) {
+        return blockCount;
+    }
+    return static_cast<ui16>(RandomNumber<ui32>(blockCount - 1) + 1);
+}
+
+void DoUpdateRandomized(ui64 diskSize, benchmark::State& state)
+{
+    constexpr size_t RangeSize = 1024;
```
SvartMetal: Isn't it the same as BlocksInRange?

Maybe just keep BlocksInRange

cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
```diff
@@ -48,6 +38,89 @@ void DoUpdate(ui64 diskSize, benchmark::State& state)
...
+            usedBlockCount,
+            0,
+            0,
+            0,
+            false,
+            score);
+
+        counters.emplace_back(i * RangeSize, rangeStat);
+
+        const ui64 rangeBase = i * RangeSize;
```
SvartMetal: rangeBlockOffset?

cloud/blockstore/libs/storage/core/compaction_map.cpp
```diff
@@ -239,19 +239,28 @@ struct TCompactionMap::TImpl
...
                 if (GetGarbageBlockCount(*group) < newGarbageBlockCount) {
                     group->GarbageBlockCount = newGarbageBlockCount;
                 }
-            } else if (prev.GarbageBlockCount() == group->GarbageBlockCount) {
+            } else if (
+                prev.GarbageBlockCount() == group->GarbageBlockCount &&
+                newGarbageBlockCount < prev.GarbageBlockCount())
+            {
+                // Garbage block count in the top range has decreased, need to
+                // recalculate the maximal garbage block count.
```
SvartMetal, line 257: It may not be fully related to this PR, but it looks like TCompactionMap lacks proper randomized tests against a simple reference implementation. At least, I cannot find such tests [here](https://github.com/ydb-platform/nbs/blob/791910e6ea6acebc7bba5f47effc02d618e05766/cloud/blockstore/libs/storage/core/compaction_map_ut.cpp) 

Such tests are becoming more relevant as the complexity of this code increases.
See examples of this approach here https://github.com/ydb-platform/nbs/blob/092390c5690663319cb62c857a58656d6e2ceefb/cloud/storage/core/libs/common/dynamic_persistent_table_ut.cpp#L194 and there https://github.com/ydb-platform/nbs/blob/092390c5690663319cb62c857a58656d6e2ceefb/cloud/storage/core/libs/common/file_ring_buffer_ut.cpp#L92

cloud/blockstore/libs/storage/core/compaction_map.cpp
```diff
@@ -239,19 +239,28 @@ struct TCompactionMap::TImpl
...
+            const auto newGarbageBlockCount =
+                compacted ? 0 : group->Stats[index].GarbageBlockCount();
 
             if (prev.GarbageBlockCount() < newGarbageBlockCount) {
                 if (GetGarbageBlockCount(*group) < newGarbageBlockCount) {
                     group->GarbageBlockCount = newGarbageBlockCount;
                 }
-            } else if (prev.GarbageBlockCount() == group->GarbageBlockCount) {
+            } else if (
+                prev.GarbageBlockCount() == group->GarbageBlockCount &&
```
SvartMetal: It is better to use GetGarbageBlockCount(*group), as on line 255, for consistency

cloud/blockstore/libs/storage/core/compaction_map.cpp
```diff
@@ -239,19 +239,28 @@ struct TCompactionMap::TImpl
...
+                newGarbageBlockCount < prev.GarbageBlockCount())
```
SvartMetal: Wouldn't it be simpler to rewrite both branches to handle equality first, and then have one branch for > while the else branch covers <?

Same for lines 237-247

gy2411: Actually, we handle the case of equality in the first branch. I.e. we just replace < by <= in if condition.

cloud/blockstore/libs/storage/core/compaction_map.cpp
gy2411: It's a good idea. I will add such test in the next pr.

cloud/blockstore/libs/storage/core/compaction_map.cpp
SvartMetal, line 295: Let's add a comment here as well?

cloud/blockstore/libs/storage/core/compaction_map_benchmark.cpp
```diff
@@ -44,10 +34,92 @@ void DoUpdate(ui64 diskSize, benchmark::State& state)
...
+{
+    constexpr size_t BlocksInRange = 1024;
+    constexpr size_t BlockSize = 4096;
+    const ui64 blockCount = diskSize / BlockSize;
+    const ui64 rangeCount = blockCount / BlocksInRange;
+
+    TCompressedBitmap usedBlocks(rangeCount * BlocksInRange);
+    TVector<TCompactionCounter> counters(Reserve(rangeCount));
+
+
```
SvartMetal: Excessive \n

SvartMetal (dismissed):

vladstepanyuk (approved):

SvartMetal (approved):

PR #5972, opened 2026-05-14
issue-5590: dont split huge blobs on small blobs

vladstepanyuk:
### Notes
Split the blob only if the original blob is large and all resulting blobs are large, or if the original is small.
### Issue
#5590

### DIFF
--- a/cloud/blockstore/libs/storage/partition/model/flush_blocks_visitor.cpp
+++ b/cloud/blockstore/libs/storage/partition/model/flush_blocks_visitor.cpp
@@ -51,6 +51,7 @@ TFlushBlocksVisitor::TFlushBlocksVisitor(
         bool readBlockMaskOnCompactionOptimizationEnabled,
         ui64 splitByCompactionRangeMaxBlobCount,
         ui64 tabletId,
+        ui64 writeBlobSizeThreshold,
         TVector<TBlob>& blobs)
     : BlockSize(blockSize)
     , FlushBlobSizeThreshold(flushBlobSizeThreshold)
@@ -63,6 +64,7 @@ TFlushBlocksVisitor::TFlushBlocksVisitor(
           readBlockMaskOnCompactionOptimizationEnabled)
     , SplitByCompactionRangeMaxBlobCount(splitByCompactionRangeMaxBlobCount)
     , TabletId(tabletId)
+    , WriteBlobSizeThreshold(writeBlobSizeThreshold)
     , Blobs(blobs)
 {}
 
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/partition/model/flush_blocks_visitor.cpp
```diff
@@ -195,7 +197,21 @@ void TFlushBlocksVisitor::FlushBlob(
     const bool splitByCompactionBorders =
         compactionRangeCount <= SplitByCompactionRangeMaxBlobCount;
 
-    if (!splitByCompactionBorders || compactionRangeCount == 1) {
+    const bool isOriginalBlobHuge =
+        blocks.size() * BlockSize >= WriteBlobSizeThreshold;
+
+    const bool splittedBlobsAreHuge = std::ranges::all_of(
```
qkrorlqr: past participle of "to split" is just "split"

cloud/blockstore/libs/storage/partition/model/flush_blocks_visitor.cpp
```diff
@@ -195,7 +197,21 @@ void TFlushBlocksVisitor::FlushBlob(
...
+        blocksByRanges,
+        [&](const auto& range)
+        {
+            return (range.End - range.Start) * BlockSize >=
+                   WriteBlobSizeThreshold;
+        });
+
+    if (!(splitByCompactionBorders &&
```
qkrorlqr: what a terrible cryptic if condition, let's simplify it

!(a && b) is equivalent to !a || !b

so !(a && b) || c is equivalent to !a || !b || c which is much easier to understand

so this condition can be changed to:
```
if (!splitByCompactionBorders
    || isOriginalBlobHuge != splitBlobsAreHuge
    || compactionRangeCount == 1)
{
    AppendDataBlob(...);
    return;
}
```

which means that we don't do splitting if any of the following is true:
* splitting is disabled
* original blob is huge but split blobs are not huge
* original blob is not huge but split blobs are huge - WTF, seems impossible
* compaction range count for the original blob is 1

so since bullet 3 looks impossible, we can further simplify it to:
```
if (!splitByCompactionBorders
    || (isOriginalBlobHuge && !splitBlobsAreHuge)
    || compactionRangeCount == 1)
{
    AppendDataBlob(...);
    return;
}
```

seems that we can actually get rid of all_of and simplify the code further:
```
const bool isOriginalBlobHuge = blocks.size() * BlockSize >= WriteBlobSizeThreshold;
if (isOriginalBlobHuge) {
    for (const auto& range: blocksByRanges) {
        if ((range.End - range.Start) * BlockSize < WriteBlobSizeThreshold) {
            AppendDataBlob(...);
            return;
        }
    }
}

if (!splitByCompactionBorders || compactionRangeCount == 1) {
    AppendDataBlob(...);
    return;
}
```

so now this if condition looks exactly the same as it used to look

and the diff just turns into adding this block of code:
```
const bool isOriginalBlobHuge = blocks.size() * BlockSize >= WriteBlobSizeThreshold;
if (isOriginalBlobHuge) {
    for (const auto& range: blocksByRanges) {
        if ((range.End - range.Start) * BlockSize < WriteBlobSizeThreshold) {
            AppendDataBlob(...);
            return;
        }
    }
}
```

much easier to comprehend and review - no cryptic if conditions, no functional template-heavy libraries like std::ranges, no changes to the existing code - just a new block of code which serves its own purpose which is not directly related to the existing blocks of code

qkrorlqr (approved):

komarevtsev-d (approved):

PR #5981, opened 2026-05-15
MAY REQUIRE CONFIG UPDATE! VolumeBalancer: Make IsVolumePreemptible independent of balancer's on/off state

efimov-ivan-yteam:
Currently `FeatureConfig::Balancer` does not affect VolumeBalancer's work:

- If VolumeBalancer is enabled (in terms of `State->GetEnabled()`)
  - Feature's status is always overwhelmed by `State->GetEnabled()` at https://github.com/ydb-platform/nbs/commit/3e2d4adac8144b82acb477f5568c88e1c93c0d5c#diff-0d0e0eee72b24144b2ef1bd00a11227d31c2f60c7a0b5cfcd14f7770f3b3d0adL283
- If VolumeBalancer is disabled
  - Feature's status affect volume nomination for push, but balancer is overall disabled, which renders `FeatureConfig::Balancer` practically useless again

Removing `IsVolumePreemptible()`'s dependence from balancer's on/off status we allow `FeatureConfig::Balancer`
to rule balancing whenever `State->GetEnabled()` is true

However, this also means that `FeatureConfig::Balancer` is now **required** to be on for balancer to work

### DIFF
--- a/cloud/blockstore/libs/storage/volume_balancer/volume_balancer_state.cpp
+++ b/cloud/blockstore/libs/storage/volume_balancer/volume_balancer_state.cpp
@@ -275,12 +275,11 @@ bool TVolumeBalancerState::IsVolumePreemptible(
     const TString& diskId,
     const TVolumeInfo& volume) const
 {
-    const bool isFeatureEnabledForFolder = StorageConfig->IsBalancerFeatureEnabled(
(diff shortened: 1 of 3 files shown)


SvartMetal:
First, we will merge changes to our configs

PR #5983, opened 2026-05-15
[Filestore] TFileRingBuffer: fire critical event on corruption

e673:
### Notes
It is needed to report critical events related to corruption of `TFileRingBuffer`.

### Issue


cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -754,7 +783,12 @@ class TFileRingBuffer::TImpl
 
     void SetCorrupted()
     {
+        if (!Corrupted) {
             Corrupted = true;
+            ReportFileBackedContainerCorruptionDetectedError(
```
neihar: Do we actually need more than this event? is it possible to come back to Corrupted=false after it became true without manual interaction? 
I would make 1 critical event here and in other places just checks or checks + regular logs

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -552,6 +554,13 @@ class TFileRingBuffer::TImpl
 
     bool PushBack(TStringBuf data)
     {
+        if (IsCorrupted()) {
+            ReportAccessToCorruptedFileBackedContainerError(
```
neihar: If we really need all of this events, probably make sense to do something like 
If(!ValidateAccess(__name__))
 return false;

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -689,8 +704,9 @@ class TFileRingBuffer::TImpl
         auto e = GetFrontEntry();
```
neihar, line 747: is the particular reason why we skip ReportCriticalEvent here as in other functions?

e673: No, it was just missed.
Fixed.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Done

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: I want to ensure that the caller code does not try to use `TFileRingBuffer` in the corrupted state.
Regular logs don't get enough attention - a firing alert is preferrable here.
Also, I don't want to use `Y_ABORT` - this will be unnecessary crash.

cloud/storage/core/libs/diagnostics/critical_events.h
```diff
@@ -18,6 +18,8 @@ namespace NCloud {
     xxx(GetConfigsFromCmsYamlParseError)                                       \
     xxx(FailedToParseRdmaError)                                                \
     xxx(FailedToSerializeRdmaError)                                            \
+    xxx(FileBackedContainerCorruptionDetectedError)                            \
```
SvartMetal: FileRingBufferCorruptionDetectedError

cloud/storage/core/libs/diagnostics/critical_events.h
```diff
@@ -18,6 +18,8 @@ namespace NCloud {
...
+    xxx(AccessToCorruptedFileBackedContainerError)                             \
```
SvartMetal: AccessToCorruptedFileRingBufferError

SvartMetal (changes requested):

cloud/storage/core/libs/diagnostics/critical_events.h
e673: Renamed

cloud/storage/core/libs/diagnostics/critical_events.h
e673: Renamed

SvartMetal (dismissed):

SvartMetal (approved):

yegorskii (approved):

Issue #5763, opened 2026-04-16
[Filestore] implemet circuit breaker for huge directory listing


neihar:
Currently we have listing for directories with 100M+ files and size of cache can consume more than expected space. To avoid OOMs for such cases the next solution can be implemented:

Add a hard limit for the total tmpfs cache size during the working phase. If the limit is reached, drop the tmpfs cache for all active handles or just big handles. New handles after the drop should work with tmpfs cache normally.

PR #5984, opened 2026-05-15
[Filestore] introducing filemap memory limiter for directory handles cache

neihar:
### Notes

Added shared memory accounting for filemap cache used by directory handle storage.

The main idea is that vhost creates a common IFileMapMemoryLimtiter from config and passes it down to directory handle storage. TDynamicPersistentTable now accounts mapped file size through this limiter and rejects later storage growth when the memory limit is exceeded.

What changed:

Added a memory limiter with file-map usage tracking.
Updated TDynamicPersistentTable to check file-map growth against the memory limiter.
Changed table allocation to return errors instead of only InvalidIndex.
Updated directory handle storage to drop the whole handle from persistent storage when table growth is rejected.
Added tests for the new behavior at each layer.

### Issue
#5763


cloud/filestore/config/vhost.proto
```diff
@@ -90,6 +90,12 @@ message TVhostServiceConfig
 
     // Max single growth step for DirectoryHandles persistent table data area.
     optional uint64 DirectoryHandlesMaxDataAreaStepSize = 22;
+
+    // Memory limit in bytes for vhost. Value 0 means that we will not use it.
+    optional uint64 MemoryLimit = 23;
```
budevg: What is this limit actually constraining ? Is it limit on process heap allocation or is it cgroup memory limit for process. Why not get the limit dynamically instead of relying on hardcoded number ?
I think it would be better to just set the TmpfsMemoryLimit to number of bytes and not introduce some global memory limit which is not connected to the process memory management system

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
```diff
@@ -18,15 +18,21 @@ class TDirectoryHandleCache
 
     TAdaptiveLock Lock;
     TDirectoryHandleMap Handles;
-    TDirectoryHandleStoragePtr Storage;
+    TDirectoryHandleStoragePtr MemoryAwareStorage;
```
budevg: Is there not memory aware storage ?

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -100,35 +98,38 @@ void TDirectoryHandleCache::ResetHandle(
...
+        MemoryAwareStorage->UpdateHandle(handleId, handleChunk);
+    }
 }
 
 void TDirectoryHandleCache::Clear()
 {
     with_lock (Lock) {
         STORAGE_DEBUG("clear directory cache of size %lu", Handles.size());
+        for (const auto& [_, handle]: Handles) {
+            DecreaseStatsForHandle(handle);
```
budevg: batch the stats update here and in other places where you iterate handles.
Calling DecreaseStatsForHandle on each handle is expensive for 1M handles since you update atomics
instead do something like this
```
  size_t totalSize = 0, totalChunks = 0;
  for (const auto& [_, handle] : Handles) {
      if (handle) {
          totalSize += handle->GetSerializedSize();
          totalChunks += handle->GetChunkCount();
      }
  }
  if (Stats) {
      Stats->DecreaseCacheSize(totalSize);
      Stats->DecreaseChunkCount(totalChunks);
  }
  Handles.clear();
```

cloud/filestore/libs/vfs_fuse/directory_handle_storage.cpp
```diff
@@ -24,24 +27,34 @@ TDirectoryHandleStorage::TDirectoryHandleStorage(
...
     const TDirectoryHandleChunk& initialHandleChunk)
 {
+    TBuffer record = SerializeHandle(handleId, initialHandleChunk);
+
+    TGuard guard(TableLock);
+
     if (HandleIdToIndices.contains(handleId)) {
         ReportDirectoryHandlesStorageError(
             "Failed to store record with existing handle id");
+        RemoveRecords(handleId);
```
budevg, line 45: is this a bug? ... how would this case happen ? Don't you want to fallback and replace old record with fresh record ?

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -1012,12 +1040,60 @@ class TDynamicPersistentTable
...
         HeaderPtr->DataAreaSize = DataAreaSize;
 
         ResizeAndRemap();
 
         ResetShrinkState();
+
+        return {};
+    }
+
+    NProto::TError CanResizeFileMap(ui64 newFileMapSize) const
```
budevg: `CanResizeFileMap` implies filemap can either grow or shrink but the function will work correctly only if it grows. So better rename the function to something `CanGrowFileMapSize` or similar

cloud/storage/core/libs/common/memory_controller.h
```diff
@@ -0,0 +1,38 @@
...
+struct IMemoryController
+{
+    virtual ~IMemoryController() = default;
+
+    // The check is advisory, not a reservation. Callers account memory only
+    // after a file mapping actually grows, so concurrent growth can temporarily
+    // overshoot the limit; the limit gates future growth attempts.
+    [[nodiscard]] virtual bool CanIncreaseFileMapUsage(
+        ui64 increaseSize) const = 0;
+    virtual void IncreaseFileMapUsage(ui64 diffSize) = 0;
```
budevg: would be great to extract some metrics from memory controller ... otherwise we can't know what's going on inside

budevg (changes requested):

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
neihar: If we provide no memory_controller then it is not memory aware. I wanted to express that this storage doesn't have guarantee to store all data.

cloud/storage/core/libs/common/memory_controller.h
neihar: yes, will do this in separate PR later

cloud/filestore/config/vhost.proto
neihar: Later I will plan to pass here real limit for container from k8s during container start. We have already similar logic for some variables.

cloud/filestore/libs/vfs_fuse/directory_handle_storage.cpp
neihar: It shouldn't happen. If it happens something completely wrong. Because such inconsistency is unlikely it is okay to just ignore storage for that handle once and look what happened

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
neihar: Do you suggest to return just Storage name?

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
neihar: Agreed, Changed

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: Fixed

cloud/filestore/config/vhost.proto
neihar: Agreed to have only TmpfsMemoryLimit as initial solution.

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
neihar: renamed back to Storage

budevg (dismissed):

cloud/storage/core/libs/common/memory_controller.cpp
```diff
@@ -0,0 +1,82 @@
...
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+constexpr TStringBuf MemoryControllerEntityType = "MemoryController";
+constexpr TStringBuf MemoryControllerEntityId = "file_map";
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TMemoryController final: public IMemoryController
```
SvartMetal: TMemoryController and IMemoryController are too generic names. These entities do not control memory in general, but rather a specific memory type (tmpfs, filemap).

Also, based on the implementation, this looks more like some kind of limiter than a controller

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -743,6 +745,7 @@ class TFileSystemLoop final
         , Timer(std::move(timer))
         , ProfileLog(std::move(profileLog))
         , Session(std::move(session))
+        , MemoryController(std::move(memoryController))
```
SvartMetal: This question was probably already asked, sorry.

Do we really need this controller as a separate entity here? 
Can we just implement a built-in memory limit directly in TDynamicPersistentTable?

cloud/filestore/libs/vfs_fuse/directory_handle_storage.cpp
```diff
@@ -61,44 +78,45 @@ void TDirectoryHandleStorage::UpdateHandle(
...
     const TBuffer& record)
 {
-    ui64 recordIndex = CreateRecord(record);
-
-    if (recordIndex == TDirectoryHandleTable::InvalidIndex) {
-        ReportDirectoryHandlesStorageError(
-            "Failed to create record for directory handle chunk");
-        return;
+    auto result = CreateRecord(record);
+    if (HasError(result)) {
```
SvartMetal, line 80: Isn't it more reasonable to update HandlesExcludedFromStorage on error here, rather than at the caller site?

cloud/storage/core/libs/common/memory_controller.cpp
neihar: Agreed, I will rename it to MemoryLimiter

cloud/filestore/libs/vfs_fuse/loop.cpp
neihar: We need it as separate entity because we can have more than 1 FS (and directory cache) in vhost and this memory limiter will guarantee that we will not consume more than X bytes.
Separate simple measure will be implemented in directory cache, like drop cache for all dirs bigger than 5 GiB for example, but it can't guarantee that at some point in time we will not have 4 GiB in 10 different FSes or 200 000 listings with 100 mb each and so on. Such events should be very rare, but worth handle them in advance instead of crash loop because of OOM.

cloud/filestore/libs/vfs_fuse/directory_handle_storage.cpp
neihar: It is trade off, in one place we don't need RemoveRecords call, but we ofc can call it with no side effects. Also I would prefer to keep different logs. 
I would say if it were more call sides or absolutely the same, I would go that way. Here I don't have preference, if you insist I can do that change.

cloud/storage/core/libs/common/memory_controller.cpp
neihar: I want  it to be as MemoryLimiter, not as TmpfsMemoryLimiter, because in another task regarding cache size of original struct in memory, I want to use the same limiter, but different counter, for controlling regular cache size and eviction.

cloud/filestore/apps/client/lib/mount.cpp
```diff
@@ -77,7 +77,8 @@ class TMountCommand final
             Scheduler,
             Timer,
             CreateProfileLogStub(),
-            session);
+            session,
+            nullptr);
```
qkrorlqr: /* paramName */

cloud/filestore/libs/vfs_fuse/fuzz/starter.cpp
```diff
@@ -122,7 +122,8 @@ TStarter::TStarter()
         Scheduler,
         Timer,
         CreateProfileLogStub(),
-        std::move(session));
+        std::move(session),
+        nullptr);
```
qkrorlqr: /* paramName */

cloud/filestore/libs/daemon/vhost/bootstrap.cpp
```diff
@@ -430,6 +431,11 @@ void TBootstrapVhost::InitEndpoints()
 
     FileStoreEndpoints = std::move(endpoints);
 
+    auto memoryController = NCloud::CreateMemoryController(
+        NCloud::TMemoryControllerConfig{
+            .TmpfsMemoryLimit =
```
qkrorlqr: we actually don't force tmpfs usage anywhere in the code - the code just works with any file paths

so it's more like PersistentMemoryLimit (wbc-style naming) or FileMemoryLimit

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -6,6 +6,33 @@ namespace NCloud::NFileStore::NFuse {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace {
+
```
qkrorlqr, line 8: `////...` goes before struct/func decl, not before ns

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -145,4 +170,42 @@ void TDirectoryHandleCache::Reset()
...
+void TDirectoryHandleCache::IncreaseStatsForHandle(
+    const std::shared_ptr<TDirectoryHandle>& handle)
+{
+    if (Stats && handle) {
+        Stats->IncreaseCacheSize(handle->GetSerializedSize());
+        Stats->IncreaseChunkCount(handle->GetChunkCount());
+    }
+}
+
+void TDirectoryHandleCache::IncreaseStatsForHandle(
```
qkrorlqr: ForHandle**s**?

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -145,4 +170,42 @@ void TDirectoryHandleCache::Reset()
...
+void TDirectoryHandleCache::DecreaseStatsForHandle(
+    const std::shared_ptr<TDirectoryHandle>& handle)
+{
+    if (Stats && handle) {
+        Stats->DecreaseCacheSize(handle->GetSerializedSize());
+        Stats->DecreaseChunkCount(handle->GetChunkCount());
+    }
+}
+
+void TDirectoryHandleCache::DecreaseStatsForHandle(
```
qkrorlqr: and here as well

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -145,4 +170,42 @@ void TDirectoryHandleCache::Reset()
...
+    if (Stats && handle) {
+        Stats->IncreaseCacheSize(handle->GetSerializedSize());
+        Stats->IncreaseChunkCount(handle->GetChunkCount());
+    }
+}
+
+void TDirectoryHandleCache::IncreaseStatsForHandle(
+    const TDirectoryHandleMap& handles)
+{
+    if (Stats) {
```
qkrorlqr: let's remove this check (to make the code a bit simpler) - `Stats` should never be null

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
```diff
@@ -216,6 +216,16 @@ void TDirectoryHandle::ConsumeChunk(TDirectoryHandleChunk& chunk)
 {
     Y_ABORT_UNLESS(Index == chunk.Index);
 
+    const auto chunkSize = chunk.GetSerializedSize();
```
qkrorlqr: ui64 or size_t

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
qkrorlqr: do we have a guarantee that chunk.Key shouldn't be already present in Content? probably we should check that emplace succeeds

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -222,7 +222,7 @@ void TFileSystem::ReadDir(
 
                 auto handleChunk = handle->UpdateContent(
                     size,
-                    offset,
+                    static_cast<size_t>(offset),
```
qkrorlqr: maybe `SafeIntegerCast()`?

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -293,33 +304,39 @@ class TDynamicPersistentTable
...
-            ExpandRecordArea();
+            if (auto error = ExpandRecordArea(); HasError(error)) {
+                return error;
+            }
             index = NextFreeRecordIndex++;
         }
 
         if (index == InvalidIndex) {
-            return InvalidIndex;
+            return MakeError(E_FAIL, "Invalid record index");
```
qkrorlqr: in which cases can we have `index == InvalidIndex`?

cloud/storage/core/libs/common/dynamic_persistent_table.h
```diff
@@ -1012,12 +1040,60 @@ class TDynamicPersistentTable
...
+            return {};
+        }
+
+        const ui64 increaseSize = newFileMapSize - fileMapSize;
+        if (MemoryController->CanIncreaseFileMapUsage(increaseSize)) {
+            return {};
+        }
+
+        return MakeError(
+            E_REJECTED,
```
qkrorlqr: doesn't look like a retriable error - if nothing else changes, then `MemoryController->CanIncreaseFileMapUsage(increaseSize)` will keep returning `false`

it should be something like `ENOSPC` or `ENOMEM`

cloud/storage/core/libs/common/memory_controller.h
```diff
@@ -0,0 +1,37 @@
...
+using IMemoryControllerPtr = std::shared_ptr<IMemoryController>;
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TMemoryControllerConfig
+{
+    ui64 TmpfsMemoryLimit = 0;
+};
+
+// Returns nullptr when memory limiting is disabled.
```
qkrorlqr: why not return a stub implementation to get rid of all the nullptr checks?

cloud/storage/core/libs/common/memory_controller.h
```diff
@@ -0,0 +1,37 @@
...
+namespace NCloud {
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct IMemoryController
+{
+    virtual ~IMemoryController() = default;
+
+    // The check is advisory, not a reservation. Callers account memory only
+    // after a file mapping actually grows, so concurrent growth can temporarily
```
qkrorlqr: an interface that has an atomic `bool TryIncreaseFileMapUsage(ui64 diffSize)` op instead of 2 ops (`CanIncrease` + `Increase`) looks cleaner and avoids such races

cloud/storage/core/libs/common/memory_controller.h
neihar: Will use stub, agree that code is cleaner without checks

cloud/storage/core/libs/common/memory_controller.h
neihar: Agreed about general idea, but for current usage it is better to have that way in my opinion. both Can and Try can exists if needed. Points behind current usage: 
1. We do not force it in all places, for example init stage can expand without checking limit, so already no strong rule. It is used only in places where payload expansion is possible.
2. It is easier to manage size changes in one place for code simplicity. Like ChangeSize (old, new) without remembering that only decreases should be handled.

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: I felt uncomfortable leaving this place without check. But you are right, we can drop this check completely.

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
neihar: added warning, I do not expect with current linux model regarding readdir that this can happen

cloud/filestore/config/vhost.proto
```diff
@@ -90,6 +90,10 @@ message TVhostServiceConfig
 
     // Max single growth step for DirectoryHandles persistent table data area.
     optional uint64 DirectoryHandlesMaxDataAreaStepSize = 22;
+
+    // Memory limit in bytes for file-map-backed cache data.
+    // Value 0 means that a stub file map memory limiter will be used.
+    optional uint64 FileMapMemoryLimit = 23;
```
SvartMetal, line 99: Let's mention examples of use cases? Write-back cache, directory handles, etc.

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
```diff
@@ -212,19 +212,35 @@ size_t TDirectoryHandle::GetChunkCount() const
...
 
-void TDirectoryHandle::ConsumeChunk(TDirectoryHandleChunk& chunk)
+void TDirectoryHandle::ConsumeChunk(TDirectoryHandleChunk& chunk, TLog& Log)
 {
     Y_ABORT_UNLESS(Index == chunk.Index);
 
+    const size_t chunkSize = chunk.GetSerializedSize();
+    size_t serializedSizeDelta = 0;
+    if (chunk.UpdateVersion == 0) {
+        serializedSizeDelta = chunkSize - BaseSerializedSize;
```
SvartMetal, line 229: Are we protected from underflow here?

SvartMetal (changes requested):

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
neihar: Added comment

SvartMetal (approved):

cloud/storage/core/libs/common/dynamic_persistent_table.h
agalibin, line 588: Внутри MoveRecordData может быть вызван ResizeAndRemap без дополнительных проверок. Это ожидаемо?

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
```diff
@@ -212,19 +212,42 @@ size_t TDirectoryHandle::GetChunkCount() const
...
+            chunkSize >= BaseSerializedSize,
+            "Chunk size %zu is smaller than base serialized size %zu",
+            chunkSize,
+            BaseSerializedSize);
+        serializedSizeDelta = chunkSize - BaseSerializedSize;
+    } else {
+        serializedSizeDelta = chunkSize;
+    }
+
+    SerializedSize += serializedSizeDelta;
```
agalibin, line 234: если Content.emplace ничего не добавит, то получим зазря увеличенный размер или я ошибаюсь?

cloud/storage/core/libs/common/dynamic_persistent_table.h
neihar: да ожидаемо, хотим ограничивать только там где рост места для payload, вспомогательные буферы не ограничиваем

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
neihar: 1. мы такого не ожидаем, поэтому стоит варнинг
2. даже если такое случится, то одинаковые кусочки всеравно сериализованные и занимают место, поэтому логично их учитывать

agalibin (approved):

PR #5992, opened 2026-05-18
Fix smooth and critical suffers calculation

komarevtsev-d:
### Notes
Smooth and Critical suffers are calculated over 15 seconds window. But the `DidSuffer()` method had a hardcoded 1 second window. This PR adds window size parameter to the `DidSuffer()` to properly calculate regular `Suffer` count and the `SmoothSuffer`/`CriticalSuffer` flags

### DIFF
--- a/cloud/blockstore/libs/diagnostics/volume_perf.cpp
+++ b/cloud/blockstore/libs/diagnostics/volume_perf.cpp
@@ -169,11 +169,12 @@ void TVolumePerformanceCalculator::OnRequestCompleted(
 }
 
 bool TVolumePerformanceCalculator::DidSuffer(
-    long expectedScore,
-    long actualScore) const
+    ui64 expectedScore,
+    ui64 actualScore,
+    TDuration window) const
 {
-    return (expectedScore < ExpectedIoParallelism * 1e6)
-        && (actualScore > expectedScore);
+    const ui64 windowCapacity = window.MicroSeconds() * ExpectedIoParallelism;
+    return (expectedScore < windowCapacity) && (actualScore > expectedScore);
 }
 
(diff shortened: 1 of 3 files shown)


qkrorlqr (changes requested):
This fix doesn't look correct because this `TVolumePerformanceCalculator::UpdateStats()` func is indeed supposed to be called once every second: https://github.com/ydb-platform/nbs/blob/86e6aa3b738e6788bc6fde0b73f8c57b31514850/cloud/storage/core/libs/diagnostics/stats_updater.cpp#L84

and then we accumulate `SampleCount` samples and calculate suffer metrics based on suffer flags for each of those samples

but it's better to use `UpdateStatsInterval` instead of a hardcoded 1e6 value, yes

qkrorlqr:
> This fix doesn't look correct because this `TVolumePerformanceCalculator::UpdateStats()` func is indeed supposed to be called once every second:
> 
> https://github.com/ydb-platform/nbs/blob/86e6aa3b738e6788bc6fde0b73f8c57b31514850/cloud/storage/core/libs/diagnostics/stats_updater.cpp#L84
> 
> and then we accumulate `SampleCount` samples and calculate suffer metrics based on suffer flags for each of those samples
> 
> but it's better to use `UpdateStatsInterval` instead of a hardcoded 1e6 value, yes

Sorry, misread the diff - yes, the fix looks correct, we do use `UpdateStatsInterval` for each sample and yes, seems that we need to use `UpdateCountersInterval` for `SmoothSuffer` and `CriticalSuffer` metrics

qkrorlqr (approved):

drbasic (approved):

PR #6008, opened 2026-05-21
fix error code when disk is not mounted

yegorskii:
### Notes
1. Both TSplitRequestService and TOverlappingRequestsGuard should reply with E_BS_INVALID_SESSION rather than E_REJECTED when disk is not mounted (since both of them rely on information returned in Mount Response). E_REJECTED is useless because it does not trigger remount.
2. DiskId in TOverlappingRequestsGuard seems to be useless because it is used only in Y_ABORT_UNLESS which is also removed. The only information we acutally need is a BlockSize.

### Issue

### DIFF
--- a/cloud/blockstore/libs/service/split_request_service.cpp
+++ b/cloud/blockstore/libs/service/split_request_service.cpp
@@ -488,7 +488,9 @@ TFuture<TResponse> TSplitRequestService::ExecuteRequestWithSplit(
 
     if (!config) {
         TResponse response;
-        *response.MutableError() = MakeError(E_BS_INVALID_SESSION, "Volume not mounted");
+        *response.MutableError() = MakeError(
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/service/split_request_service.cpp
```diff
@@ -492,7 +488,7 @@ TFuture<TResponse> TSplitRequestService::ExecuteRequestWithSplit(
 
     if (!config) {
         TResponse response;
-        *response.MutableError() = MakeError(E_REJECTED, "Volume not mounted");
+        *response.MutableError() = MakeError(E_BS_INVALID_SESSION, "Volume not mounted");
```
drbasic: clang-format

drbasic (approved):

SvartMetal (approved):

Issue #6011, opened 2026-05-21
[Filestore] reduce the number of zero metrics reported by filestore-vhost


debnatkh:
The main idea is:
1. Stop reporting root/ssd/hdd level requests for non-essential requests (say, `GetNodeXAttr` or `KickEndpoint`) and make them lazy
2. For non-essential requests, do not report the time histogram. For non-throttlable requests, do not report throttling-related params
3. Do not report `PostponedQueueSize`, `MaxPostponedQueueSize`, `PostponedCount`, `PostponedQueueSizeGrpc`, `MaxPostponedQueueSizeGrpc`, `PostponedCountGrpc` for most of the requests

PR #6012, opened 2026-05-21
issue-6011: [Filestore] enable lazy initialization for all non-IO requests in filestore

debnatkh:
### Notes
First step described in the original issue: _Stop reporting root/ssd/hdd-level requests for non-essential requests (e.g., GetNodeXAttr or KickEndpoint) and make them lazy_.

+stop registering the time histogram for all 60+ requests

In the case of a stale service, the number of counters went from:

```
$ curl -s localhost:8768/counters/prometheus | wc -l
20456
```
to 
```
$  curl -s localhost:8768/counters/prometheus | wc -l                                    
8012
```

### Issue
#6011

### DIFF
--- a/cloud/storage/core/libs/user_stats/counter/user_counter.cpp
+++ b/cloud/storage/core/libs/user_stats/counter/user_counter.cpp
@@ -230,28 +230,24 @@ private:
 
     const TBuckets Buckets;
     const TString Units;
-    TVector<TBaseDynamicCounters> BaseCounters;
+    // Original inputs. Kept for lazy re-resolution if base subgroups are
+    // not yet registered.
+    const TVector<TBaseDynamicCounters> OriginalInputs;
+    // Empty if source histograms are lazily registered (not yet created at
+    // construction time).
+    const TVector<TBaseDynamicCounters> BaseCounters;
+    // True if some of the subgroups were not yet registered at construction.
+    // GetValue re-resolves from OriginalInputs on every call in this case.
+    const bool IsLazy;
     const size_t HistogramSize;
     TIntrusivePtr<TExplicitHistogramSnapshot> Histogram;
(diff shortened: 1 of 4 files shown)


cloud/storage/core/libs/user_stats/counter/user_counter.cpp
```diff
@@ -296,9 +312,14 @@ class TUserSumHistogramWrapper
         TInstant time,
         NMonitoring::IMetricConsumer* consumer) const override
     {
+        // Re-resolve on every call when lazy: histogram subgroups may not have
```
qkrorlqr, line 315: isn't it very expensive?

debnatkh: This happens every 15 seconds, FindSubgroup is just a map lookup

cloud/storage/core/libs/user_stats/counter/user_counter.cpp
```diff
@@ -296,9 +312,14 @@ class TUserSumHistogramWrapper
...
+        // existed at construction but are created once the first request fires.
+        const auto lazyCounters =
+            IsLazy ? ResolveBaseCounters() : TVector<TBaseDynamicCounters>{};
+        const auto& counters = IsLazy ? lazyCounters : BaseCounters;
```
qkrorlqr, line 319: these 2 lines can be simplified to just one:
```
const auto& counters = IsLazy ? ResolveBaseCounters() : BaseCounters;
```

debnatkh: I was doing it to avoid copying BaseCounters to the local variable, since ResolveBaseCounters will be prvalue and I don't think I can make a const reference to it

qkrorlqr (approved):

komarevtsev-d (approved):

PR #6038, opened 2026-05-27
issue-4649: Add E_TRANSPORT_ERROR to the aborted error kinds

debnatkh:
### Notes
This error is expected after a filestore restart and should not be treated as a fatal one

### Issue
#4649


SvartMetal (dismissed):

SvartMetal (dismissed):

cloud/storage/core/libs/common/error.cpp
```diff
@@ -56,6 +56,7 @@ EErrorKind GetErrorKind(const NProto::TError& e)
         case E_BS_THROTTLED:
         case E_FS_THROTTLED:
         case E_RDMA_UNAVAILABLE:
+        case E_TRANSPORT_ERROR:
             return EErrorKind::ErrorRetriable;
```
SvartMetal, line 59: https://github.com/ydb-platform/nbs/blob/3d2dbad0b1d21323b56fc8a354189892f2d7e181/cloud/filestore/libs/server/server_memory_state.cpp#L145

Maybe it should not be retriable in the sense that the request cannot be retried as is. Instead, the whole request chain should be retried, more like with an E_ABORTED error code https://github.com/ydb-platform/nbs/blob/main/doc/blockstore/diagnostics/result-codes.md

https://github.com/ydb-platform/nbs/blob/3d2dbad0b1d21323b56fc8a354189892f2d7e181/cloud/storage/core/libs/common/error.cpp#L65

debnatkh: Sounds clean and reasonable, yes

SvartMetal (approved):

agalibin (approved):

PR #6040, opened 2026-05-27
issue-4961: add restart of main tablet to mitigate filesystem destroy hangs on active sessions; make ForceDestroySizeThreshold ui64

WilyTiger:
### Notes
Expose index tablet uptime via GetStorageStats and add configurable RestartTabletUptimeThresholdDuringDestroy. When destroy is blocked by active sessions, restart the index tablet only if the threshold is set and the tablet has been running long enough, allowing sessions to become orphan and later be cleaned up. 

When RestartTabletUptimeThresholdDuringDestroy is enabled, it disables AllowFileStoreDestroyWithOrphanSessions to avoid destroying a filesystem with orphan sessions created by the tablet restart.

### Issue
https://github.com/ydb-platform/nbs/issues/4961


debnatkh (dismissed):

cloud/filestore/libs/storage/core/config.cpp
```diff
@@ -333,7 +333,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(ResponseLogEntryTTL,                TDuration,  TDuration::Hours(1)   )\
     xxx(TabletRegularTasksSchedulePeriod,   TDuration,  TDuration::Minutes(1) )\
                                                                                \
-    xxx(ForceDestroySizeThreshold,          ui32,       0                     )\
+    xxx(ForceDestroySizeThreshold,          ui64,       0                     )\
+    xxx(RestartTabletUptimeThreshold,       TDuration,  TDuration::Zero()     )\
```
debnatkh: Let us make it a reasonable value by default? Say, 1h?

cloud/filestore/libs/storage/service/service_actor_destroyfs.cpp
```diff
@@ -198,13 +246,59 @@ void TDestroyFileStoreActor::HandleDescribeSessionsResponse(
         for (const auto& sessionInfo: msg->Record.GetSessions()) {
             message << " " << sessionInfo.GetClientId();
         }
+
+        if (RestartTabletUptimeThreshold &&
+            TabletUptime > RestartTabletUptimeThreshold)
```
qkrorlqr: instead of uptime check we can just check whether all sessions are orphans

if all of them are orphans then there's no point in restarting the tablet

this makes the change less complex and removes the need to keep IdleSessionsTimeout and RestartTabletUptimeThreshold in sync

cloud/filestore/libs/storage/core/config.cpp
WilyTiger: No, this is needed so that, by default, this restart logic doesn’t run; that’s why the default is set to 0

cloud/filestore/libs/storage/service/service_actor_destroyfs.cpp
WilyTiger: If we do it this way, then for file systems that have active sessions, tablets will keep getting killed on retries. We agreed to add to the current logic a check that we shouldn’t kill the tablet if all sessions are orphaned

qkrorlqr (dismissed):

agalibin (dismissed):

agalibin (dismissed):

nisqatsi (approved):

qkrorlqr (approved):

PR #6060, opened 2026-05-29
Read BlockMasks and BlobMetas in separate transaction

vladstepanyuk:
### Notes
Moved reading block masks  and blob metas into separate transactions to be able to execute them in parallel with the readblob/writeblob, which can speed up compacting in case blob splitting optimization(ReadBlockMaskOnCompactionOptimizationEnabled + SplitByCompactionRangeMaxBlobCount features) is not working.

### Issue
#5590

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor.h
+++ b/cloud/blockstore/libs/storage/partition/part_actor.h
@@ -130,13 +130,6 @@ private:
     const ui64 ChannelHistorySize;
     const NBlockCodecs::ICodec* BlobCodec;
     const ui64 VolumeTabletId;
-    const bool SplitCompactionTxEnabled =
-        Config->GetSplitCompactionTxEnabled() ||
-        Config->IsSplitCompactionTxFeatureEnabled(
-            PartitionConfig.GetCloudId(),
-            PartitionConfig.GetFolderId(),
-            PartitionConfig.GetDiskId());
-
     TLogTitle LogTitle;
 
     std::unique_ptr<TPartitionState> State;
--- a/cloud/blockstore/libs/storage/partition/part_actor.h
+++ b/cloud/blockstore/libs/storage/partition/part_actor.h
@@ -130,6 +130,12 @@ private:
     const ui64 ChannelHistorySize;
     const NBlockCodecs::ICodec* BlobCodec;
     const ui64 VolumeTabletId;
+    const bool SplitCompactionTxEnabled =
+        Config->GetSplitCompactionTxEnabled() ||
+        Config->IsSplitCompactionTxFeatureEnabled(
+            PartitionConfig.GetCloudId(),
+            PartitionConfig.GetFolderId(),
+            PartitionConfig.GetDiskId());
 
     TLogTitle LogTitle;
 
--- a/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
@@ -88,6 +88,15 @@ private:
     TVector<TRangeCompactionInfo> RangeCompactionInfos;
     TVector<TRequest> Requests;
 
+    const bool SplitTxEnabled;
+    TVector<TPartialBlobId> BlobsToReadBlockMasks;
+    TVector<TPartialBlobId> BlobsToReadBlobMetas;
+
+    THashMap<TPartialBlobId, size_t, TPartialBlobIdHash>
+        BlockMaskResponseIndex;
+    THashMap<TPartialBlobId, size_t, TPartialBlobIdHash>
+        BlobMetaResponseIndex;
+
     TVector<IProfileLog::TBlockInfo> AffectedBlockInfos;
 
     TVector<TBatchRequest> BatchRequests;
@@ -132,7 +141,10 @@ public:
         TVector<TRangeCompactionInfo> rangeCompactionInfos,
         TVector<TRequest> requests,
         TDuration compactionTxTime,
-        TChildLogTitle logTitle);
+        TChildLogTitle logTitle,
+        bool splitTxEnabled,
+        TVector<TPartialBlobId> blobsToReadBlockMasks,
+        TVector<TPartialBlobId> blobsToReadBlobMetas);
 
     void Bootstrap(const TActorContext& ctx);
 
@@ -143,6 +155,8 @@ private:
     void WriteBlobs(const TActorContext& ctx);
     void AddBlobs(const TActorContext& ctx);
     void MakeDiffs(TRangeCompactionInfo& rc);
+
+    void ContinueAfterBlobInfo(const TActorContext& ctx);
 
     void NotifyCompleted(const TActorContext& ctx, const NProto::TError& error);
     bool HandleError(const TActorContext& ctx, const NProto::TError& error);
@@ -175,6 +189,10 @@ private:
         const TEvPartitionPrivate::TEvAddBlobsResponse::TPtr& ev,
         const TActorContext& ctx);
 
+    void HandleCompactionReadBlobInfoResponse(
+        const TEvPartitionPrivate::TEvCompactionReadBlobInfoResponse::TPtr& ev,
+        const TActorContext& ctx);
+
     void HandlePoisonPill(
         const TEvents::TEvPoisonPill::TPtr& ev,
         const TActorContext& ctx);
@@ -198,7 +216,10 @@ TCompactionActor::TCompactionActor(
         TVector<TRangeCompactionInfo> rangeCompactionInfos,
         TVector<TRequest> requests,
         TDuration compactionTxTime,
-        TChildLogTitle logTitle)
+        TChildLogTitle logTitle,
+        bool splitTxEnabled,
+        TVector<TPartialBlobId> blobsToReadBlockMasks,
+        TVector<TPartialBlobId> blobsToReadBlobMetas)
     : RequestInfo(std::move(requestInfo))
     , TabletId(tabletId)
     , DiskId(std::move(diskId))
@@ -214,6 +235,9 @@ TCompactionActor::TCompactionActor(
     , CommitId(commitId)
     , RangeCompactionInfos(std::move(rangeCompactionInfos))
     , Requests(std::move(requests))
+    , SplitTxEnabled(splitTxEnabled)
+    , BlobsToReadBlockMasks(std::move(blobsToReadBlockMasks))
+    , BlobsToReadBlobMetas(std::move(blobsToReadBlobMetas))
     , CompactionTxTime(compactionTxTime)
 {}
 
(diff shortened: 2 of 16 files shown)


cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -2409,4 +2641,113 @@ void TPartitionActor::CompleteCompaction(
...
+            continue;
+        }
+        STORAGE_VERIFY_C(
+            mask.Defined(),
+            TWellKnownEntityTypes::TABLET,
+            TabletID(),
+            TStringBuilder()
+                << "Could not read block mask for blob: "
+                << MakeBlobId(TabletID(), args.BlobsToReadBlockMasks[i]));
+        args.BlockMasks[i] = std::move(*mask);
```
komarevtsev-d: Мув не нужен

komarevtsev-d (dismissed):

komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition/part_compaction_logic_ut.cpp
SvartMetal, line 1: Not added to ya.make

cloud/blockstore/libs/storage/partition/part_actor.h
```diff
@@ -130,6 +130,12 @@ class TPartitionActor final
     const ui64 ChannelHistorySize;
     const NBlockCodecs::ICodec* BlobCodec;
     const ui64 VolumeTabletId;
+    const bool SplitCompactionTxEnabled =
+        Config->GetSplitCompactionTxEnabled() ||
+        Config->IsSplitCompactionTxFeatureEnabled(
+            PartitionConfig.GetCloudId(),
+            PartitionConfig.GetFolderId(),
+            PartitionConfig.GetDiskId());
```
SvartMetal: We usually compute feature enablement in TPartitionActor methods, see for example https://github.com/ydb-platform/nbs/blob/09558ff3d0102ab5f555a8230a519eda1e75e7d5/cloud/blockstore/libs/storage/partition/part_actor_writemergedblocks.cpp#L557

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -87,6 +87,15 @@ class TCompactionActor final
     TVector<TRangeCompactionInfo> RangeCompactionInfos;
     TVector<TRequest> Requests;
 
+    const bool SplitTxEnabled;
```
SvartMetal: Let's add a detailed comment describing how the feature works

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -426,33 +434,18 @@ void ReadAffectedBlobsForCompaction(
             !readBlockMaskOnCompactionOptimizationEnabled)
         {
             state.IncrementBlockMaskReadDuringCompaction();
```
SvartMetal: Should it be there, especially given the possibility of transaction restarts (when read == false)?

SvartMetal (changes requested):

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -89,6 +92,20 @@ TBlobCompactionRequest::TBlobCompactionRequest(
...
 {}
 
+void ApplyChecksumFixups(TRangeCompactionInfo& rc)
+{
+    for (const auto& fixup: rc.ChecksumFixups) {
+        const auto* ab = rc.AffectedBlobs.FindPtr(fixup.BlobId);
+        Y_ABORT_UNLESS(ab);
+        auto* meta = ab->BlobMeta.Get();
+        Y_ABORT_UNLESS(meta);
+        if (fixup.BlobOffset < meta->BlockChecksumsSize()) {
```
SvartMetal: Is it expected to happen?

Should we at least have debug abort for that?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -87,6 +87,15 @@ class TCompactionActor final
...
+    TVector<TPartialBlobId> BlobsToReadBlockMasks;
+    TVector<TPartialBlobId> BlobsToReadBlobMetas;
+
+    THashMap<TPartialBlobId, size_t, TPartialBlobIdHash>
+        BlockMaskResponseIndex;
+    THashMap<TPartialBlobId, size_t, TPartialBlobIdHash>
+        BlobMetaResponseIndex;
```
SvartMetal: Maybe just save `TVector<TPartialBlobId>` for each request?

We can search AffectedBlobs by those blob ids afterwards

Blob ids positions in the vector should match corresponding records in the TEvCompactionReadBlobInfoResponse

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -525,23 +519,14 @@ TBuildBlobContentAndMasksResult BuildBlobContentAndMasks(
...
-                    }
-
-                    result.BlockChecksums.push_back(blockChecksum);
+                    // checksums will be moved from the affected blob's meta
+                    // later
+                    result.ChecksumFixups.emplace_back(
+                        result.BlockChecksums.size(),   // ChecksumIndex
+                        mark.BlobId,
+                        mark.BlobOffset);
+                    result.BlockChecksums.push_back(0);
```
SvartMetal: How do we ensure that zero checksums will not be read later (when writing them to profile logs)?

Maybe add a safe guard for that?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -228,6 +252,35 @@ void TCompactionActor::Bootstrap(const TActorContext& ctx)
...
+        auto request = std::make_unique<
+            TEvPartitionPrivate::TEvCompactionReadBlobInfoRequest>(
+            std::move(BlobsToReadBlockMasks),
+            std::move(BlobsToReadBlobMetas));
+
+        NCloud::Send(ctx, Tablet, std::move(request));
+        return;
+    }
+
+    ContinueAfterBlobInfo(ctx);
```
SvartMetal: CompactionReadBlobInfosFinished

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
SvartMetal: It looks like an invalid checksum (0) can be propagated further for blobs that were not read, for example when the read was optimized by patching

Because HandleReadBlobResponse (and thus ApplyChecksumFixups) will not be called for such blobs

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
vladstepanyuk: Yes, checksums are filled only for some disk prefix. Mixed blobs can be very wide, and some blocks from these blobs may not be under the checksum prefix.

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
vladstepanyuk: This is used only for testing and was added to the tests in which transactions are not restarting. So, I think it doesn't really matter if it's incremented or not on transaction restarts.

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -32,6 +32,19 @@ namespace {
...
 ////////////////////////////////////////////////////////////////////////////////
 
+TVector<ui32> EnsureBlockChecksums(
+    const TVector<std::optional<ui32>>& blockChecksums,
+    ui64 tabletId)
+{
+    TVector<ui32> result;
+    result.reserve(blockChecksums.size());
+    for (const auto& checksum: blockChecksums) {
+        STORAGE_VERIFY(checksum, TWellKnownEntityTypes::TABLET, tabletId);
```
SvartMetal: Let's fire an impossible critical event here, but avoid crashing the process in release

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
SvartMetal: But it does not belong here anymore. It seems better to keep it next to `db.ReadBlockMask`, as it was before.

SvartMetal (changes requested):

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -837,13 +901,17 @@ void TCompactionActor::HandleReadBlobResponse(
...
     Y_ABORT_UNLESS(batchIndex < BatchRequests.size());
     auto& batch = BatchRequests[batchIndex];
-    const auto& rc = *batch.RangeCompactionInfo;
+    auto& rc = *batch.RangeCompactionInfo;
 
     const auto n = Min(batch.Requests.size(), msg->BlockChecksums.size());
     for (ui32 i = 0; i < n; ++i) {
         const auto* r = batch.Requests[i];
+        STORAGE_VERIFY(
+            rc.BlockChecksums[r->IndexInBlobContent],
```
SvartMetal: Let's fire an impossible event instead of crash in release

And ignore empty checksum

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -525,23 +519,14 @@ TBuildBlobContentAndMasksResult BuildBlobContentAndMasks(
...
-                    }
-
-                    result.BlockChecksums.push_back(blockChecksum);
+                    // checksums will be moved from the affected blob's meta
+                    // later
+                    result.ChecksumFixups.emplace_back(
+                        result.BlockChecksums.size(),   // ChecksumIndex
+                        mark.BlobId,
+                        mark.BlobOffset);
+                    result.BlockChecksums.push_back(std::nullopt);
```
SvartMetal, line 520: Let's not change the default behavior. When `SplitCompactionTxEnabled = false` and we have `BlobMeta`s, we can build checksums here.

vladstepanyuk: I think the Prepare and Complete Compaction code is already very complicated. So, it would be better not to add new if statements for every new feature. Instead, we should try to unify the logic. And we have impossible events to ensure that all checksums are filled.

cloud/blockstore/libs/storage/partition/part_actor_readblobinfo.cpp
```diff
@@ -51,6 +51,9 @@ bool TPartitionActor::PrepareCompactionReadBlobInfo(
 
     bool ready = true;
 
+    State->IncrementBlockMaskReadDuringCompaction(
+        args.BlobsToReadBlockMasks.size());
```
SvartMetal: Let's move it to CompleteCompactionReadBlobInfo?

SvartMetal (dismissed):

komarevtsev-d (dismissed):

SvartMetal (approved):

agalibin (approved):

PR #6061, opened 2026-05-29
[Filestore] suicide if blob storage return error during load confirmation

neihar:
### Notes
If blob storage return error we want to suicide and try again instead of hang in broken state and wait manual restart.
We expect that blob storage should answer for requests without error eventually

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_confirmblobs.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_confirmblobs.cpp
@@ -263,7 +263,7 @@ void TIndexTabletActor::HandleConfirmBlobsCompleted(
             TStringBuilder()
             << "tabletId: " << TabletID() << ", error: " << errorMessage);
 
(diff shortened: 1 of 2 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_confirmblobs.cpp
```diff
@@ -263,7 +263,7 @@ void TIndexTabletActor::HandleConfirmBlobsCompleted(
             TStringBuilder()
             << "tabletId: " << TabletID() << ", error: " << errorMessage);
 
-        BecomeAux(ctx, STATE_BROKEN);
+        Suicide(ctx);
```
SvartMetal, line 266: Test?

neihar: added

SvartMetal (approved):

agalibin (approved):

PR #6064, opened 2026-05-29
[Blockstore] TDurableClient should initialize RequestId if it is missing

SvartMetal:
### Notes
Noticed that RequestId is sometimes missing in logs:
```
2026/05/29 12:05:32.461657 ReleaseNVMeDevice retry request (retries: 2, timeout: 5m0s, error: E_GRPC_DEADLINE_EXCEEDED Deadline Exceeded)
```

RequestId initialization is currently present only in TClient code: https://github.com/ydb-platform/nbs/blob/d1c9ed4b9680da97f760dc4fe12a3d7cf44cde63/cloud/blockstore/libs/client/client.cpp#L435

This change makes the logic more robust by ensuring that RequestId is set before TDurableClient starts request execution.

### Issue

### DIFF
--- a/cloud/blockstore/libs/client/durable.cpp
+++ b/cloud/blockstore/libs/client/durable.cpp
@@ -188,6 +188,8 @@ private:
             return;
         }
 
+        EnsureRequestId(*request);
+
         TMethod::Execute(Client.get(), state->CallContext, std::move(request))
             .Subscribe(
                 [state = std::move(state), weakSelf = this->weak_from_this()](
--- a/cloud/blockstore/libs/client/durable_ut.cpp
+++ b/cloud/blockstore/libs/client/durable_ut.cpp
@@ -7,7 +7,9 @@
 #include <cloud/blockstore/libs/diagnostics/request_stats.h>
 #include <cloud/blockstore/libs/diagnostics/volume_stats.h>
 #include <cloud/blockstore/libs/service/context.h>
+#include <cloud/blockstore/libs/service/request_helpers.h>
 #include <cloud/blockstore/libs/service/service_test.h>
+
 #include <cloud/storage/core/libs/common/error.h>
 #include <cloud/storage/core/libs/common/helpers.h>
 #include <cloud/storage/core/libs/common/scheduler.h>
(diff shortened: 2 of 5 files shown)


cloud/filestore/libs/client/bench/ya.make
```diff
@@ -1,5 +1,7 @@
 Y_BENCHMARK()
 
+SIZE(medium)
```
SvartMetal, line 3: Times out locally

cloud/blockstore/libs/client/durable.cpp
neihar, line 233: Do we have correct logs? I see that we use state->requests, but Ensure gives for each retry separate one?

neihar: do we need to update state->CallContext->RequestId as well?

cloud/blockstore/libs/client/durable_ut.cpp
```diff
@@ -1432,7 +1435,7 @@ Y_UNIT_TEST_SUITE(TDurableClientTest)
 
         client->PingHandler =
             [&] (std::shared_ptr<NProto::TPingRequest> request) {
-                Y_UNUSED(request);
+                UNIT_ASSERT_UNEQUAL(0, GetRequestId(*request));
```
neihar, line 1438: can we check it also for something like WriteBlocksLocal

cloud/blockstore/libs/client/durable.cpp
SvartMetal: >  Ensure gives for each retry separate one?

Ensure initializes request id only once

SvartMetal: > do we need to update state->CallContext->RequestId as well?

I think it is unrelated to this PR

cloud/blockstore/libs/client/durable_ut.cpp
SvartMetal: I think current checks are sufficient, otherwise problem is not so critical

neihar (approved):

komarevtsev-d (approved):

PR #6070, opened 2026-05-30
support zero blocks in snapshot encryption client

EvgeniyKozev:
### Problem

Creating a disk from an encrypted snapshot could fail when the restore flow issued `ZeroBlocks` through `TSnapshotEncryptionClient`.

Example error:

```text
commited fatal error for disks.CreateDiskFromSnapshot with task id 8de84cb7-bdda-48b5-a240-08d6b1accd88: Detailed error, Details=<nil>, Silent=false: rpc error: code = Unknown desc = E_NOT_IMPLEMENTED ZeroBlocks requests not supported by snapshot encryption client
```

There is no linked issue/ticket for this fix.

### Solution

Support `ZeroBlocks` in `TSnapshotEncryptionClient` by forwarding the request to the underlying blockstore client.

The snapshot encryption client does not encrypt writes or decrypt reads. Its read path only masks blocks marked by `UnencryptedBlockMask`. Because of that, forwarding native `ZeroBlocks` is preferable to materializing zero-filled `WriteBlocksLocal` requests: it preserves snapshot sparseness and keeps the underlying storage semantics for zeroed/unwritten blocks.

### Tests

```bash
./ya make -tt cloud/blockstore/libs/encryption/ut
```

### DIFF
--- a/cloud/blockstore/libs/encryption/encryption_client.cpp
+++ b/cloud/blockstore/libs/encryption/encryption_client.cpp
@@ -884,6 +884,9 @@ class TSnapshotEncryptionClient final
 private:
     const NProto::TEncryptionDesc EncryptionDesc;
 
+    ui32 BlockSize = 0;
+    TString ZeroBlock;
+
     TLog Log;
 
 public:
@@ -908,11 +911,18 @@ public:
         TCallContextPtr callContext,
         std::shared_ptr<NProto::TReadBlocksLocalRequest> request) override;
 
+    TFuture<NProto::TWriteBlocksLocalResponse> WriteBlocksLocal(
+        TCallContextPtr callContext,
+        std::shared_ptr<NProto::TWriteBlocksLocalRequest> request) override;
+
     TFuture<NProto::TZeroBlocksResponse> ZeroBlocks(
         TCallContextPtr callContext,
         std::shared_ptr<NProto::TZeroBlocksRequest> request) override;
 
 private:
+    void HandleMountVolumeResponse(
+        const NProto::TMountVolumeResponse& response);
+
     static NProto::TReadBlocksResponse HandleReadBlocksResponse(
         NProto::TReadBlocksResponse response);
 
@@ -942,9 +952,26 @@ TFuture<NProto::TMountVolumeResponse> TSnapshotEncryptionClient::MountVolume(
     encryption.SetMode(EncryptionDesc.GetMode());
     encryption.SetKeyHash(EncryptionDesc.GetKeyHash());
 
-    return Client->MountVolume(
+    auto future = Client->MountVolume(
         std::move(callContext),
         std::move(request));
+
+    return future.Apply([this] (const auto& f) {
+        const auto& response = f.GetValue();
+        if (!HasError(response)) {
+            HandleMountVolumeResponse(response);
+        }
+        return response;
+    });
+}
+
+void TSnapshotEncryptionClient::HandleMountVolumeResponse(
+    const NProto::TMountVolumeResponse& response)
+{
+    if (BlockSize == 0) {
+        BlockSize = response.GetVolume().GetBlockSize();
+        ZeroBlock = TString(BlockSize, 0);
+    }
 }
 
 TFuture<NProto::TReadBlocksResponse> TSnapshotEncryptionClient::ReadBlocks(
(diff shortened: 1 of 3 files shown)


agalibin:
Please add a description and justification
The PR template is empty — no description and no linked issue. Before review, could you fill in:

What problem does this solve? A concrete scenario where TSnapshotEncryptionClient::ZeroBlocks is called and currently fails with E_NOT_IMPLEMENTED (who is the caller — disk_manager snapshot transfer/restore? fill flow?).
Link the issue/ticket.
Design questions
1. Why materialize physical zero writes instead of delegating to Client->ZeroBlocks?
For the regular TEncryptionClient the write-conversion is needed because a native zero would later be decrypted into garbage. But the snapshot client neither encrypts writes nor decrypts reads — so that reasoning doesn't apply here. Forwarding to the underlying ZeroBlocks (like the EZP_WRITE_ZERO_BLOCKS policy path) looks more natural and:

preserves snapshot sparseness — WriteBlocks(zeros) forces block allocation and physically stores zero blocks, which can significantly inflate the snapshot when zeroing large sparse ranges;
keeps the UnencryptedBlockMask semantics that this client's own read path relies on — after a zero-write the blocks are recorded as "written/encrypted" data rather than unencrypted/holes. Is the mask consumed anywhere else (incremental snapshots / CBT / dedup)? If so, this changes its meaning.
Could you describe why the physical-write approach was chosen? Does the underlying snapshot storage not support native ZeroBlocks?

2. Error code for the not-mounted case.
When BlockSize == 0 (mount hasn't completed) ZeroBlocks returns E_ARGUMENT. That's a state error, not an argument error — E_INVALID_STATE / E_REJECTED would be more accurate, and it would let callers distinguish "called before mount" from the genuine "BlocksCount == 0" case. The renamed test currently bakes in E_ARGUMENT for the pre-mount path.

Notes / nits
Thread-safety: BlockSize/ZeroBlock are written in the MountVolume callback and read in ZeroBlocks with no synchronization. Not a regression (same pattern as TEncryptionClient), but it's a data race if mount and requests land on different threads (TSAN).
BlockSize is set once (if (BlockSize == 0)); a remount with a different block size leaves it stale. Edge case, also inherited.
Test coverage: good happy-path + pre-mount tests. Please add a case where the underlying WriteBlocksLocal returns an error, to verify it's propagated correctly into TZeroBlocksResponse.
The unrelated filestore CI failure is a flake (passed on rerun).
Summary
The change is functionally sound and well-tested, and is self-consistent within the client (zeros read back as zeros). The main open point is the architectural choice of physical zero-writes vs. delegating native ZeroBlocks (sparseness + mask semantics). Please add the description/justification and consider the E_INVALID_STATE tweak.

EvgeniyKozev:
> Please add a description and justification The PR template is empty — no description and no linked issue. Before review, could you fill in:
> 
> What problem does this solve? A concrete scenario where TSnapshotEncryptionClient::ZeroBlocks is called and currently fails with E_NOT_IMPLEMENTED (who is the caller — disk_manager snapshot transfer/restore? fill flow?). Link the issue/ticket. Design questions
> 
> 1. Why materialize physical zero writes instead of delegating to Client->ZeroBlocks?
>    For the regular TEncryptionClient the write-conversion is needed because a native zero would later be decrypted into garbage. But the snapshot client neither encrypts writes nor decrypts reads — so that reasoning doesn't apply here. Forwarding to the underlying ZeroBlocks (like the EZP_WRITE_ZERO_BLOCKS policy path) looks more natural and:
> 
> preserves snapshot sparseness — WriteBlocks(zeros) forces block allocation and physically stores zero blocks, which can significantly inflate the snapshot when zeroing large sparse ranges; keeps the UnencryptedBlockMask semantics that this client's own read path relies on — after a zero-write the blocks are recorded as "written/encrypted" data rather than unencrypted/holes. Is the mask consumed anywhere else (incremental snapshots / CBT / dedup)? If so, this changes its meaning. Could you describe why the physical-write approach was chosen? Does the underlying snapshot storage not support native ZeroBlocks?
> 
> 2. Error code for the not-mounted case.
>    When BlockSize == 0 (mount hasn't completed) ZeroBlocks returns E_ARGUMENT. That's a state error, not an argument error — E_INVALID_STATE / E_REJECTED would be more accurate, and it would let callers distinguish "called before mount" from the genuine "BlocksCount == 0" case. The renamed test currently bakes in E_ARGUMENT for the pre-mount path.
> 
> Notes / nits Thread-safety: BlockSize/ZeroBlock are written in the MountVolume callback and read in ZeroBlocks with no synchronization. Not a regression (same pattern as TEncryptionClient), but it's a data race if mount and requests land on different threads (TSAN). BlockSize is set once (if (BlockSize == 0)); a remount with a different block size leaves it stale. Edge case, also inherited. Test coverage: good happy-path + pre-mount tests. Please add a case where the underlying WriteBlocksLocal returns an error, to verify it's propagated correctly into TZeroBlocksResponse. The unrelated filestore CI failure is a flake (passed on rerun). Summary The change is functionally sound and well-tested, and is self-consistent within the client (zeros read back as zeros). The main open point is the architectural choice of physical zero-writes vs. delegating native ZeroBlocks (sparseness + mask semantics). Please add the description/justification and consider the E_INVALID_STATE tweak.

Thanks for the review. I agree with the main point about preserving native `ZeroBlocks` semantics in the snapshot encryption client.

What changed since the previous version:

- `TSnapshotEncryptionClient::ZeroBlocks` now forwards the request directly to the underlying blockstore client instead of converting it into `WriteBlocksLocal` with a zero-filled sglist.
- Removed the snapshot-client-local `BlockSize` / `ZeroBlock` state and the `MountVolume` callback that initialized it. This also removes the related pre-mount `BlockSize == 0` path and avoids adding new state/race concerns to the snapshot client.
- Removed the extra `WriteBlocksLocal` override from `TSnapshotEncryptionClient`; the default wrapper behavior is enough now.
- Updated tests to verify that snapshot `ZeroBlocks` is forwarded with the original request fields intact.
- Added coverage for propagation of an error returned by the underlying `ZeroBlocks` into `TZeroBlocksResponse`.
- Updated the pre-mount test to match the new behavior: the snapshot client no longer rejects locally with `E_ARGUMENT`; it forwards the request and lets the underlying client return the appropriate state error.
- Updated the destroy-client test to expect forwarded `ZeroBlocks` to remain pending until the underlying future completes, matching the other forwarded requests.

Rationale:

`TSnapshotEncryptionClient` does not encrypt writes and does not decrypt reads. Its read path only zeroes blocks marked in `UnencryptedBlockMask`. Because of that, materializing zero writes is unnecessary here and can change storage behavior by allocating physical zero blocks. Forwarding native `ZeroBlocks` preserves snapshot sparseness and keeps the underlying storage semantics for zeroed/unwritten ranges.

Issue/ticket:

There is no linked issue/ticket. The original user-visible failure was observed when creating a disk from a snapshot:

```text
commited fatal error for disks.CreateDiskFromSnapshot with task id 8de84cb7-bdda-48b5-a240-08d6b1accd88: Detailed error, Details=<nil>, Silent=false: rpc error: code = Unknown desc = E_NOT_IMPLEMENTED ZeroBlocks requests not supported by snapshot encryption client
```

Tests:

```bash
./ya make -tt cloud/blockstore/libs/encryption/ut
```

Result: 43 tests passed.

cloud/blockstore/libs/encryption/encryption_client.cpp
```diff
@@ -1025,12 +1025,9 @@ TFuture<NProto::TZeroBlocksResponse> TSnapshotEncryptionClient::ZeroBlocks(
     TCallContextPtr callContext,
     std::shared_ptr<NProto::TZeroBlocksRequest> request)
 {
-    Y_UNUSED(callContext);
-    Y_UNUSED(request);
-
-    return MakeFutureErrorResponse<NProto::TZeroBlocksResponse>(
-        E_NOT_IMPLEMENTED,
-        "ZeroBlocks requests not supported by snapshot encryption client");
+    return Client->ZeroBlocks(
```
drbasic: Тут нужно учитывать флаг EncryptZeroPolicy: EZP_WRITE_ZERO_BLOCKS

cloud/blockstore/libs/encryption/encryption_client_ut.cpp
```diff
@@ -534,6 +534,113 @@ Y_UNIT_TEST_SUITE(TEncryptionClientTest)
         }
     }
 
+    Y_UNIT_TEST(ShouldForwardZeroBlocksInSnapshotEncryptionClient)
```
drbasic: а точно нужен тест ShouldForwardZeroBlocksInSnapshotEncryptionClient, второго не достаточно?

cloud/blockstore/libs/encryption/encryption_client.cpp
EvgeniyKozev: Спасибо, комментарий валидный. Поправил в `22cfd5b96d`.

Теперь `TSnapshotEncryptionClient` учитывает `EncryptZeroPolicy`:
- `CreateSnapshotEncryptionClient` принимает policy явно.
- `TEncryptionClientFactory` передает текущий `EncryptZeroPolicy` в snapshot client.
- При `EZP_WRITE_ZERO_BLOCKS` snapshot client форвардит `ZeroBlocks` в underlying client.
- При `EZP_WRITE_ENCRYPTED_ZEROS` snapshot client материализует zero blocks через `WriteBlocksLocal`.
- Для pre-mount случая в ветке `EZP_WRITE_ENCRYPTED_ZEROS` возвращается `E_INVALID_STATE`, а не `E_ARGUMENT`.

Проверки:
`./ya make -tt cloud/blockstore/libs/encryption/ut` — 44 tests GOOD.
`git --no-pager diff --check` — clean.

cloud/blockstore/libs/encryption/encryption_client_ut.cpp
EvgeniyKozev: Согласен, отдельный `ShouldForwardZeroBlocksInSnapshotEncryptionClient` оказался избыточен: forwarding path уже покрыт тестом на propagation error от underlying `ZeroBlocks`.

Удалил этот тест в `22cfd5b96d`. Вместо него добавил недостающее покрытие для второй policy-ветки: `ShouldWriteZeroBlocksInSnapshotEncryptionClient` проверяет `EZP_WRITE_ENCRYPTED_ZEROS`, где snapshot client должен писать физические zero blocks через `WriteBlocksLocal`.

cloud/blockstore/libs/encryption/encryption_client.cpp
```diff
@@ -1025,12 +1056,69 @@ TFuture<NProto::TZeroBlocksResponse> TSnapshotEncryptionClient::ZeroBlocks(
     TCallContextPtr callContext,
     std::shared_ptr<NProto::TZeroBlocksRequest> request)
 {
-    Y_UNUSED(callContext);
-    Y_UNUSED(request);
+    if (EncryptZeroPolicy == NProto::EZP_WRITE_ZERO_BLOCKS) {
```
drbasic: это не решает проблему с EZP_WRITE_ZERO_BLOCKS потому что мы продолжаем писать блоки с нулями вместо зашифрованных блоков. 
В TSnapshotEncryptionClient невозможна стратегия, когда пишутся зашифрованные блоки, потому что TSnapshotEncryptionClient не имеет ключа шифрования. И правильно просто убрать TSnapshotEncryptionClient::ZeroBlocks, оставив ZeroBlocks по-умолчанию (как это сделано для WriteBlocks)

PR #6088, opened 2026-06-02
[Filestore] cleanup directory handles filemap if feature is disabled, but file exists

neihar:
### Notes
It is possible that the feature can be disabled right after startup or after a crash. In this case, we do not want to keep stale filemaps, so we clean up the directory handles filemap because it is no longer needed.

### Issue

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/loop.cpp
+++ b/cloud/filestore/libs/vfs_fuse/loop.cpp
@@ -1089,12 +1089,12 @@ private:
 
             IDirectoryHandleStorageStatsPtr directoryHandleStorageStats;
             TDirectoryHandleStoragePtr directoryHandleStorage;
-            if (FileSystemConfig->GetDirectoryHandlesStorageEnabled()) {
-                if (Config->GetDirectoryHandlesStoragePath()) {
-                    auto path =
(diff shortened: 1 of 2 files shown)


cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1121,14 +1120,27 @@ class TFileSystemLoop final
...
-                        "is not set",
-                        Config->GetFileSystemId().Quote().c_str(),
-                        Config->GetClientId().Quote().c_str());
+                } else if (path.Exists()) {
+                    // The feature is disabled but a file from a previous
+                    // session with it enabled is still on disk. The file
+                    // holds only a derived view of the directory listing,
+                    // so it can be removed without any drain.
+                    try {
+                        NFs::RemoveRecursive(path);
```
e673: This is extremely dangerous because it shares the directory with WriteBackCache.
You may delete WriteBackCache state as well.

e673 (approved):

e673 (dismissed):

yegorskii (dismissed):

e673 (approved):

yegorskii (approved):

PR #6132, opened 2026-06-05
[Disk Agent, Disk Registry] Show device model in monitoring pages

zavik001:
This PR is a rework of #5555.
  
  The original PR had a compilation error: NVMeIdentifyCtrl was called without the required AdminCmdTimeout argument. Since I don't have write access to that branch, I'm duplicating the
  implementation here with the fix applied. #5555 will be closed in favour of this PR.

  Disk Agent reads and exposes the hardware model of each device during initialization.

  - For NVMe devices, the model is read from the Identify Controller data structure (ctrl.mn)
  - For rotational devices (HDD), it is read from the ATA IDENTIFY response (hd.model)
  - The model is stored in TDeviceConfig.DeviceModel and propagated through the registration flow
  - The Disk Agent monitoring page gains a "Model" column next to the serial number
  - The Disk Registry monitoring page gains a "Model" column in the devices table and a "Model" field on the per-device info page

image
image

### DIFF
--- a/cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_ut.cpp
+++ b/cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_ut.cpp
@@ -98,10 +98,13 @@ TVector<ui64> FindProcessesWithOpenFile(const TString& targetPath)
 struct TTestNvmeManager: NNvme::INvmeManager
 {
     THashMap<TString, TString> PathToSerial;
+    THashMap<TString, TString> PathToModel;
 
-    explicit TTestNvmeManager(
-            const TVector<std::pair<TString, TString>>& pathToSerial)
+    TTestNvmeManager(
+            const TVector<std::pair<TString, TString>>& pathToSerial,
+            const TVector<std::pair<TString, TString>>& pathToModel)
         : PathToSerial{pathToSerial.cbegin(), pathToSerial.cend()}
+        , PathToModel{pathToModel.cbegin(), pathToModel.cend()}
     {}
 
     void Start() final
@@ -143,6 +146,23 @@ struct TTestNvmeManager: NNvme::INvmeManager
 
         auto it = PathToSerial.find(device);
         if (it == PathToSerial.end()) {
+            return MakeError(MAKE_SYSTEM_ERROR(42), path);
+        }
+
+        return it->second;
+    }
+
+    TResultOrError<TString> GetDeviceModel(const TString& path) override
+    {
+        TString device;
+        try {
+            device = NFs::ReadLink(path);
+        } catch (...) {
+            return MakeError(E_FAIL, CurrentExceptionMessage());
+        }
+
+        auto it = PathToModel.find(path);
+        if (it == PathToModel.end()) {
             return MakeError(MAKE_SYSTEM_ERROR(42), path);
         }
 
@@ -5502,6 +5522,13 @@ Y_UNIT_TEST_SUITE(TDiskAgentTest)
             {Devices[3], "Z"},   // nvme3n1
         };
 
+        TVector<std::pair<TString, TString>> pathToModel{
+            {Devices[0], "A"},   // nvme0n1
+            {Devices[1], "B"},   // nvme1n1
+            {Devices[2], "C"},   // nvme2n1
+            {Devices[3], "D"},   // nvme3n1
+        };
+
         // build the config cache
         {
             NProto::TDiskAgentConfig config;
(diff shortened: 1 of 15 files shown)


drbasic (dismissed):

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
budevg, line 411: shouldn't we set device model for static config as well ?

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
```diff
@@ -365,6 +369,33 @@ void TInitializer::SetupSerialNumbers(
     }
 }
 
+TString TInitializer::GetDeviceModel(const TString& path)
+{
+    auto it = PathToModel.find(path);
+    if (it == PathToModel.end()) {
+        auto [model, error] = NvmeManager->GetDeviceModel(path);
+        it = PathToModel.emplace(path, model).first;
```
budevg: in case of one time error the model will be cached as empty string ... maybe it's better to avoid caching in this case ? Or if it's on purpose let's add comment explaining why we want to cache empty string

budevg (changes requested):

cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_ut.cpp
```diff
@@ -149,6 +152,23 @@ struct TTestNvmeManager: NNvme::INvmeManager
...
+        TString device;
+        try {
+            device = NFs::ReadLink(path);
+        } catch (...) {
+            return MakeError(E_FAIL, CurrentExceptionMessage());
+        }
+
+        auto it = PathToModel.find(path);
+        if (it == PathToModel.end()) {
+            return MakeError(MAKE_SYSTEM_ERROR(42), path);
```
qkrorlqr: 42?

cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_ut.cpp
```diff
@@ -149,6 +152,23 @@ struct TTestNvmeManager: NNvme::INvmeManager
         return it->second;
     }
 
+    TResultOrError<TString> GetDeviceModel(const TString& path) override
+    {
+        TString device;
+        try {
+            device = NFs::ReadLink(path);
+        } catch (...) {
+            return MakeError(E_FAIL, CurrentExceptionMessage());
```
qkrorlqr, line 161: use a more specific error code please

if link reading fails because path doesn't exist, it should be something like `MAKE_SYSTEM_ERROR(ENOENT)`, etc

qkrorlqr: sorry, it's ut, not that important actually

cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_ut.cpp
zavik001: > 42?

replaced with EIO

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
zavik001: > shouldn't we set device model for static config as well ?

added device model setup for static config devices alongside serial number

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
zavik001: > in case of one time error the model will be cached as empty string ... maybe it's better to avoid caching in this case ? Or if it's on purpose let's add comment explaining why we want to cache empty string

removed caching on error, the model will be re-fetched on the next call instead of returning an empty string

drbasic (dismissed):

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
budevg, line 547: should we refresh the models as well here if we fallback to cached config ?

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
budevg: Actually there is similar caching pattern in GetSerialNumber let's fix it as well

budevg (changes requested):

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
zavik001: > should we refresh the models as well here if we fallback to cached config ?

yes

cloud/blockstore/libs/storage/disk_agent/storage_initializer.cpp
zavik001: > Actually there is similar caching pattern in GetSerialNumber let's fix it as well

done

budevg (approved):

drbasic (approved):

Issue #6142, opened 2026-06-05
[Disk Manager] Allow to specify S3 storage class for images


jkuradobery:
1) Allow to specify S3 storage class for images.
2) If we exceed the quota - use default storage.
3) Do not affect other oss users.

PR #6143, opened 2026-06-05
issue-6142: [Disk Manager] Allow to specify S3 storage class for images

jkuradobery:
### Issue
#6142

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/create_snapshot_from_legacy_snapshot_task.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/create_snapshot_from_legacy_snapshot_task.go
@@ -82,6 +82,7 @@ func (t *createSnapshotFromLegacySnapshotTask) Run(
 		t.storage,
 		ignoreZeroChunks,
 		t.request.UseS3,
+		"", // storageClass
 	)
 	defer target.Close(ctx)
 
--- a/cloud/disk_manager/internal/pkg/dataplane/migrate_snapshot.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/migrate_snapshot.go
@@ -82,6 +82,7 @@ func (t *migrateSnapshotTask) Run(
 		t.dstStorage,
 		true, // ignoreZeroChunks
 		t.useS3,
+		"", // storageClass
 	)
 	defer target.Close(ctx)
 
--- a/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
@@ -6,6 +6,8 @@ import (
 	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/facade/testcommon"
 )
 
+////////////////////////////////////////////////////////////////////////////////
+
 func TestImageServiceCreateImageFromDiskWithS3StorageClassQuotaFallback(t *testing.T) {
 	ctx := testcommon.NewContext()
 
--- a/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/facade/testcommon"
 )
 
@@ -19,4 +20,20 @@ func TestImageServiceCreateImageFromDiskWithS3StorageClassQuotaFallback(t *testi
 		"folder",
 		false, // pooled
 	)
+
+	quotaExceededCounters := testcommon.GetCountersDataplane(
+		t,
+		"errors_quotaExceeded",
+		map[string]string{
+			"call":      "PutObject",
+			"component": "s3_client",
+		},
+	)
+
+	var quotaExceededCount float64
+	for _, counter := range quotaExceededCounters {
+		quotaExceededCount += counter
+	}
+
+	require.NotZero(t, quotaExceededCount)
 }
--- a/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
@@ -1,0 +1,20 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/facade/testcommon"
+)
+
+func TestImageServiceCreateImageFromDiskWithS3StorageClassQuotaFallback(t *testing.T) {
+	ctx := testcommon.NewContext()
+
+	testcommon.CreateImage(
+		t,
+		ctx,
+		t.Name(),
+		4*1024*1024, // imageSize
+		"folder",
+		false, // pooled
+	)
+}
--- a/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
@@ -3,8 +3,18 @@ package tests
 import (
 	"testing"
 
+	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/stretchr/testify/require"
+	disk_manager "github.com/ydb-platform/nbs/cloud/disk_manager/api"
+	internal_client "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/client"
 	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/facade/testcommon"
+)
+
+////////////////////////////////////////////////////////////////////////////////
+
+const (
+	imageSize = 4 * 1024 * 1024
+	zoneID    = "zone-a"
 )
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -12,14 +22,58 @@ import (
 func TestImageServiceCreateImageFromDiskWithS3StorageClassQuotaFallback(t *testing.T) {
 	ctx := testcommon.NewContext()
 
-	testcommon.CreateImage(
-		t,
-		ctx,
-		t.Name(),
-		4*1024*1024, // imageSize
-		"folder",
-		false, // pooled
-	)
+	client, err := testcommon.NewClient(ctx)
+	require.NoError(t, err)
+	defer client.Close()
+
+	diskID := t.Name() + "_disk"
+
+	reqCtx := testcommon.GetRequestContext(t, ctx)
+	operation, err := client.CreateDisk(reqCtx, &disk_manager.CreateDiskRequest{
+		Src: &disk_manager.CreateDiskRequest_SrcEmpty{
+			SrcEmpty: &empty.Empty{},
+		},
+		Size: int64(imageSize),
+		Kind: disk_manager.DiskKind_DISK_KIND_SSD,
+		DiskId: &disk_manager.DiskId{
+			ZoneId: zoneID,
+			DiskId: diskID,
+		},
+	})
+	require.NoError(t, err)
+	require.NotEmpty(t, operation)
+	err = internal_client.WaitOperation(ctx, client, operation.Id)
+	require.NoError(t, err)
+
+	nbsClient := testcommon.NewNbsTestingClient(t, ctx, zoneID)
+	data := make([]byte, imageSize)
+	for i := range data {
+		data[i] = 1
+	}
+
+	err = nbsClient.Write(diskID, 0, data)
+	require.NoError(t, err)
+
+	reqCtx = testcommon.GetRequestContext(t, ctx)
+	operation, err = client.CreateImage(reqCtx, &disk_manager.CreateImageRequest{
+		Src: &disk_manager.CreateImageRequest_SrcDiskId{
+			SrcDiskId: &disk_manager.DiskId{
+				ZoneId: zoneID,
+				DiskId: diskID,
+			},
+		},
+		DstImageId: t.Name(),
+		FolderId:   "folder",
+		Pooled:     false,
+	})
+	require.NoError(t, err)
+	require.NotEmpty(t, operation)
+
+	response := disk_manager.CreateImageResponse{}
+	err = internal_client.WaitResponse(ctx, client, operation.Id, &response)
+	require.NoError(t, err)
+	require.Equal(t, int64(imageSize), response.Size)
+	require.NotZero(t, response.StorageSize)
 
 	quotaExceededCounters := testcommon.GetCountersDataplane(
 		t,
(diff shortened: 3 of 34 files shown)


cloud/disk_manager/internal/pkg/dataplane/create_snapshot_from_legacy_snapshot_task.go
```diff
@@ -82,6 +82,7 @@ func (t *createSnapshotFromLegacySnapshotTask) Run(
 		t.storage,
 		ignoreZeroChunks,
 		t.request.UseS3,
+		"", // storageClass
```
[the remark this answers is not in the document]
jkuradobery: Hell no.

cloud/disk_manager/test/mocks/s3-quota-proxy/main.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func parseQuotaSpecs(specs []string) map[string]uint64 {
+	quota := make(map[string]uint64)
+
+	for _, spec := range specs {
+		class, sizeStr, _ := strings.Cut(spec, ":")
+		size, _ := strconv.ParseUint(strings.TrimSpace(sizeStr), 10, 64)
+		quota[normalizeStorageClass(class)] = size
+	}
```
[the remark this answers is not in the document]
jkuradobery: I do that explicitly for the sake of simplicity.

cloud/tasks/persistence/s3.go
```diff
@@ -225,9 +229,15 @@ func (c *S3Client) GetObject(
 		return S3Object{}, errors.NewRetriableError(err)
 	}
 
+	storageClass := aws.StringValue(res.StorageClass)
+	if len(storageClass) == 0 {
+		storageClass = aws_s3.StorageClassStandard
+	}
```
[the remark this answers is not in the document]
jkuradobery: unrelated to this review

jkuradobery:
Results of actual fallback tests on our cluster:
https://gist.github.com/jkuradobery/46523fc13256279d43c953bdc9ac9183

cloud/disk_manager/internal/pkg/facade/image_service_s3_quota_test/image_service_s3_quota_test.go
```diff
@@ -0,0 +1,20 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/facade/testcommon"
+)
+
```
SvartMetal, line 12: /////...

cloud/tasks/persistence/s3.go
```diff
@@ -246,13 +256,34 @@ func (c *S3Client) PutObject(
...
 		ContentEncoding: aws.String("application/octet-stream"),
-	})
+	}
+	if len(object.StorageClass) != 0 {
+		input.StorageClass = aws.String(object.StorageClass)
+	}
+
+	_, err = c.s3.PutObjectWithContext(ctx, input)
+	if aerr, ok := err.(awserr.Error); ok &&
+		aerr.Code() == s3ErrCodeQuotaLimitExceeded &&
```
SvartMetal: Let's use this constant? https://github.com/ekkx/yaylib/blob/3be1f408f59810e1d4ac3af9e6901f786d5fb93a/error_codes.go#L64

cloud/tasks/persistence/s3.go
```diff
@@ -246,13 +256,34 @@ func (c *S3Client) PutObject(
...
+		len(object.StorageClass) != 0 {
```
SvartMetal: Better put the code below under the `if err != nil` https://github.com/ydb-platform/nbs/pull/6143/changes#diff-58c84c2229f1c9fdf2bb754a1fc096be73aa06ebd10fafab30735f8b2a2247edL256

cloud/tasks/persistence/s3.go
```diff
@@ -246,13 +256,34 @@ func (c *S3Client) PutObject(
...
+
+		logging.Warn(
+			ctx,
+			"s3 put object with storage class %q failed with quota "+
+				"error, retrying without storage class",
+			object.StorageClass,
+		)
+
+		input.Body = bytes.NewReader(object.Data)
+		input.StorageClass = nil
```
SvartMetal: Maybe create a new input object? Looks clearer

cloud/disk_manager/test/recipe/s3_launcher.py
```diff
@@ -24,28 +25,70 @@ def __init__(self, port, working_dir):
...
+            if quotas else self.__moto_port
+        )
 
         working_dir = get_unique_path_for_current_test(
             output_path=yatest_common.output_path(),
             sub_folder=""
         )
         ensure_path_exists(working_dir)
-        self.__daemon = S3Service(self.__port, working_dir)
+        self.__moto = S3Service(self.__moto_port, working_dir)
```
SvartMetal: self.__service = ...

cloud/disk_manager/test/recipe/s3_launcher.py
```diff
@@ -24,28 +25,70 @@ def __init__(self, port, working_dir):
             service_name=SERVICE_NAME)
 
 
+class S3QuotaProxyService(Daemon):
+    def __init__(self, port, s3_port, quotas, working_dir):
+        command = [
+            yatest_common.binary_path(
+                'cloud/disk_manager/test/mocks/s3_quota_proxy/s3quota-proxy'
```
SvartMetal: cloud/disk_manager/test/mocks/s3-quota-proxy

?

cloud/disk_manager/test/recipe/s3_launcher.py
```diff
@@ -9,6 +9,7 @@
 _logger = logging.getLogger(__file__)
 
 SERVICE_NAME = "s3"
+S3_QUOTA_PROXY_SERVICE_NAME = "s3-quota-proxy"
```
SvartMetal: QUOTA_PROXY_SERVICE_NAME

as SERVICE_NAME has no "S3" prefix

cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -669,6 +668,17 @@ def __init__(
...
+            if image_s3_default_storage_class:
+                image_s3_default_storage_class_config = (
+                    IMAGE_S3_DEFAULT_STORAGE_CLASS_CONFIG_TEMPLATE.format(
+                        storage_class=image_s3_default_storage_class,
+                    )
+                )
+
+            image_use_s3_percentage = "0"
+            if image_s3_default_storage_class:
+                image_use_s3_percentage = "100"
```
SvartMetal, line 669: Let's move it under `if image_s3_default_storage_class:` on 671 line?

cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -280,6 +282,10 @@
 >
 """
 
+IMAGE_S3_DEFAULT_STORAGE_CLASS_CONFIG_TEMPLATE = """\
```
SvartMetal: Do we need '\' after """?

cloud/disk_manager/test/recipe/disk_manager_launcher.py
```diff
@@ -669,6 +668,17 @@ def __init__(
                         cert_file=cert_file,
                     )
                 )
+            image_s3_default_storage_class_config = ""
+            if image_s3_default_storage_class:
+                image_s3_default_storage_class_config = (
+                    IMAGE_S3_DEFAULT_STORAGE_CLASS_CONFIG_TEMPLATE.format(
```
SvartMetal: Why do we need template here?

Maybe add a parameter `S3DefaultStorageClass: {image_s3_default_storage_class}`?

cloud/disk_manager/test/recipe/s3_launcher.py
```diff
@@ -24,28 +25,70 @@ def __init__(self, port, working_dir):
...
 class S3Launcher:
-    def __init__(self):
+    def __init__(self, quotas=None):
+        if quotas is None:
+            quotas = []
+
         self.__port_manager = yatest_common.PortManager()
-        self.__port = self.__port_manager.get_port()
+        self.__moto_port = self.__port_manager.get_port()
+        self.__port = (
```
SvartMetal, line 55: Let's add a comment that s3 quota proxy wraps s3 service and injects quota limit errors

cloud/disk_manager/test/mocks/s3-quota-proxy/main.go
```diff
@@ -0,0 +1,259 @@
...
+			if resp.StatusCode < http.StatusBadRequest {
+				return nil
+			}
+
+			p.releaseReservation(resp.Request.Context())
+			return nil
+		},
+		ErrorHandler: func(w http.ResponseWriter, r *http.Request, err error) {
+			p.releaseReservation(r.Context())
+			http.Error(w, err.Error(), http.StatusBadGateway)
```
SvartMetal, line 66: Looks clearer if we copy original error code instead of hardcoding StatusBadGateway here

cloud/disk_manager/test/mocks/s3_quota_proxy/main.go
```diff
@@ -0,0 +1,259 @@
...
+
+func main() {
+	log.Println("launching s3quota-proxy")
+
+	var port string
+	var s3Port string
+	var quotaSpecs []string
+
+	rootCmd := &cobra.Command{
+		Use:   "s3quota-proxy",
```
SvartMetal: s3-quota-proxy?

cloud/disk_manager/test/mocks/s3-quota-proxy/main.go
```diff
@@ -0,0 +1,259 @@
...
+	if len(storageClass) == 0 {
+		return nil, true
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	quota, ok := p.quota[storageClass]
+	if !ok {
+		return nil, true
```
SvartMetal: It is better return an error for unknown storage classes, treating default storage class as a well known one (with quota limiting disabled)

SvartMetal (changes requested):

cloud/disk_manager/test/recipe/disk_manager_launcher.py
jkuradobery: We would like to enable it conditionally. So we are sure we maintain the backwards compat. with empty config value.

cloud/disk_manager/test/mocks/s3-quota-proxy/main.go
jkuradobery: It is not called on http code, it is called on connection error or proxy missconfig.
See:
https://go.dev/src/net/http/httputil/reverseproxy.go
I think it is a common practice to return 502 in case of some upstream error (not http error code).

cloud/tasks/persistence/s3.go
```diff
@@ -246,13 +256,34 @@ func (c *S3Client) PutObject(
...
+	if aerr, ok := err.(awserr.Error); ok &&
+		aerr.Code() == s3ErrCodeQuotaLimitExceeded &&
+		len(object.StorageClass) != 0 {
+
+		logging.Warn(
+			ctx,
+			"s3 put object with storage class %q failed with quota "+
+				"error, retrying without storage class",
+			object.StorageClass,
+		)
```
SvartMetal: It is possible to get an error message flood if we hit this issue.

Let's add metrics instead, and validate in the test that the quota limit was reached

SvartMetal (changes requested):

cloud/tasks/persistence/s3.go
jkuradobery: S3 reports their own metrics which we can rely on (e.g. count of non 2xx errors.

SvartMetal: > S3 reports their own metrics which we can rely on (e.g. count of non 2xx errors

These errors should be seen in our monitoring also

cloud/disk_manager/test/recipe/disk_manager_launcher.py
SvartMetal: > So we are sure we maintain the backwards compat. with empty config value

Empty config value will be handled when image_s3_default_storage_class == ""

cloud/tasks/persistence/s3.go
SvartMetal: It is also needed to validate that quota error actually happened in test

cloud/disk_manager/test/recipe/disk_manager_launcher.py
jkuradobery: We do not want to rely on missing field being equal to "", although we use Get...() after the protobuf is parsed, the field is currently missing and will be added, so it is more realistic to test it as is. 

However, I do not think we should use this style of protobuf configs, we should use modular proto configs merger like we do in filestore. However now we use the same method (put string template into a separate variable). Using j2 here would be both overkill and unnecessary complexity.

cloud/tasks/persistence/s3.go
jkuradobery: Disagreed. This constant is specific to our S3 implementation and I have tested against it. It does not affect other S3 client users since they do not use storage classes.

cloud/tasks/persistence/s3.go
```diff
@@ -246,27 +258,101 @@ func (c *S3Client) PutObject(
...
+	if err == nil {
+		return nil
+	}
+
+	if !isQuotaLimitExceeded(err) {
+		return wrapError(err)
+	}
+
+	if len(object.StorageClass) == 0 {
+		return wrapError(err)
```
gy2411, line 271: I don't uderstand how this error could happen and what does it mean. Object has no StorageClass but we still get QuotaLimitExceeded error? Maybe we should add some extra error message here?

SvartMetal (approved):

cloud/disk_manager/test/recipe/__main__.py
```diff
@@ -70,6 +70,13 @@ def parse_args(args):
     parser.add_argument("--without-shadow-disks", action='store_true', default=False)
     parser.add_argument("--filesystem-dataplane-enabled", action='store_true', default=False)
     parser.add_argument("--list-nodes-max-bytes", type=int, default=0)
+    parser.add_argument(
+        "--image-s3-storage-class",
```
gy2411, line 74: Why not `--image-s3-default-storage-class`?

cloud/disk_manager/internal/pkg/dataplane/migrate_snapshot.go
```diff
@@ -82,6 +82,7 @@ func (t *migrateSnapshotTask) Run(
 		t.dstStorage,
 		true, // ignoreZeroChunks
 		t.useS3,
+		"", // storageClass
```
gy2411, line 85: Why don't we use StorageClass for this task?

cloud/tasks/persistence/s3.go
jkuradobery: >  Object has no StorageClass but we still get QuotaLimitExceeded error
This can happen in our storage when the quota limit is exceeded (for standard class). There is not point to fallback & retry,

cloud/disk_manager/test/recipe/__main__.py
jkuradobery: I will fix it separately

cloud/disk_manager/internal/pkg/dataplane/migrate_snapshot.go
jkuradobery: Migration is a manual operation. We do not consider using storage class for migration for the sake of simplicity.

gy2411 (approved):

PR #6150, opened 2026-06-08
[Blockstore] Force overlapping requests detection for remotely forwarded write requests

zavik001:
For write requests that arrived via pipe (IsForwardedEvent returns true Recipient != RecipientRewrite), the policy is forced to ORP_ENABLE regardless of the configured policy, because the message arrived from a remote disk and there is no guarantee of message ordering. This also means that if ORP_ENABLE_WITH_CRIT_EVENT is configured, the crit event will not be fired for forwarded requests.

### DIFF
--- a/cloud/blockstore/libs/storage/volume/volume_ut.cpp
+++ b/cloud/blockstore/libs/storage/volume/volume_ut.cpp
@@ -9239,7 +9239,7 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
         TVolumeClient volume(*runtime);
 
         bool gotVolumeActorId = false;
-        TActorId volumeActor;
+        TActorId volumeActorId;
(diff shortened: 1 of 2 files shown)


cloud/blockstore/libs/storage/volume/volume_ut.cpp
```diff
@@ -9283,13 +9370,11 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
             "AppCriticalEvents/OverlappingRequestsDetected",
             true);
 
-        // Send first request
```
drbasic, line 9286: зачем удалять комменты?

zavik001: > зачем удалять комменты?

Случайно удалил при рефакторинге, вернул обратно

PR #6145, opened 2026-06-05
issue-4961: Keep service session actors per subsession to align with tablet state

WilyTiger:
### Notes
The tablet tracks ownership per subsession: each `(sessionId, seqNo)` has its own owner actor. The service state used to keep only one `SessionActor` for the whole `TSessionInfo`, so creating or recovering one subsession could replace or kill the actor that still belonged to another active subsession.

This change stores `SessionActor` inside `TSubSessionInfo` and makes service-side validation and routing use the actor for the exact `(sessionId, seqNo)`. This keeps the service state consistent with the tablet state and prevents recovery of one subsession from invalidating other active subsessions.

detailed explanation of problem the https://github.com/ydb-platform/nbs/issues/4961#issuecomment-4606262675

### Issue
#4961

PR #6176, opened 2026-06-10
issue-4961: Track session owner generation for CreateSession for diagnostics

WilyTiger:
### Notes
Add tablet-side owner generation for subsessions and return it in CreateSession responses. The generation is initialized when a subsession owner is accepted and is incremented when the owner changes for the same session seqno.

The returned owner generation also includes the tablet generation in the high bits and the per-subsesson owner generation in the low bits. This makes the value monotonic across tablet restarts: even if the per-subsession counter starts from 1 again after a tablet reboot, the full generation still increases because the tablet generation has advanced.

Propagate this value through TCreateSessionActor to the storage service state. For now this does not change the existing actor replacement behavior: if the service receives a CreateSession notification from an actor different from the currently stored session actor, it still keeps the current actor and poisons the notification sender.

The generation is used only for diagnostics in this change. The service logs stale/newer CreateSession owner generations and actor mismatches, so we can observe whether delayed CreateSession notifications or owner ordering races happen before changing the service-side session model. 

This potential problem was found while implementing the initial plan from https://github.com/ydb-platform/nbs/issues/4961#issuecomment-4614804880 in PR https://github.com/ydb-platform/nbs/pull/6145#discussion_r3373425885.

Before changing the CreateSession actor replacement logic, we want to make sure that the current ordering assumptions are valid and that the change will not break migration/recovery scenarios. This PR adds per-subsession owner generation to CreateSession responses and propagates it to the service side, so we can detect stale or out-of-order CreateSession notifications and use this information later to keep tablet and service session state consistent.

### Issue
#4962

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_createsession.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_createsession.cpp
@@ -56,6 +56,7 @@ private:
     bool ReadOnly;
     TString SessionId;
     TString SessionState;
+    ui64 OwnerGeneration = 0;
     NProto::TFileStore FileStore;
 
     ui64 TabletId = -1;
@@ -428,6 +429,7 @@ void TCreateSessionActor::HandleCreateSessionResponse(
     }
 
     SessionState = msg->Record.GetSessionState();
+    OwnerGeneration = msg->Record.GetOwnerGeneration();
     FileStore = msg->Record.GetFileStore();
 
     // Some of the features of the filestore are set not by the tablet but also
@@ -588,6 +590,7 @@ void TCreateSessionActor::Notify(
     response->SessionSeqNo = SeqNo;
     response->ReadOnly = ReadOnly;
     response->TabletId = TabletId;
+    response->OwnerGeneration = OwnerGeneration;
     response->FileStore = FileStore;
     response->RequestInfo = std::move(RequestInfo);
     response->Shutdown = shutdown;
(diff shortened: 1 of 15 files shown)


WilyTiger:
@codex review pls and check typos

cloud/filestore/libs/storage/service/service_actor_createsession.cpp
```diff
@@ -846,14 +849,71 @@ void TStorageServiceActor::HandleSessionCreated(
     const auto* msg = ev->Get();
 
     auto* session = State->FindSession(msg->SessionId);
+    const TSessionInfo::TSubSessionInfo* subSession = nullptr;
+    if (session) {
+        subSession = session->FindSubSession(msg->SessionSeqNo);
+    }
+    const auto storedOwnerGeneration =
```
qkrorlqr: nit: no `auto` for primitive types plz

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -11,21 +11,47 @@ namespace {
 ////////////////////////////////////////////////////////////////////////////////
 
 constexpr size_t MaxSubSessions = 2;
+constexpr ui64 SubSessionOwnerGenerationBits = 32;
+constexpr ui64 SubSessionOwnerGenerationMask =
+    (ui64(1) << SubSessionOwnerGenerationBits) - 1;
```
qkrorlqr: nit: `ui64(1)` -> `1ULL`

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -11,21 +11,47 @@ namespace {
...
 
+}   // namespace
+
+////////////////////////////////////////////////////////////////////////////////
+
+ui64 MakeSubSessionOwnerGeneration(ui32 tabletGeneration, ui32 ownerGeneration)
+{
+    return (ui64(tabletGeneration) << SubSessionOwnerGenerationBits) |
```
qkrorlqr: 1ULL

cloud/filestore/libs/storage/tablet/subsessions.cpp
```diff
@@ -11,21 +11,47 @@ namespace {
...
     MaxSeenSeqNo = std::max(MaxSeenSeqNo, seqNo);
     if (!readOnly) {
         MaxSeenRwSeqNo = std::max(MaxSeenRwSeqNo, seqNo);
     }
-    SubSessions.push_back({seqNo, readOnly, owner});
+    SubSessions.push_back({
+        seqNo,
+        readOnly,
+        owner,
+        MakeSubSessionOwnerGeneration(tabletGeneration, 1)});
```
qkrorlqr: nit: `1 /* ownerGeneration */`

cloud/filestore/libs/storage/tablet/subsessions_ut.cpp
```diff
@@ -8,26 +8,46 @@ using namespace NActors;
 
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace {
+
```
qkrorlqr, line 10: nit: `///...` goes before decl/def, not before `namespace`

cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
```diff
@@ -296,6 +296,19 @@ void TIndexTabletActor::ExecuteTx_CreateSession(
                 ctx);
             args.SessionInterrupted = true;
             if (toKill != owner) {
+                const auto subSession =
```
qkrorlqr: nit: `auto*`

cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
```diff
@@ -341,6 +354,19 @@ void TIndexTabletActor::ExecuteTx_CreateSession(
                 readOnly,
                 owner,
                 ctx);
+            const auto subSession =
```
qkrorlqr: nit: `auto*`

cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
```diff
@@ -377,6 +403,19 @@ void TIndexTabletActor::ExecuteTx_CreateSession(
         readOnly,
         owner,
         sessionOptions);
+    const auto subSession =
```
qkrorlqr: nit: `auto*`

qkrorlqr (commented):
LGTM

cloud/filestore/libs/storage/service/service_state.h
```diff
@@ -364,10 +377,16 @@ struct TSessionInfo
         return SubSessions.count(seqNo) > 0;
     }
 
+    const TSubSessionInfo* FindSubSession(ui64 seqNo) const
+    {
+        auto it = SubSessions.find(seqNo);
```
agalibin, line 382: return SubSessions.FindPtr(seqNo);

agalibin (approved):

qkrorlqr (approved):

cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
```diff
@@ -377,6 +403,19 @@ void TIndexTabletActor::ExecuteTx_CreateSession(
...
+        args.OwnerGeneration = subSession->OwnerGeneration;
+    } else {
+        LOG_ERROR(ctx, TFileStoreComponents::TABLET,
+            "%s CreateSession c:%s, s:%s, seqno:%lu created new session, "
+            "but subsession was not found",
+            LogTag.c_str(),
+            clientId.c_str(),
+            newSession->GetSessionId().c_str(),
+            seqNo);
+    }
```
neihar, line 418: looks like this block identical with 2 blocks in changes above, probably we can somehow reuse the code?

WilyTiger: I would keep this code inline for now.
The repeated block is small and local, and extracting it into a helper makes the code longer overall and slightly harder to read, because the reader has to jump away from the CreateSession flow to understand a very simple action: read the accepted subsession and copy its OwnerGeneration into the transaction args.

The three call sites also have slightly different context in the error message (`recovered by session`, `recovered by client`, `created new session`). Keeping the code inline preserves that context directly at the point where each branch handles its own CreateSession result.

Issue #5722, opened 2026-04-13
Blockstore/Filestore/DiskManager require restart to update GRPC certificates


yegorskii:
Currently to update server certificates we need to restart out services which sometimes is not desirable. However GRPC provides interfaces allow to dynamically update server certificates

PR #6190, opened 2026-06-11
introduce certificate provider with autmatic certifcates refresh

yegorskii:
### Notes
Implemented TPeriodicCertiicateProvider class for dynamic TLS certificates update. New functionality implements GRPC distributor interface which allows to dynamically change TLS certificates for the channels. So by periodically re-reading TLS certificates we can update them "on-line" without restarting our services.

### Issue


cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
```diff
@@ -983,9 +984,24 @@ void TBootstrapYdb::SetupCellManager()
...
+            ythrow yexception()
+                << "Secure cells port is configured without certificates";
+        }
+
+        auto cellCertProvider = CreateCertificateProvider(
+            Logging,
+            GetComponentName(
+                TBlockStoreComponents::TLS_CERTIFICATE_PROVIDER),
+            Scheduler,
+            CreateLongRunningTaskExecutor("CertRefresh"),
```
qkrorlqr, line 998: do we need multiple such executors? why not just one?

I mean, why not just use BackgroundScheduler and schedule cert updates as asks in it?

cloud/filestore/libs/daemon/server/bootstrap.cpp
```diff
@@ -139,12 +140,22 @@ void TBootstrapServer::InitComponents()
...
-    if (!certPathList.empty()) {
-        CertificateProvider = CreateStaticCertificateProvider(
-            Configs->ServerConfig->GetRootCertsFile(),
-            std::move(certPathList));
+    if (Configs->ServerConfig->GetSecurePort() && certPathList.empty()) {
+        ythrow yexception()
+            << "Secure port is configured without certificates";
     }
 
+    CertificateProvider = CreateCertificateProvider(
```
qkrorlqr: let's integrate this into filestore in a separate PR

cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
yegorskii: Unsing BackgroundScheduler seems to bad idea, because under the hood it uses single thread and if certificates reading stucks then we will loose the whole background scheduler. LongRunningTaskExecutor seems to be more appropriate solution.

cloud/filestore/libs/daemon/server/bootstrap.cpp
yegorskii: ok........

cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
qkrorlqr: ok

PR #5308, opened 2026-02-24
issue-5293: fixed caching of stale GetNodeAttr responses in case of concurrent WriteData/SetNodeAttr/AllocateData requests

qkrorlqr:
### Notes
* introducing a per-fs atomic `GlobalAttrVersion` counter - incrementing it every time upon attr-modifying operation completion
* invalidating node attr in `TNodeCache` upon each attr-modifying operation completion and setting the just-incremented `GlobalAttrVersion` as this node's version
* loading current node attr version from `TNodeCache` upon attr-reading operation start
* resetting attr_timeout in the attrs returned to the guest if upon attr-reading operation completion we find out that the version of the node in `TNodeCache` has changed since the start of the operation 

extra:
* moved `TDirectoryBuilder` to a separate .h/.cpp file pair and added unit tests for it
* moved `NodeCacheLock` into `TNodeCache` so `TNodeCache` is now thread-safe
* `TNodeCache` benchmark
* implemented `TNodeCache` sharding by node id to reduce contention (the results can be checked via the added benchmark)

```
// 16 client threads, 1 shard (equivalent to non-sharded version)
----------- UpdateNode16 ---------------
 samples:       4364
 iterations:    11758825
 iterations hr:    11.8M
 run time:      5.000951105
 per iteration: 1235.532165 cycles

// 16 client threads, 16 shards
----------- UpdateNode16_16 ---------------
 samples:       10080
 iterations:    62736801
 iterations hr:    62.7M
 run time:      5.001321319
 per iteration: 156.7914719 cycles
```

### Issue
https://github.com/ydb-platform/nbs/issues/5293

PR #6203, opened 2026-06-12
[Filestore] fix unlink vs readdir data race (#5679)

tindarid:
### Notes
* Fix unlink vs readdir data race in a similar way to https://github.com/ydb-platform/nbs/pull/5308
* Reuse the same global counter for both purposes: entry version and node version
* Logic is the same: we invalidate an entry (child) in the `readdir` response (for the parent), if captured version is less than version during start of `readdir` call

#### Tests
* Add another variant to `fmdtest` (more likely to reproduce the subject) and convert warning to error
* Add unit and integration tests

### Bench results
| Name | Samples | Iterations | Per_iteration_sec |
|---|---:|---:|---:|
| UpdateNode1 | 5043 | 15705210 | 381.3295052 cycles |
| UpdateNode2 | 3215 | 6384951 | 974.4663193 cycles |
| UpdateNode4 | 2773 | 4750903 | 1324.970686 cycles |
| UpdateNode8 | 2934 | 5315430 | 1168.003143 cycles |
| UpdateNode16 | 2779 | 4769416 | 1284.114966 cycles |
| UpdateNode1_16 | 4726 | 13794378 | 426.6229458 cycles |
| UpdateNode2_16 | 4663 | 13429153 | 423.2701204 cycles |
| UpdateNode4_16 | 5558 | 19074576 | 268.2531732 cycles |
| UpdateNode8_16 | 6246 | 24085270 | 162.4875869 cycles |
| UpdateNode16_16 | 5715 | 20170776 | 136.5057092 cycles |
| AdvanceEntryVersion1 | 4129 | 10527166 | 424.1007809 cycles |
| AdvanceEntryVersion2 | 3660 | 8272278 | 595.338656 cycles |
| AdvanceEntryVersion4 | 3085 | 5877306 | 894.2590632 cycles |
| AdvanceEntryVersion8 | 3187 | 6274653 | 803.7219196 cycles |
| AdvanceEntryVersion16 | 2956 | 5397255 | 1003.919736 cycles |
| AdvanceEntryVersion1_16 | 4002 | 9890128 | 437.1417586 cycles |
| AdvanceEntryVersion2_16 | 4097 | 10367181 | 457.0654566 cycles |
| AdvanceEntryVersion4_16 | 4984 | 15337491 | 274.0974006 cycles |
| AdvanceEntryVersion8_16 | 5482 | 18559278 | 163.1958863 cycles |
| AdvanceEntryVersion16_16 | 5157 | 16425046 | 130.8882172 cycles |

### Issue
https://github.com/ydb-platform/nbs/issues/5679


cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -151,6 +151,17 @@ void TDirectoryHandleCache::AppendChunk(
     }
 }
 
+void TDirectoryHandleCache::InvalidateEntries(fuse_ino_t directoryIno, ui64 version)
+{
+    with_lock (Lock) {
+        for (const auto& [_, handle]: Handles) {
```
tindarid: here I am not sure, what is the usual size of this cache (len(handles))

if it is ~100, I think it is fine to loop over and invalidate

if it is > 10k, I think we need to have a mapping `ino -> {handle}`

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
```diff
@@ -129,6 +129,7 @@ void TFileSystem::MkDir(
             self->FSyncQueue->Dequeue(reqId, error, TNodeId {parent});
 
             if (CheckResponse(self, *callContext, req, response)) {
+                self->InvalidateDirectoryEntriesInCache(parent);
```
neihar: add test for mkdir

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
neihar: With current clients we sometimes have 150k open directory handles, so per-ino hashmap would of course help here. But more fundamentally, this version is property of the node, not of the handle, so tracking it per handle just doesn't look right. Better to track per node and reset it same way as NodeCache does for attrs, probably reusing NodeCache itself.
And handle->Index already gives us the node anyway, so the ReadDir reply becomes one O(1) sharded lookup instead of this O(n) scan under the global lock.

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
```diff
@@ -220,6 +226,22 @@ TDirectoryHandleStats TDirectoryHandle::GetStats() const
...
 }
 
+ui64 TDirectoryHandle::GetEntryVersion() const
+{
+    with_lock (Lock) {
+        return EntryVersion;
+    }
+}
+
+void TDirectoryHandle::InvalidateEntries(ui64 version)
```
neihar: AdvanceEntryVersion?

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -183,15 +188,42 @@ bool TFileSystem::UpdateNodeAttrsInCache(
...
+
+    STORAGE_TRACE("invalidating directory entries: " << nodeId
+        << ", version: " << version);
+
+    DirectoryHandleCache->InvalidateEntries(nodeId, version);
+}
+
+void TFileSystem::InvalidateDirectoryEntriesInCache(
+    fuse_ino_t nodeId,
+    fuse_ino_t anotherNodeId)
```
neihar: can we make better name for this var?

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -93,13 +93,18 @@ void TFileSystem::ReadDir(
...
-        auto error = ResetAttrTimeout(c.Data(), c.Size(), [&] (ui64 ino) {
-            return NodeCache.GetNodeVersion(ino) > content.AttrVersion;
-        });
+        auto error = ResetCacheTimeouts(
+            c.Data(),
+            c.Size(),
+            [&] (ui64 ino) {
+                return NodeCache.GetNodeVersion(ino) > content.CacheVersion;
+            },
+            handle->GetEntryVersion() > content.CacheVersion);
```
neihar: version belongs to node entity, it doesn't look right to track it in handle

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
```diff
@@ -318,6 +322,7 @@ void TFileSystem::Rename(
             self->FSyncQueue->Dequeue(reqId, error, TNodeId {parent});
 
             if (CheckResponse(self, *callContext, req, response)) {
+                self->InvalidateDirectoryEntriesInCache(parent, newparent);
```
neihar: test?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -5388,6 +5405,64 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
 
         UNIT_ASSERT(stopFuture.Wait(WaitTimeout));
     }
+
+    Y_UNIT_TEST(ShouldNotAllowGuestEntryCachingUponListNodesAndUnlinkRace)
```
neihar: add tests for other operations that were touched

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
```diff
@@ -170,6 +171,7 @@ void TFileSystem::RmDir(
             self->FSyncQueue->Dequeue(reqId, error, TNodeId {parent});
 
             if (CheckResponse(self, *callContext, req, response)) {
+                self->InvalidateDirectoryEntriesInCache(parent);
```
neihar: should we also invalidate parent attrs? https://github.com/ydb-platform/nbs/pull/6203/changes#r3404572762
rewiring tracking to node cache can help to solve this problem, thought probably stale attrs for dir not so important

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
neihar: do we need it at all for mkdir?

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
```diff
@@ -359,6 +364,7 @@ void TFileSystem::SymLink(
...
```
neihar: do we need it?

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
```diff
@@ -403,6 +409,7 @@ void TFileSystem::Link(
             self->FSyncQueue->Dequeue(reqId, error, TNodeId {ino});
 
             if (auto self = ptr.lock(); CheckResponse(self, *callContext, req, response)) {
+                self->InvalidateDirectoryEntriesInCache(newparent);
```
neihar: do we really need it?

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
```diff
@@ -233,6 +235,7 @@ void TFileSystem::MkNode(
             self->FSyncQueue->Dequeue(reqId, error, TNodeId {parent});
 
             if (CheckResponse(self, *callContext, req, response)) {
+                self->InvalidateDirectoryEntriesInCache(parent);
```
neihar: do we really need it?

cloud/filestore/libs/vfs_fuse/fs_directory_handle.h
```diff
@@ -71,12 +71,13 @@ class TDirectoryHandle
...
     {
         TBufferPtr Buffer;
-        ui64 AttrVersion = 0;
+        ui64 CacheVersion = 0;
     };
 
     TString Cookie;
     TMap<ui64, TContent> Content;
     ui64 UpdateVersion = 0;
+    ui64 EntryVersion = 1;
```
neihar: probably worth comment why 1 and describe whole logic in comment

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
neihar: with suggested implementation in other comment, we need handle recover case for cache version carefully here

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -170,6 +170,9 @@ void TFileSystem::Create(
             self->FSyncQueue->Dequeue(reqId, error, TNodeId {parent});
 
             if (CheckResponse(self, *callContext, req, response)) {
+                if (HasFlag(flags, NProto::TCreateHandleRequest::E_CREATE)) {
+                    self->InvalidateDirectoryEntriesInCache(parent);
```
neihar: do we need it?

cloud/filestore/libs/vfs_fuse/fs_impl.h
```diff
@@ -419,6 +419,11 @@ class TFileSystem final
         ui64 version);
 
     void InvalidateNodeInCache(ui64 nodeId);
+    void InvalidateDirectoryEntriesInCache(fuse_ino_t nodeId);
+    // atomic for rename usecase
```
neihar: what is exactly atomic here?  and do we really need it? looks misleading to me.
Also, I don't see a problem to have 2 different versions for src and dst, but we can keep one if we really want it and just fix comment

neihar (changes requested):

cloud/filestore/libs/vfs_fuse/fs_impl.h
tindarid: N/A

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
tindarid: removed, creations do not violate the behaviour

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
tindarid: removed, creations do not violate the behaviour

cloud/filestore/libs/vfs_fuse/fs_directory_handle.h
tindarid: removed

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
tindarid: removed, creations do not violate the behaviour

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
tindarid: removed, creations do not violate the behaviour

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
tindarid: removed

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
tindarid: changed

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
tindarid: removed, creations do not violate the behaviour

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: Changed logic as discussed offline

* introduced TDirectoryEntryVersionCache to track {directoryIno -> {childName -> version}}
* the worst case in terms of memory usage is when somebody keeps directory opened and actively unlinks/renames. however, it should be a very weird pattern with long living handles

tindarid:
TBD:
* Lookup shouldn't return stale data
* Investigate serializaiton of cache version
* More tests

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        auto& state = it->second;
+        if (--state.RefCount == 0) {
+            Directories.erase(it);
+        }
+    }
+}
+
+void TDirectoryEntryVersionCache::ChangeVersion(
```
neihar: can it be lower version than before or it can only go up? AdvanceVersion?

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -93,13 +93,22 @@ void TFileSystem::ReadDir(
...
-        });
+        auto error = ResetCacheTimeouts(
+            c.Data(),
+            c.Size(),
+            [&] (ui64 nodeId) {
+                return fs.NodeCache.GetNodeVersion(nodeId) >
+                       content.CacheVersion;
+            },
+            [&] (ui64, TStringBuf name) {
+                return fs.DirectoryEntryVersionCache.GetVersion(ino, name) >
```
neihar: for every GetVersion call we will lock common Mutex across all inodes, we can do sharding and bench as for nodecache

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -93,13 +93,22 @@ void TFileSystem::ReadDir(
...
-            return NodeCache.GetNodeVersion(ino) > content.AttrVersion;
-        });
+        auto error = ResetCacheTimeouts(
+            c.Data(),
+            c.Size(),
+            [&] (ui64 nodeId) {
+                return fs.NodeCache.GetNodeVersion(nodeId) >
+                       content.CacheVersion;
+            },
+            [&] (ui64, TStringBuf name) {
```
neihar: do we need this ui64?

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
tindarid: Good catch, I think we need to invalidate attributes on unlink for example: for both parent and child. Also for other cases ...

I think this PR can fully cover ENTRY semantics and we can fully tackle ATTRIBUTE semantics in a different task, wdyt?

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
```diff
@@ -129,18 +129,24 @@ TDirectoryHandleChunk TDirectoryHandle::UpdateContent(
...
     size_t end = offset + content->size();
     TDirectoryHandleChunk chunk{
         .Key = end,
         .Index = Index,
-        .DirectoryContent = {content, 0, size}};
+        .DirectoryContent = {
+            .Content = content,
+            .Offset = 0,
+            .Size = size,
+            .CacheVersion = cacheVersion,
```
tindarid, line 143: We were not setting the version for the fresh chunks before, seems like our behavior improved here, but please take a look

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
tindarid: done

cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp
tindarid: done

tindarid:
Current status:
1. I think we need to investigate how we handle attribute changes in case of unlink for example
* Also, double check why we are setting version of the newly created nodes to 1 (instead of GlobalCacheVersion) https://github.com/ydb-platform/nbs/blob/08013a8ec741a874d15908e7489b9cdd89b89b69/cloud/filestore/libs/vfs_fuse/fs_impl_node.cpp#L138
* This can be a separate task
2. I think we need to extend this PR to handle Lookup/Mkdir/etc methods (not only readdir)
* This can be a continuation PR
* We will extend new cache with in flight requests too

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,62 @@
...
+            oldVersion = version;
+        }
+    }
+}
+
+ui64 TDirectoryEntryVersionCache::GetVersion(
+    fuse_ino_t directory,
+    TStringBuf name) const
+{
+    with_lock (Lock) {
```
neihar, line 72: one lock for all ino can be a problem, we can shard it as for nodecache

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -29,15 +29,23 @@ TDirectoryHandleStats SumDirectoryHandlesStats(
...
+    , DirectoryEntryVersionCache(directoryEntryVersionCache)
     , Stats(std::move(stats))
 {
     if (Storage) {
         Storage->LoadHandles(Handles);
     }
 
+    if (DirectoryEntryVersionCache) {
+        for (const auto& [_, handle]: Handles) {
+            DirectoryEntryVersionCache->RegisterHandle(handle->Index);
```
neihar: we do not persist  (ino, name)->version across restarts. After restart I do not think that we need to refcount directory caches, we need to reset timeout for all of such dir entries

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
neihar, line 145: during reset, we drop all data, probably we can unsubscribe and subscribe again and refresh version

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
```diff
@@ -30,7 +32,8 @@ class TDirectoryHandleCache
     TDirectoryHandleCache(
         TLog log,
         TDirectoryHandleModuleStatsPtr stats,
-        TDirectoryHandleStoragePtr storage);
+        TDirectoryHandleStoragePtr storage,
+        TDirectoryEntryVersionCache* directoryEntryVersionCache = nullptr);
```
neihar: we try to avoid default params and better to pass explicit in all places

cloud/filestore/libs/vfs_fuse/fs_directory_handle.cpp
neihar: yes, this is correct and needed change

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -93,13 +93,22 @@ void TFileSystem::ReadDir(
...
+                return fs.DirectoryEntryVersionCache.GetVersion(ino, name) >
+                       content.CacheVersion;
```
neihar, line 106: after restart DirectoryEntryVersionCache is empty, so we need to be sure that we reset for restored caches all entires timeouts. In currect code, can we have 0 > 0 treating like all is correct, but in reality we need to invalidate all data if content version is 0

If node cache is empty, looks like we have the same problem for attrs as well.

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,62 @@
+#include "directory_entry_version_cache.h"
+
+namespace NCloud::NFileStore::NFuse {
+
+void TDirectoryEntryVersionCache::RegisterHandle(fuse_ino_t directory)
```
neihar: looks like in current implementation we do not take into account, that we can have
open dir 1
unlink node A
open dir 2
close dir 1
--- dir 2 will hold this ref, when in reality we don't need it---
so, theoretically we can prolong cache and grow in to unlimited size

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
neihar: we do not track version for subscription, wrote about it in other place, in current implementation it doesn't make sense.

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: This applies, only if dir 1 == dir 2, if I read this correctly (same dir)

I think it is fine, otherwise we need to segment this inside directory: dir 2 should have a different subset of names, but any new name should be added to all subsets (dir 1, dir 2) or in the other direction every name should know to which prefix of dir handles it belongs

neihar: I mean, you can have N different handles on the same directory, all of them will have the same inode to register. so, we can prolong inode cache by openening new handle before old one closes

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: no, changed, thx

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
tindarid: good catch, removed

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
tindarid: changed

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: removed

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
tindarid: added comments and also check for cache version == 0

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: Yes, I got it

On the other hand, indeed if we track subscriber versions, when subscriber unsubscribes, we can go through the map and invalidate all names with version < min(subscribers_versions)

I think this complicates current implementation and can be solved later (if it is a real issue)

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: added sharding

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
tindarid: added sharding

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -29,9 +29,11 @@ TDirectoryHandleStats SumDirectoryHandlesStats(
 TDirectoryHandleCache::TDirectoryHandleCache(
     TLog log,
     TDirectoryHandleModuleStatsPtr stats,
-    TDirectoryHandleStoragePtr storage)
+    TDirectoryHandleStoragePtr storage,
+    TDirectoryEntryVersionCache* directoryEntryVersionCache)
```
neihar: can we avoid raw pointers?

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
```diff
@@ -19,6 +20,7 @@ class TDirectoryHandleCache
     TAdaptiveLock Lock;
     TDirectoryHandleMap Handles;
     TDirectoryHandleStoragePtr Storage;
+    TDirectoryEntryVersionCache* DirectoryEntryVersionCache = nullptr;
```
neihar: don't use raw pointer

cloud/filestore/libs/vfs_fuse/fs_directory_handle.h
```diff
@@ -25,7 +25,9 @@ struct TDirectoryContent
     TBufferPtr Content = nullptr;
     size_t Offset = 0;
     size_t Size = 0;
-    ui64 AttrVersion = 0;
+    // GlobalCacheVersion starts from 1
+    // so this starting value is used to identify values loaded from disk
```
neihar: not from disk, "values recovered after restart from persistent storage". for regular vhost use tmpfs

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
neihar: Agree in general. Maybe just add some metric counter for amount of entires in that cash, just to see fluctuations in real workload. Even if it happens somewhere at least we will know in fact why

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
neihar: separate test for that case will be good, also to check if we have the same problem for attr

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: will add proposed implementation and this note in the final conclusion

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
tindarid: I've implemented a test which validates that CacheVersion loaded from persistent storage == 0. It is not checking full e2e flow (as we need to pass different lambdas to ResetCacheVersions, so I think this doesn't add a lot of value)

For attrs - I will add a note to final conclusion

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: fixed (shared_ptr)

cloud/filestore/libs/vfs_fuse/directory_handle_cache.h
tindarid: fixed (shared_ptr)

cloud/filestore/libs/vfs_fuse/fs_directory_handle.h
tindarid: changed, ty

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: Added metrics

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -19,27 +21,41 @@ void TDirectoryEntryVersionCacheShard::UnregisterHandle(fuse_ino_t directory)
...
+        auto& childVersions = it->second.ChildVersions;
+        inserted = !childVersions.contains(name);
+
+        auto& oldVersion = childVersions[name];
         if (oldVersion < version) {
             oldVersion = version;
         }
     }
+
+    if (inserted && Stats) {
```
neihar: we do not pass nullptr for stats, we use Stub with no-op logic instead, you can check in ctor that it is not nullptr Y_ABORT_UNLESS

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache_ut.cpp
```diff
@@ -10,7 +10,7 @@ Y_UNIT_TEST_SUITE(TDirectoryEntryVersionCacheTest)
 {
     Y_UNIT_TEST(ShouldNotRememberChangesWithoutRegisteredHandle)
     {
-        TDirectoryEntryVersionCache cache;
+        TDirectoryEntryVersionCache cache(1, nullptr);
```
neihar: pass Stub version of stats

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -19,27 +21,41 @@ void TDirectoryEntryVersionCacheShard::UnregisterHandle(fuse_ino_t directory)
...
         auto it = Directories.find(directory);
         if (it == Directories.end()) {
             return;
         }
 
-        auto& oldVersion = it->second.ChildVersions[name];
+        auto& childVersions = it->second.ChildVersions;
+        inserted = !childVersions.contains(name);
+
+        auto& oldVersion = childVersions[name];
```
neihar: nit: do we have insert_or_assign here?

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -19,27 +21,41 @@ void TDirectoryEntryVersionCacheShard::UnregisterHandle(fuse_ino_t directory)
 
         auto& state = it->second;
         if (--state.RefCount == 0) {
+            removedCount = static_cast<i64>(state.ChildVersions.size());
             Directories.erase(it);
         }
     }
+
+    if (removedCount != 0 && Stats) {
+        Stats->ChangeEntryVersionCacheEntryCount(-removedCount);
```
neihar: probably we can add it on dctor as well, for the sake of consistency

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache_ut.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: replaced with `try_emplace`

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: added + test

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -29,9 +29,11 @@ TDirectoryHandleStats SumDirectoryHandlesStats(
 TDirectoryHandleCache::TDirectoryHandleCache(
     TLog log,
     TDirectoryHandleModuleStatsPtr stats,
-    TDirectoryHandleStoragePtr storage)
+    TDirectoryHandleStoragePtr storage,
+    std::shared_ptr<TDirectoryEntryVersionCache> directoryEntryVersionCache)
```
neihar: nit: Better to make namePtr aliases as for other things

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
neihar, line 47: I was thinking, that we still need to register here handles. It is not actual for all records that were loaded, but we can append new chunks with newer versions, and for them it can be actual to track invalid names

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: done

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: done

cloud/filestore/libs/vhost/request.h
```diff
@@ -539,4 +539,56 @@ struct TSetAttrRequest
...
+    TRmDirRequest(const TString& name, ui64 parentNodeId)
+    {
+        In = TIn::Create(name.size() + 1);
+        In->Header.opcode = FUSE_RMDIR;
+        In->Header.nodeid = parentNodeId;
+        strcpy((char*)In->Data(), name.c_str());
+    }
+};
+
+
```
neihar: excess line

cloud/filestore/libs/vhost/request.h
```diff
@@ -539,4 +539,56 @@ struct TSetAttrRequest
...
+    TUnlinkRequest(const TString& name, ui64 parentNodeId)
+    {
+        In = TIn::Create(name.size() + 1);
+        In->Header.opcode = FUSE_UNLINK;
+        In->Header.nodeid = parentNodeId;
+        strcpy((char*)In->Data(), name.c_str());
+    }
+};
+
+
```
neihar: excess line

cloud/filestore/libs/vhost/request.h
```diff
@@ -539,4 +539,56 @@ struct TSetAttrRequest
...
+struct TRmDirRequest
+    : public TRequestBase<void, void, void>
+{
+    TRmDirRequest(const TString& name, ui64 parentNodeId)
+    {
+        In = TIn::Create(name.size() + 1);
+        In->Header.opcode = FUSE_RMDIR;
+        In->Header.nodeid = parentNodeId;
+        strcpy((char*)In->Data(), name.c_str());
```
neihar: why don't use staitic cast instead of C cast as below?

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace NCloud::NFileStore::NFuse {
+
+// Tracks per-name changes while a directory handle is open.
+// ReadDir uses these versions to avoid returning cacheable entries when
+// a listing races with unlink or rename of the same child.
+//
+// DirectoryHandleCache registers handles here, so when we don't have
+// handles pointing to directory, we erase all information regarding
+// entry changes to avoid unbound memory usage.
```
neihar: unbound -> unbounded

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,91 @@
...
+        auto& childVersions = it->second.ChildVersions;
+        auto [versionIt, newEntry] = childVersions.try_emplace(name, version);
+        inserted = newEntry;
+
+        if (versionIt->second < version) {
+            versionIt->second = version;
+        }
+    }
+
+    if (inserted) {
```
neihar: we need to do it under lock, otherwise it can be race to -1

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+        auto& state = it->second;
+        if (--state.RefCount == 0) {
+            removedCount = static_cast<i64>(state.ChildVersions.size());
+            Directories.erase(it);
+        }
+    }
+
+    if (removedCount != 0) {
+        Stats->ChangeEntryVersionCacheEntryCount(-removedCount);
```
neihar: move under lock, atomic will help with consistency across shards, but will not help for unregister and advance race

neihar:
@codex review

cloud/filestore/libs/vhost/request.h
tindarid: fixed

cloud/filestore/libs/vhost/request.h
tindarid: fixed

cloud/filestore/libs/vhost/request.h
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.h
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: fixed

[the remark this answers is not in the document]
tindarid: fixed

cloud/filestore/tests/directory_handles/test.py
```diff
@@ -63,7 +64,7 @@ def create_file(ssh: SshToGuest, dir: str, file_name: str):
     return ssh(f"sudo touch {dir}/{file_name}")
 
 
-@retry(stop_max_attempt_number=RETRY_COUNT, wait_fixed=WAIT_TIMEOUT_MS)
+@retry(stop_max_attempt_number=WAIT_FILE_RETRY_COUNT, wait_fixed=WAIT_TIMEOUT_MS)
```
neihar, line 71: If time allows, I would like to see bench(how resetting affects) for that as well.

neihar (dismissed):

cloud/filestore/tests/directory_handles/test.py
neihar: probably mutex for every get is too slow

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,83 @@
...
+        auto& state = it->second;
+        if (--state.RefCount == 0) {
+            i64 removedCount = static_cast<i64>(state.ChildVersions.size());
+            Directories.erase(it);
+            Stats->ChangeEntryVersionCacheEntryCount(-removedCount);
+        }
+    }
+}
+
+void TDirectoryEntryVersionCacheShard::AdvanceVersion(
```
qkrorlqr, line 44: nit: mb AdvanceVersion -> UpdateVersion?

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -80,8 +91,12 @@ void TDirectoryHandleCache::RemoveHandle(ui64 handleId)
     with_lock (Lock) {
         auto it = Handles.find(handleId);
         if (it != Handles.end()) {
+            const auto ino = it->second->Index;
```
qkrorlqr: nit: we try not to use `auto` for primitive types

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
```diff
@@ -97,10 +112,14 @@ bool TDirectoryHandleCache::RemoveHandle(ui64 handleId, fuse_ino_t ino)
     with_lock (Lock) {
         auto it = Handles.find(handleId);
         if (it != Handles.end()) {
-            isConsistent = it->second->Index == ino;
+            const auto handleIno = it->second->Index;
```
qkrorlqr: nit: we try not to use `auto` for primitive types

cloud/filestore/libs/vfs_fuse/directory_handle_cache_ut.cpp
```diff
@@ -92,7 +92,8 @@ struct TDirectoryHandleCacheTestFixture: public NUnitTest::TBaseFixture
         return TDirectoryHandleCache(
             Log,
             stats,
-            CreateStorage(stats, persistentHandleMaxSize));
+            CreateStorage(stats, persistentHandleMaxSize),
+            nullptr);
```
qkrorlqr: nit: `/* paramName */` plz

cloud/filestore/libs/vfs_fuse/directory_handle_stats.cpp
```diff
@@ -94,4 +105,9 @@ TDirectoryHandleModuleStatsPtr CreateDirectoryHandleStats(
         std::move(storageStats));
 }
 
+TDirectoryHandleModuleStatsPtr CreateDirectoryHandleStatsStub()
+{
+    return CreateDirectoryHandleStats(CreateWallClockTimer(), nullptr);
```
qkrorlqr: nit: `nullptr /* paramName */` plz

cloud/filestore/libs/vfs_fuse/fs_directory_content_format_ut.cpp
```diff
@@ -190,12 +193,15 @@ Y_UNIT_TEST_SUITE(TDirectoryContentFormatTest)
...
 
-        auto error = ResetAttrTimeout(
+        auto error = ResetCacheTimeouts(
             buffer->Data(),
             buffer->Size(),
             [] (ui64 ino) {
                 Y_UNUSED(ino);
                 return true;
+            },
+            [] (TStringBuf) {
```
qkrorlqr: nit: let's format this cb (in all `ResetCacheTimeouts()` calls) similarly to the other cbs in this file - `TStringBuf` -> `TStringBuf name` + `Y_UNUSED(name)`

cloud/filestore/libs/vfs_fuse/fs_directory_content_format_ut.cpp
qkrorlqr, line 306: let's test that our direntry visitor is not called in this case?

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
```diff
@@ -93,13 +93,23 @@ void TFileSystem::ReadDir(
...
-        });
+        auto error = ResetCacheTimeouts(
+            c.Data(),
+            c.Size(),
+            [&] (ui64 nodeId) {
+                return fs.NodeCache.GetNodeVersion(nodeId) >
+                       content.CacheVersion;
+            },
+            [&] (TStringBuf name) {
+                return content.CacheVersion == 0 || // loaded from persistent storage
```
qkrorlqr, line 104: why do we need this check in the dentry visitor whereas we don't need it in the attr visitor?

at a first glance it seems that we don't need it because if any update happens, then `fs.DirectoryEntryVersionCache->GetVersion(ino, name)` will be nonzero and then `fs.DirectoryEntryVersionCache->GetVersion(ino, name) > content.CacheVersion` will be true for all dentries loaded upon filestore-vhost startup

neihar: after restart we lost all info about entry cache/node cache, but we recover dir cache. In this case for many ino, name or just ino (for attr) we will get 0 and 0 > 0 equation. Do I miss something?

qkrorlqr: yes, you are right, we can still have pre-unlink/rename dir handle state loaded after restart but we won't know that unlink/rename happened

but then it seems that we need the same check in the attr visitor, right?

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: I think AdvanceVersion better reflects the behaviour

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_handle_cache.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_handle_cache_ut.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/directory_handle_stats.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/fs_directory_content_format_ut.cpp
tindarid: fixed

cloud/filestore/libs/vfs_fuse/fs_directory_content_format_ut.cpp
tindarid: made the additional check, adde comment

cloud/filestore/libs/vfs_fuse/fs_impl_list.cpp
tindarid: Yes, we should revisit how we handle attributes and in this case I think we need to follow the same logic (cacheversion == 0 => invalidate)

But I think this can be fixed in a different PR

cloud/filestore/tests/directory_handles/test.py
tindarid: I've run 2 different benchmarks:

1. Just benching readdir call -> effect of querying EntryVersionCache is minimal (around 2% change)
2. Benching e2e (with FUSE-driver, qemu, etc): reading directory after restart, while invalidating all entries (current logic), leads to a lot of manipulations with FUSE cache and can slow down the read by x5. On my machine reading directory (backend call) is x8 slower then reading from DirectoryHandleCache, so the performance stays within boundaries [cache, backend]

neihar (dismissed):

cloud/filestore/tests/directory_handles/test.py
neihar: Okay, ok, let's see how it goes in real clusters and tests. If it will provide problems, we can consider preserving EntryCache across restarts  for example

qkrorlqr (dismissed):

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
```diff
@@ -0,0 +1,83 @@
+#include "directory_entry_version_cache.h"
+
+namespace NCloud::NFileStore::NFuse {
```
SvartMetal, line 3: We usually place ////... here

See for example https://github.com/ydb-platform/nbs/blob/a6ab0f2daaf658b1977caa077c2b6ac3f77a41c7/cloud/filestore/libs/vfs_fuse/config.h#L13

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.h
```diff
@@ -0,0 +1,98 @@
...
+#include <util/generic/hash.h>
+#include <util/generic/strbuf.h>
+#include <util/generic/string.h>
+#include <util/generic/vector.h>
+#include <util/system/spinlock.h>
+#include <util/system/yassert.h>
+
+#include <utility>
+
+namespace NCloud::NFileStore::NFuse {
```
SvartMetal, line 17: ////... also here

yegorskii (dismissed):

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.cpp
tindarid: fixed + rebase

cloud/filestore/libs/vfs_fuse/directory_entry_version_cache.h
tindarid: fixed + rebase

neihar (dismissed):

yegorskii (dismissed):

neihar (approved):

yegorskii (approved):

PR #6207, opened 2026-06-12
fix destruction race in tchunk

yegorskii:
### Notes
There is race in TChunk destruction. First we deallocate memory and then unmap it from rdma. However there a chance the between freen and unmapping someone alse allocate and map this memory.

### Issue

### DIFF
--- a/cloud/storage/core/libs/rdma/impl/buffer.cpp
+++ b/cloud/storage/core/libs/rdma/impl/buffer.cpp
@@ -54,6 +54,7 @@ public:
 
     ~TChunk()
     {
+        MemoryRegion.reset();
         free(Address);
     }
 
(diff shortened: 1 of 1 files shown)


budevg (approved):

cloud/storage/core/libs/rdma/impl/buffer.cpp
```diff
@@ -240,8 +241,9 @@ class TBufferPool::TImpl
                 addr.get(),
                 chunkSize,
                 Flags);
-            Y_UNUSED(addr.release());
-            return new TChunk(std::move(mr), custom);
+            auto* chunk = new TChunk(std::move(mr), custom);
```
tpashkin, line 244: but why?

tpashkin (approved):

komarevtsev-d (approved):

PR #6219, opened 2026-06-15
[Blockstore] Replace SglistOwner with OwnsSglist flag in TWriteBlocksLocalRequest

zavik001:
This is a rework of #5466.

  The previous PR replaced SglistOwner with an OwnsSglist flag but kept std::optional<TGuardedSglistOwner> as an intermediate ownership mechanism. This PR completes the cleanup by removing
  TGuardedSglistOwner entirely and redesigning the ownership model from scratch.

  What changed vs #5466:

  - Removed TGuardedSglistOwner from guarded_sglist.h/cpp — no longer needed since ownership is tracked by OwnsSglist bool directly on the request.
  - Replaced CopySglistIntoBuffers() → TakeDataOwnership() — clearer name; now idempotent (early return if already owner or Sglist is empty).
  - Replaced CreateDependentRequest() with TDependentTag constructor — TWriteBlocksLocalRequest(source, TDependentTag{}) allows callers to construct a dependent copy directly in-place (e.g. via
  make_shared) without an intermediate stack copy.
  - Explicit move constructor and move assignment with SSO-safe Sglist rebuild — the default = default implementations were unsafe: after a protobuf move, short strings (SSO) are relocated to a
  new address, leaving raw Sglist pointers dangling. Both now call RebuildSglistFromBlocks() after moving proto fields.
  - Explicit destructor — calls CloseOwnedSglist() which spin-waits until all Acquire() guards are released before Blocks memory is freed, preventing use-after-free.
  - Added unit tests covering all ownership scenarios.

### DIFF
--- a/cloud/blockstore/libs/service/request.cpp
+++ b/cloud/blockstore/libs/service/request.cpp
@@ -14,11 +14,7 @@ TWriteBlocksLocalRequest::TWriteBlocksLocalRequest(
     , Sglist(std::move(other.Sglist))
     , BlocksCount(std::exchange(other.BlocksCount, 0))
     , OwnsSglist(std::exchange(other.OwnsSglist, false))
-{
-    if (OwnsSglist) {
-        RebuildSglistFromBlocks();
-    }
-}
+{}
 
 TWriteBlocksLocalRequest::TWriteBlocksLocalRequest(
     const TWriteBlocksLocalRequest& source,
@@ -49,9 +45,6 @@ TWriteBlocksLocalRequest& TWriteBlocksLocalRequest::operator=(
     BlocksCount = std::exchange(other.BlocksCount, 0);
     OwnsSglist = std::exchange(other.OwnsSglist, false);
     TWriteBlocksRequest::operator=(std::move(other));
-    if (OwnsSglist) {
-        RebuildSglistFromBlocks();
-    }
     return *this;
 }
 
@@ -91,17 +84,6 @@ void TWriteBlocksLocalRequest::CloseOwnedSglist()
         Sglist.Close();
         OwnsSglist = false;
     }
-}
-
-void TWriteBlocksLocalRequest::RebuildSglistFromBlocks()
-{
-    TSgList newSgList;
-    const auto& buffers = GetBlocks().buffers();
-    newSgList.reserve(buffers.size());
-    for (const auto& buffer: buffers) {
-        newSgList.emplace_back(buffer.data(), buffer.size());
-    }
-    Sglist = TGuardedSgList(std::move(newSgList));
 }
 
 TWriteBlocksLocalRequest CopyRequest(const TWriteBlocksLocalRequest& request)
--- a/cloud/blockstore/libs/service/request.cpp
+++ b/cloud/blockstore/libs/service/request.cpp
@@ -45,10 +45,10 @@ TWriteBlocksLocalRequest& TWriteBlocksLocalRequest::operator=(
 {
     CloseOwnedSglist();
 
-    TWriteBlocksRequest::operator=(std::move(other));
     Sglist = std::move(other.Sglist);
     BlocksCount = std::exchange(other.BlocksCount, 0);
     OwnsSglist = std::exchange(other.OwnsSglist, false);
+    TWriteBlocksRequest::operator=(std::move(other));
     if (OwnsSglist) {
         RebuildSglistFromBlocks();
     }
--- a/cloud/blockstore/libs/service/request.cpp
+++ b/cloud/blockstore/libs/service/request.cpp
@@ -8,45 +8,43 @@ namespace NProto {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-TWriteBlocksLocalRequest TWriteBlocksLocalRequest::Clone() const
+TWriteBlocksLocalRequest::~TWriteBlocksLocalRequest()
 {
-    auto request = CreateDependentRequest();
-    if (SglistOwner) {
-        request.CopySglistIntoBuffers();
-    }
-
-    return request;
+    CloseOwnedSglist();
 }
 
-TWriteBlocksLocalRequest TWriteBlocksLocalRequest::CreateDependentRequest() const
+TWriteBlocksLocalRequest::TWriteBlocksLocalRequest(
+    const TWriteBlocksLocalRequest& source,
+    TDependentTag)
 {
-    Y_ABORT_UNLESS(GetDescriptor()->field_count() == 8);
-
-    TWriteBlocksLocalRequest copiedRecord;
+    Y_ABORT_UNLESS(source.GetDescriptor()->field_count() == 8);
 
     // Copy all protobuf fields except Blocks (field 4)
-    copiedRecord.MutableHeaders()->CopyFrom(GetHeaders());
-    copiedRecord.SetDiskId(GetDiskId());
-    copiedRecord.SetStartIndex(GetStartIndex());
-    copiedRecord.SetFlags(GetFlags());
-    copiedRecord.SetSessionId(GetSessionId());
-    copiedRecord.SetBlockSize(GetBlockSize());
-    copiedRecord.MutableChecksums()->CopyFrom(GetChecksums());
+    MutableHeaders()->CopyFrom(source.GetHeaders());
+    SetDiskId(source.GetDiskId());
+    SetStartIndex(source.GetStartIndex());
+    SetFlags(source.GetFlags());
+    SetSessionId(source.GetSessionId());
+    SetBlockSize(source.GetBlockSize());
+    MutableChecksums()->CopyFrom(source.GetChecksums());
 
     // Copy non-protobuf fields
-    copiedRecord.Sglist = Sglist;
-    copiedRecord.BlocksCount = BlocksCount;
-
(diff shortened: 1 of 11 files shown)


cloud/blockstore/libs/service/request_ut.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    Y_UNIT_TEST(TakeOwnershipCopiesData)
+    {
+        TString data = "hello";
+        TWriteBlocksLocalRequest request;
+        SetupRequest(request, data);
+        UNIT_ASSERT(!request.IsOwner());
+
+        request.TakeDataOwnership();
```
drbasic, line 90: data = "world";

cloud/blockstore/libs/service/request_ut.cpp
```diff
@@ -0,0 +1,160 @@
...
+        TWriteBlocksLocalRequest request;
+        SetupRequest(request, data);
+        request.TakeDataOwnership();
+
+        data = "world";
+
+        auto guard = request.Sglist.Acquire();
+        UNIT_ASSERT_VALUES_EQUAL(
+            TStringBuf("hello"),
+            TStringBuf(guard.Get()[0].Data(), guard.Get()[0].Size()));
```
drbasic: guard.Get()[0].AsStringBuf()

cloud/blockstore/libs/service/request_ut.cpp
zavik001: > data = "world";

done

cloud/blockstore/libs/service/request_ut.cpp
zavik001: > guard.Get()[0].AsStringBuf()

done

cloud/blockstore/libs/service/request.cpp
```diff
@@ -8,45 +8,70 @@ namespace NProto {
...
+        RebuildSglistFromBlocks();
     }
-
-    return request;
 }
 
-TWriteBlocksLocalRequest TWriteBlocksLocalRequest::CreateDependentRequest() const
+TWriteBlocksLocalRequest::TWriteBlocksLocalRequest(
+    const TWriteBlocksLocalRequest& source,
+    TDependentTag)
```
SvartMetal, line 21: Why do we need to change the signature here? It looked fine before

cloud/blockstore/libs/service/request.cpp
```diff
@@ -56,15 +81,34 @@ void TWriteBlocksLocalRequest::CopySglistIntoBuffers()
         memcpy(buffer.begin(), block.Data(), block.Size());
         newSgList.emplace_back(buffer.data(), buffer.size());
     }
+    Sglist = TGuardedSgList(std::move(newSgList));
+    OwnsSglist = true;
+}
+
+void TWriteBlocksLocalRequest::CloseOwnedSglist()
+{
+    if (OwnsSglist && !Sglist.Empty()) {
```
SvartMetal: Codex has concern https://github.com/ydb-platform/nbs/pull/6219#discussion_r3412231802

Let's discuss it

cloud/blockstore/libs/service/request.h
```diff
@@ -56,49 +56,66 @@ struct TReadBlocksLocalResponse: public TReadBlocksResponse
...
+
+    // Copies data from Sglist into Blocks, rebuilds Sglist to point into those
+    // Blocks buffers, and sets OwnsSglist=true.
+    // No-op if already owner (OwnsSglist=true) or if the object was moved from.
+    void TakeDataOwnership();
+
+    bool IsOwner() const
+    {
+        return OwnsSglist;
+    }
```
SvartMetal, line 108: The need to expose it looks a bit weird.

cloud/blockstore/libs/service/request.cpp
SvartMetal: It looks like we had a similar `if` statement before https://github.com/ydb-platform/nbs/pull/6219/changes#diff-a5754f6ce2b4c1e9c8ae5192a295afe16b9a7a1442c075aedf322ce629cbe4c2L308

cloud/blockstore/libs/service/request.cpp
zavik001: > Why do we need to change the signature here? It looked fine before

CreateDependentRequest() returned by value, so callers always paid for at least one extra object construction (default-ctor + move-assign, or temporary + move into make_shared) https://github.com/ydb-platform/nbs/pull/6219/changes#diff-00637dbd7e0c10248f5f86ac97557dd33fd8115a3b55b815b71bc2debdca7e71L445 . A constructor eliminates the intermediate object and lets make_shared construct in-place directly

cloud/blockstore/libs/service/request.h
zavik001: > The need to expose it looks a bit weird.

This is useful for tests

cloud/blockstore/libs/service/request.cpp
zavik001: > It looks like we had a similar `if` statement before https://github.com/ydb-platform/nbs/pull/6219/changes#diff-a5754f6ce2b4c1e9c8ae5192a295afe16b9a7a1442c075aedf322ce629cbe4c2L308


The bot's concern doesn't apply here. The !Sglist.Empty() guard in CloseOwnedSglist() is only reached when OwnsSglist=true dependent requests (including those in TNonreplicatedPartitionActor) have wnsSglist=false and skip the entire close path. The behavior is equivalent to the old code: SglistOwner was std::optional and simply wasn't set for dependents, so ~TGuardedSglistOwner never ran for them either

cloud/blockstore/libs/service/request.cpp
```diff
@@ -8,45 +8,70 @@ namespace NProto {
...
-    return copiedRecord;
+TWriteBlocksLocalRequest& TWriteBlocksLocalRequest::operator=(
+    TWriteBlocksLocalRequest&& other) noexcept
+{
+    CloseOwnedSglist();
+
+    TWriteBlocksRequest::operator=(std::move(other));
+    Sglist = std::move(other.Sglist);
+    BlocksCount = std::exchange(other.BlocksCount, 0);
+    OwnsSglist = std::exchange(other.OwnsSglist, false);
```
SvartMetal: It looks weird that we apply `std::move` to `other` and then use its fields, such as `BlocksCount`, right after that

zavik001: > It looks weird that we apply `std::move` to `other` and then use its fields, such as `BlocksCount`, right after that

changed the order

drbasic (dismissed):

cloud/blockstore/libs/service/request.cpp
```diff
@@ -8,45 +8,70 @@ namespace NProto {
...
+    : TWriteBlocksRequest(std::move(other))
+    , Sglist(std::move(other.Sglist))
+    , BlocksCount(std::exchange(other.BlocksCount, 0))
+    , OwnsSglist(std::exchange(other.OwnsSglist, false))
 {
-    auto request = CreateDependentRequest();
-    if (SglistOwner) {
-        request.CopySglistIntoBuffers();
+    if (OwnsSglist) {
+        RebuildSglistFromBlocks();
```
SvartMetal: What will happen to dependent requests if the source request that owns the `sglist` is moved?

Are we sure that dependents will not continue using the old, invalid `sglist`?

At the very least, we should have a test for this.

cloud/blockstore/libs/service/request.h
```diff
@@ -56,49 +56,66 @@ struct TReadBlocksLocalResponse: public TReadBlocksResponse
...
-    // will be called) after this object destruction. Doesn't copy Blocks field
-    // from TWriteBlocksLocalRequest, because all blocks can be accessed via
-    // Sglist.
-    TWriteBlocksLocalRequest CreateDependentRequest() const;
+    void CloseOwnedSglist();
 
-    void CopySglistIntoBuffers();
+    // After a protobuf move, short strings (SSO) may be relocated to a new
+    // address, so raw pointers in Sglist become stale. This method rebuilds
+    // Sglist from the current buffer addresses in the Blocks field.
```
SvartMetal: Claude says this:

Image

cloud/blockstore/libs/service/request.cpp
zavik001: > What will happen to dependent requests if the source request that owns the `sglist` is moved?
> 
> Are we sure that dependents will not continue using the old, invalid `sglist`?
> 
> At the very least, we should have a test for this.

RebuildSglistFromBlocks was based on a wrong premise RepeatedPtrField<string> move transfers the pointer array, not the string objects, so buffer.data() is stable. Worse, the rebuild was creating the very roblem it meant to prevent: it forked TGuardedSgList, orphaning dependents with a guard that was never closed. After removing it, owner and dependents share the same TGuardedSgList through the move, and CloseOwnedSglist() on destruction correctly invalidates all of them

SvartMetal (approved):

cloud/blockstore/libs/service/request.cpp
SvartMetal: Looks like Codex's concern was valid

drbasic (approved):

Issue #6220, opened 2026-06-15
use-after-free in rdma target


tpashkin:
ExecuteRequest passes ownership of the it's buffer to the user. If session gets disconnected during request execution, the buffer will get freed along with the pool it was allocated from

This affects both cloud/blockstore/libs/service_rdma/rdma_target.cpp and cloud/blockstore/libs/storage/disk_agent/rdma_target.cpp and probably cloud/blockstore/libs/endpoints_rdma/rdma_server.cpp

cloud/blockstore/libs/storage/disk_agent/rdma_target.cpp holds a weak_ptr to the endpoint, so it won't try to send the response if endpoint is down, but may still use the buffer

PR #6222, opened 2026-06-15
fix use-after free in RDMA

yegorskii:
### Notes
It is possible that async handlers has access to memory after it was freed after RDMA disconnect. In this case we can reallocate all the buffers but handler reading/writing to this memory will be still executing

### Issue
https://github.com/ydb-platform/nbs/issues/6220


cloud/blockstore/libs/service_rdma/rdma_target.cpp
```diff
@@ -146,14 +136,23 @@ class TRequestHandler final
 
         STORAGE_TRACE("Processing req with msgId %u", parseResult.MsgId);
 
+#define BLOCKSTORE_HANDLE_REQUEST(name, ...)                                   \
+    case TBlockStoreServerProtocol::Ev##name##Request:                         \
+        return Handle##name##Request(                                          \
+            ep,                                                                \
+            context,                                                           \
+            std::move(callContext),                                            \
+            static_cast<NProto::T##name##Request*>(&*parseResult.Proto),      \
```
tpashkin: indent

cloud/blockstore/libs/client_rdma/rdma_client_ut.cpp
```diff
@@ -346,6 +346,40 @@ Y_UNIT_TEST_SUITE(TRdmaClientTest)
         }
     }
 
+    Y_UNIT_TEST(ShouldReadIntoLargerSglist)
```
SvartMetal: Let's also add a randomized test that reproduces the use-after-free

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1484,8 +1535,23 @@ class TCompletionPoller final
...
             Config->AdaptiveWaitSleepDelay);
+        bool flushStarted = false;
+
+        while (true) {
+            if (ShouldStop()) {
+                if (!flushStarted) {
+                    if (WaitMode == EWaitMode::Poll) {
+                        StopEvent.Clear();
+                    }
+                    FlushSessions();
```
tpashkin, line 1535: let's do this after the main loop

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1484,8 +1535,23 @@ class TCompletionPoller final
...
+                    flushStarted = true;
+                }
+
+                if (!HasSessions()) {
```
tpashkin, line 1539: we need to spin here

cloud/blockstore/libs/client_rdma/rdma_client.cpp
```diff
@@ -249,7 +249,7 @@ class TReadBlocksHandler final
             dst);
 
         STORAGE_VERIFY(
-            bytesRead == dstSize,
+            bytesRead == srcSize,
```
tpashkin: if these 2 differ, it's a bug in itself. but regardless, it has little to do with use-after-free we are trying to fix, so let's remove it from this PR

cloud/blockstore/libs/endpoints_rdma/rdma_server.cpp
```diff
@@ -150,12 +151,7 @@ void TRdmaEndpoint::HandleRequest(
                 });
 
             if (HasError(error)) {
-                if (auto p = self.lock()) {
-                    p->Endpoint->SendError(
-                        context,
-                        error.GetCode(),
-                        error.GetMessage());
-                }
+                ep->SendError(context, error.GetCode(), error.GetMessage());
```
tpashkin: these changes are irrelevant to the bug in question. agreed to remove these

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1467,14 +1515,17 @@ class TCompletionPoller final
...
-    void DisconnectFlushed()
+    void DisconnectFlushed(bool stopRequested)
     {
         auto sessions = Sessions.Get();
 
         for (const auto& session: *sessions) {
-            if (session->IsFlushed()) {
-                Release(session.get());
+            if (!session->IsFlushed() &&
+                !(stopRequested && session->CanReleaseAfterStop()))
```
tpashkin: i think we should stick to IsFlushed in both cases. it would actually be a good idea to wait for WRs on stop

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -627,6 +640,10 @@ void TServerSession::Flush() noexcept
 {
     RDMA_DEBUG("flush queues");
 
+    if (FlushStarted) {
```
tpashkin, line 642: i suppose we should move this before the RDMA_DEBUG call

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -295,6 +302,11 @@ class TServerSession final
     TSimpleList<TRequest> QueuedRequests;
     TEventHandle RequestEvent;
     bool FlushStarted = false;
+    // Count of requests currently transferred to the handler (via req.release()
+    // in ExecuteRequest) whose SendResponse/SendError has not yet been called.
+    // IsFlushed() waits for this to reach zero so session buffers are not freed
+    // while the handler or its async callbacks still hold in/out TStringBufs.
+    std::atomic<ui32> ExecutingRequests{0};
```
tpashkin: this variable only gets accessed from the CQ thread, it doesn't really need to be atomic

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -800,6 +830,9 @@ void TServerSession::FreeRequest(TRequestPtr req, TSendWr* send) noexcept
     if (req == nullptr) {
         return;
     }
+    if (req->TransferredToHandler) {
+        ExecutingRequests.fetch_sub(1, std::memory_order_release);
```
tpashkin: i would prefer this decrement to happen closer to the actual transfer of the ownership from the client back to us. EnqueueRequest or thereabouts

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1010,6 +1043,9 @@ void TServerSession::ExecuteRequest(TRequestPtr req) noexcept
         req->CallContext->LWOrbit,
         req->CallContext->RequestId);
 
+    req->TransferredToHandler = true;
```
tpashkin, line 1038: if ExecutingRequests is atomic, shouldn't TransferredToHandler also be atomic?

cloud/blockstore/libs/client_rdma/rdma_client_ut.cpp
yegorskii: decided to move this to a separete pr since this test is not releated to use-after-free

cloud/storage/core/libs/rdma/impl/server.cpp
SvartMetal, line 1: What about tests?

We need a "concrete scenario" and randomized test as well

yegorskii: will do, but as you  see the current implementation is ruining the existing ones

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1473,7 +1513,7 @@ class TCompletionPoller final
 
         for (const auto& session: *sessions) {
             if (session->IsFlushed()) {
-                Release(session.get());
+                 Release(session.get());
```
tpashkin: i don't think this extra space is necessary

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -800,6 +821,9 @@ void TServerSession::FreeRequest(TRequestPtr req, TSendWr* send) noexcept
     if (req == nullptr) {
         return;
     }
+    if (req->TransferredToHandler) {
```
tpashkin, line 825: come to think about it. there is actually another way to check this – test req->State

cloud/storage/core/libs/rdma/impl/server_ut.cpp
```diff
@@ -175,12 +234,18 @@ TEST(TRdmaServerTest, ShouldUseConfiguredQpParamsOnAcceptAndSetupQp)
     {
         Y_UNUSED(qp);
 
+        if ((mask & IBV_QP_STATE) && attr->qp_state == IBV_QPS_ERR) {
+            NVerbs::Flush(testContext);
+            return;
+        }
+
         const int expectedMask = IBV_QP_TIMEOUT | IBV_QP_MIN_RNR_TIMER;
+        if (mask != expectedMask) {
```
tpashkin: are there any other ModifyQP operations, except this one and transitioning QP into the ERR state, which we handled above?

cloud/storage/core/libs/rdma/impl/server_ut.cpp
```diff
@@ -378,26 +444,32 @@ TEST(TRdmaServerTest, ShouldHandleErrors)
...
         context->HandleCompletionEvent = [&](ibv_wc* wc) {
-            // good id, good opcode, error status
-            if (wc->wr_id == recv[0]->wr_id) {
-                wc->status = IBV_WC_RETRY_EXC_ERR;
+            // bad id, good opcode
+            if (wc->wr_id == badWr.wr_id) {
                 return;
             }
-            // bad id, good opcode
+
```
tpashkin: let's add newlines consistently – either remove extra newlines in the modified code or add newlines between if's

cloud/storage/core/libs/rdma/impl/server_ut.cpp
```diff
@@ -506,6 +578,191 @@ TEST(TRdmaServerTest, ShouldRejectConnectionOnConfigMismatchInStrictValidation)
...
+    NThreading::TPromise<void> flushTriggered =
+        NThreading::NewPromise<void>();
+
+    // Combine flush-on-error with a signal so the test can wait until Stop()
+    // has actually entered the flush phase before calling SendResponse.
+    auto flushHook = SetupFlushOnQpError(context);
+    context->ModifyQP = [flushTriggered, flushHook = std::move(flushHook)](
+        ibv_qp* qp, ibv_qp_attr* attr, int mask) mutable
+    {
+        flushHook(qp, attr, mask);
```
tpashkin: i think calling Flush directly looks cleaner

cloud/storage/core/libs/rdma/impl/test_verbs.cpp
```diff
@@ -73,6 +73,33 @@ void EnqueueConnectionEvent(
...
+        if (remaining > 0) {
+            context->HandleCompletionEvent =
+                [ctx = context.Get(), remaining](ibv_wc* wc) mutable {
+                    if (wc->opcode != IBV_WC_RECV) {
+                        return;  // let send WR completions through unchanged
+                    }
+                    wc->status = IBV_WC_WR_FLUSH_ERR;
+                    wc->opcode = static_cast<ibv_wc_opcode>(0);
+                    if (--remaining == 0) {
+                        ctx->HandleCompletionEvent = nullptr;
```
tpashkin: why would you need to do this? looks like something that would be needed if we wanted to complete requests after the disconnect. i don't see you using Flush in ModifyQP, so i assume it's not necessary

cloud/storage/core/libs/rdma/impl/server_ut.cpp
```diff
@@ -506,6 +578,191 @@ TEST(TRdmaServerTest, ShouldRejectConnectionOnConfigMismatchInStrictValidation)
...
+
+        context->ProcessedRecvEvents.push_back(recv);
+        context->CompletionHandle.Set();
+    }
+
+    handler->RequestReceived.GetFuture().GetValueSync();
+    ASSERT_NE(nullptr, handler->Context);
+
+    // ExecutingRequests == 1: session cannot be released yet.
+    ASSERT_FALSE(sessionDestroyed.GetFuture().HasValue());
```
tpashkin: i think there is no point checking it before Stop has been called. better yet, we should wait until WRs are flushed like in the second test (where we call disconnect) and check it then

cloud/storage/core/libs/rdma/impl/server_ut.cpp
```diff
@@ -506,6 +578,191 @@ TEST(TRdmaServerTest, ShouldRejectConnectionOnConfigMismatchInStrictValidation)
...
+    Y_DEFER {
+        if (stopThread.joinable()) {
+            stopThread.join();
+        }
+    };
+
+    // Wait until Stop() has actually reached the flush phase (ModifyQP hook
+    // fired), guaranteeing it is now blocked on IsFlushed().
+    flushTriggered.GetFuture().GetValueSync();
+
```
tpashkin, line 663: i think that's the appropriate place to test sessionDestroyed

cloud/storage/core/libs/rdma/impl/server_ut.cpp
tpashkin: it wouldn't hurt, so we can leave it as is. but also should check after the Stop has been called and WRs have been flushed

cloud/storage/core/libs/rdma/impl/test_verbs.cpp
```diff
@@ -73,6 +73,33 @@ void EnqueueConnectionEvent(
...
 
+void Flush(TTestContextPtr context)
+{
+    with_lock (context->CompletionLock) {
+        auto remaining = context->RecvEvents.size();
+        if (remaining > 0) {
+            context->HandleCompletionEvent =
+                [ctx = context.Get(), remaining](ibv_wc* wc) mutable {
+                    if (wc->opcode != IBV_WC_RECV) {
+                        return;  // let send WR completions through unchanged
```
tpashkin: but why?

cloud/storage/core/libs/rdma/impl/server.cpp
tpashkin: RejectRequest also transitions request to ExecuteRequest state, so we would also have to increment ExecutingRequests, which is undesirable

tpashkin (dismissed):

cloud/storage/core/libs/rdma/impl/server_ut.cpp
```diff
@@ -506,6 +571,206 @@ TEST(TRdmaServerTest, ShouldRejectConnectionOnConfigMismatchInStrictValidation)
...
+               c->GetCounter("ActiveSend")->Val()  == 0 &&
+               c->GetCounter("ActiveWrite")->Val() == 0 &&
+               c->GetCounter("ActiveRead")->Val()  == 0;
+    }));
+
+    ASSERT_FALSE(sessionDestroyed.GetFuture().HasValue());
+
+    endpoint->SendResponse(handler->Context, 0);
+
+    sessionDestroyed.GetFuture().GetValueSync();
```
tpashkin, line 778: nit: maybe do `ASSERT_TRUE(sessionDestroyed.GetFuture().GetValueSync())` like in the other test

tpashkin (dismissed):

komarevtsev-d (dismissed):

cloud/storage/core/libs/rdma/impl/server.cpp
```diff
@@ -1,5 +1,7 @@
 #include "server.h"
 
+#include <atomic>
```
tpashkin: let's move it next to `include <variant>`

komarevtsev-d (approved):

tpashkin (approved):

PR #6249, opened 2026-06-17
[NBS] Fix blobs cleanup for disks with not default block size with Block Mask optimization

vladstepanyuk:
### Notes
Let's say we have a disk with a 32 KB block size. Then IsBlockMaskFull returns true only if the first 128 bits are set to ones and the rest are not. If BlockMask has only ones, IsBlockFullMask will return false. We use IsBlockMaskFul to determine whether to clean up a blob. Therefore, blobs are not cleaned even if BlockMask is full.

In this pr i corrected IsBlockMaskFull function: now it returns true if the first blockCount bits are set and doesn't check other bits. And corrected GetFullBlockMask function, so now it accepts MaxBlocksInBlob as an argument and fills only the first MaxBlocksInBlob bits in the mask.

### Issue
https://github.com/ydb-platform/nbs/issues/5590


SvartMetal:
Let's explain the bug in more detail in the description: the root cause, the impact, and the fix.

cloud/blockstore/tests/loadtest/local-newfeatures/test.py
```diff
@@ -336,11 +361,156 @@ def __init__(
...
+            storage_config_with_new_features_enabled(),
+        ],
+        None,
+        check_disk_size=True,
+        disk_id="vol0",
+    ),
 ]
 
 
+def _int_field(value, default=0):
```
SvartMetal: _int_or_default

cloud/blockstore/tests/loadtest/local-newfeatures/test.py
```diff
@@ -336,11 +361,156 @@ def __init__(
...
+    fresh = 0
+    bytes_count = 0
+
+    for partition_id in range(partition_count):
+        try:
+            partition_info = get_partition_info(client, disk_id, partition_id)
+        except AssertionError:
+            if partition_id == 0:
+                raise
+            break
```
SvartMetal: Could you please explain why is it needed?

cloud/blockstore/tests/loadtest/local-newfeatures/test.py
```diff
@@ -336,11 +361,156 @@ def __init__(
...
+def get_disk_blocks_count(client, disk_id, max_partitions=1):
+    blocks_count = 0
+
+    for partition_id in range(max_partitions):
+        try:
+            partition_info = get_partition_info(client, disk_id, partition_id)
+        except AssertionError:
+            if partition_id == 0:
+                raise
+            break
```
SvartMetal, line 459: Maybe extract into a func?

cloud/blockstore/libs/storage/partition/part_ut.cpp
```diff
@@ -13897,12 +13922,30 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
...
     }
 
-    Y_UNIT_TEST(ShouldCleanupMergedBlobsWithoutReadingBlockMasks)
+    void DoTestShouldCleanupMergedBlobsWithoutReadingBlockMasks(
+        ui32 blockSize = DefaultBlockSize)
     {
+        constexpr ui32 MaxBlobSize = 4_MB;
+        const ui32 maxBlocksPerBlob =
+            CalculateMaxBlocksInBlob(MaxBlobSize, blockSize);
+        const ui32 halfMaxBlocksPerBlob = maxBlocksPerBlob / 2;
```
SvartMetal: maxBlocksInBlob

halfMaxBlocksInBlob

cloud/blockstore/libs/storage/partition/model/block_mask.cpp
```diff
@@ -31,8 +31,12 @@ bool IsBlockMaskFull(const TBlockMask& mask, ui32 blockCount)
...
         if (blockCount < blocksInChunk) {
             const TBitMap<blocksInChunk> m(chunk);
-            Y_DEBUG_ABORT_UNLESS(m.Count() <= blockCount);
-            return m.Count() == blockCount;
+            for (size_t j = 0; j < blockCount; ++j) {
+                if (!m.Test(j)) {
+                    return false;
+                }
+            }
+            return true;
```
SvartMetal: Maybe we can calculate it in a more efficient way?

cloud/blockstore/libs/storage/partition/model/block_mask.cpp
```diff
@@ -2,6 +2,27 @@
 
 namespace NCloud::NBlockStore::NStorage::NPartition {
 
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <typename T>
+T GetIntWithNBits(unsigned n)
```
SvartMetal, line 13: Let's static assert that T is also an unsigned type, for safety

SvartMetal: Even stricter: we need T to be an unsigned Integer-like type

SvartMetal:
```
static_assert(std::is_integral_v<T>);
static_assert(std::is_unsigned_v<T>);
```

SvartMetal: Looks clearer:
```
template <typename T>
T GetIntWithNBits(unsigned n)
{
    static_assert(std::is_integral_v<T>);
    static_assert(std::is_unsigned_v<T>);
    static_assert(!std::is_same_v<T, bool>);

    constexpr unsigned bits = std::numeric_limits<T>::digits;

    if (n == 0) {
        return T{0};
    }

    if (n >= bits) {
        return std::numeric_limits<T>::max();
    }

    return static_cast<T>(
        std::numeric_limits<T>::max() >> (bits - n));
}
```

SvartMetal (approved):

komarevtsev-d (approved):

PR #6253, opened 2026-06-17
CSI driver should support read-only mode + MRSW E2E test

faucct:
### Notes
Fixing that a read-only attach of PVC ends up writeable. Also adding a E2E test for multi-read-single-write attaches.

### DIFF
--- a/cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
+++ b/cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
@@ -119,13 +119,14 @@ def test_publish_distinguishes_remote_mount_access_modes(
             instance_id=instance_id,
             access_mode=access_mode)
 
-    def publish(pod_id: str, access_mode: str):
+    def publish(pod_id: str, access_mode: str, readonly: bool = False):
         env.csi.publish_volume(
             pod_id=pod_id,
             volume_id=volume_name,
             pod_name=pod_name,
             access_type=access_type,
             instance_id=instance_id,
+            readonly=readonly,
             access_mode=access_mode)
         published_pod_ids.append(pod_id)
 
@@ -137,7 +138,7 @@ def test_publish_distinguishes_remote_mount_access_modes(
         env.csi.create_volume(name=volume_name, size=volume_size)
 
         if vm_mode and instance_id == "":
-            stage("single-node-writer")
+            stage("single-node-reader-only")
 
             publish(single_reader_pod_id, "single-node-reader-only")
 
@@ -151,11 +152,14 @@ def test_publish_distinguishes_remote_mount_access_modes(
                 remote=0)
 
             unpublish(single_reader_pod_id)
+            env.csi.unstage_volume(volume_name)
+
+            stage("multi-node-single-writer")
 
             publish(writer_pod_id, "multi-node-single-writer")
 
             for pod_id in reader_pod_ids:
-                publish(pod_id, "multi-node-reader-only")
+                publish(pod_id, "multi-node-single-writer", readonly=True)
 
             assert "rw" == get_access_mode(str(target_path(writer_pod_id)))
             for pod_id in reader_pod_ids:
--- a/cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
+++ b/cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
@@ -63,76 +63,158 @@ def test_readonly_volume(mount_path, access_type, vm_mode, gid):
         csi.cleanup_after_test(env, volume_name, access_type, [pod_id])
 
 
-def test_vm_mode_legacy_publish_distinguishes_reader_access_modes():
-    env, run = csi.init(vm_mode=True)
+def get_target_path(pod_id: str, volume_name: str, access_type: str) -> Path:
+    if access_type == "block":
+        return (
+            Path("/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish")
+            / pod_id
+            / volume_name
+        )
+
+    return (
+        Path("/var/lib/kubelet/pods")
+        / pod_id
+        / "volumes/kubernetes.io~csi"
+        / volume_name
+        / "mount"
+    )
+
+
+def assert_endpoints(run, volume_name: str, total: int, read_only: int, remote: int):
+    endpoints = run("listendpoints", "--proto").stdout
+    assert total == endpoints.count(f'DiskId: "{volume_name}"')
+    assert read_only == endpoints.count("VolumeAccessMode: VOLUME_ACCESS_READ_ONLY")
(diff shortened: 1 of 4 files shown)


cloud/blockstore/tools/csi_driver/client/main.go
```diff
@@ -275,19 +330,23 @@ func newNodeStageVolumeCommand(endpoint *string) *cobra.Command {
...
-			accessMode := csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER
 			response, err := client.NodeStageVolume(
 				ctx,
 				&csi.NodeStageVolumeRequest{
 					VolumeId:          volumeId,
 					StagingTargetPath: stagingTargetPath,
-					VolumeCapability:  createVolumeCapability(accessMode, accessType, ""),
-					VolumeContext:     volumeContext,
+					VolumeCapability: createVolumeCapability(
+						csi.VolumeCapability_AccessMode_Mode(accessMode),
```
jkuradobery, line 346: Зачем тут нужен cast?

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -250,6 +250,42 @@ func parseVolumeId(volumeId string) (nbsId string, instanceId string) {
...
+	}
+
+	return nbsapi.EVolumeAccessMode_VOLUME_ACCESS_READ_WRITE
+}
+
+func nbsVolumeMountMode(
+	accessMode nbsapi.EVolumeAccessMode,
+) nbsapi.EVolumeMountMode {
+	if accessMode == nbsapi.EVolumeAccessMode_VOLUME_ACCESS_READ_ONLY {
+		return nbsapi.EVolumeMountMode_VOLUME_MOUNT_REMOTE
```
tpashkin: RO access does not imply REMOTE

faucct: MRSW won’t work without this, I guess. Do you want the kubernetes user to somehow request remote explicitly?

faucct: I've changed it to be distinguished by `SINGLE_NODE_READER_ONLY`/`MULTI_NODE_READER_ONLY`:
https://github.com/container-storage-interface/spec/blob/master/spec.md

tpashkin: what about `MULTI_NODE_SINGLE_WRITER`? it should also use remote mount

cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
```diff
@@ -63,6 +63,84 @@ def test_readonly_volume(mount_path, access_type, vm_mode, gid):
         csi.cleanup_after_test(env, volume_name, access_type, [pod_id])
 
 
+def test_vm_mode_legacy_publish_distinguishes_reader_access_modes():
```
antonmyagkov: Why does this test cover only the legacy mode? Most PVs already include instance_id, so the legacy path is becoming less representative. Let's parameterize the test with instanceId and verify both scenarios.

antonmyagkov: You also changed the start endpoint logic for NBD devices (vm_mode=false).

Could you parameterize the test with vm_mode as well so that both scenarios are covered?

antonmyagkov (changes requested):

cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
```diff
@@ -63,6 +63,166 @@ def test_readonly_volume(mount_path, access_type, vm_mode, gid):
...
+    assert remote == endpoints.count("VolumeMountMode: VOLUME_MOUNT_REMOTE")
+
+
+@pytest.mark.parametrize(
+    "vm_mode,instance_id,access_type",
+    [
+        (True, "", "mount"),
+        (True, "example-instance-id", "mount"),
+        (False, "", "block"),
+        (False, "example-instance-id", "block"),
```
antonmyagkov: False, "", mount

instance_id is empty for non-vm mode

cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
```diff
@@ -63,6 +63,166 @@ def test_readonly_volume(mount_path, access_type, vm_mode, gid):
...
+    ],
+    ids=["vm-legacy", "vm-instance-id", "pod-legacy", "pod-nbd"])
```
antonmyagkov: pod-legacy -> nbd_block_mode
pod-nbd -> nbd_mount_mode

cloud/blockstore/tests/csi_driver/e2e_tests_part2/test.py
```diff
@@ -63,6 +63,166 @@ def test_readonly_volume(mount_path, access_type, vm_mode, gid):
...
+
+    def unpublish(pod_id: str):
+        env.csi.unpublish_volume(pod_id, volume_name, access_type)
+        published_pod_ids.remove(pod_id)
+
+    try:
+        env.csi.create_volume(name=volume_name, size=volume_size)
+
+        if vm_mode and instance_id == "":
+            stage("single-node-writer")
```
antonmyagkov: Here we stage volume with single-node-writer access mode and publish with single-node-reader-only access mode.

According csi spec it should be fine to have different access modes on stage/publish. However I'm not sure that kuberentes implementation allows it as kuberentes rely on NodePublishVolumeRequest.Readonly

We should check kuberenetes implementation and provide the link on implementation/tests if it is possible to have different access mode for NodeStageVolumeRequest/NodePublushVolumeRequest for the same pvc

antonmyagkov (changes requested):

cloud/blockstore/tools/csi_driver/internal/driver/node.go
```diff
@@ -250,6 +250,46 @@ func parseVolumeId(volumeId string) (nbsId string, instanceId string) {
...
+	return nbsapi.EVolumeAccessMode_VOLUME_ACCESS_READ_WRITE
+}
+
+func nbsVolumeMountMode(
+	accessMode *csi.VolumeCapability_AccessMode,
+) nbsapi.EVolumeMountMode {
+	if accessMode != nil {
+		switch accessMode.GetMode() {
+		case csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY,
+			csi.VolumeCapability_AccessMode_MULTI_NODE_SINGLE_WRITER:
```
antonmyagkov: It is not possible to set MULTI_NODE_SINGLE_WRITER access mode in kubernetes. Let's remove it form this pull request.

[MULTI_NODE_SINGLE_WRITER](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes)

antonmyagkov (approved):

tpashkin (approved):

PR #6262, opened 2026-06-18
[NBS] Fix PG and disk routing to different NBS cells

agalibin:
### Solution
'CreatePlacementGroup' now uses 'CellSelector.SelectCellForPlacementGroup' to pick the cell. The resolved cell ID is stored in 'PlacementGroupMeta.ZoneID' instead of the raw zone ID from the request.
'SelectCellForDisk' now reads 'PlacementGroupMeta.ZoneID' from YDB and routes directly to that cell, bypassing 'CellSelector'.

### Notes
In the next iteration, we should implement group creation across all shards simultaneously. This will eliminate the need for group selection procedures. However, before doing so, we need to ensure there are no shard‑rack overlaps and agree to prevent such overlaps in the future.

### Issue
When NBS zone has multiple cells (e.g. 'zone-a' and 'zone-a-shard1'), 'CreatePlacementGroup' bypassed 'CellSelector' and sent the request using the raw 'ZoneId' from the API. 'CreateDisk' went through 'CellSelector', which could pick a different cell (e.g. via 'MAX_FREE_BYTES' policy). Since each cell is an independent NBS cluster with its own DiskRegistry, the disk creation failed with 'E_NOT_FOUND' placement group.


cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -51,7 +53,7 @@ func (t *createPlacementGroupTask) Run(
 
 	placementGroupMeta, err := t.storage.CreatePlacementGroup(ctx, resources.PlacementGroupMeta{
 		ID:                      t.request.GroupId,
-		ZoneID:                  t.request.ZoneId,
+		ZoneID:                  client.ZoneID(),
```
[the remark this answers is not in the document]
agalibin: Valid concern. Migration flow uses DstZoneId directly and doesn't resolve PG cell. This is a pre-existing issue — before this PR there was no sharding for PGs at all. It should be fixed in next PR.

cloud/disk_manager/internal/pkg/services/disks/create_disk_common.go
```diff
@@ -36,7 +38,25 @@ func SelectCellForDisk(
 			return nil, err
 		}
 	} else {
-		if common.IsLocalDiskKind(params.Kind) {
+		if len(params.PlacementGroupId) > 0 {
```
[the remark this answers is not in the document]
agalibin: This is by design. Once a placement group exists on a specific cell, all disks in that PG must go to the same cell - otherwise DiskRegistry returns E_NOT_FOUND.

agalibin:
@codex review

agalibin:
@codex review

agalibin:
@codex review

agalibin:
@codex review

cloud/disk_manager/internal/pkg/cells/interface.go
```diff
@@ -46,4 +53,8 @@ type CellSelector interface {
 	) (nfs.Client, error)
 
 	ZoneContainsCell(zoneID string, cellID string) bool
+
+	// Returns all cell IDs that belong to the given zone. If no sharding
```
gy2411: Les't write 'cells config' instead of 'sharding config'

agalibin:
@codex review

cloud/disk_manager/internal/pkg/services/disks/service.go
```diff
@@ -728,7 +729,7 @@ func (s *service) MigrateDisk(
 			req,
 		)
 	}
-	if req.DiskId.ZoneId == req.DstZoneId {
+	if req.DiskId.ZoneId == req.DstZoneId && len(req.DstPlacementGroupId) == 0 {
```
gy2411: Why do we check `len(req.DstPlacementGroupId) == 0` here?

cloud/disk_manager/internal/pkg/services/disks/service.go
```diff
@@ -740,6 +741,41 @@ func (s *service) MigrateDisk(
 		return "", err
 	}
 
+	dstZoneID := req.DstZoneId
+
+	if len(req.DstPlacementGroupId) > 0 {
```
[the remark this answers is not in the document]
gy2411: I think this argument is not valid.

Migration between cells is internal operation for NBS. It is not a problem for us to point exact cells in the request if needed.

cloud/disk_manager/internal/pkg/services/disks/service.go
```diff
@@ -740,6 +741,51 @@ func (s *service) MigrateDisk(
...
+	}
+
+	if zoneID == dstZoneID {
+		return "", common.NewInvalidArgumentError(
+			"cannot migrate disk to the same zone, req=%v",
+			req,
+		)
+	}
+
+	isMigrationBetweenCells := req.IsMigrationBetweenCells ||
```
gy2411: isMigrationBetweenCells must remain unchanged.

The reason is the following. If isMigrationBetweenCells=false, the migration task communicates with the caller (and it doesn't if isMigrationBetweenCells=ture). So the value of isMigrationBetweenCells must be determined by the caller.

cloud/disk_manager/internal/pkg/services/placementgroup/alter_placement_group_membership_task.go
gy2411, line 65: What will happen if we add disks from another cell to the placement group? Will NBS return error? Will NBS get to some inconsistent state?

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -51,7 +72,7 @@ func (t *createPlacementGroupTask) Run(
 
 	placementGroupMeta, err := t.storage.CreatePlacementGroup(ctx, resources.PlacementGroupMeta{
```
gy2411: Now it is possible that two different generations of the task select different cells and both call storage.CreatePlacementGroup. This will lead to the task failure.

Les't use the same pattern that we use for cell selection of disks -- save the cell id to task state:
- If the state already has the cell, then just use the cell from the state.
- Select cell
- Save the cell to the task state (and call t.saveState)
 
See https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/services/disks/create_empty_disk_task.go#L50 and then https://github.com/ydb-platform/nbs/blob/3b27b0a8a6a97dfa307874384f738f736ecaf214/cloud/disk_manager/internal/pkg/services/disks/create_disk_common.go#L20 -- note that we save task state here https://github.com/ydb-platform/nbs/blob/3b27b0a8a6a97dfa307874384f738f736ecaf214/cloud/disk_manager/internal/pkg/services/disks/create_disk_common.go#L76.

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -90,7 +119,17 @@ func (t *createPlacementGroupTask) Cancel(
 	execCtx tasks.ExecutionContext,
 ) error {
 
-	client, err := t.nbsFactory.GetClient(ctx, t.request.ZoneId)
+	pgMeta, err := t.storage.GetPlacementGroupMeta(ctx, t.request.GroupId)
```
gy2411: We already get placementGroupMeta in the line 139 -- no need to call GetPlacementGroupMeta.

cloud/disk_manager/internal/pkg/services/placementgroup/delete_placement_group_task.go
```diff
@@ -43,7 +46,29 @@ func (t *deletePlacementGroupTask) deletePlacementGroup(
 	execCtx tasks.ExecutionContext,
 ) error {
 
-	client, err := t.nbsFactory.GetClient(ctx, t.request.ZoneId)
+	pgMeta, err := t.storage.GetPlacementGroupMeta(ctx, t.request.GroupId)
```
gy2411: Let's use placementGroupMeta returned from torage.DeletePlacementGroup. No need to call GetPlacementGroupMeta.

cloud/disk_manager/internal/pkg/cells/cells.go
```diff
@@ -206,6 +219,15 @@ func (s *cellSelector) ZoneContainsCell(zoneID string, cellID string) bool {
 	return slices.Contains(s.getCells(zoneID), cellID)
 }
 
+func (s *cellSelector) ResolveCells(zoneID string) ([]string, error) {
```
gy2411, line 223: Let's write a comment explaining what this function do.

cloud/disk_manager/internal/pkg/cells/cells.go
```diff
@@ -326,6 +366,71 @@ func (s *cellSelector) selectCellForDisk(
...
+	if len(cells) == 0 {
+		if s.isCell(zoneID) {
+			return zoneID, nil
+		}
+
+		return "", errors.NewNonCancellableErrorf(
+			"incorrect zone ID provided: %q",
+			zoneID,
+		)
+	}
```
gy2411, line 422: This looks almost like a copy-paste from ResolveCells function. Can we deduplicate this code?

cloud/disk_manager/internal/pkg/services/disks/create_disk_common.go
```diff
@@ -36,7 +39,73 @@ func SelectCellForDisk(
...
+				if localClient.ZoneID() != pgMeta.ZoneID {
+					return nil, errors.NewNonRetriableErrorf(
+						"agent for local disk is in cell %v, "+
+							"but placement group %v is in cell %v",
+						localClient.ZoneID(),
+						params.PlacementGroupId,
+						pgMeta.ZoneID,
+					)
+				}
+				client = localClient
```
gy2411, line 104: SelectCellForDisk function looks very complicated now. Can we move some part(s) of this code to separate function(s)?

cloud/disk_manager/internal/pkg/services/disks/service.go
agalibin: replace len(req.DstPlacementGroupId) == 0 with !req.IsMigrationBetweenCells

cloud/disk_manager/internal/pkg/services/disks/service.go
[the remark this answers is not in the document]
agalibin: if migration between cells we need to throw error if PG cell and dst zone id not equal

cloud/disk_manager/internal/pkg/services/placementgroup/alter_placement_group_membership_task.go
agalibin: looks like will be returned error E_ARGUMENT from libs/storage/disk_registry/disk_registry_state.cpp:4915

cloud/disk_manager/internal/pkg/cells/cells.go
agalibin: it has comment-description in interface.go - https://github.com/ydb-platform/nbs/blob/bbb8e992dfd19b503274e0d44ce4e9caa63a7f43/cloud/disk_manager/internal/pkg/cells/interface.go#L57

cloud/disk_manager/internal/pkg/services/disks/service.go
```diff
@@ -377,6 +378,42 @@ func (s *service) isOverlayDiskAllowed(
 	return !s.config.GetDisableOverlayDisks(), nil
 }
 
+func (s *service) getPlacementGroupZoneId(
+	ctx context.Context,
+	req *disk_manager.MigrateDiskRequest,
```
gy2411, line 383: style: need blank line after multi-line function signature

cloud/disk_manager/internal/pkg/cells/cells.go
gy2411: Can we pass id of the cell as zoneID? What will be returned? Let's write about this in the comment.

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -42,16 +44,49 @@ func (t *createPlacementGroupTask) Run(
 	execCtx tasks.ExecutionContext,
 ) error {
 
-	client, err := t.nbsFactory.GetClient(ctx, t.request.ZoneId)
+	var client nbs.Client
+
+	// On retry, check if PG was already persisted so we use its stored cell
+	// instead of running cell selection that may fail with changed config.
+	existingMeta, err := t.storage.GetPlacementGroupMeta(
```
gy2411: I don't understart why we need to to call `GetPlacementGroupMeta` here.

Such code should be ok:
```
if len(t.state.GetSelectedCellId()) > 0 {
    client, err = t.nbsFactory.GetClient(ctx, t.state.GetSelectedCellId())
    ...
} else {
    client, err = t.cellSelector.SelectCellForPlacementGroup(
			ctx,
			t.request.ZoneId,
		)
    ...
    t.state.SelectedCellId = client.ZoneID()
	err = execCtx.SaveState(ctx)
}
```

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -42,16 +44,49 @@ func (t *createPlacementGroupTask) Run(
...
+		)
+
+		if err == nil {
+			t.state.SelectedCellId = client.ZoneID()
+			err = execCtx.SaveState(ctx)
+		}
+
+		if err != nil {
+			return err
+		}
```
gy2411, line 59: Why 2 if's? This looks better:

```
		if err != nil {
			return err
		}

		t.state.SelectedCellId = client.ZoneID()
		err = execCtx.SaveState(ctx) 
```

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -42,16 +44,49 @@ func (t *createPlacementGroupTask) Run(
 	execCtx tasks.ExecutionContext,
 ) error {
 
-	client, err := t.nbsFactory.GetClient(ctx, t.request.ZoneId)
+	var client nbs.Client
```
gy2411, line 47: There is also non-trivial point concerning backwards compatibility.

cloud/disk_manager/internal/pkg/cells/storage/storage_ydb.go
```diff
@@ -177,3 +197,42 @@ func (s *storageYDB) getRecentClusterCapacities(
...
+			FROM cluster_capacity AS t
+			WHERE t.zone_id = $zone_id
+		);
+		SELECT
+			cell_id,
+			SUM(free_bytes) AS free_bytes,
+			SUM(total_bytes) AS total_bytes
+		FROM $recent
+		WHERE row_number = 1
+		GROUP BY cell_id
```
SvartMetal: It would be simpler to select recent capacities for each disk kind and aggregate them by disk kind in simple Go code.

It looks like `scanAggregatedClusterCapacities` would not be needed in this case.

cloud/disk_manager/internal/pkg/cells/cells.go
```diff
@@ -255,6 +277,24 @@ func (s *cellSelector) isAgentAvailable(
 	return true
 }
 
+func (s *cellSelector) resolveCells(zoneID string) ([]string, error) {
+
```
SvartMetal: \n is only needed after multi-line function signature

cloud/disk_manager/internal/pkg/cells/cells.go
```diff
@@ -326,6 +366,63 @@ func (s *cellSelector) selectCellForDisk(
...
+		capacities = slices.DeleteFunc(capacities, func(c storage.ClusterCapacity) bool {
+			return !slices.Contains(cells, c.CellID)
+		})
+
+		if len(capacities) == 0 {
+			logging.Warn(
+				ctx,
+				"no capacities found for zone %v, "+
+					"using first cell in config: %v",
```
SvartMetal, line 441: Maybe we should return an error if this is a configuration mismatch.

Otherwise, we might make the Compute and Disk Manager states inconsistent.

@gy2411 WDYT?

cloud/disk_manager/internal/pkg/services/placementgroup/delete_placement_group_task.go
```diff
@@ -43,11 +46,6 @@ func (t *deletePlacementGroupTask) deletePlacementGroup(
 	execCtx tasks.ExecutionContext,
 ) error {
 
-	client, err := t.nbsFactory.GetClient(ctx, t.request.ZoneId)
-	if err != nil {
-		return err
-	}
```
SvartMetal, line 49: This had a special purpose here: it validated `ZoneId` before passing it to `t.storage.DeletePlacementGroup`

cloud/disk_manager/internal/pkg/services/placementgroup/alter_placement_group_membership_task.go
```diff
@@ -38,7 +44,29 @@ func (t *alterPlacementGroupMembershipTask) Run(
...
+		if !slices.Contains(zoneCells, pgMeta.ZoneID) {
+			return errors.NewNonRetriableErrorf(
+				"placement group %v is in zone %v, not in requested zone %v",
+				t.request.GroupId,
+				pgMeta.ZoneID,
+				t.request.ZoneId,
+			)
+		}
+		zoneID = pgMeta.ZoneID
+	}
```
SvartMetal: Maybe we should extract this into a separate function?

I see the same code copied in several places.

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -90,11 +125,6 @@ func (t *createPlacementGroupTask) Cancel(
 	execCtx tasks.ExecutionContext,
 ) error {
 
-	client, err := t.nbsFactory.GetClient(ctx, t.request.ZoneId)
-	if err != nil {
-		return err
-	}
```
SvartMetal, line 96: Same comment as in https://github.com/ydb-platform/nbs/pull/6262/changes#r3455008415

cloud/disk_manager/pkg/app/controlplane.go
```diff
@@ -537,7 +538,7 @@ func initControlplane(
 	facade.RegisterPlacementGroupService(
 		server,
 		taskScheduler,
-		placementgroup.NewService(taskScheduler, nbsFactory),
+		placementgroup.NewService(taskScheduler, nbsFactory, resourceStorage, cellSelector),
```
SvartMetal: 80 symbols

SvartMetal (changes requested):

cloud/disk_manager/internal/pkg/services/disks/service.go
agalibin: Do we have some auto-formatter for GO?

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
agalibin: I agree, but if already have meta why don't use it?
As result, I will remove this code and will use from task state or from selector.

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
agalibin: ?

cloud/disk_manager/internal/pkg/cells/cells.go
agalibin: old logic, we could change it in separate PR if needed

cloud/disk_manager/pkg/app/controlplane.go
agalibin: It is not problem to change but we have huge amount of strings more than 80 symbols in our GO-code.

cloud/disk_manager/internal/pkg/services/placementgroup/delete_placement_group_task.go
agalibin: Ok, I will return it back to my initial code, but where is a problem if we called 't.storage.DeletePlacementGroup' and exited from function with error? (@gy2411 )

SvartMetal: > if we called 't.storage.DeletePlacementGroup' and exited from function with error?

For example, if ZoneId in the request is invalid, the placement group should not be deleted by its ID, because the whole request seems to be invalid and violates the contract

cloud/disk_manager/pkg/app/controlplane.go
SvartMetal: Let's at least avoid increasing entropy by adding new cases

cloud/disk_manager/internal/pkg/services/placementgroup/delete_placement_group_task.go
agalibin: As I understood "t.storage" is internal for task and will removed together with it, am I wrong?
And anyway I already fixed this code ).

cloud/disk_manager/internal/pkg/cells/cells.go
```diff
@@ -326,6 +364,111 @@ func (s *cellSelector) selectCellForDisk(
...
+	diskKinds := []types.DiskKind{
+		types.DiskKind_DISK_KIND_SSD,
+		types.DiskKind_DISK_KIND_HDD,
+		types.DiskKind_DISK_KIND_SSD_NONREPLICATED,
+		types.DiskKind_DISK_KIND_SSD_MIRROR2,
+		types.DiskKind_DISK_KIND_SSD_LOCAL,
+		types.DiskKind_DISK_KIND_SSD_MIRROR3,
+		types.DiskKind_DISK_KIND_HDD_NONREPLICATED,
+		types.DiskKind_DISK_KIND_HDD_LOCAL,
+	}
```
SvartMetal: Let's introduce a `getAllDiskKinds` function somewhere here:

`cloud/disk_manager/internal/pkg/util`

Alternatively, you can still write a YQL query with `diskKind == *`.

My previous point was not about completely eliminating the new YQL query, but rather about keeping the queries simple and easy to understand.

agalibin: Ok, I will add new function "getAllDiskKinds" and will not add new YQL request.

cloud/disk_manager/internal/pkg/cells/cells.go
```diff
@@ -326,6 +365,102 @@ func (s *cellSelector) selectCellForDisk(
 	}
 }
 
+func (s *cellSelector) getRecentAggregatedClusterCapacities(
+	ctx context.Context,
+	zoneID string,
+) ([]storage.ClusterCapacity, error) {
+
+	diskKinds := util.GetAllDisksKind()
```
SvartMetal: GetAllDiskKinds

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -60,6 +81,7 @@ func (t *createPlacementGroupTask) Run(
 		CreatingAt:    time.Now(),
 		CreatedBy:     "", // TODO: Extract CreatedBy from execCtx
 	})
+
```
SvartMetal: We intentionally keep `if err != nil {` close to the corresponding call site

cloud/disk_manager/internal/pkg/services/placementgroup/service.go
```diff
@@ -46,11 +49,49 @@ func getPlacementStrategy(
 	}
 }
 
+func getClientByGroupSelector(
```
SvartMetal: selectClientByGroup

cloud/disk_manager/internal/pkg/util/proto.go
```diff
@@ -37,3 +38,16 @@ func ParseProto(
 
 	return nil
 }
+
+func GetAllDisksKind() []types.DiskKind {
```
SvartMetal: Let's add a comment stating that this should be kept in sync with https://github.com/ydb-platform/nbs/blob/61f175ab8b6f8373539e918bf089d04a569aec3c/cloud/disk_manager/internal/pkg/types/types.proto#L10

And a corresponding comment to types.proto as well

cloud/disk_manager/internal/pkg/services/disks/service.go
gy2411: Hm, I'm not sure about this.

cloud/disk_manager/internal/pkg/services/placementgroup/delete_placement_group_task.go
gy2411: > This had a special purpose here: it validated `ZoneId` before passing it to `t.storage.DeletePlacementGroup`

Ok, I agree. Thank you for pointing out this!

gy2411: > Ok, I will return it back to my initial code, but where is a problem if we called 't.storage.DeletePlacementGroup' and exited from function with error? (@gy2411 )

Now code looks ok, I don't see any problems anymore.

cloud/disk_manager/internal/pkg/services/disks/service.go
kobzonega: `ya make -t` should format with `gofmt` and run linter `govet`

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
gy2411: There is a small chance that task will fail during release (if it start on the new version and continues on the old versoin).

We discussed it and decided to accept this risk.

cloud/disk_manager/internal/pkg/cells/cells.go
gy2411: > Otherwise, we might make the Compute and Disk Manager states inconsistent.

I'm not sure that i understand this concern well. Compute doesn't know anything about cells. 

Do you mean scenario like this?
1. We misconfigure DM and add a cell from zone-b to cells config of zone-a
2. We get to this branch of code and fall back to the first cell from config
3. Disk is created in zone-b instead of zone-a.

cloud/disk_manager/internal/pkg/services/disks/service.go
agalibin: it is not auto-formatter, I know that we have possibility to call "ya tool go fmt <file>" but it is not convenient

cloud/disk_manager/internal/pkg/services/placementgroup/service.go
```diff
@@ -46,11 +49,49 @@ func getPlacementStrategy(
 	}
 }
 
+func selectClientByGroup(
+	ctx context.Context,
+	storage resources.Storage,
+	nbsFactory nbs.Factory,
+	cellSelector cells.CellSelector,
+	groupId string,
+	reqZoneId string,
```
SvartMetal: Linter can fail because of these `Id`, change the suffix to ID please

groupID, reqZoneID

cloud/disk_manager/internal/pkg/types/types.proto
```diff
@@ -7,6 +7,7 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 ////////////////////////////////////////////////////////////////////////////////
 
 // These values are stored in DB, do not shuffle them around.
+// Keep in sync with "internal/pkg/util/proto.go : GetAllDiskKind"
```
SvartMetal: GetAllDiskKinds

cloud/disk_manager/internal/pkg/util/proto.go
```diff
@@ -37,3 +38,17 @@ func ParseProto(
 
 	return nil
 }
+
+// Keep in sync with "internal/pkg/types/types.proto : enum DiskKind"
+func GetAllDiskKind() []types.DiskKind {
```
SvartMetal: GetAllDiskKinds

cloud/disk_manager/internal/pkg/services/placementgroup/create_placement_group_task.go
```diff
@@ -42,16 +44,35 @@ func (t *createPlacementGroupTask) Run(
...
+	var err error
+
+	if len(t.state.GetSelectedCellId()) > 0 {
+		client, err = t.nbsFactory.GetClient(ctx, t.state.GetSelectedCellId())
+	} else {
+		client, err = t.cellSelector.SelectCellForPlacementGroup(
+			ctx,
+			t.request.ZoneId,
+		)
+
```
SvartMetal: Excessive \n

cloud/disk_manager/internal/pkg/cells/cells.go
SvartMetal: > Do you mean scenario like this?

Yes, or Compute might be misconfigured and use a wrong cell

SvartMetal (approved):

gy2411 (approved):

PR #6265, opened 2026-06-18
[Filestore] TFileRingBuffer: Remove default constructor values

e673:
### Notes

Force users of `TFileRingBuffer` to explicitly specify version instead of relating on the default parameter.

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
@@ -4106,7 +4106,12 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
                     sessionId / "write_back_cache";
 
         {
-            TFileRingBuffer ringBuffer(path, WriteBackCacheCapacity);
+            TFileRingBuffer ringBuffer(
+                path,
+                WriteBackCacheCapacity,
+                0,
+                EFileRingBufferVersion::V6);
+
             UNIT_ASSERT(!ringBuffer.Empty());
         }
 
@@ -4147,7 +4152,12 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
         }
 
         {
-            TFileRingBuffer ringBuffer(path, WriteBackCacheCapacity);
+            TFileRingBuffer ringBuffer(
+                path,
+                WriteBackCacheCapacity,
+                0,
+                EFileRingBufferVersion::V6);
+
             UNIT_ASSERT(ringBuffer.Empty());
         }
 
--- a/cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
+++ b/cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
@@ -9,11 +9,29 @@
 #include <util/system/filemap.h>
 #include <util/system/tempfile.h>
 
-namespace NCloud {
+namespace NCloud::NTesting {
 
 using EVersion = EFileRingBufferVersion;
 
 namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TFileRingBuffer: public NCloud::TFileRingBuffer
+{
+public:
+    TFileRingBuffer(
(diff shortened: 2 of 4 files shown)


SvartMetal (approved):

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
```diff
@@ -9,14 +9,32 @@
 #include <util/system/filemap.h>
 #include <util/system/tempfile.h>
 
-namespace NCloud {
+namespace NCloud::NTesting {
```
e673, line 12: Note: this change is temporary (is needed to reduce diff in this PR)

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -4106,7 +4106,12 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
                     sessionId / "write_back_cache";
 
         {
-            TFileRingBuffer ringBuffer(path, WriteBackCacheCapacity);
+            TFileRingBuffer ringBuffer(
+                path,
+                WriteBackCacheCapacity,
+                0,
+                EFileRingBufferVersion::V6);
```
neihar, line 4113: I can see why do we need this, but definitely don't like this and think that we should not do this

neihar: if you need some new functionality, that shouldn't be enabled by default, it can be under more explicit param name, like "bool featureX". From my perspective, version in current usage it is Umbrella token to hide a bunch of changes under the hood, which are actually not critical and can be disabled

e673: File formats of different versions are mutually incompatible — I want to explicitly control the version number instead of the feature set.

> if you need some new functionality, that shouldn't be enabled by default, it can be under more explicit param name, like "bool featureX".

Do you mean a mapping (requested feature set) -> version number that can support these features?

Also, there may be changes that add no new functionality. For example, V6 adds only stability features (data alignment, checksum calculation). 

> it is Umbrella token to hide a bunch of changes under the hood, which are actually not critical and can be disabled

They cannot be disabled — the capabilities are defined by version.

P.S. This is unrelated to the refactoring in the PR. Let's not dig into unnecessary overengineering.

neihar: I mean, even if it is not possible to disable. From my perspective it is better to have para or tempalte arg to specify which strategy to use for FilrRingBuffer like
`FileRingBuffer"<AlignedDataStrategy>". FileRingBuffer"<DefaultDataStrategy>". `
What I afraid, that with this appraoch we can endup with usage in code like
FileRubgBuffer(ver_X) ... FileRubgBuffer(ver_X_N) wich is clearly not very descriptive and requires more effort to understand what is going on.

neihar: version it is something, that should not last long and eventually needs to be migrated. If it is only temporary stuff and we want controllable migration, than it is ok. maybe add static assert that we can't have more than 2 versions at the same time

e673: > version it is something, that should not last long and eventually needs to be migrated. If it is only temporary stuff and we want controllable migration, than it is ok.

Yes, you are right — versions is a temporary stuff.
Eventually, a migration will happen everywhere, and support for old versions will be removed.

The idea is to make it stable to a release downgrade:

1. Add support for a new version.
2. Wait until the code supporting it is deployed everywhere.
3. Gradually start using the new version — control it manually and explicitly.
4. Wait until the old version is no more used including the previous releases.
5. Remove support for the old version.

> maybe add static assert that we can't have more than 2 versions at the same time

Not needed - there may be more than 2 versions in support.

drbasic (approved):

PR #6292, opened 2026-06-21
Fix HandlePollEvent / StopEndpoint race; resolves issue-6233

### LINKED ISSUE #6233: use after free in RDMA client
Step 1. Each endpoint is registered in epoll with event.data.ptr set to a raw TClientEndpoint* via PtrEventTag().
Step 2. epoll_wait() returns an event whose event.data.ptr points to ep_c.
Step 3. Concurrently, another thread calls StopEndpoint():
```
Poller->Detach(endpoint   // removes fd from epoll
Poller->Release(endpoint)  // drops the shared_ptr; ep_c is destroyed
```
Step 4. The CQ thread dereferences the now-dangling pointer in HandlePollEvent:
```
auto* endpoint = PtrFromTag<TClientEndpoint>(event.data.ptr);
endpoint->HandleCompletionEvents(); 
```

tpashkin:
Reproduction steps
1. Disconnect gets triggered and proceeds to dismantle endpoint
2. Flush times out
3. Completion gets triggered, but CQ stalls for any reason
3. Endpoint gets destroyed by CM
4. CQ continues and dereferences raw endpoint pointer

It's quite hard to reproduce – I actually had to modify handlers to be less restrictive about the state in a poll mode and also insert a lot of manual waiting. Resulting test is somewhat convoluted, but reliably triggers the bug under asan

### DIFF
--- a/cloud/storage/core/libs/rdma/impl/client.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client.cpp
@@ -894,10 +894,6 @@ bool TClientEndpoint::HandleInputRequests() noexcept
         RequestEvent.Clear();
     }
 
-    if (!CheckState(EEndpointState::Connected)) {
-        return false;
-    }
-
     auto requests = InputRequests.DequeueAll();
     if (!requests) {
         return false;
@@ -967,12 +963,6 @@ bool TClientEndpoint::HandleCancelRequests() noexcept
         CancelRequestEvent.Clear();
     }
 
-    if (!CheckState(EEndpointState::Connected) &&
-        !CheckState(EEndpointState::Disconnecting))
-    {
-        return false;
-    }
-
     THashSet<ui64> clientRequestIdToCancel;
     for (auto req: CancelRequests.DequeueAll()) {
         clientRequestIdToCancel.emplace(req.ClientRequestId);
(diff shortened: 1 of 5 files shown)


yegorskii (approved):

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -2068,9 +2068,6 @@ void TClient::StopEndpoint(TClientEndpoint* endpoint) noexcept
 {
     RDMA_INFO(endpoint->Log, "release resources");
     ConnectionPoller->Detach(endpoint);
-    if (endpoint->CompletionChannel) {
```
budevg, line 2071: what if `StopEndpoint` called from "Connecting" state ? `endpoint->Poller->Detach` won't be called at all

budevg (changes requested):

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: It's a valid point, and I will fix this

However, I think it's not a problem *right now*. In the Connecting state no WRs are posted yet and no requests are accepted, cancelled or aborted, so StopEndpoint can't race with HandlePollEvents in the same way as in Disconnected state. Aside from that, closing event handle without detaching it is valid

budevg: ok I am approving the PR let's fix the StopEndpoint in other PR

budevg (approved):

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: https://github.com/ydb-platform/nbs/pull/6309

PR #6027, opened 2026-05-25
[Blockstore] Add new compaction mode by garbage ignoring zeroed blocks

gy2411:
### Notes

We want to add new compaction type to avoid burst of compaction after discards of large ranges. See https://github.com/ydb-platform/nbs/issues/5815 for details. This is the final step for it. The previous step was here: https://github.com/ydb-platform/nbs/pull/5865.

In this pr we:
- Add NewlyZeroedBlocks counter to part state.
- Add the new compaction type: `IgnoringZeroedCompaction` (unter the feature flag)
- Add a new parameter for compaction throttling: `MaxCompactionExecTimePerSecondForZeroed`. Together with `MaxCompactionExecTimePerSecond`, It allows to throttle `GarbageCompaction` more severely then `IgnoringZeroedCompaction`.

### Issue
https://github.com/ydb-platform/nbs/issues/5815

Issue #5815, opened 2026-04-23
Discard of large range should not trigger compaction too aggressively


gy2411:
If a large range on disk is discarded, a large amount of garbage appears at once. This can trigger a burst of compaction by garbage. This creates a risk of performance degradation.

We want to adjust the compaction trigger logic to make these bursts smoother.

gy2411:
This issue is a part of #4823

gy2411:
# Proposed solution

Key idea:

We introduce a new type of compaction. It works similarly to garbage compaction, but it ignores zeroed blocks. When a large discard occurs, it behaves the same way garbage compaction would have behaved before the discard.

If there is nothing left to compact while ignoring zeroed blocks, we then start compacting zeroed blocks as well, but we can do so more gradually by using compaction throttling.

### Introduce a new parameter: `NewlyZeroedBlocks`

This is a lower-bound estimate of the number of zeroed blocks, i.e. blocks that were previously written and later became zeroed due to a `ZeroBlocks` request, while still having garbage underneath them.

The `NewlyZeroedBlocks` counter is maintained per compaction range.

- When previously written blocks are overwritten via `ZeroBlocks`, `NewlyZeroedBlocks` is increased by the number of blocks that were used became zeroed.
- On any `WriteBlocks` request, the counter is decreased by the number of blocks that were previously unused or zeroed and become used again.
- The counter is reset to zero during compaction.

A disk-level `NewlyZeroedBlocks` counter is also introduced as the sum of `NewlyZeroedBlocks` across all ranges.

Note that this is only a lower-bound estimate of the number of zeroed blocks. A precise counter is not easy to maintain, because on writes we cannot distinguish whether a block was previously zeroed or had never been written at all.

Also note that this counter is not persisted, so it will be reset to zero after a tablet restart.

Related parameters:
GarbageIgnoringZeroed := GarbageBlockCount - NewlyZeroedBlocks
UsedBlocksIgnoringZeroed := UsedBlockCount + NewlyZeroedBlocks

Physical meaning: these values estimate the amount of garbage and used blocks, respectively, after adjusting for zeroed blocks — as if we “ignored” the fact that some blocks had been zeroed.

The advantage of these derived metrics is that they change smoothly, whereas the regular garbage and used-block counters may jump sharply when a large number of blocks is zeroed at once.

---

### Introduce a new compaction type

Currently, we have two compaction types:

- `RangeCompaction` — based on the number of blobs
[...]

PR #6315, opened 2026-06-23
[Blockstore] remove vefify on UsedBlocksCount <= BlockCount; add comment; add unit test

gy2411:
Remove the verify previosly added in https://github.com/ydb-platform/nbs/pull/6027. The reason is that UsedBlocksCount might be greater then BlocksCount.

For example, we may write a mixed/merged blob and then add a zero blob covering the same blocks to fresh. After compaction, BlockCount drops but UsedBlockCount stays the same. This mismatch is corrected only on the next flush.

This problem does not seem to be easily solvable. We can not just set/unset used blocks on compaction because WriteBlocks request with greater commit id could set its blocks earlier.

But this problem does not seem to be a blocker. So we remove the verify and add unit test instead. The test ensures that compaction goes normally even if `UsedBlocksCount > BlocksCount`.

#5815

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_ut.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_ut.cpp
@@ -15095,6 +15095,144 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
             UNIT_ASSERT_VALUES_EQUAL(i, response->Record.GetBlocks(i).GetBlobOffset());
         }
     }
+
+    void DoShouldRunCompactionWhenUsedBlocksCountGreaterThanBlockCount(
+        bool ignoringZeroedCompactionEnabled)
+    {
+        auto config = DefaultConfig();
+        config.SetFreshChannelWriteRequestsEnabled(true);
+        config.SetFreshChannelZeroRequestsEnabled(true);
+        config.SetWriteBlobThresholdSSD(1_MB);
+        config.SetCleanupThreshold(1000);
+        config.SetSSDMaxBlobsPerRange(1000);
+        config.SetFlushThreshold(1000_MB);
+        config.SetV1GarbageCompactionEnabled(true);
+        config.SetIgnoringZeroedCompactionEnabled(ignoringZeroedCompactionEnabled);
+        config.SetCompactionGarbageThreshold(999999);
+        config.SetCompactionRangeGarbageThreshold(20);
+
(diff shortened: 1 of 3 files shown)


komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/partition/part_ut.cpp
```diff
@@ -15095,6 +15095,145 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
...
+            auto response = partition.StatPartition();
+            const auto& stats = response->Record.GetStats();
+            UNIT_ASSERT_VALUES_EQUAL(stats.GetMixedBlocksCount(), 0);
+            UNIT_ASSERT_VALUES_EQUAL(stats.GetMergedBlocksCount(), 128 + 512);
+            UNIT_ASSERT_VALUES_EQUAL(stats.GetUsedBlocksCount(), 512 + 512);
+        }
+
+        partition.Flush();
+        runtime->DispatchEvents(TDispatchOptions(), TDuration::Seconds(1));
+        runtime->DispatchEvents(TDispatchOptions(), TDuration::Seconds(1));
```
SvartMetal: Why twice?

cloud/blockstore/libs/storage/partition/part_ut.cpp
```diff
@@ -15095,6 +15095,145 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
...
+
+        {
+            auto response = partition.StatPartition();
+            const auto& stats = response->Record.GetStats();
+            UNIT_ASSERT_VALUES_EQUAL(stats.GetMixedBlocksCount(), 0);
+            UNIT_ASSERT_VALUES_EQUAL(stats.GetMergedBlocksCount(), 128 + 512);
+            UNIT_ASSERT_VALUES_EQUAL(stats.GetUsedBlocksCount(), 128 + 512);
```
SvartMetal: expected vs actual

SvartMetal: And in other places

cloud/blockstore/libs/storage/partition/part_ut.cpp
gy2411: Accidently

SvartMetal (approved):

komarevtsev-d (approved):

PR #6319, opened 2026-06-24
[Disk Registry] Add dirty online devices monitoring

zavik001:
image

Columns

UUID — Unique device identifier. Clickable — opens the device detail page with full hardware info (name, serial number, model, block size).

Agent — Hostname of the disk agent node where the device is located. Shown in red for the "Agent unavailable" table. Empty if the agent is not found in DiskRegistry.

Pool — Storage pool the device belongs to (e.g. SSD, HDD, local). Determined by device.GetPoolName().

State message — Human-readable message attached to the current device state. Typically contains the reason for transitioning into the current state (e.g. error description, replacement reason).

In state since — Timestamp of the last device state change (TDeviceConfig::StateTs). Note: this reflects when the device last changed its state (e.g. became ONLINE or ERROR), not necessarily when it entered the dirty list.

Time in state — Duration elapsed since "In state since". Computed as Now() - StateTs. Gives a quick sense of how long the device has been stuck in its current state.

  Tables

Erasing now — Devices whose pool currently has an active TSecureEraseActor running. All dirty devices of a pool are sent as a single batch, so presence here means the device's erase is in progress.

Ready for cleanup — Devices that passed CanSecureErase but whose pool has not yet been picked up by SecureErase(). Either the trigger has not fired since the last state change, or the device was skipped due to per-pool rate limiting (MaxDevicesToErasePerDeviceName).

Waiting for cleanup — Devices in ONLINE or WARNING state for which no agent was found by NodeId. The node hosting this device is not registered in DiskRegistry. Cleanup cannot proceed until the agent appears.

Agent unavailable — Devices in ONLINE or WARNING state whose agent is registered but in  AGENT_STATE_UNAVAILABLE. The host is temporarily unreachable. Cleanup will resume automatically once the agent recovers.

Cleanup blocked — Devices in ONLINE or WARNING state with a reachable agent, but CanSecureErase returned false for another reason: the device was recently auto-replaced (IsAutomaticallyReplaced), its path is detached (AttachDetachPathsEnabled), or the device is suspended without a pending resume. These require manual investigation.

Device broken — Devices in DEVICE_STATE_ERROR. CanSecureErase explicitly rejects error-state devices, so no erase will be attempted. Requires investigation via the device detail page.

### DIFF
--- a/cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_monitoring.cpp
+++ b/cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_monitoring.cpp
@@ -2252,16 +2252,7 @@ void TDiskRegistryActor::HandleHttpInfo_RenderDirtyOnlineDeviceList(
 void TDiskRegistryActor::RenderDirtyOnlineDeviceList(IOutputStream& out) const
 {
     auto dirtyDevices = State->GetDirtyDevices();
-    size_t onlineCount = 0;
-    for (const auto& device: dirtyDevices) {
-        const auto state = device.GetState();
-        if (state == NProto::DEVICE_STATE_ONLINE ||
-            state == NProto::DEVICE_STATE_WARNING)
-        {
-            ++onlineCount;
-        }
-    }
-    if (onlineCount == 0) {
+    if (dirtyDevices.empty()) {
         return;
     }
     DumpActionLink(
@@ -2269,7 +2260,7 @@ void TDiskRegistryActor::RenderDirtyOnlineDeviceList(IOutputStream& out) const
         TabletID(),
         "RenderDirtyOnlineDeviceList",
         "Dirty online devices",
-        onlineCount);
+        dirtyDevices.size());
 }
 
 void TDiskRegistryActor::RenderDirtyOnlineDeviceListDetailed(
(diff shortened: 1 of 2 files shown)


zavik001:
@codex review

cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_monitoring.cpp
```diff
@@ -2216,6 +2237,168 @@ void TDiskRegistryActor::RenderDirtyDeviceListDetailed(IOutputStream& out) const
...
+    size_t onlineCount = 0;
+    for (const auto& device: dirtyDevices) {
+        const auto state = device.GetState();
+        if (state == NProto::DEVICE_STATE_ONLINE ||
+            state == NProto::DEVICE_STATE_WARNING)
+        {
+            ++onlineCount;
+        }
+    }
+    if (onlineCount == 0) {
```
drbasic: зачем это условие?

zavik001: > зачем это условие?

поправил через State->GetDirtyDevices()

drbasic (approved):

sharpeye (approved):

PR #6320, opened 2026-06-24
[NBS] Update stored endpoint request after successful alter

agalibin:
### Notes
- endpoint_manager.cpp: update endpoint.Request and re-persist to EndpointStorage after successful AlterEndpoint;
- session_manager.cpp: remove const from TEndpoint::StartRequest and update it in Alter() after successful MountVolume;
- new tests.

### Issue
When AlterEndpoint changes VolumeAccessMode, VolumeMountMode, or MountSeqNumber (e.g. during VM's migration), the stored TStartEndpointRequest was never updated. This caused:
- Subsequent StartEndpoint calls with the same (already altered) parameters to trigger redundant alters instead of returning S_ALREADY, because the comparison used the stale original request;
- ListEndpoints to report stale access/mount modes;
- persistent endpoints to restore with stale parameters after restart;
- session switching (SwitchSessionForEndpoint) to create new sessions with stale access/mount modes from the const StartRequest.

For the last reason, we sent an error to the guest after the disk copying was finished, because we tried to mount it with a request that was stale after the migration (RO-Local instead of RW-Local).

### DIFF
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -179,6 +179,15 @@ bool CompareRequests(
     return left.GetUnixSocketPath() == right.GetUnixSocketPath();
 }
 
+void UpdateRequest(
+    NProto::TStartEndpointRequest& dstReq,
+    const NProto::TStartEndpointRequest& srcReq)
+{
+    dstReq.SetVolumeAccessMode(srcReq.GetVolumeAccessMode());
+    dstReq.SetVolumeMountMode(srcReq.GetVolumeMountMode());
+    dstReq.SetMountSeqNumber(srcReq.GetMountSeqNumber());
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
 class TDeviceManager
@@ -913,9 +922,7 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
         }
 
         // update in‑memory cached request parameters for the endpoint
-        endpoint.Request->SetVolumeAccessMode(request->GetVolumeAccessMode());
-        endpoint.Request->SetVolumeMountMode(request->GetVolumeMountMode());
-        endpoint.Request->SetMountSeqNumber(request->GetMountSeqNumber());
+        UpdateRequest(*endpoint.Request, *request);
 
         // Try to persist the updated request parameters for the endpoint.
         // If persistence fails, do not report an error: AlterEndpoint was
@@ -1036,9 +1043,7 @@ NProto::TError TEndpointManager::AlterEndpoint(
             << " has already been started");
     }
 
-    oldReq.SetVolumeAccessMode(newReq.GetVolumeAccessMode());
-    oldReq.SetVolumeMountMode(newReq.GetVolumeMountMode());
-    oldReq.SetMountSeqNumber(newReq.GetMountSeqNumber());
+    UpdateRequest(oldReq, newReq);
 
     if (!CompareRequests(newReq, oldReq)) {
         return MakeError(E_INVALID_STATE, TStringBuilder()
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -912,6 +912,22 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
             return TErrorResponse(error);
         }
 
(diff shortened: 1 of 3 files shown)


cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -912,6 +912,18 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
...
+        endpoint.Request->SetVolumeMountMode(request->GetVolumeMountMode());
+        endpoint.Request->SetMountSeqNumber(request->GetMountSeqNumber());
+
+        if (auto storageError = AddEndpointToStorage(*endpoint.Request);
+            HasError(storageError))
+        {
+            STORAGE_ERROR(
+                "Failed to update endpoint in storage: "
+                << FormatError(storageError));
+        }
```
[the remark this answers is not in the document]
agalibin: It is enough to have message in this place (identical to StopEndpointImpl if RemoveEndpoint failed).

agalibin:
@codex review

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -912,6 +912,31 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
...
+        if (auto updateError = Executor->WaitFor(updateFuture);
+            HasError(updateError))
+        {
+            STORAGE_ERROR(
+                "Failed to update session access params: "
+                << FormatError(updateError));
+        }
+
+        if (auto storageError = AddEndpointToStorage(*endpoint.Request);
+            HasError(storageError))
```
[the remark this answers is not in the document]
agalibin: it is ok in this case

nisqatsi (commented):
Добавил tsan тесты

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -912,6 +912,18 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
             return TErrorResponse(error);
         }
 
+        endpoint.Request->SetVolumeAccessMode(request->GetVolumeAccessMode());
```
drbasic: в 899 строке для RestartListenerEndpoint не нужно заранее поправить все параметры запроса?

drbasic (dismissed):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
agalibin: обсудили: не надо, тут создается эндпоинт со старыми параметрами, а затем делается Alter на новый

nisqatsi (dismissed):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -912,6 +912,22 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
             return TErrorResponse(error);
         }
 
+        // update in‑memory cached request parameters for the endpoint
+        endpoint.Request->SetVolumeAccessMode(request->GetVolumeAccessMode());
```
antonmyagkov: The same logic already exists in AlterEndpoint. Could you refactor the code so it's implemented only once?

https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/endpoints/endpoint_manager.cpp#L1023

antonmyagkov (changes requested):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -912,6 +921,20 @@ NProto::TStartEndpointResponse TEndpointManager::StartEndpointImpl(
             return TErrorResponse(error);
         }
 
+        // update in‑memory cached request parameters for the endpoint
```
antonmyagkov, line 924: Is it possible to call UpdateRequest before AlterEndpoint and remove UpdateRequest call from AlterEndpoint?

antonmyagkov (changes requested):

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
agalibin: No, it is wrong, because Alter can receive error and we don't need to update parameters

antonmyagkov (approved):

nisqatsi (approved):

drbasic (approved):

PR #6274, opened 2026-06-18
[Disk Manager] Enable periodic running and metrics for remote test

jkuradobery:

PR #6321, opened 2026-06-24
Add SkipTests config for Disk Manager remote tests

kobzonega:
### Notes
Add `SkipTests` to the remote test TestConfig so tests can be skipped via config instead of changing the test runner

Changes:
- Add repeated string `SkipTests` to `cloud/disk_manager/test/remote/cmd/config/config.proto`
- In `runTests`, build a set from `testConfig.GetSkipTests()` and skip matching tests by `test.Name`, with a log line for each skipped test

### Issue
Some remote tests assume resources that are not available in every test environment. For example, certain tests require HDD disks, but not all labs provide them. In environments like `hw-nbs-stable-lab`, running the full test suite may fail because the required disk types or other resources are missing.

`SkipTests` lets us exclude incompatible tests per environment via config, without modifying the test runner code.

### DIFF
--- a/cloud/disk_manager/test/remote/cmd/config/config.proto
+++ b/cloud/disk_manager/test/remote/cmd/config/config.proto
@@ -11,4 +11,5 @@ message TestConfig {
     required string ImageURL = 2;
     required uint64 ImageSize = 3;
     required uint32 ImageCrc32 = 4;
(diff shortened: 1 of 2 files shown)


kobzonega:
Implemented `SkippedTests` in https://github.com/ydb-platform/nbs/pull/6274

PR #6325, opened 2026-06-24
[Disk Manager]  Add hanging tasks timeouts by task type

jkuradobery:
### Notes
For ready_to_run tasks, which are enqueued, we won't notice them being hanging until they are executed by the runner. If we want to enable estimates, we will have to set hanging tasks timeout to (max disk size / estimated transfer bandwidth), which would cap at days. 
This would prevent us from seing tasks expected to be executed quickly as hanging.

### DIFF
--- a/cloud/tasks/config/config.proto
+++ b/cloud/tasks/config/config.proto
@@ -82,4 +82,7 @@ message TasksConfig {
 
     // Overrides MaxRetriableErrorCount.
     map<string, uint64> MaxRetriableErrorCountByTaskType = 40;
+
+    // Overrides HangingTaskTimeout by task type.
+    map<string, string> HangingTasksTimeoutByType = 41;
 }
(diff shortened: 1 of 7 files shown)


gy2411 (dismissed):

cloud/tasks/config/config.proto
```diff
@@ -82,4 +82,7 @@ message TasksConfig {
 
     // Overrides MaxRetriableErrorCount.
     map<string, uint64> MaxRetriableErrorCountByTaskType = 40;
+
+    // Overrides HangingTaskTimeout by task type.
+    map<string, string> HangingTasksTimeoutByType = 41;
```
SvartMetal: HangingTaskTimeoutByType

for consitency

cloud/tasks/storage/common.go
```diff
@@ -46,6 +46,30 @@ func strListValue(strings []string) persistence.Value {
 	return result
 }
 
+func durationByTaskTypeListValue(
+	valuesByTaskType map[string]time.Duration,
```
SvartMetal: valueByTaskType

SvartMetal (changes requested):

SvartMetal (approved):

gy2411 (approved):

PR #6272, opened 2026-06-18
[Filestore] Throttle number of in-flight shard management ops

Maximus5:
### Notes
Closes #6310

When huge number of shard requests issued simultaneously we faced a problem: Tablet OOMed while resizing FS from 1,000 to 25,000 shards.

This PR implements throttling of create/resize/destroy shard requests ensuring we don't have more than `MaxShardManagementRequestsInFlight` in-flight operations. New behavior is under feature flag `ShardManagementRequestThrottlingEnabled`.

### Testing
```
./ya test --build=profile -- cloud/filestore/libs/storage/service/ut

Total 1 suite:
        1 - GOOD
Total 227 tests:
        227 - GOOD
Ok
```

PR #6335, opened 2026-06-25
[Filestore] Fix Alter requests throttling

### LINKED ISSUE #6310: [Filestore] Send requests to SchemeShard in batches for CreateFileStore, ResizeFileStore, and DestroyFileStore
We faced a problem with SchemeShard being overloaded by Describe requests because ResizeFileStore from 1000 to 25000 shards sends all requests to SchemeShard simultaneously. We observed high CPU utilization, OOMs, and had to manually restart tablets.

While sending shard requests during FS Create/Resize/Destroy we need to cap max number of requests in-the-flight.

## MaxShardManagementRequestsInFlight

Implement new storage config `MaxShardManagementRequestsInFlight`, when it's zero (default) we keep default behavior, no limitation.

When `MaxShardManagementRequestsInFlight > 0` we cap max number of in-flight requests. When we receive a response, we emit a new request.

Maximus5:
### Notes
Closes #6310

Main functionality implemented in #6272.

Add missing AlterShards throttling.

### Testing
```
./ya test -F '*TStorageServiceShardingTest*' -- cloud/filestore/libs/storage/service/
...
Total 1 suite:
        1 - GOOD
Total 90 tests:
        90 - GOOD
Ok
```

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
@@ -141,6 +141,10 @@ NProtoPrivate::TUnsafeChangeTabletStateResponse SetCompressNodeRef(
     return response;
 }
 
+// Helper class to count Describe/Create/Alter/Destroy/Configure
+// shard requests. We intentionally skip Main FS requests and requests
+// made by proxies. That's because we want to process only original shard
+// requests to check max in-flight and request/response counters.
 class TShardRequestCounter
 {
 private:
@@ -203,6 +207,7 @@ private:
             }
 
             case TEvIndexTablet::EvConfigureAsShardRequest: {
+                // Skipping calls made by Proxies.
                 if (ev->Recipient != MakeIndexTabletProxyServiceId()) {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


cloud/filestore/libs/storage/service/service_ut_sharding.cpp
```diff
@@ -234,11 +245,23 @@ class TShardRequestCounter
             }
 
             case TEvSSProxy::EvDescribeFileStoreResponse: {
+                if (ev->Cookie == Max<ui64>()) {
```
SvartMetal, line 262: Why is it possible?

Maximus5: It's a MainFileStoreCookie called from TAlterFileStoreActor::DescribeMainFileStore.
https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/libs/storage/service/service_actor_alterfs.cpp#L266

I don't want to count this single operation as throttling should apply only on `DescribeShards` op.

SvartMetal: Can you add a comment explaining that this is the main Filestore cookie?

SvartMetal (approved):

yegorskii (approved):

PR #6356, opened 2026-06-26
[Disk Manager] Handle res.Err() in scanFilesystemSnapshotStates

jkuradobery:
### Notes
We use CheckFilesystemSnapshotReady in snapshot transfer tasks (snapshot <-> fs). 
And when we cancel the context within tests, it returns nil state, which leads to silent non-retriable error, which fails tests.

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/common.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/common.go
@@ -134,7 +134,7 @@ func scanFilesystemSnapshotStates(ctx context.Context, res persistence.Result) (
 		}
 	}
 
-	// NOTE: always check stream query result after iteration.
+	// NOTE: always check query result after iteration.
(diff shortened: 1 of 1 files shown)


cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/common.go
```diff
@@ -134,6 +134,11 @@ func scanFilesystemSnapshotStates(ctx context.Context, res persistence.Result) (
 		}
 	}
 
+	// NOTE: always check stream query result after iteration.
```
SvartMetal: I don't see that it is a stream query result https://github.com/search?q=repo%3Aydb-platform%2Fnbs+scanFilesystemSnapshotStates&type=code

Could you please point it out?

SvartMetal (approved):

yudov-maksim (approved):

kobzonega (approved):

PR #6364, opened 2026-06-29
[NBS] Added decrementing WriteOrZeroRequestsInProgress on fresh hard limit exceeded; Added processing drain requests on zero blocks completed;

vladstepanyuk:
### Notes
Stuck Drain requests casing stuck for the CreateCheckpoint requests for multipartition disks.
In the same time wile checkpoint is created, volume will reject all write requests, so user request will stuck too.

Now we have some cases in which drain requests can stuck and trigger max time alert:
1. if we exceed fresh bytes count hard limit
2. if last request  was fresh zero request and Fresh Blocks Writer enabled
  
in first case we increment requests counter before calling WriteFreshBlock
https://github.com/ydb-platform/nbs/blob/25ac27219d5a060a37c132bf690de1799f360dab/cloud/blockstore/libs/storage/partition/part_actor_writeblocks.cpp#L270
and if fresh bytes count is exceed we will reject request but will not decrement counter
https://github.com/ydb-platform/nbs/blob/7087fbf57e80ba077fc34168074d451ca828ed1d/cloud/blockstore/libs/storage/partition/part_actor_writefreshblocks.cpp#L43-L67
so next checkpoint creation will stuck forever.

In second case we just not calling ProcessDrainRequests after zero blocks request ended.

### DIFF
--- a/cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
+++ b/cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
@@ -631,8 +631,7 @@ Y_UNIT_TEST_SUITE(TFreshBlocksWriterTest)
                     UNIT_ASSERT(addFreshBlocksRequestObserved);
                 }
 
-                return event->GetTypeRewrite() ==
-                       TEvPartitionCommonPrivate::EvTrimFreshLogRequest;
+                return false;
             });
 
         auto partition = testEnv.GetPartitionClient();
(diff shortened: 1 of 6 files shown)


komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
```diff
@@ -662,8 +661,7 @@ Y_UNIT_TEST_SUITE(TFreshBlocksWriterTest)
 
                     return false;
```
qkrorlqr: now this is redundant

cloud/blockstore/libs/storage/fresh_blocks_writer/fresh_blocks_writer_actor_ut.cpp
qkrorlqr: this comment is out of date now

qkrorlqr (commented):
LGTM

qkrorlqr (approved):

komarevtsev-d (approved):

Issue #1541, opened 2024-07-03
[Filestore] non-blocking open/close


qkrorlqr:
open (CreateHandle) and close (DestroyHandle) do the following expensive things (taking DestroyHandle as an example):
* https://github.com/ydb-platform/nbs/blob/bc2f8e0966f19b88d9c3eb8fbc940d3a4d9b0871/cloud/filestore/libs/storage/tablet/tablet_actor_destroyhandle.cpp#L97 - deleting the handle from the session structure
* https://github.com/ydb-platform/nbs/blob/bc2f8e0966f19b88d9c3eb8fbc940d3a4d9b0871/cloud/filestore/libs/storage/tablet/tablet_actor_destroyhandle.cpp#L107 - deleting the inode (if this handle was holding the last ref to this inode)

Both operations make modifications to the index stored persistently in blobstorage. There is a way to actually make these modifications in background and ack the corresponding syscall to the guest quickly. The open call is a bit more complex - I won't describe the solution for it right now (but I have some ideas), but the close call seems to be simple to implement: we can simply ack the call straight away and issue the cleanup (handle deletion and possibly node deletion) in the background. If this process fails, we should retry it from the client. If the client dies, client session dies together with the client after a timeout of inactivity and the cleanup happens automatically.

qkrorlqr:
Currently we use the client's virtio queue as our redo log. If we ack the request before its actual completion, we need another redo log. We can create a log file on top of tmpfs local to filestore-vhost. open/close request processing may look like this:
1. Send CreateHandleStage1/DestroyHandleStage1 to the tablet which would perform basic checks over the tablet's inmemory state and return a preliminary result (including HandleId for CreateHandle)
2. If the preliminary result is OK, we will write this result to the redo log
3. We respond to the client
4. We send CreateHandleStage2/DestroyHandleStage2 requests which wait for the actual completion of the operation and retry those requests upon errors 
5. We should delete the corresponding op from the redo log upon Stage2 success

If filestore-vhost restarts, it should reread its redo log to continue retrying the operations. The handles are local to the client so no other client can interfere with them and cause nonretriable errors between Stage1 and Stage2.

qkrorlqr:
[...]

PR #6375, opened 2026-06-29
issue-1541: process only read-only handles asynchronously during close

WilyTiger:
### Notes
We want to enable async destroy only for files opened read-only, as a safer option

### Issue
#1541

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_impl.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_impl.cpp
@@ -99,14 +99,17 @@ void TFileSystem::Reset()
 
 void TFileSystem::ScheduleProcessHandleOpsQueue()
 {
-    if (Config->GetAsyncDestroyHandleEnabled()) {
+    if (Config->GetAsyncDestroyHandleEnabled() ||
+        Config->GetAsyncDestroyReadOnlyHandleEnabled())
+    {
         Scheduler->Schedule(
-        Timer->Now() + Config->GetAsyncHandleOperationPeriod(),
-        [=, ptr = weak_from_this()] () {
-            if (auto self = ptr.lock()) {
-                self->ProcessHandleOpsQueue();
-            }
-        });
+            Timer->Now() + Config->GetAsyncHandleOperationPeriod(),
+            [=, ptr = weak_from_this()]()
+            {
+                if (auto self = ptr.lock()) {
+                    self->ProcessHandleOpsQueue();
+                }
+            });
     }
 }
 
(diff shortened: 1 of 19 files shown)


cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
SvartMetal, line 351: It looks like there is no need to release a read-only handle through the write-back cache.

@e673 what do you think?

Can be fixed in a separate PR though

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -99,14 +99,17 @@ void TFileSystem::Reset()
 
 void TFileSystem::ScheduleProcessHandleOpsQueue()
 {
-    if (Config->GetAsyncDestroyHandleEnabled()) {
+    if (Config->GetAsyncDestroyHandleEnabled() ||
+        Config->GetAsyncDestroyReadOnlyHandleEnabled())
+    {
```
SvartMetal: Maybe it would be better to use this `if` instead?

```
if (HandleOpsQueue) {
```

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
e673: If there are no cached WriteData requests associated with the handle (this is true for read-only handles), `TWriteBackCache::ReleaseHandle` will be completed immediately so this fix will only affect metrics.

However, I'm for applying this fix — it will make intentions more clear.

SvartMetal (dismissed):

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -329,6 +330,9 @@ void TFileSystem::Release(
     fuse_file_info* fi)
 {
     auto handle = fi->fh;
+    const bool readOnly = (fi->flags & O_ACCMODE) == O_RDONLY;
```
qkrorlqr, line 336: let's sync this code with this: https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp#L525 - to have a single way of checking whether the user's opening the file in read-only mode

qkrorlqr (dismissed):

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
qkrorlqr: actually the check in the tablet code is just more strict but it's okay to have a weaker check here because it's about closing the file, not opening it, so truncate and create flags are irrelevant

agalibin (approved):

SvartMetal (approved):

PR #6376, opened 2026-06-29
[Filestore] [1/N] Document the filestore resize flow in legacy and throttled modes

### LINKED ISSUE #6310: [Filestore] Send requests to SchemeShard in batches for CreateFileStore, ResizeFileStore, and DestroyFileStore

Maximus5:
### Notes
* Main issue: #6310
* Implementation: #7256

The document describes the main resize state machine, shard creation throttling, persistent shard creation state, CAS-based bitmap update handling, retry behavior, and the reason why shard creation state is retained after a successful resize. This should make the upcoming implementation PRs easier to review and preserve the intended invariants around retry-safe shard creation.

---

### Stack

0. #6376
1. #7247
2. #7249
3. #7253
4. #7255
5. #7256


cloud/filestore/libs/storage/service/helpers.h
```diff
@@ -19,4 +21,26 @@ void Convert(
     const NKikimrFileStore::TConfig& config,
     NProto::TFileStore& fileStore);
 
+class TInplaceBitmap
```
qkrorlqr, line 24: why not use something from https://github.com/ydb-platform/nbs/blob/main/util/generic/bitmap.h ?

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -358,7 +376,81 @@ void TAlterFileStoreActor::HandleDescribeFileStoreResponse(
...
+
+void TAlterFileStoreActor::UpdateShardCreatedState(
+    const TActorContext& ctx,
+    const ui32 shardIndex)
+{
+    TMutableInplaceBitmap(*FileSystemResizeState.MutableCreatedShardBitmap())
+        .Set(shardIndex);
+
+    auto request =
+        std::make_unique<TEvIndexTablet::TEvUnsafeChangeTabletStateRequest>();
```
qkrorlqr, line 412: we probably need some kind of a `ChangeStateCookie` in this request to avoid ChangeState-races for concurrent requests from multiple filestore-servers

e.g. use `ChangeStateCookie = hash(PrevState)` to be able to check in the tablet code that the caller actually knows the most recent state

qkrorlqr (commented):
are you planning to implement this logic separately for `CreateFileSystem` requests?

cloud/filestore/libs/storage/service/helpers.cpp
```diff
@@ -24,4 +25,48 @@ void Convert(
     fileStore.SetConfigVersion(config.GetVersion());
 }
 
+////////////////////////////////////////////////////////////////////////////////
+
+TInplaceBitmap::TInplaceBitmap(const TString& data)
```
SvartMetal, line 30: Let's use TCompressedBitmap?

For examle we use it for used blocks map https://github.com/ydb-platform/nbs/blob/1a36de1d909dd3d7f7957a1b6a75cb124fbd8ab9/cloud/blockstore/libs/storage/partition/part_schema.h#L351

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -724,18 +816,38 @@ void TAlterFileStoreActor::CreateShards(const TActorContext& ctx)
...
+                Y_DEBUG_ABORT_UNLESS(ShardsToCreate);
+                --ShardsToCreate;
+                continue;
+            }
+        }
+
+        CreateShard(ctx, NextShardToCreate);
+        ++NextShardToCreate;
+        ++requests;
     }
```
SvartMetal, line 851: Maybe it would be less error-prone to keep this code here as well:

```cpp
if (ShardsToCreate == 0) {
    ConfigureShards(ctx);
    return;
}
```

The caller code could also be simplified this way.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
SvartMetal, line 1: Can we add early `Resize`/`Alter` retries to our medium tests?

For example, we can add them to this test, which we use for feature adoption: https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/tests/scheme_cache/test.py

It would also be good to run two `filestore-server` instances and send `Resize` retries to them in a round-robin manner.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
Maximus5: With ResizeState.Version and updated Merge/Verify/Retry logic in `TAlterFileStoreActor::HandleResizeStateResponse` protect us from races.

1. Actors A and B both read ResizeState.Version = 1.
2. A creates shard X and sends bitmap {X} with Version = 1.
3. B creates shard Y and sends bitmap {Y} with Version = 1.
4. The tablet handles e.g. A first, stores bitmap {X}, and bumps Version to 2.
5. The tablet handles B next. B's Version = 1 is stale, so the tablet does not
   apply bitmap {Y}. The response contains the current persisted state:
   Version = 2, bitmap {X}.
6. B merges current persisted state with its own local state, and if this merged state differs from persisted state, does retry to store bitmap {Y'}

Maximus5: Anyway. When two concurrent resize requests were started, both actors have the same initial topology, and they both tries to create shard{curSize+1}. Where only one actor wins, another got StatusAlreadyExists error and stop processing.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -724,18 +816,38 @@ void TAlterFileStoreActor::CreateShards(const TActorContext& ctx)
...
+    {
+        // When MaxShardManagementRequestsInFlight is Zero we do not use
+        // persistent storage for resize state and do not limit in-flight
+        // shard requests.
+        if (limit != 0) {
+            if (IsShardCreated(NextShardToCreate)) {
+                ++NextShardToCreate;
+                Y_DEBUG_ABORT_UNLESS(ShardsToCreate);
+                --ShardsToCreate;
+                continue;
```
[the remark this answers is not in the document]
Maximus5: Now `ConfigureShards` is called from `TAlterFileStoreActor::CreateShards` after `ContinuteCreateShards`. So if `ShardsToCreate` was decremented to 0 without any create calls - we are safe now, process is advanced.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
Maximus5: I would keep single responsibility for `ContinueCreateShards` method.

We want to run `ConfigureShards` only if no create calls were initiated by this actor. I.e. if there are in-flight requests to update bitmap, we want first to wait until it's stored durably.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -789,14 +901,22 @@ void TAlterFileStoreActor::HandleCreateFileStoreResponse(
...
+    --ShardsToCreate;
+
+    if (StorageConfig->GetMaxShardManagementRequestsInFlight()) {
+        UpdateShardCreatedState(ctx, ev->Cookie);
+
+        if (ShardsToCreate > 0 &&
+            NextShardToCreate < FileStoreConfig.ShardConfigs.size())
+        {
+            // Set limit to 1 as only one response was received now.
+            ContinueCreateShards(ctx, 1);
```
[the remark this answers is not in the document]
Maximus5: We are safe now. Even if bitmap stored by last call is not synced with our current state, we merge and recheck the bitmap, and if there is a diff, we initiate `UpdateResizeState` once more.
Ref: `TAlterFileStoreActor::HandleResizeStateResponse`.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -358,7 +376,81 @@ void TAlterFileStoreActor::HandleDescribeFileStoreResponse(
...
+    }
+
+    if (!msg->Record.HasResizeState()) {
+        ReplyAndDie(
+            ctx,
+            MakeError(
+                E_INVALID_STATE,
+                TStringBuilder()
+                    << "Missing resize state: cookie=" << ev->Cookie));
+        return;
```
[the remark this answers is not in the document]
Maximus5: If older FS versions responds with no ResizeState, we have nullptr in `CreatedShardBitmap` and continure with `GetStorageStats`.
Ref: `TAlterFileStoreActor::HandleResizeStateResponse`

cloud/filestore/libs/storage/service/helpers.h
Maximus5: Switched to TCompressedBitmapData

cloud/filestore/libs/storage/service/helpers.cpp
Maximus5: Done

Maximus5:
> are you planning to implement this logic separately for `CreateFileSystem` requests?

I want to keep this PR explicitly for `ResizeFileStore` request. And explore possibilities for `Create` next.

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -1182,6 +1434,9 @@ STFUNC(TAlterFileStoreActor::ResizeStateWork)
         HFunc(
             TEvIndexTablet::TEvConfigureShardsResponse,
             HandleConfigureMainFileStoreResponse);
+        HFunc(
+            TEvIndexTablet::TEvUnsafeChangeTabletStateResponse,
```
debnatkh, line 1357: Originally the Unsafe* set of operations was intended for manual interventions into the tablet state. Routing ordinary requests through it is a bit unusual

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -375,7 +405,183 @@ void TAlterFileStoreActor::HandleDescribeFileStoreResponse(
...
+{
+    Y_DEBUG_ABORT_UNLESS(CreatedShardBitmap);
+
+    for (const auto& chunk: bitmap.GetChunks()) {
+        CreatedShardBitmap->Merge(
+            {.ChunkIdx = chunk.GetChunkIdx(), .Data = chunk.GetData()});
+    }
+}
+
+bool TAlterFileStoreActor::HasUnpersistedCreatedShards() const
```
debnatkh: This is identical code across two actors. Probably worth unifying into a common method

cloud/filestore/private/api/protos/tablet.proto
```diff
@@ -1088,6 +1088,31 @@ message TUnsafeCreateHandleResponse
     NCloud.NProto.TError Error = 1;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// FileSystem Resize State.
+
+message TCompressedBitmapChunk
+{
+    uint32 ChunkIdx = 1;
```
debnatkh, line 1096: I might be missing something, but what is the value of chunking here as opposed to just single `bytes` or `repeated bool`?

cloud/filestore/libs/storage/service/service_actor_createfs.cpp
```diff
@@ -378,6 +541,85 @@ void TCreateFileStoreActor::HandleCreateFileStoreResponse(
...
+{
+    auto* msg = ev->Get();
+    if (HasError(msg->GetError())) {
+        LOG_WARN(
+            ctx,
+            TFileStoreComponents::SERVICE,
+            "[%s] UnsafeChangeTabletState failed: %s",
+            LogTag.c_str(),
+            FormatError(msg->GetError()).Quote().c_str());
+        return;
```
debnatkh, line 460: 1. We can get a retriable error in case of an ongoing tablet restart. It is probably worth retrying retriable errors (see `GetErrorKind` function)

2. If we reach this place, won't this actor just leak even without replying to the caller? 

3. It is both reasonable to eeither propagate non-retiriable error to the caller, or raise a critical event if this is supposed to be an Impossible event

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
```diff
@@ -375,7 +405,183 @@ void TAlterFileStoreActor::HandleDescribeFileStoreResponse(
...
+        LOG_WARN(
+            ctx,
+            TFileStoreComponents::SERVICE,
+            "[%s] FS topology not yet read, shard creation state unavailable",
+            FileSystemId.c_str());
+        return;
+    }
+
+    if (shardCreationState.GetVersion() < ShardCreationStateVersion) {
+        return;
```
debnatkh, line 483: Will this line also lead to an actor leak?

cloud/filestore/libs/storage/service/service_actor_alterfs.cpp
Maximus5: Decided to use `Usafe` after a talk with @qkrorlqr. Reasoning - reuse already existing API to avoid reserving new operation IDs just to get/put state of the tablet.
Probably, we may rename this operation to just `ChangeTabletState`.

I'm not sure if this could cause any incompatibilities during rolling updates?

PR #6388, opened 2026-06-30
fix: not included paths to ya make build graph

librarian:
PoC of script that finds missed ya.make files that are not reachable from root (technically from cloud/* dirs).

This also adds missed dirs with ones that are missed intentionally.

### DIFF
--- a/cloud/storage/core/libs/ya.make
+++ b/cloud/storage/core/libs/ya.make
@@ -9,6 +9,7 @@ RECURSE(
     diagnostics
     endpoints
     features
+    file_backed_containers
     grpc
     hive_proxy
     http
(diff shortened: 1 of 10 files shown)


kobzonega (dismissed):

kobzonega (approved):

cloud/storage/core/libs/ya.make
```diff
@@ -9,6 +9,7 @@ RECURSE(
     diagnostics
     endpoints
     features
+    file_backed_containers
```
SvartMetal, line 12: @e673

e673: Yes, it should be added here.

SvartMetal (approved):

Issue #6385, opened 2026-06-30
[Filestore] If file size is decreased and then immediately increased during FlushBytes then old bytes are still readable until FlushBytes completes


qkrorlqr:
Affects the unaligned writes layer (aka "fresh bytes").
 
1. `FlushBytes` starts - current bytes are sealed and a new bytes chunk is created: https://github.com/ydb-platform/nbs/blob/0c354d65b3155feb9729519a9a4661887525fdc5/cloud/filestore/libs/storage/tablet/model/fresh_bytes.cpp#L160
2. File size is decreased - e.g. via `SetNodeAttr` op, fresh bytes deletion marker gets added into the new chunk: https://github.com/ydb-platform/nbs/blob/0c354d65b3155feb9729519a9a4661887525fdc5/cloud/filestore/libs/storage/tablet/model/fresh_bytes.cpp#L137
3. File size is increased back to the original size
4. `ReadData` arrives here https://github.com/ydb-platform/nbs/blob/0c354d65b3155feb9729519a9a4661887525fdc5/cloud/filestore/libs/storage/tablet/model/fresh_bytes.cpp#L274 and reads the old bytes that were supposed to be erased on step 2 **<--- BUG**
5. `FlushBytes` completes - after this step those old bytes disappear because old bytes get trimmed and they're also not visible in the blobs added by `FlushBytes` (due to block-level deletion markers) **<--- THIS IS CORRECT**

So the bug is in not applying fresh bytes deletion markers to the bytes in sealed fresh bytes chunks. The logic here was that those bytes would get trimmed eventually and the blob wouldn't have those bytes as well because of block-level deletion markers (applied in `AddBlob`). But there's this time window between the start of the `FlushBytes` op and its completion and if an explicit deletion (not overwriting, but deletion - e.g. via `SetNodeAttr` to smaller size) happens in this window those bytes will remain visible after this operation completes and until `FlushBytes` (its `TrimBytes` stage) completes

PR #6389, opened 2026-06-30
issue-6385: fix for truncated bytes visibility bug upon FlushBytes vs Truncate race

qkrorlqr:
### Notes
A detailed description of the problematic scenario can be found in the linked issue.

Next `TFreshBytes` chunk's deleted ranges should be applied when the previous chunk is read. Implementing this via:
* adding a `DeletedRanges` map to each chunk where for each byte range we track the first `CommitId` at which it was deleted
* applying `DeletedRanges` of the next chunk to the ranges read from the previous chunk upon `FindBytes`
* added unit tests both at the `TFreshBytes` level and at the tablet level
* added many comments describing the behaviour and intent of `TFreshBytes` components

Most of the PR is unit tests + comments + cleanup. The core part of the fix is localized in `fresh_bytes.h/cpp` and consists of 3 things:
* implemented `TDeletedRangeMap` and added it to each `TFreshBytes` chunk
* adding each deleted range into this map upon `AddDeletionMarker()`
* looking into `DeletedRanges` of each of the chunks that go after the chunk that we scan upon `FindBytes()`

Will add more comments in a separate PR:
* will document `TFreshBytes` public methods
* will add some comments for block/byte visitors describing the guarantees they rely on

### Issue
https://github.com/ydb-platform/nbs/issues/6385

### DIFF
--- a/cloud/filestore/libs/storage/tablet/model/fresh_bytes.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/fresh_bytes.cpp
@@ -327,7 +327,7 @@ void TFreshBytes::AddDeletionMarker(
 
     //
     // We need to mark this range as deleted to make the corresponding range
-    // invisible in the previous chunk if read at this commitId or higher.
+    // invisible in the previous chunks if read at this commitId or higher.
     //
 
     c.DeletedRanges.AddRange(nodeId, offset, len, commitId);
@@ -460,8 +460,7 @@ void TFreshBytes::FindBytes(
     ui64 commitId) const
 {
     for (ui64 i = 0; i < Chunks.size(); ++i) {
-        const auto& c = Chunks[i];
-        if (c.FirstCommitId > commitId) {
+        if (Chunks[i].FirstCommitId > commitId) {
             //
             // Chunk CommitId ranges do not overlap and are ordered by
             // FirstCommitId so all the remaining ranges are not visible.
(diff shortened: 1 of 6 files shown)


cloud/filestore/libs/storage/tablet/model/fresh_bytes.cpp
```diff
@@ -262,17 +459,28 @@ void TFreshBytes::FindBytes(
...
+
             break;
         }
 
-        FindBytes(c, visitor, nodeId, byteRange, commitId);
+        const TChunk* nextC = nullptr;
+        if (i + 1 < Chunks.size() && Chunks[i + 1].FirstCommitId <= commitId) {
+            nextC = &Chunks[i + 1];
+        }
+        FindBytes(c, nextC, visitor, nodeId, byteRange, commitId);
```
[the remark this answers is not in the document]
qkrorlqr: Yes, I thought about it. I'll do it, it will definitely make the semantics of `TFreshBytes::FindBytes()` clearer

qkrorlqr:
@codex review

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
```diff
@@ -107,6 +202,9 @@ class TFreshBytes
     void UpdateLogTag(TString logTag)
     {
         LogTag = std::move(logTag);
+        for (auto& c: Chunks) {
+            c.DeletedRanges.UpdateLogTag(LogTag);
```
neihar, line 206: will it update only existing chunks? Is it possible to have cases with empty log tag?

neihar (dismissed):

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
```diff
@@ -138,16 +236,29 @@ class TFreshBytes
     bool Intersects(ui64 nodeId, TByteRange byteRange) const;
```
neihar, line 236: should we add test for this function?

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
qkrorlqr: I don't understand the question. It will update all chunks in this data structure. LogTag should not be empty. If it's empty, then upon `TABLET_VERIFY` we won't see fs id.

yegorskii (dismissed):

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
qkrorlqr: this is actually a good catch

this function is supposed to find all ranges - even those covered by deleted ranges of the next chunks

it's used to prevent rebasing block commit ids past fresh bytes commit ids (which is needed to simplify the logic - we don't need to rebase fresh bytes alongside blocks)

I'll change this function to ignore `DeletedRangeMap` added in this PR and will write a test that verifies this behaviour

qkrorlqr: this could've been a bug for checkpoint reads - we could've made some fresh bytes invisible in a checkpoint in which they should be visible by rebasing an older block past a newer fresh bytes range within the commit id range of this checkpoint

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
neihar: I mean, if I understand correctly, we run UpdateLogTag only once and we can add new chunks without LogTag

qkrorlqr: aaah, yes
unfortunately it's hard to test this because `LogTag` affects only `TABLET_VERIFY`s in `TFreshBytes`

yes, correct, will fix

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
qkrorlqr: fixed

cloud/filestore/libs/storage/tablet/model/fresh_bytes.h
qkrorlqr: fixed

qkrorlqr:
@codex review

yegorskii (approved):

debnatkh (approved):

PR #6393, opened 2026-07-01
[Filestore] TFileRingBuffer: multi-version unit tests

e673:
### Notes
Since TFileRingBuffer now supports multiple versions, it is needed to unit-test each version

### DIFF
--- a/cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
+++ b/cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
@@ -9,7 +9,7 @@
 #include <util/system/filemap.h>
 #include <util/system/tempfile.h>
 
-namespace NCloud::NTesting {
+namespace NCloud {
 
 using EVersion = EFileRingBufferVersion;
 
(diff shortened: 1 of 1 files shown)


SvartMetal (dismissed):

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
```diff
@@ -365,16 +357,25 @@ Y_UNIT_TEST_SUITE(TFileRingBufferTest)
         char* data = static_cast<char*>(m.Ptr());
         data[260] = 'A';
 
-        UNIT_ASSERT_VALUES_EQUAL(
-            "data=vasya ecsum=3387363649 csum=3387363646",
-            Dump(rb.Validate()));
+        if (ver >= EVersion::V6) {
```
neihar, line 360: I like the idea with macro for tests, but in scenario like this, looks like better to have individual tests macro
FILE_RING_BUFFER_TEST_V4
FILE_RING_BUFFER_TEST_V5
to keep different behaviour in different tests and not with IFs

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
```diff
@@ -9,29 +9,30 @@
...
+        TestImpl##name(EVersion::V5);                                          \
+    }                                                                          \
+    Y_UNIT_TEST(name##V6)                                                      \
+    {                                                                          \
+        TestImpl##name(EVersion::V6);                                          \
+    }                                                                          \
+    void TestImpl##name(EVersion ver)                                          \
+// FILE_RING_BUFFER_TEST
 
 ////////////////////////////////////////////////////////////////////////////////
```
neihar, line 37: Also, probably make sense somehow visually separate CommonTests and version specific for better readability and future removal/migration

neihar (approved):

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
e673: Not needed - in the future PRs, I will use an accessor that will correct the checksum.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
e673: Method reordering will bloat the diff.
Also I hope that all tests will become mult-version after adding an accessor.

yegorskii (approved):

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
neihar: maybe we can reorder in separate PR for future readability

Issue #6279, opened 2026-06-19
[NBS] Get rid of reading blobs index on cleanup


vladstepanyuk:
Top transactions by ExecuteCPUTime on host with eternal test (ReadBlockMaskOnCompactionOptimizationEnabled: true, so compaction trnsactions doesn't take so much CPU)

<img width="1788" height="539" alt="Image" src="https://github.com/user-attachments/assets/45aeb867-1383-4032-96cc-cabd62558e9d" />

As we can see, there is a significant amount of CPU time spent on Cleanup transactions, as reading BlobIndex is expensive. We have already seen that removing the need to read these indexes can significantly speed up transactions (see https://github.com/ydb-platform/nbs/issues/5590).

But we can get rid of any reads in cleanup transaction, because we already do all needed reeds in compaction transaction:
1. If we are not reading BlobIndex for blob on compaction, then we can get all the necessary information by reading the mixed/merged block index. For merged blobs, everything is easy. For mixed blobs, we can recreate the BlobMeta if it is completely available for compaction: all the blocks from the blob are in one compacted range, and there are no commit ids larger then the compaction commitId.
2. If we reading BlobIndex for blob on compaction, then we can also read BlobMetas not only BlockMasks. 

So, we can read or recreate all the blobMetas in compaction, and then just pass them to the cleanup queue stored in RAM. If we use that meta for cleanup, we can perform zero localdb reads on the cleanup transaction.

PR #6394, opened 2026-07-01
[NBS] issue-6297: read blob metas along with block masks for cleanup optimization

vladstepanyuk:
### Notes
If we are going to read a block mask for a blob on compaction, we can also read the blob meta for it. Reading one more field by key should be almost free, but this meta can be used later for cleanup, so we only do one blob index read for this blob instead of two.
 
### Issue
#6279

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
@@ -1540,6 +1540,49 @@ private:
     }
 };
 
+void FillBlobsInfo(
+    TPartitionDatabase& db,
+    TTxPartition::TCompaction& args,
+    bool& ready,
+    ui64 tabletId)
+{
+    auto blobsToReadBlockMasks = TVector<TPartialBlobId>(
+        args.BlobsToReadBlockMasks.begin(),
+        args.BlobsToReadBlockMasks.end());
+    auto blobsToReadBlobMetas = TVector<TPartialBlobId>(
+        args.BlobsToReadBlobMetas.begin(),
+        args.BlobsToReadBlobMetas.end());
+
+    auto blobsToOutputIndices = DeduplicateBlobInfos(
+        tabletId,
+        blobsToReadBlockMasks,
+        blobsToReadBlobMetas);
+
+    TVector<TBlockMask> blockMasks;
+    blockMasks.resize(args.BlobsToReadBlockMasks.size());
+    TVector<NProto::TBlobMeta> blobMetas;
+    blobMetas.resize(args.BlobsToReadBlobMetas.size());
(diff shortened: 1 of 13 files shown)


qkrorlqr:
@codex review

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1540,6 +1540,49 @@ class TCompactionTriggerer
...
+        args.BlobsToReadBlobMetas.begin(),
+        args.BlobsToReadBlobMetas.end());
+
+    auto blobsToOutputIndices = DeduplicateBlobInfos(
+        tabletId,
+        blobsToReadBlockMasks,
+        blobsToReadBlobMetas);
+
+    TVector<TBlockMask> blockMasks;
+    blockMasks.resize(args.BlobsToReadBlockMasks.size());
```
qkrorlqr: nit: we can make this a couple lines shorter:
```
TVector<TBlockMask> blockMasks(args.BlobsToReadBlockMasks.size());
TVector<NProto::TBlobMeta> blobMetas(args.BlobsToReadBlobMetas.size());
```

cloud/blockstore/libs/storage/partition/part_database.h
```diff
@@ -36,6 +36,13 @@ class TPartitionDatabase
         Partial
     };
 
+    THashMap<TString, ui64> MethodCallCounts;
```
qkrorlqr: let's remove this from production builds

it's better to inject this via a template parameter and use a no-op counter outside of the ut suite where you need this

cloud/blockstore/libs/storage/partition/part_database.h
```diff
@@ -36,6 +36,13 @@ class TPartitionDatabase
...
+
+    void IncrementMethodCallCount(const TString& methodName)
```
qkrorlqr: let's not use `TString` here - you're copying function name into another buffer for seemingly no reason

just use `TStringBuf`

cloud/blockstore/libs/storage/partition/part_readblobinfo_logic.cpp
```diff
@@ -5,50 +5,153 @@
...
+
+        indexes.BlobMetaIndex = i;
+    }
+
+    return blobsToOutputIndices;
+}
+
+bool ReadBlobsInfo(
+    TPartitionDatabase& db,
+    const THashMap<
```
qkrorlqr: let's add `using xxx` for such structures with long signatures, otherwise it's harder to read

TBlobId2IndexMap

qkrorlqr (approved):

komarevtsev-d (approved):

PR #6407, opened 2026-07-02
fix static provider creation

yegorskii:
### Notes
Fix broken static provider creation where we forgot to provide keys content to GRPC

### Issue

### DIFF
--- a/cloud/storage/core/libs/grpc/tls_utils.cpp
+++ b/cloud/storage/core/libs/grpc/tls_utils.cpp
@@ -396,8 +396,16 @@ TVector<TCertificatePair> LoadCertificatePairs(
     TVector<TCertificatePair> result;
     result.reserve(prepared.size());
     for (auto& cert: prepared) {
+        auto keyCertPair = ReadAndValidateIdentityPair(cert);
+        if (HasError(keyCertPair.GetError())) {
+            ythrow yexception() << keyCertPair.GetError().GetMessage();
+        }
+
+        const auto& keyCert = keyCertPair.GetResult().front();
         result.push_back({
             .Files = std::move(cert),
+            .PrivateKey = TString(keyCert.private_key()),
+            .CertChain = TString(keyCert.cert_chain()),
         });
     }
     return result;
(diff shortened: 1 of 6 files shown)


cloud/storage/core/libs/grpc/tls_utils.cpp
```diff
@@ -396,8 +396,16 @@ TVector<TCertificatePair> LoadCertificatePairs(
     TVector<TCertificatePair> result;
     result.reserve(prepared.size());
     for (auto& cert: prepared) {
+        auto keyCertPair = ReadAndValidateIdentityPair(cert);
```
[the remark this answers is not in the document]
yegorskii: TPeriodicCertificateProvider is expected to validate initial TLS materials at creation/startup time (same fail-fast behavior as static), and then continue strict revalidation on refresh with fallback to last known-good materials.

To make this explicit and avoid regressions, I aligned loading/validation in LoadRootCaPair / LoadCertificatePairs and kept UT coverage for invalid/missing initial certs plus refresh behavior.

cloud/storage/core/libs/grpc/tls_utils.cpp
```diff
@@ -396,8 +396,16 @@ TVector<TCertificatePair> LoadCertificatePairs(
...
+        if (HasError(keyCertPair.GetError())) {
+            ythrow yexception() << keyCertPair.GetError().GetMessage();
+        }
```
SvartMetal, line 402: Should we move it to PrepareAndValidateCertificates?

yegorskii: I assume no, PrepareAndValidateCertificates just checks if cert contains both pricate key and cert filenames. It does not check if contests is cryptographicaly correct

yegorskii: Ok, may be function name is a little bit confusing

SvartMetal (approved):

komarevtsev-d (approved):

PR #6370, opened 2026-06-29
periodic TLS certificate provider.

yegorskii:
### Notes
Implemented TPeriodicCertiicateProvider class for dynamic TLS certificates update. New functionality implements GRPC distributor interface which allows to dynamically change TLS certificates for the channels. So by periodically re-reading TLS certificates we can update them "on-line" without restarting our services.

### Issue
https://github.com/ydb-platform/nbs/issues/5722

PR #6408, opened 2026-07-02
fix: trying to fix reporting on timed out/cancelled/failed workflows

librarian:
Addresses issues with not reporting plaque and jobs status in https://github.com/ydb-platform/nbs/pull/6370 because of issue with infinitely running recipes.

### DIFF
--- a/.github/workflows/on_pooled_build_and_test_ya.yaml
+++ b/.github/workflows/on_pooled_build_and_test_ya.yaml
@@ -110,7 +110,7 @@ jobs:
           fi
           echo "LABELS_LIST=${LABELS_LIST}" | tee -a "$GITHUB_OUTPUT"
   build-and-test:
-    name: Build and test ${{ inputs.build_preset }}
+    name: Build and test [build_preset=${{ inputs.build_preset }} component=${{ inputs.component == '' && 'all' || inputs.component }}]
     runs-on: ${{ fromJson(needs.calculate-runs-on.outputs.runs-on) }}
     timeout-minutes: 480
     needs: calculate-runs-on
@@ -196,7 +196,7 @@ jobs:
         target_platform: ${{ inputs.target_platform }}
         component: ${{ inputs.component == '' && 'all' || inputs.component }}
         workload_check_status: running
-        current_job_name: Build and test ${{ inputs.build_preset }}
+        current_job_name: Build and test [build_preset=${{ inputs.build_preset }} component=${{ inputs.component == '' && 'all' || inputs.component }}]
         runner_name: ${{ runner.name }}
         add_summary_link: yes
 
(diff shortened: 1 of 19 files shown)


librarian:
> [!TIP]
> Nightly workflows are **finished**.

Collector: [nightly-builds job](https://github.com/ydb-platform/nbs/actions/runs/28623127901/job/84883484708)

- :red_circle: [`nightly-asan.yaml`](https://github.com/ydb-platform/nbs/actions/runs/28623154080) - `collector timed out; GitHub accepted cancellation, but the nightly run stayed in_progress`

librarian:
@codex review

librarian:
@codex review

WilyTiger:
It looks like the finalizer can determine a job’s actual conclusion only when the workload row already contains a job  url.
In finalize_workload_comments.py, resolve_job_conclusion() returns None immediately when parse_actions_job_url(job_url) fails. That None is then mapped to report_failed in workload_check_status_for_job_conclusion().
This can misreport jobs that fail, are cancelled, or are skipped before reaching the mark workload as running step. In those cases, no job URL is added to the PR comment, so the finalizer produces: "job finished but report update failed" even though the actual conclusion may be failure, cancelled, or skipped.

Could we add a fallback that resolves the job from the matrix entry, for example using build_preset and component? It may also help to include these values explicitly in the job name, making the mapping unambiguous. Moving mark workload as running earlier would reduce the number of affected cases, but would not cover skipped jobs or startup failures

.github/workflows/on_pooled_build_and_test_ya.yaml
```diff
@@ -269,8 +269,17 @@ jobs:
         SLEEP_AFTER_TESTS: ${{ inputs.sleep_after_tests || '7200' }}
       run: |
         sleep "$SLEEP_AFTER_TESTS"
+    - name: mark workload as cancelled
+      if: ${{ cancelled() && steps.build.outcome != 'failure' && steps.test.outputs.report_generated != 'yes' && (github.event_name == 'pull_request' || github.event_name == 'pull_request_target') }}
+      uses: ./.github/actions/workload_comment
+      with:
+        command: update
+        build_preset: ${{ inputs.build_preset }}
```
WilyTiger, line 283: Don't we also need `target_platform: ${{ inputs.target_platform }}` here? It seems necessary for cross-platform builds. 
Otherwise, the workload comment may use the runner architecture instead of the target architecture

librarian: This PR was created well before we introduced cross platform builds...

librarian: Fixed

WilyTiger (approved):

robot-nbs (approved):

PR #6411, opened 2026-07-02
feat: calculate and apply upper boundary for test runs

librarian:
Addresses issues with infinitely starting recipes due to misconfiguration here https://github.com/ydb-platform/nbs/pull/6370 

It sets up upper boundary on how long tests can run while reserving some time for s3 sync etc.

### DIFF
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -338,6 +338,10 @@ runs:
                 FAIL_CHECKER_RC=1
             fi
             DO_RETRY=0
+            ATTEMPT_TIMED_OUT=0
+            if [ "$RC" -eq 124 ]; then
+                ATTEMPT_TIMED_OUT=1
+            fi
             case $RC in
                 0)
                     echo "ya test successfully finished, no errors"
@@ -352,6 +356,7 @@ runs:
                 137)
                     if [ "$YA_MAKE_ATTEMPT_ELAPSED_SECONDS" -ge "$ATTEMPT_TIMEOUT_SECONDS" ]; then
                         echo "ya test returned $RC after timeout killed it with SIGKILL"
+                        ATTEMPT_TIMED_OUT=1
                         DO_RETRY=1
                         RETRY_FLAG="-X"
                     else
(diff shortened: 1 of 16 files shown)


.github/actions/test/action.yaml
```diff
@@ -238,13 +258,33 @@ runs:
...
+            now=$(date +%s)
+            elapsed_test_seconds=$((now - TEST_BUDGET_STARTED_AT))
+            echo $((TEST_BUDGET_MINUTES * 60 - TEST_BUDGET_RESERVE_MINUTES * 60 - BUILD_DURATION_SECONDS - elapsed_test_seconds))
+        }
+
+        if [ "$TEST_BUDGET_RESERVE_MINUTES" -ge "$TEST_BUDGET_MINUTES" ]; then
+            echo "Invalid test budget reserve (${TEST_BUDGET_RESERVE_MINUTES}m) for total budget (${TEST_BUDGET_MINUTES}m), using 30m reserve"
+            TEST_BUDGET_RESERVE_MINUTES=30
+        fi
+        TEST_BUDGET_STARTED_AT=$(date +%s)
```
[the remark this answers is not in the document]
librarian: This is already accounted for in 30m saved budget.

librarian:
> [!TIP]
> Nightly workflows are **finished**.

Collector: [nightly-builds job](https://github.com/ydb-platform/nbs/actions/runs/28623138141/job/84884745309)

- :red_circle: [`nightly-tsan.yaml`](https://github.com/ydb-platform/nbs/actions/runs/28623548076) - `failure`

librarian:
@codex review

librarian:
@codex review

.github/workflows/build_and_test_proxy.yaml
```diff
@@ -42,6 +42,10 @@ on:
       number_of_retries:
         required: true
         type: number
+      test_timeout_minutes:
```
WilyTiger, line 57: The explicit test_timeout_minutes input is never used because every generated matrix entry already contains matrix.test_timeout_minutes, which takes precedence here.
For example, if a workflow-dispatch caller requests 45 minutes and the matrix calculates 90, this expression always selects 90, because of
````
${{ matrix.test_timeout_minutes || inputs.test_timeout_minutes || '390' }}
````
Could we distinguish an explicit caller override from the calculated default and give the explicit value precedence? Otherwise, the newly exposed workflow input is effectively a no-op in all three execution modes

librarian: Valid comment and after this change it exposed that we also set test_timeout_minutes externally where we calculated what we need to test in pr_build_and_test_matrix_plan

WilyTiger (approved):

robot-nbs (approved):

Issue #5726, opened 2026-04-13
[Filestore] Optimize NodeRefs cache performance


qkrorlqr:
Currently we can read approx 1-1.5M `NodeRef`s from cache per second and get bottlenecked by `TIndexTabletActor` which starts to consume 100% CPU. Most of the time is spent iterating over `std::map` which backs our current cache implementation. `std::map` uses a red-black tree which is not the best data structure for the case of range reads - and upon listing we read thousands of `NodeRef`s per request. B-tree suits this scenario much better.

The second inefficient thing is how we build the response - we build it node by node, name by name, we don't use arenas for allocation and as a result spend a lot of time allocating and filling those nodes / names.

qkrorlqr:
Will use `absl::btree_map`, added to our repo here: https://github.com/ydb-platform/nbs/pull/5724

qkrorlqr:
btree_map + no eviction + ListNodesInterval give us approx x3 better listing performance - tested on 64 dirs x 1M files each x 32 lister threads (via fmdtest), all these 64M node refs were stored in 1 tablet

before:
```
cat baseline-report.json 
{
    "create":0,
    "create-lat-us":0,
    "unlink":0,
    "unlink-lat-us":0,
    "rename":0,
    "rename-lat-us":0,
    "stat":0,
    "stat-lat-us":0,
    "list":896,
    "list-lat-us":21516058,
    "list-w":896000000,
    "list-wlat-us":21,
    "errors":0
}
```

after:
```
cat new-report.json 
{
    "create":0,
    "create-lat-us":0,
    "unlink":0,
    "unlink-lat-us":0,
    "rename":0,
    "rename-lat-us":0,
    "stat":0,
    "stat-lat-us":0,
    "list":2425,
    "list-lat-us":7946465,
    "list-w":2425000000,
    "list-wlat-us":7,
    "errors":0
}
```

4M node refs per second

PR #6416, opened 2026-07-02
issue-5726: log cookies for ListNodes in profile log

WilyTiger:
### Notes
Add more informative profile logs for list operations to help with debugging

### Issue
#5726

### DIFF
--- a/cloud/filestore/libs/diagnostics/profile_log_events.cpp
+++ b/cloud/filestore/libs/diagnostics/profile_log_events.cpp
@@ -480,8 +480,10 @@ void InitProfileLogRequestInfo(
     NProto::TProfileLogRequestInfo& profileLogRequest,
     const NProto::TListNodesRequest& request)
 {
-    auto* nodeInfo = profileLogRequest.MutableNodeInfo();
-    nodeInfo->SetNodeId(request.GetNodeId());
+    auto* listNodesInfo = profileLogRequest.MutableListNodesInfo();
+    listNodesInfo->SetNodeId(request.GetNodeId());
+    listNodesInfo->SetMaxBytes(request.GetMaxBytes());
+    listNodesInfo->SetRequestCookie(request.GetCookie());
 }
 
 template <>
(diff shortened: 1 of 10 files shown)


cloud/filestore/libs/diagnostics/profile_log_events.cpp
```diff
@@ -731,8 +733,9 @@ void FinalizeProfileLogRequestInfo(
     NProto::TProfileLogRequestInfo& profileLogRequest,
     const NProto::TListNodesResponse& response)
 {
-    auto* nodeInfo = profileLogRequest.MutableNodeInfo();
-    nodeInfo->SetSize(response.GetNames().size());
+    auto* listNodesInfo = profileLogRequest.MutableListNodesInfo();
+    listNodesInfo->SetSize(response.GetNames().size());
```
qkrorlqr: let's call it `NameCount`, otherwise when you look at the schema and see `Size` you'd probably think that it's measured in bytes

qkrorlqr (dismissed):

cloud/filestore/libs/diagnostics/profile_log_events.cpp
```diff
@@ -480,8 +480,10 @@ void InitProfileLogRequestInfo(
     NProto::TProfileLogRequestInfo& profileLogRequest,
     const NProto::TListNodesRequest& request)
 {
-    auto* nodeInfo = profileLogRequest.MutableNodeInfo();
-    nodeInfo->SetNodeId(request.GetNodeId());
+    auto* listNodesInfo = profileLogRequest.MutableListNodesInfo();
```
debnatkh, line 483: This will move the place where we store NodeId, if we do so, the requests printer should support the legacy format

debnatkh: Same goes for the Size field

WilyTiger: It will work because the old format stored this data in TNodeInfo, so profile-tool will handle it through this path https://github.com/ydb-platform/nbs/pull/6416/changes#diff-ff11f6be8a843087dfac560d03ba130be3d8925f9a793ecc745f095506221fcaL337

WilyTiger: https://github.com/ydb-platform/nbs/blob/2ffc34141784dc3fafab9ad64a41541800adccf9/cloud/filestore/tools/analytics/libs/event-log/request_printer.cpp#L384

debnatkh (dismissed):

debnatkh (approved):

agalibin (approved):

PR #6423, opened 2026-07-03
[Filestore] emulate page faults in tests

tdiff:
### Notes
This change enables randomly restarting transactions in all tests using TTestEnv.  Additionally, it allows to do the same in larger/integration tests and exposes new configuration in storage config.

Currently there is some gap in testing transactions restarted due to "page faults" in YDB, which reschedule and restart read transactions after some delay. As we did not find a way to trigger page faults in YDB, I extended previously existing failure injection to all Read* methods of TInMemoryIndexState. 

### Issue
#6467

### DIFF
--- a/cloud/filestore/config/storage.proto
+++ b/cloud/filestore/config/storage.proto
@@ -845,4 +845,8 @@ message TStorageConfig
 
     // Async processing of destroy handle requests only for read only handles.
     optional bool AsyncDestroyReadOnlyHandleEnabled = 525;
+    
+    // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
+    optional bool FakeReadFailuresEnabled = 526;
+    optional float FakeReadFailuresProbabilityPercentage = 527;
 }
--- a/cloud/filestore/config/storage.proto
+++ b/cloud/filestore/config/storage.proto
@@ -847,6 +847,6 @@ message TStorageConfig
     optional bool AsyncDestroyReadOnlyHandleEnabled = 525;
     
     // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
-    optional bool FakeReadFailuresEnabled = 526;
-    optional float FakeReadFailuresProbabilityPercentage = 527;
+    optional bool FakeTxPageFaultsEnabled = 526;
+    optional float FakeTxPageFaultsProbabilityPercentage = 527;
 }
--- a/cloud/filestore/libs/storage/core/tablet.h
+++ b/cloud/filestore/libs/storage/core/tablet.h
@@ -143,26 +143,6 @@ protected:
             Generation = tx.Generation;
             Step = tx.Step;
 
-            // Reschedule (restart) the transaction from PrepareTx when the
-            // injected rescheduler asks for it. In production this is a no-op
-            // stub; tests inject an implementation to make the transaction
-            // restart/reorder path.
-            if (Self->TxRescheduler->ShouldReschedule()) {
-                LOG_INFO(
-                    ctx,
-                    T::LogComponent,
-                    "[%lu] Forced reschedule of %s (gen: %u, step: %u)",
-                    Self->TabletID(),
-                    TTx::Name,
-                    Generation,
-                    Step);
-
-                Args.Clear();
-                Args.OnRestart();
-                tx.Reschedule();
-                return false;
-            }
-
             TX_TRACK(TxPrepare);
             LOG_TRACE(ctx, T::LogComponent,
                 "[%lu] PrepareTx %s (gen: %u, step: %u)",
@@ -176,6 +156,22 @@ protected:
 
                 Args.Clear();
                 Args.OnRestart();
+
+                // Reschedules the transaction from PrepareTx assuming the rescheduler
+                // was triggered during PrepareTx() call.
+                if (Y_UNLIKELY(Self->TxRescheduler && Self->TxRescheduler->IsTriggered())) {
+                    Self->TxRescheduler->Reset();
+                    LOG_INFO(
+                        ctx,
+                        T::LogComponent,
+                        "[%lu] Forced reschedule of %s (gen: %u, step: %u)",
+                        Self->TabletID(),
+                        TTx::Name,
+                        Generation,
+                        Step);
+                    tx.Reschedule();
+                }
+
                 return false;
             }
 
@@ -355,7 +351,7 @@ protected:
         {                                                                      \
             TCPUUsageTimerGuard t(target.AccessCPUUsageTimer());               \
             if (target.ValidateTx_##name(ctx, args)) {                         \
-                dbType db(tx.DB, args.NodeUpdates);                            \
+                dbType db(tx.DB, args.NodeUpdates, target.TxRescheduler);      \
                 return target.PrepareTx_##name(ctx, db, args);                 \
             }                                                                  \
             return true;                                                       \
--- a/cloud/filestore/libs/storage/core/tablet.h
+++ b/cloud/filestore/libs/storage/core/tablet.h
@@ -164,11 +164,12 @@ protected:
                     LOG_INFO(
                         ctx,
                         T::LogComponent,
-                        "[%lu] Forced reschedule of %s (gen: %u, step: %u)",
+                        "[%lu] Forced reschedule of %s (gen: %u, step: %u, random seed=%lu)",
                         Self->TabletID(),
                         TTx::Name,
                         Generation,
-                        Step);
+                        Step,
+                        Self->TxRescheduler->GetSeed());
                     tx.Reschedule();
                 }
 
--- a/cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
+++ b/cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
@@ -1,5 +1,11 @@
 #include "tablet_tx_rescheduler.h"
 
+#include <cloud/filestore/libs/storage/api/components.h>
+
+#include <contrib/ydb/library/actors/core/actor.h>
+#include <contrib/ydb/library/actors/core/log.h>
+
+#include <util/random/mersenne.h>
 #include <util/random/random.h>
 
 namespace NCloud::NFileStore::NStorage {
@@ -18,15 +24,21 @@ class TRandomTxRescheduler final
 private:
     const float Probability = 0;
     bool Triggered = false;
+    ui64 Seed;
+    TMersenne<ui64> RandomGen;
 
 public:
-    explicit TRandomTxRescheduler(float probabilityPercentage)
+    explicit TRandomTxRescheduler(float probabilityPercentage,
+                                  std::optional<ui64> randomSeed)
         : Probability(probabilityPercentage / 100.0F)
-    {}
+        , Seed(randomSeed ? *randomSeed : RandomNumber<ui64>())
+        , RandomGen(Seed)
+    {
+    }
 
     bool ShouldReschedule() override
     {
-        const bool ret = RandomNumber<float>() < Probability;
+        const bool ret = RandomGen.GenRandReal4() < Probability;
         Triggered |= ret;
         return ret;
     }
@@ -40,15 +52,22 @@ public:
     {
         Triggered = false;
     }
+
+    ui64 GetSeed() const override
+    {
+        return Seed;
+    }
 };
 
 }   // namespace
 
 ////////////////////////////////////////////////////////////////////////////////
 
-ITxReschedulerPtr CreateRescheduler(float probabilityPercentage)
+ITxReschedulerPtr CreateRescheduler(TReschedulerParams params)
 {
-    return std::make_shared<TRandomTxRescheduler>(probabilityPercentage);
+    return std::make_shared<TRandomTxRescheduler>(
+        params.ProbabilityPercentage,
+        params.RandomSeed);
 }
 
 }   // namespace NCloud::NFileStore::NStorage
--- a/cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
+++ b/cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
@@ -1,18 +1,44 @@
 #include "tablet_tx_rescheduler.h"
 
+#include <util/random/random.h>
+
 namespace NCloud::NFileStore::NStorage {
 
 namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-class TNoOpTxRescheduler final
+// With a given probability decides if a transaction should be rescheduled.
+// Used in tests to exercise behaviour in case of transaction restarts/reorderings.
+// Note that it does not reschedule anything by itself and only serves for communication
+// between TIndexActorDatabase and Tx::Execute().
+class TRandomTxRescheduler final
     : public ITxRescheduler
 {
+private:
+    const float Probability = 0;
+    bool Triggered = false;
+
 public:
+    explicit TRandomTxRescheduler(float probabilityPercentage)
+        : Probability(probabilityPercentage / 100.0F)
+    {}
+
     bool ShouldReschedule() override
     {
-        return false;
+        const bool ret = RandomNumber<float>() < Probability;
+        Triggered |= ret;
+        return ret;
+    }
+
+    bool IsTriggered() const override
+    {
+        return Triggered;
+    }
+
+    void Reset() override
+    {
+        Triggered = false;
     }
 };
 
@@ -20,9 +46,9 @@ public:
 
 ////////////////////////////////////////////////////////////////////////////////
 
-ITxReschedulerPtr CreateNoOpTxRescheduler()
+ITxReschedulerPtr CreateRescheduler(float probabilityPercentage)
 {
-    return std::make_shared<TNoOpTxRescheduler>();
+    return std::make_shared<TRandomTxRescheduler>(probabilityPercentage);
 }
 
 }   // namespace NCloud::NFileStore::NStorage
--- a/cloud/filestore/libs/storage/tablet/bench/tablet_bench.cpp
+++ b/cloud/filestore/libs/storage/tablet/bench/tablet_bench.cpp
@@ -64,7 +64,7 @@ struct TTabletSetup
             .LogPriority_NFS = NActors::NLog::PRI_ALERT,
             .LogPriority_KiKiMR = NActors::NLog::PRI_ALERT,
             .LogPriority_Others = NActors::NLog::PRI_ALERT,
-            .PageFaultFakesEnabled = false})
+            .FakePageFaultsEnabled = false})
     {
         NCloud::NFileStore::NProto::TStorageConfig storageConfig;
         storageConfig.SetInMemoryIndexCacheEnabled(true);
--- a/cloud/filestore/libs/storage/tablet/bench/tablet_bench.cpp
+++ b/cloud/filestore/libs/storage/tablet/bench/tablet_bench.cpp
@@ -63,7 +63,8 @@ struct TTabletSetup
             // Turn off logging in order to reduce performance overhead
             .LogPriority_NFS = NActors::NLog::PRI_ALERT,
             .LogPriority_KiKiMR = NActors::NLog::PRI_ALERT,
-            .LogPriority_Others = NActors::NLog::PRI_ALERT})
+            .LogPriority_Others = NActors::NLog::PRI_ALERT,
+            .PageFaultFakesEnabled = false})
     {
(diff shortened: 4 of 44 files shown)


cloud/filestore/libs/storage/testlib/test_env.cpp
```diff
@@ -130,6 +130,10 @@ TTestEnv::TTestEnv(
         Runtime.GetAppData().Counters);
 
     SetupSubDomain();
+
+    // Inject a rescheduler so recovery confirmation is exercised under
+    // transaction restarts/reorderings.
+    SetTxRescheduler(CreateRescheduler(config.PageFaultProbabilityPct));
```
neihar: we need isntance per tablet, not global for env, currently not so many tests or none with multiple tablets, but it is possible

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -14,12 +16,60 @@ class TNoOpTxRescheduler final
...
+    bool Triggered = false;
+
+public:
+    explicit TRandomTxRescheduler(ui32 probabilityPct)
+        : ProbabilityPct(probabilityPct)
+    {}
+
+    bool ShouldReschedule() override
+    {
+        const bool ret = RandomNumber<ui32>(100) < ProbabilityPct;
```
neihar: with current implementation it is probability per Read, not per Tx, so probably it is to high and for heavy TXes which have for loops it can be hard to pass, I suggest to add support for more granular precent enablement, like 0.01%. Or probably we can make something like cooldown or other mechanic, that will allow after trigger some reads to pass without problem.

cloud/filestore/libs/storage/tablet/tablet_database.cpp
```diff
@@ -42,6 +42,11 @@ bool TIndexTabletDatabase::ReadFileSystem(NProto::TFileSystem& fileSystem)
 {
     using TTable = TIndexTabletSchema::FileSystem;
 
+    if (Y_UNLIKELY(TestReadRescheduler && TestReadRescheduler->ShouldReschedule()))
```
neihar: It will work not so good with Stub I guess, better to have nullptr for prod

cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
```diff
@@ -258,7 +258,7 @@ bool TIndexTabletActor::PrepareTx_WriteData(
         args.NodeId,
         args.ByteRange.Describe().c_str());
 
-    TIndexTabletDatabaseProxy db(tx.DB, args.NodeUpdates);
+    TIndexTabletDatabaseProxy db(tx.DB, args.NodeUpdates, TxRescheduler);
```
neihar, line 251: also, probably we can use here ref and avoid ref+-

cloud/filestore/libs/storage/tablet/tablet_database.cpp
neihar, line 476: We should use for all read operations rescheduler

cloud/filestore/libs/storage/tablet/tablet_actor_updateconfig.cpp
```diff
@@ -515,7 +515,7 @@ void TIndexTabletActor::ExecuteTx_ConfigureAsShard(
 {
     Y_UNUSED(ctx);
 
-    TIndexTabletDatabase db(tx.DB);
+    TIndexTabletDatabase db(tx.DB, TxRescheduler);
```
neihar: for db in Execute we can pass nullptr all the time, because we do not expect restarts here

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -14,12 +16,60 @@ class TNoOpTxRescheduler final
...
+    void Reset() override
+    {
+    }
+};
+
+// With given probability decides if a transaction should be rescheduled.
+// Used in tests to exercise behaviour in case of transaction restarts/reorderings.
+// Note that it does not reschedule anything by itself and only serves for communication
+// between TIndexActorDatabase and Tx::Execute().
+class TRandomTxRescheduler final
```
neihar, line 21: now it looks like it is TxReadFaultInjector and not Recheduler :)

cloud/filestore/libs/storage/tablet/tablet_database.cpp
```diff
@@ -329,6 +359,11 @@ bool TIndexTabletDatabase::ReadNodeAttr(
 {
     using TTable = TIndexTabletSchema::NodeAttrs;
 
+    if (Y_UNLIKELY(TestReadRescheduler && TestReadRescheduler->ShouldReschedule()))
```
neihar: if (Y_UNLIKELY(ShouldFailRead())

cloud/filestore/config/storage.proto
```diff
@@ -834,4 +834,8 @@ message TStorageConfig
     // Sets the minimum shard count that each filesystem would have regardless
     // of its size.
     optional uint32 MinShardCount = 522;
+
+    // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
+    optional bool UnsafeTestReadFailuresEnabled = 523;
```
neihar: FakeReadFailuresEnabled

cloud/filestore/config/storage.proto
```diff
@@ -834,4 +834,8 @@ message TStorageConfig
...
+    optional uint32 UnsafeTestReadFailuresProbabilityPct = 524;
```
neihar: FakeReadFailuresProbabilityPercent

neihar:
https://github-actions-s3.storage.eu-north2.nebius.cloud/ydb-platform/nbs/PR-check/28689646890/1/nebius-x86-64/summary/filestore/2/ya-test.html#FAIL

sharding tests are failing, looks like after TX restart we just choose different shard. Doesn't look important, we can use nullptr rescheduler for sharding tests

cloud/filestore/libs/storage/tablet/tablet_actor_updateconfig.cpp
tdiff: thanks, just an accidental change here

cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
tdiff: But we need one incr anyway in the end to create a new reference, and here I just create it upfront and move() it further, no?

cloud/filestore/libs/storage/tablet/tablet_database.cpp
tdiff: I agree, I think we don't need a noop-rescheduler (and probably we don't need that interface either)

cloud/filestore/libs/storage/tablet/tablet_database.cpp
tdiff: thanks, missed

cloud/filestore/config/storage.proto
tdiff: honestly, a matter of taste, there is no objective measure which option is better except for some existing practice in the repo.

for me personally it is kind of more obvious that "Unsafe"  should not be used in prod, and "Fake" is somewhat less clear.

in any case, I won't insist :)

neihar: yes, agree, I mostly don't like combination of Unsafe+Test, just Unsafe is probably fine, but from my perspective it doesn't not describe well what this param enabled. Fake ilustrates this clearly - we will have fake page failures

tdiff: UsafeFake* would combine best of two worlds?

neihar: I don't want to produce more "patterns" for such functionality. we already have for example "FakeDescribeDataEnabled"

tdiff: thats fair

cloud/filestore/config/storage.proto
```diff
@@ -834,4 +834,8 @@ message TStorageConfig
     // Sets the minimum shard count that each filesystem would have regardless
     // of its size.
     optional uint32 MinShardCount = 522;
+
+    // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
+    optional bool FakeReadFailuresEnabled = 523;
+    optional float FakeReadFailuresProbabilityPct = 524;
```
neihar: we usually use whole words like Percentage, see in this file for example.

cloud/filestore/libs/storage/core/tablet.h
```diff
@@ -176,6 +156,24 @@ class TTabletBase
 
                 Args.Clear();
                 Args.OnRestart();
+
+                // Reschedules the transaction from PrepareTx assuming the rescheduler
+                // was triggered during PrepareTx() call. In production a a no-op
```
neihar: in production it is a nullptr now

cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
neihar: we can use Ptr& here and it will bond to nullptr arg just fine, creating empty shared_ptr. which is cheaper than refcount.

cloud/filestore/libs/storage/core/tablet.h
```diff
@@ -176,6 +156,24 @@ class TTabletBase
...
+                // stub is used; tests inject an implementation to make the transaction
+                // restart/reorder path.
+                if (Self->TxRescheduler->IsTriggered()) {
```
neihar: should we use unlikely and nullptr instead of stub here as well?

cloud/filestore/config/storage.proto
tdiff: changed

cloud/filestore/libs/storage/core/tablet.h
tdiff: added

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -1,28 +1,54 @@
...
     : public ITxRescheduler
 {
+private:
+    const float Probability = 0;
+    bool Triggered = false;
+
 public:
+    explicit TRandomTxRescheduler(float probabilityPct)
```
neihar: let's use Percentage everywhere instead of Pct

cloud/filestore/libs/storage/tablet/bench/tablet_bench.cpp
```diff
@@ -63,7 +63,9 @@ struct TTabletSetup
             // Turn off logging in order to reduce performance overhead
             .LogPriority_NFS = NActors::NLog::PRI_ALERT,
             .LogPriority_KiKiMR = NActors::NLog::PRI_ALERT,
-            .LogPriority_Others = NActors::NLog::PRI_ALERT})
+            .LogPriority_Others = NActors::NLog::PRI_ALERT,
+            // Turn off random pagefaults
+            .PageFaultProbabilityPercentage = 0})
```
neihar: let's disable it specifically with enabled flag

neihar:
@codex review

cloud/filestore/libs/storage/testlib/test_env.h
```diff
@@ -76,6 +76,10 @@ struct TTestEnvConfig
     NActors::NLog::EPriority LogPriority_NFS = NActors::NLog::PRI_TRACE;
     NActors::NLog::EPriority LogPriority_KiKiMR = NActors::NLog::PRI_WARN;
     NActors::NLog::EPriority LogPriority_Others = NActors::NLog::PRI_WARN;
+
+    // This controls probability that read transaction will be restarted
```
neihar: that read will fail and tx restarted (not read transaction)

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -1,28 +1,54 @@
...
+
 public:
+    explicit TRandomTxRescheduler(float probabilityPercentage)
+        : Probability(probabilityPercentage / 100.0F)
+    {}
+
     bool ShouldReschedule() override
     {
-        return false;
+        const bool ret = RandomNumber<float>() < Probability;
```
neihar: Nit: probably logging seed somewhere could be nice

cloud/filestore/libs/storage/testlib/test_env.cpp
tdiff: done

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
tdiff: changed to float

cloud/filestore/libs/storage/tablet/tablet_database.cpp
tdiff: removed no-op class

cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
tdiff: as discussed, after removing noop stub there is no issue here.

cloud/filestore/libs/storage/tablet/tablet_database.cpp
tdiff: done

cloud/filestore/config/storage.proto
tdiff: done

cloud/filestore/config/storage.proto
tdiff: done

cloud/filestore/libs/storage/core/tablet.h
tdiff: updated comment

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
tdiff: done

cloud/filestore/libs/storage/tablet/bench/tablet_bench.cpp
tdiff: added

cloud/filestore/libs/storage/testlib/test_env.h
tdiff: fixed

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -1,28 +1,74 @@
...
+            Cout << "Initializing test env with random seed=" << seed << "\n";
+            return seed;
+        }())
+    {
+    }
+
+    TRandomTxRescheduler(float probabilityPercentage, ui64 seed)
+        : Probability(probabilityPercentage / 100.0F)
+        , RandomGen([=](){
+            Cout << "Initializing test env with provided seed=" << seed << "\n";
```
neihar: let's return GetSeed() and log it when reschedule TX

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -1,28 +1,78 @@
...
+    explicit TRandomTxRescheduler(float probabilityPercentage,
+                                  std::optional<ui64> randomSeed)
+        : Probability(probabilityPercentage / 100.0F)
+        , RandomGen([=](){
+            ui64 seed = RandomNumber<ui64>();
+            if (randomSeed) {
+                seed = *randomSeed;
+            }
+            LOG_INFO(
+                *NActors::TlsActivationContext,
```
neihar: First, it can be called only inside executor thread, if we can tx rescheduler outside of it, as for example it was for test env - we will crash.
Secondonly it will not log tablet id, without it, it will be hard to tell on node with multiple tablets to which fs this seed belongs. We need to log it from some reacher and safer context (at least with tablet id)

neihar: I still suggest to use GetSeed and log during triggered reschedules

tdiff: ok, done

neihar (dismissed):

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
tdiff: done

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
tdiff: done

cloud/filestore/config/storage.proto
```diff
@@ -834,4 +834,8 @@ message TStorageConfig
     // Sets the minimum shard count that each filesystem would have regardless
     // of its size.
     optional uint32 MinShardCount = 522;
+
+    // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
+    optional bool FakeReadFailuresEnabled = 523;
```
SvartMetal: Maybe name it `FakeTxPageFaultsEnabled`?

tdiff: renamed

SvartMetal (dismissed):

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -1,28 +1,73 @@
...
+        , Seed(randomSeed ? *randomSeed : RandomNumber<ui64>())
+        , RandomGen(Seed)
+    {
+    }
+
     bool ShouldReschedule() override
     {
-        return false;
+        const bool ret = RandomGen.GenRandReal4() < Probability;
+        Triggered |= ret;
```
agalibin, line 42: It’s strange that the value of Triggered can change from false to true, but not vice versa.

tdiff: It is for handling transactions doing multiple reads, where not-the-last-one read is failed. Assumption is made here that if any read() in transaction fails, the whole transaction should be rescheduled.

Transition true->false happens in Reset().

cloud/filestore/libs/storage/tablet/tablet_database.cpp
```diff
@@ -42,6 +42,11 @@ bool TIndexTabletDatabase::ReadFileSystem(NProto::TFileSystem& fileSystem)
 {
     using TTable = TIndexTabletSchema::FileSystem;
 
+    if (Y_UNLIKELY(ShouldFailReadInTest()))
```
qkrorlqr, line 45: looks very ugly

let's just make another implementation of `IIndexTabletDatabase` that would wrap another `IIndexTabletDatabase` with these `if xxx -> return false` things

qkrorlqr (changes requested):

cloud/filestore/libs/storage/tablet/tablet_database.cpp
tdiff: I was thinking about that, but there is also TIndexTabletDatabaseProxy inherited from TIndexTabletDatabase used in 50% of all places.

Also, TIndexTabletDatabaseProxy is used in some macros magic for method gen in PP that would be harder to customize:
```
    FILESTORE_TABLET_INDEX_RO_TRANSACTIONS(
        FILESTORE_IMPLEMENT_RO_TRANSACTION,
        TTxIndexTablet,
        TIndexTabletDatabaseProxy, <<<< 
        IIndexTabletDatabase);
```

So alternatives were:
- add new "FailingTabledDatabase" and parametrize TIndexTabletDatabaseProxy with base class
- re-implement TIndexTabletDatabaseProxy in a new class inherited from FailingTabledDatabase (copying its current logic).
- just making the base class uglier. 

The last option seemed to be less intrusive in the end, but we can reconsider this, wyt?

neihar: then in every place where it is created we will need ugly IF, so it will just change the place. Or do I miss something?

neihar: we can do something like CreateIndexTabletDatabase and hide it in creation function. that way it will be nice in all sides. Will look definitely better. So, I am not sure that it worth it. But I would say, both options work for me, existing and proposed

SvartMetal: I don't like these `if` statements either.

Let's do it the other way around.

PR #6426, opened 2026-07-05
Filestore] add SIGUSR1 retries to FUSE loop thread shutdown to mitigate single lost SIGUSR1 wakeup

neihar:
### Problem
https://github.com/ydb-platform/nbs/actions/runs/27998974260/job/82867193543

### Notes
TFuseLoopThread::Stop() sends one pthread_kill(SIGUSR1), then joins. The loop thread runs while(!fuse_session_exited(se)) read(/dev/fuse). A signal only breaks read() with EINTR if the thread is already sleeping inside it. There is a short window between the exit-flag check and the start of read(). If the signal arrives in that window, the no-op handler eats it. The thread then enters read() with nothing pending and blocks forever. No more FUSE requests come. No more signals are sent. The unmount that would break read() with ENODEV runs only after the join, and the join is already stuck on this thread. This window widens under host load. The bug predates the multi-queue refactor (#4045).

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/loop.cpp
+++ b/cloud/filestore/libs/vfs_fuse/loop.cpp
@@ -580,12 +580,14 @@ public:
             "stopping FUSE loop thread " << FuseLoopIndex << "."
                                          << BackendQueueIndex);
 
-        // A single signal is not enough: it may be delivered when the loop
-        // thread is not blocked on sem_wait/read yet (e.g. right after it has
-        // checked the session exit flag but before it has entered the
-        // syscall). Such a signal interrupts nothing and the loop thread then
-        // blocks indefinitely, so keep signalling until the loop actually
-        // finishes.
+        // A single signal is not enough. The loop thread runs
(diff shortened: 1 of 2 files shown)


cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -582,7 +580,16 @@ class TFuseLoopThread final
             "stopping FUSE loop thread " << FuseLoopIndex << "."
                                          << BackendQueueIndex);
 
+        // A single signal is not enough: it may be delivered when the loop
+        // thread is not blocked on sem_wait/read yet (e.g. right after it has
```
budevg: let's fix the comment to reference the `fuse_session_receive_buf` call directly which can be blocked indefinitely if we miss the signal

budevg (changes requested):

qkrorlqr (dismissed):

budevg (approved):

agalibin (approved):

SvartMetal (approved):

PR #6429, opened 2026-07-06
chore: rework Github client init

librarian:
Refactoring includes trying to init GH client once, at script start and passing it down. It solves problem where you can't really tell what is needed for this specific script to work.

### DIFF
--- a/.github/scripts/helpers.py
+++ b/.github/scripts/helpers.py
@@ -471,7 +471,7 @@ def github_client(github_token: str | None = None) -> Github:
 
 def github_client_from_env() -> Github:
     github_token = os.environ.get("GITHUB_TOKEN")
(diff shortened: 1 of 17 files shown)


librarian:
@codex review

.github/scripts/helpers.py
```diff
@@ -454,6 +469,18 @@ def github_client(github_token: str | None = None) -> Github:
     return Github()
 
 
+def github_client_from_env() -> Github:
+    github_token = os.environ.get("GITHUB_TOKEN")
```
WilyTiger, line 473: maybe check that GITHUB_TOKEN is "" also?

WilyTiger (approved):

robot-nbs (approved):

PR #6432, opened 2026-07-06
filestore: TEndpointManager::Drain() shouldn't crash for non-started TEndpointManager

qkrorlqr:
### Notes
Currently `TEndpointManager::Drain()` tries to access `TEndpointManager::Executor::Thread::Dispatcher` which is `nullptr` before `Start()` is called. At the same time we call `Stop()` whenever `Init()` or `Start()` fail and sometimes that means that `TEndpointManager` is created but not started yet: https://github.com/ydb-platform/nbs/blob/da706bc3ae9fe45768abc0d3902ca8a2c9a16ce4/cloud/storage/core/libs/daemon/app.h#L40

The cleanest way to fix this seems to actually turn `Drain()` into a no-op if `TEndpointManager` is not started yet.

### Issue
none

### DIFF
--- a/cloud/filestore/libs/endpoint/endpoint_manager.cpp
+++ b/cloud/filestore/libs/endpoint/endpoint_manager.cpp
@@ -103,6 +103,7 @@ private:
     const ui32 SocketAccessMode;
 
     TExecutorPtr Executor = TExecutor::Create("SVC");
+    bool Started = false;
     TLog Log;
 
     // Fields may only be accessed from the Executor thread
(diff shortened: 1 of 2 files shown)


debnatkh (approved):

cloud/filestore/libs/endpoint/endpoint_manager.cpp
agalibin, line 140: Started = false;

agalibin (approved):

cloud/filestore/libs/endpoint/endpoint_manager.cpp
qkrorlqr: It’s not really important in this case

PR #6437, opened 2026-07-06
[Filestore] Narrow cachebypass overlap, add metrics 

neihar:
### Notes
Added metrics for cache bypass during unconfirmed writes
Narrowed cache bypass for ReadAheadCache to overlapping range
Remove ReadXAttr from cache bypass participation
Added ut

### Issue
#2293

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_addblob.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_addblob.cpp
@@ -178,39 +178,42 @@ private:
         UpdateNodeAttrs(db, args);
     }
 
-    std::pair<ui64, ui64> CalculateCacheBypassRange(
+    TByteRange CalculateCacheBypassRange(
         const TTxIndexTablet::TAddBlob& args) const
     {
+        const ui32 blockSize = Tablet.GetBlockSize();
         const auto& writeRange = args.WriteRanges.front();
+        const auto node = args.Nodes.find(writeRange.NodeId);
 
         // On CommitIdOverflow no data is written, but the tablet is about to
-        // be stopped, so no assumptions are made and all reads for the node
+        // be stopped; a missing node means that the old file size is unknown.
+        // In both cases no assumptions are made and all reads for the node
         // are bypassed.
-        if (args.CommitId == InvalidCommitId) {
-            return {0, Max<ui64>()};
+        if (args.CommitId == InvalidCommitId || node == args.Nodes.end()) {
+            return TByteRange::MaxEnd(0, blockSize);
         }
-
-        const ui64 blockSize = Tablet.GetBlockSize();
 
         ui64 begin = writeRange.MaxOffset;
         for (const auto& blob: args.MergedBlobs) {
-            begin = Min(begin, blob.Block.BlockIndex * blockSize);
+            begin = Min(
+                begin,
+                static_cast<ui64>(blob.Block.BlockIndex) * blockSize);
         }
         for (const auto& part: args.UnalignedDataParts) {
-            begin =
-                Min(begin, part.BlockIndex * blockSize + part.OffsetInBlock);
+            begin = Min(
+                begin,
+                static_cast<ui64>(part.BlockIndex) * blockSize +
+                    part.OffsetInBlock);
         }
 
-        auto node = args.Nodes.find(writeRange.NodeId);
-        if (node == args.Nodes.end()) {
-            return {0, Max<ui64>()};
+        const ui64 fileSize = node->Attrs.GetSize();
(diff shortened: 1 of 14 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_addblob.cpp
```diff
@@ -178,6 +178,41 @@ class TAddBlobsExecutor
...
+                Min(begin, part.BlockIndex * blockSize + part.OffsetInBlock);
+        }
+
+        auto node = args.Nodes.find(writeRange.NodeId);
+        if (node == args.Nodes.end()) {
+            return {0, Max<ui64>()};
+        }
+
+        if (writeRange.MaxOffset > node->Attrs.GetSize()) {
+            return {Min(begin, node->Attrs.GetSize()), Max<ui64>()};
```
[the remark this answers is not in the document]
neihar: added comment

cloud/filestore/libs/storage/tablet/tablet_actor_addblob.cpp
```diff
@@ -178,6 +178,41 @@ class TAddBlobsExecutor
         UpdateNodeAttrs(db, args);
     }
 
+    std::pair<ui64, ui64> CalculateCacheBypassRange(
```
qkrorlqr: let's not use `std::pair` - here `TByteRange` looks better

cloud/filestore/libs/storage/tablet/tablet_actor_addblob.cpp
```diff
@@ -178,6 +178,41 @@ class TAddBlobsExecutor
...
+        for (const auto& blob: args.MergedBlobs) {
+            begin = Min(begin, blob.Block.BlockIndex * blockSize);
+        }
+        for (const auto& part: args.UnalignedDataParts) {
+            begin =
+                Min(begin, part.BlockIndex * blockSize + part.OffsetInBlock);
+        }
+
+        auto node = args.Nodes.find(writeRange.NodeId);
+        if (node == args.Nodes.end()) {
```
qkrorlqr: how can this be possible?

cloud/filestore/libs/storage/tablet/tablet_cache_read_bypass.h
```diff
@@ -15,21 +15,50 @@ namespace NCloud::NFileStore::NStorage {
  */
 class TCacheReadBypass
 {
+private:
+    struct TActiveWrite
+    {
+        ui64 CommitId = 0;
+        // Affected byte range, End is exclusive. Writes that may change the
```
qkrorlqr: looks like something LLM-generated without having this https://github.com/ydb-platform/nbs/blob/main/cloud/storage/core/libs/common/byte_range.h in its context window

neihar: no, I explicitly used begin/end for better readability. Because we need end as Max for some cases, and with byte rage we need to do math, that looks ugly to my opinion, but if you want, I can do this options and we will see how it looks

qkrorlqr: What kind of math do you mean? You do explicit `<` / `>` comparisons here, `Max`, `Min`, etc whereas you could've used methods like `TByteRange::Overlaps()` instead and could've added other methods that you need (probably something like `TByteRange SuperRange(TByteRange x, TByteRange y)` which would return a minimum range containing both of those ranges)

neihar: I can use theoretically byte range + additional flag for size changing ops, but it will be also not so readable as this variant

cloud/filestore/libs/storage/tablet/tablet_cache_read_bypass.h
```diff
@@ -15,21 +17,51 @@ namespace NCloud::NFileStore::NStorage {
...
+
+    ui64 GetBypassedNodeReadCount() const;
+    ui64 GetBypassedRangeReadCount() const;
+
+private:
+    // range == nullptr checks reads of the whole node.
+    bool ShouldBypassReadImpl(
+        ui64 nodeId,
+        ui64 commitId,
+        const TByteRange* range) const;
```
neihar, line 57: I can do optional instead instead of pointer if needed

qkrorlqr (approved):

cloud/filestore/libs/storage/tablet/tablet_cache_read_bypass.h
qkrorlqr: IMO pointer is ok

yegorskii (approved):

Issue #6438, opened 2026-07-07
Feature request: expose cumulative per-volume availability counters


apliev:
## Motivation

blockstore-server currently exposes per-volume **instantaneous** health signals (e.g. `HasDowntime`, `CriticalSuffer` under the `server_volume` component). These tell you the state of a disk *right now*, but they don't directly answer questions about behavior *over time*, such as:

- What fraction of time has a disk been available over the last hour / day?
- How much time did a disk spend degraded?
- How do these numbers look per disk, for dashboards, reporting, or alerting?

To answer these today, a consumer has to integrate the 0/1 signals over time on its own (outside the server). That requires extra tooling, and integrating instantaneous point samples is sensitive to scrape/poll timing.

## Proposal

It would be useful if the server itself exposed **cumulative, time-based counters** per volume, accumulating how long each disk spent in each state:

- **observed time** - how long the disk was served (a natural denominator);
- **available time** - how long the disk was available (not in downtime);
- **healthy time** - how long the disk was available and not degraded.

These are nested "positive" counters (`observed >= available >= healthy`), so any SLI is a direct ratio of two counters with no inversion needed. Being monotonic counters, they let any monitoring/TSDB system compute availability directly from metrics, over an arbitrary window, e.g.:

- `availability = available / observed`
- `quality      = healthy / observed` (or `healthy / available`)
- degraded time, if needed, is simply `available - healthy`

## Why this is useful

- Availability / degradation over any time window becomes a simple ratio of two counters - no external integration step needed.
- Makes it straightforward to build uptime dashboards, availability reporting, and SLI/SLO-style alerting on top of blockstore metrics.
- A single server-side source of truth, consistent across consumers.
- Cumulative counters are far less sensitive to scrape timing than integrating instantaneous 0/1 signals externally.

## Behavior notes (high level)

- Counters are monotonic and accumulate **only while the volume is being served**. When a disk is not served, the counters simply don't advance - a gap means "no data for this period", not "0% available".
- A server restart resets the counters, which is the normal, expected behavior for cumulative counters and is handled by monitoring backends.

PR #6447, opened 2026-07-08
issue-6438: add per-volume cumulative availability counters

apliev:
### Notes
Add three monotonic derivative counters to TVolumeInfo in volume_stats.cpp: ObservedSeconds, AvailableSeconds, HealthySeconds. They advance once per finished update interval and are nested (observed >= available >= healthy), reflecting downtime (HasDowntime) and critical suffering (PerfCalc.IsSufferingCritically()) state at the time of each tick. Interval duration is measured via Timer->Now() to avoid assuming a fixed tick period. Add unit tests covering the basic accumulation, downtime, critical-suffering, and unmounted-volume cases.

### Issue
https://github.com/ydb-platform/nbs/issues/6438

### DIFF
--- a/cloud/blockstore/libs/diagnostics/volume_stats.cpp
+++ b/cloud/blockstore/libs/diagnostics/volume_stats.cpp
@@ -257,6 +257,14 @@ private:
     TRequestCounters RequestCounters;
     TDynamicCounters::TCounterPtr HasDowntimeCounter;
 
+    // Cumulative per-volume availability counters (derivative/RATE, seconds).
+    // Nested: ObservedSeconds >= AvailableSeconds >= HealthySeconds. Consumers
+    // compute availability = Available/Observed and quality = Healthy/Observed
+    // over a window. Advance only while the volume is served.
+    TDynamicCounters::TCounterPtr ObservedSecondsCounter;
+    TDynamicCounters::TCounterPtr AvailableSecondsCounter;
+    TDynamicCounters::TCounterPtr HealthySecondsCounter;
+
     TInstant LastRemountTime;
 
     // Number of pins on the object.
@@ -545,6 +553,11 @@ private:
     >;
     TDownDisksCounters DownDisksCounters;
     TDynamicCounters::TCounterPtr TotalDownDisksCounter;
+
+    // Wall-clock time of the last finished update interval, used to measure the
+    // real elapsed duration for the cumulative availability counters. Accessed
+    // only from UpdateStats (single periodic updater).
+    TInstant LastUpdateStatsTime;
 
 public:
     TVolumeStats(
@@ -884,6 +897,18 @@ public:
         ui32 totalDownDisks = 0;
         std::array<ui32, NProto::EStorageMediaKind_ARRAYSIZE> downDisksCounters{};
 
+        // Real duration of the finished update interval, used to advance the
+        // cumulative availability counters. Zero on the very first interval
+        // (no prior timestamp), which is fine for RATE metrics.
+        ui64 intervalSeconds = 0;
+        if (updateIntervalFinished) {
+            const auto now = Timer->Now();
+            if (LastUpdateStatsTime != TInstant::Zero()) {
+                intervalSeconds = (now - LastUpdateStatsTime).Seconds();
+            }
+            LastUpdateStatsTime = now;
+        }
+
         for (auto& [logicalDiskId, holder]: Volumes) {
             TVolumeInfoBase& volumeBase = *holder.VolumeBase;
 
@@ -905,12 +930,30 @@ public:
                     : EDowntimeStateChange::UP);
             }
 
+            const bool isSufferingCritically =
+                volumeBase.PerfCalc.IsSufferingCritically();
+            const bool isAvailable = !hasDowntime;
+            const bool isHealthy = isAvailable && !isSufferingCritically;
+
             for (auto& [key, instance]: holder.VolumeInfos) {
                 instance->RequestCounters.UpdateStats(updateIntervalFinished);
                 if (updateIntervalFinished) {
                     Y_DEBUG_ABORT_UNLESS(instance->HasDowntimeCounter);
                     if (instance->HasDowntimeCounter) {
                         *instance->HasDowntimeCounter = hasDowntime;
+                    }
+
+                    if (intervalSeconds) {
+                        if (instance->ObservedSecondsCounter) {
+                            *instance->ObservedSecondsCounter += intervalSeconds;
+                        }
+                        if (isAvailable && instance->AvailableSecondsCounter) {
+                            *instance->AvailableSecondsCounter +=
+                                intervalSeconds;
+                        }
+                        if (isHealthy && instance->HealthySecondsCounter) {
+                            *instance->HealthySecondsCounter += intervalSeconds;
+                        }
                     }
                 }
             }
@@ -1075,6 +1118,12 @@ private:
                 ->GetSubgroup("folder", volumeConfig.GetFolderId());
         info->RequestCounters.Register(*countersGroup);
         info->HasDowntimeCounter = countersGroup->GetCounter("HasDowntime");
+        info->ObservedSecondsCounter =
+            countersGroup->GetCounter("ObservedSeconds", true);
+        info->AvailableSecondsCounter =
+            countersGroup->GetCounter("AvailableSeconds", true);
+        info->HealthySecondsCounter =
+            countersGroup->GetCounter("HealthySeconds", true);
 
         auto reportZeroBlocksMetrics =
             !DiagnosticsConfig
--- a/cloud/blockstore/libs/diagnostics/volume_stats.cpp
+++ b/cloud/blockstore/libs/diagnostics/volume_stats.cpp
@@ -542,6 +542,14 @@ private:
         DiagnosticsConfig->GetExecutionTimeSizeClasses();
 
     TDynamicCountersPtr Counters;
+    // Separate, narrow counters tree (component=sli_volume) for the
+    // cumulative availability counters (ObservedSeconds/AvailableSeconds/
+    // HealthySeconds) only. component=server_volume also carries ~200-280
+    // other per-volume perf counters, so a UA scrape of the whole subtree
+    // just to reach our ~3 sensors is wasteful; this sibling group lets UA
+    // pull only what SLI needs. Only populated for EServerStats - client-side
+    // stats keep the previous (nested) placement, see RegisterInstance().
+    TDynamicCountersPtr SliCounters;
     std::shared_ptr<NUserCounter::IUserCounterSupplier> UserCounters;
     std::unique_ptr<TSufferCounters> SufferCounters;
     std::unique_ptr<TSufferCounters> SmoothSufferCounters;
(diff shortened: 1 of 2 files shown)


cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -905,13 +930,31 @@ class TVolumeStats final
...
                 instance->RequestCounters.UpdateStats(updateIntervalFinished);
                 if (updateIntervalFinished) {
                     Y_DEBUG_ABORT_UNLESS(instance->HasDowntimeCounter);
                     if (instance->HasDowntimeCounter) {
                         *instance->HasDowntimeCounter = hasDowntime;
                     }
+
+                    if (intervalSeconds) {
+                        if (instance->ObservedSecondsCounter) {
+                            *instance->ObservedSecondsCounter += intervalSeconds;
```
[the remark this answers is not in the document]
apliev: Confirmed, but this is existing by design behavior of this class, not something specific to the new counters: `UnmountVolume()` is intentionally a no-op here (an instance can back several endpoints / survive live migration), and instances are only removed by TrimVolumes() once InactiveClientsTimeout elapses. HasDowntimeCounter and RequestCounters already accrue during this same grace period. The new counters simply inherit that behavior. Didn't change the accrual logic and fixed the unit test instead -  it incorrectly asserted the counter freezes right after `UnmountVolume()` (using `InactiveClientsTimeout = TDuration::Max()`, which never trims). The test now calls `TrimVolumes()` with a finite timeout and asserts the counter stops advancing only after the instance is actually trimmed.

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -905,13 +930,31 @@ class TVolumeStats final
...
+                        }
+                        if (isAvailable && instance->AvailableSecondsCounter) {
```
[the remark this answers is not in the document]
apliev: Fixed. Moved accrual out of the if (`updateIntervalFinished`) gate so it now runs on every `UpdateStats` tick (~1s, `UpdateStatsInterval`) instead of only once per 15s publish tick (`UpdateCountersInterval`). This matches the actual sampling resolution of `DowntimeCalculator/PerfCalc`, which are already updated every tick. Short outages/degradations that recover within a 15s window are no longer swallowed. Added `TVolumeInfo::AvailabilityLastUpdateTime` (per-instance) to track exactly how much time has been credited; it advances only by the whole seconds credited each tick, avoiding sub-second drift.

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -884,6 +897,18 @@ class TVolumeStats final
...
         std::array<ui32, NProto::EStorageMediaKind_ARRAYSIZE> downDisksCounters{};
 
+        // Real duration of the finished update interval, used to advance the
+        // cumulative availability counters. Zero on the very first interval
+        // (no prior timestamp), which is fine for RATE metrics.
+        ui64 intervalSeconds = 0;
+        if (updateIntervalFinished) {
+            const auto now = Timer->Now();
+            if (LastUpdateStatsTime != TInstant::Zero()) {
+                intervalSeconds = (now - LastUpdateStatsTime).Seconds();
```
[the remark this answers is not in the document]
apliev: Fixed. Replaced the single global `TVolumeStats::LastUpdateStatsTime` with a per-instance `TVolumeInfo::AvailabilityLastUpdateTime`, seeded at construction time (`VolumeBase->Timer->Now()`, i.e. mount time). Each instance now only gets credited for the time it has actually existed, instead of being backdated to the last global finished tick. Added `ShouldCreditAvailabilityFromMountTime` covering a volume mounted mid-interval - it now correctly accrues only the ~5s it was actually served instead of the full ~15s tick.

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -913,6 +938,32 @@ class TVolumeStats final
...
+                // downtime/suffering signal at tick resolution; the per-instance
+                // timestamp (seeded at mount) avoids crediting time before the
+                // volume was served; advancing the timestamp only by the whole
+                // seconds credited keeps the sub-second remainder and avoids
+                // drift.
+                if (instance->ObservedSecondsCounter &&
+                    now > instance->AvailabilityLastUpdateTime)
+                {
+                    const ui64 seconds =
+                        (now - instance->AvailabilityLastUpdateTime).Seconds();
```
SvartMetal: If `now` jumps too far forward, for example because statistics have not been updated for a long time, we may incorrectly count all this "stall" time as availability time.

Maybe it would be better to add a safeguard and drop the current statistics increment if `now - instance->AvailabilityLastUpdateTime` is greater than `UpdateCountersInterval` (15 seconds), for example?

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -1075,6 +1126,12 @@ class TVolumeStats final
                 ->GetSubgroup("folder", volumeConfig.GetFolderId());
         info->RequestCounters.Register(*countersGroup);
         info->HasDowntimeCounter = countersGroup->GetCounter("HasDowntime");
+        info->ObservedSecondsCounter =
+            countersGroup->GetCounter("ObservedSeconds", true);
+        info->AvailableSecondsCounter =
+            countersGroup->GetCounter("AvailableSeconds", true);
+        info->HealthySecondsCounter =
+            countersGroup->GetCounter("HealthySeconds", true);
```
SvartMetal: These counters will not be reported during cyclic `blockstore-server` crashes.

This means we need to adjust the reported availability duration to account for `NODATA` intervals.

Maybe it would be simpler to calculate availability based on the number of reported "down" sensor `1` points versus the total number of `1`, `0`, and `NODATA` points?

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -884,6 +898,12 @@ class TVolumeStats final
         ui32 totalDownDisks = 0;
         std::array<ui32, NProto::EStorageMediaKind_ARRAYSIZE> downDisksCounters{};
 
+        // Wall-clock time of this stats tick. The cumulative availability
+        // counters advance on every tick (not only on the publish tick) by the
+        // real per-volume elapsed time, so state changes are sampled at tick
+        // resolution and newly mounted volumes are not over-credited.
+        const auto now = Timer->Now();
```
SvartMetal, line 916: Maybe we should use a monotonic clock here?

SvartMetal: But we definitely do not want to change the `DowntimeHistory` semantics.

cloud/blockstore/libs/diagnostics/volume_stats_ut.cpp
```diff
@@ -1194,6 +1194,315 @@ Y_UNIT_TEST_SUITE(TVolumeStatsTest)
...
+        auto observed = volumeCounters->GetCounter("ObservedSeconds");
+
+        // One accounted interval while served.
+        timer->AdvanceTime(TDuration::Seconds(15));
+        volumeStats->UpdateStats(true);
+        UNIT_ASSERT_VALUES_EQUAL(15, observed->Val());
+
+        // Unmount is a no-op; the instance is only removed by TrimVolumes once
+        // it has been inactive longer than the timeout (now - LastRemountTime
+        // = 15s > 10s).
```
SvartMetal: Then should we check that the counters are not removed until `TrimVolumes`?

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -257,6 +257,19 @@ class TVolumeInfo final
     TRequestCounters RequestCounters;
     TDynamicCounters::TCounterPtr HasDowntimeCounter;
 
+    // Cumulative per-volume availability counters (derivative/RATE, seconds).
+    // Nested: ObservedSeconds >= AvailableSeconds >= HealthySeconds. Consumers
+    // compute availability = Available/Observed and quality = Healthy/Observed
+    // over a window. Advance only while the volume is served.
```
SvartMetal: ```Advance only while the volume is being served, until the counters are trimmed.```

cloud/blockstore/libs/diagnostics/volume_stats.cpp
apliev: Agreed this is real, but it's not specific to how these counters are shaped. If `blockstore-server` is fully down, it can't report any signal - neither a duration counter nor a raw 1/0 "down" gauge, so the point-counting approach would hit the same NODATA gap during a full crash. 
Closing this really needs an independent observer that survives the NBS process dying (e.g. a second observed_seconds source), which is a separate architectural item outside this PR's scope. This PR targets "service alive but degraded/slow", not "service fully dead".

cloud/blockstore/libs/diagnostics/volume_stats_ut.cpp
apliev: Good point. I extended `ShouldStopAccruingAvailabilityAfterVolumeTrimmed` to assert the intermediate state: after `UnmountVolume()` (but before `TrimVolumes()`) the counter is still present (`FindCounter`) and keeps advancing on the next tick (15 → 30). Only after `TrimVolumes() `does it stop. This now explicitly covers that unmount alone does not remove the counters.

cloud/blockstore/libs/diagnostics/volume_stats.cpp
apliev: Done. Added the safeguard: if now - `AvailabilityLastUpdateTime > UpdateCountersInterval` the increment is dropped and the timestamp is just resynced to `now`, so a long stall / forward clock jump is not counted as availability. The threshold is a strict `>`, so a normal 15s publish tick is still credited. Covered by a new test `ShouldNotCreditLargeGapAsAvailability` (a 60s gap credits 0, the next normal 10s tick credits 10).

cloud/blockstore/libs/diagnostics/volume_stats.cpp
apliev: The tricky part here is that this same `now` also feeds `DowntimeHistory.PushBack(now, ...)`, whose timestamps are wall-clock and externally consumed - and as you noted we don't want to change that semantics. The counters, on the other hand, only need deltas, and the wall-clock hazards for them are already bounded on both sides: backward jumps are ignored by the existing `now > AvailabilityLastUpdateTime` guard (we don't credit and don't advance until the clock catches up), and forward jumps are now capped by the `UpdateCountersInterval` safeguard added above. So the residual error is at most sub-tick.

Going fully monotonic would mean keeping wall-clock for `DowntimeHistory` and introducing a second (monotonic) timer plus a second per-instance timestamp purely for the accrual delta - extra moving parts for a small gain. I'd prefer to keep the single wall-clock timer (`DowntimeHistory` untouched) and rely on the guard + cap.

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -1076,6 +1149,35 @@ class TVolumeStats final
...
+        // SliCounters comment) register the cumulative counters there;
+        // otherwise fall back to the nested per-volume group.
+        // "type" uses MediaKindToComputeType (not MediaKindToStatsString, the
+        // convention used by the DownDisks aggregate below) to produce the
+        // same disk-type spelling ("network-ssd", ...) this helper already
+        // produces elsewhere in the codebase.
+        // StorageMediaKind is immutable for the lifetime of a volume (unlike
+        // cloud/folder, which AlterVolume can change), so this extra level
+        // needs no re-registration/aliasing handling.
+        auto sliCountersGroup = SliCounters
```
nisqatsi: Не совсем понял вариант создания в countersGroup - SliCounters сейчас безусловно создается в InitCounters(). Т.е. время жизни такое же, как и у Counters. Проверки существования в строках 1230 и 1209 вроде как тоже не нужны

cloud/blockstore/libs/diagnostics/volume_stats.cpp
```diff
@@ -528,6 +543,16 @@ class TVolumeStats final
...
+    // Separate, narrow counters tree (component=sli_volume) for the
+    // cumulative availability counters (ObservedSeconds/AvailableSeconds/
+    // HealthySeconds) only. The main per-volume tree (component=
+    // server_volume / client_volume) also carries ~200-280 other per-volume
+    // perf counters, so scraping the whole subtree just to reach our ~3
+    // sensors is wasteful; this sibling group lets a monitoring agent pull
+    // only these specific counters. Populated for both EServerStats and
+    // EClientStats in InitCounters; see RegisterInstance() for the per-volume
+    // subgroup layout.
+    TDynamicCountersPtr SliCounters;
```
SvartMetal: AvailabilityCounters

looks clearer

nisqatsi: > AvailabilityCounters
>
> looks clearer



В эту группу возможно будут добавлены и другие виды SLI-метрик (индикаторов), так что я бы тут оставил привязку к назначению метрик, а не к их виду/типу, ведь именно из-за особого назначения - SLI - они и вынесены в эту отдельную группу.

apliev: Renamed `SliCounters` → `AvailabilityCounters` (and the local `sliCountersGroup` → `availabilityCountersGroup`, plus the test-local variables) for clarity. Kept the metric group name `component=sli_volume` unchanged. Thanks!

cloud/blockstore/libs/diagnostics/volume_stats.cpp
apliev: Спасибо, убрал устаревший код. Проверки действительно не нужны.

SvartMetal (approved):

nisqatsi (approved):

PR #6464, opened 2026-07-09
[Filestore] add direct RDMA flag

neihar:
### Notes
Added flag for enabling direct RDMA between vhost and server. This flag will be used for RDMA direct connection between vhost and filestore-server

### Issue
#5388

### DIFF
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -29,6 +29,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(PipeClientMinRetryTime,        TDuration, TDuration::Seconds(1)       )\
     xxx(PipeClientMaxRetryTime,        TDuration, TDuration::Seconds(4)       )\
                                                                                \
+    xxx(TabletDirectRdmaEnabled,       bool,      false                       )\
+                                                                               \
     xxx(EstablishSessionTimeout,       TDuration, TDuration::Seconds(30)      )\
     xxx(IdleSessionTimeout,            TDuration, TDuration::Hours(1)         )\
                                                                                \
--- a/cloud/filestore/config/storage.proto
+++ b/cloud/filestore/config/storage.proto
@@ -838,4 +838,8 @@ message TStorageConfig
     // If enabled, tablet overload is ignored when deciding whether to use
     // the unconfirmed data flow.
     optional bool AllowTabletOverload = 523;
+
+    // Enables the RDMA side channel for tablet-proxied requests in
+    // filestore-server and filestore-vhost.
+    optional bool TabletDirectRdmaEnabled = 524;
 }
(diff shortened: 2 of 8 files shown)


cloud/filestore/libs/vfs_fuse/config.cpp
```diff
@@ -56,6 +56,7 @@ namespace {
     xxx(GuestHandleKillPrivV2Enabled, bool,     false                         )\
     xxx(GuestPosixAclEnabled,         bool,     false                         )\
     xxx(ZeroCopyReadEnabled,          bool,     false                         )\
+    xxx(TabletDirectRdmaEnabled,            bool,     false                   )\
```
nisqatsi: выравнивание съехало

debnatkh (changes requested):
Link issue pls

cloud/filestore/libs/storage/core/config.cpp
```diff
@@ -29,6 +29,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(PipeClientMinRetryTime,        TDuration, TDuration::Seconds(1)       )\
     xxx(PipeClientMaxRetryTime,        TDuration, TDuration::Seconds(4)       )\
                                                                                \
+    xxx(TabletDirectRdmaEnabled,             bool,      false                  )\
```
drbasic: align

debnatkh (dismissed):

drbasic (dismissed):

cloud/filestore/libs/vfs_fuse/config.cpp
```diff
@@ -56,6 +56,7 @@ namespace {
     xxx(GuestHandleKillPrivV2Enabled, bool,     false                         )\
     xxx(GuestPosixAclEnabled,         bool,     false                         )\
     xxx(ZeroCopyReadEnabled,          bool,     false                         )\
+    xxx(TabletDirectRdmaEnabled,      bool,     false                         )\
```
antonmyagkov: Why do you need this flag on fuse level?

neihar: It is the existing per-fs propagation channel to the vhost side,  no? Also it can be need for fuse request size limits alignment

antonmyagkov: As I see from the ticket https://github.com/ydb-platform/nbs/issues/5388 flag is required only In TStorageServiceActor

>Also it can be need for fuse request size limits alignment

Can you provide more details?

neihar: We have limit on fuse level for max request size, also, RDMA will have it's own limit for request size. So, we need at least to check that limits and fuse limit is not bigger than RDMA. (theoretically RDMA can support any size, but in 1st implementation it will not be that way). But anyway, I removed it for now and will add in corresponding PR where it will be needed

antonmyagkov (dismissed):

drbasic (dismissed):

debnatkh (approved):

drbasic (approved):

Issue #6467, opened 2026-07-09
[Filestore] Enhance tests to cover more scenarios that emulate underlying system errors


neihar:
Enhance tests to cover more scenarios that emulate underlying system errors and other non-happy-path API variations.

The next measure should be implemented:

1) Add tablet restarts to integration tests.

Tablet restart already supported, see example https://github.com/ydb-platform/nbs/blob/ffbe22077efcb4148f8bbe272f7bb9d4a7473ec9/cloud/filestore/tests/fio_index/qemu-kikimr-multishard-tablets-restart-test/ya.make#L37
Tablet restarts should not affect operations. We can enable this for all integration tests by default. As initial step, enable only for cloud/filestore/tests/loadtest and tests which use nfs-storage-newfeatures-patch.txt 

2) Add page fault emulation to unit and integration tests

Currently we have [simple TX rescheduler that reschedules TX before PrepareTx is even started](https://github.com/ydb-platform/nbs/blob/ffbe22077efcb4148f8bbe272f7bb9d4a7473ec9/cloud/filestore/libs/storage/core/tablet.h#L150). That tests reorder case, but misses side effects from part of PrepareTx before it returns. 
Look if we can make some injection to [TIndexTabletDatabase](https://github.com/ydb-platform/nbs/blob/ffbe22077efcb4148f8bbe272f7bb9d4a7473ec9/cloud/filestore/libs/storage/tablet/tablet_database.h#L67) and run part of code before making reschedule for TX and make it identical to page fault. This injection should be zero cost for regular flows.

Rescheduling (with exclusion for page fault free TXes like [DeleteUnconfirmedData](https://github.com/ydb-platform/nbs/blob/f9e1d076cd9c3b08cbf5a74941f7807fdb8e44ad/cloud/filestore/libs/storage/tablet/tablet_actor_confirmadddata.cpp#L514)) can be enabled for all UTs and integration tests. 
As initial step, enable for UTs, loadtests and tests that use [nfs-storage-newfeatures-patch.txt ](https://github.com/ydb-platform/nbs/blob/ffbe22077efcb4148f8bbe272f7bb9d4a7473ec9/cloud/filestore/tests/common_configs/nfs-storage-newfeatures-patch.txt#L1)

3) Add BS errors injection
Filestore works with ydb Blob Storage (BS) in 2 places: vhost and tablet. 

- tablet
   on tablet we have 2 parts:
[...]

PR #6359, opened 2026-06-27
[Filestore] fix data loss if FlushBytes receives an error during ReadBlob or WriteBlob

SvartMetal:
### Notes
If `FlushBytesActor` receives an error during `ReadBlob` or `WriteBlob`, it unconditionally calls `TrimBytes`, which results in data loss.

### Issue
#6361

PR #6472, opened 2026-07-10
PoC bs interceptor

neihar:
### Notes

### Issue
#6467


SvartMetal:
Let's try reverting the change from https://github.com/ydb-platform/nbs/pull/6359, enable the BS interceptor by default, and see whether any of our tests reproduce the corruption

@tdiff

PR #6491, opened 2026-07-13
[NBS] Add grace period before stopping partitions after GC

agalibin:
During mass NBS node restarts, volume tablets reboot and start partitions for garbage collection (GC). Once GC completes, partitions are stopped immediately. If a client attempts to mount the disk moments after GC finishes, it has to wait for partitions to boot again from scratch — adding latency and contributing to mount timeouts, especially for encrypted disks that depend on synchronous KMS/IAM/Compute calls.

Added a configurable grace period (`StopPartitionsAfterGcGracePeriod`, milliseconds, default 0 = disabled) that defers partition stop after all partitions report GC completion.


agalibin:
@codex review

agalibin:
@codex review

Issue #6208, opened 2026-06-12
[Filestore] observability/tooling improvements


SvartMetal:
Metrics:
* Report the O_CREAT flag in CreateHandle metrics
* Split CreateHandle metrics into "by (parent_id, name)", "by node_id"

Tooling:
* Implement a filesystem status tool that outputs the state of all filesystem shards, including uptime, load, and other relevant information (whether the shard is active, state of the bootstrap process(whether its still loading the compaction map), whether writes are allowed, whether the compaction map is loaded, compaction progress, etc.)
* Per-file: most loaded files per shard and slowest (needs to be collected in tablet code) -> output for the whole filesystem
* Implement a tool (or extend the same one) that outputs recent slow requests for a given filesystem

PR #6493, opened 2026-07-13
Introducing a DiagnoseFilesystem command

szumal:
### Notes
Added a new FileStore command that gets the StorageStats on the main tablet to fetch an overview of current load on all of the shards

### Issue

### Follow-up
Following offline discussion, the next PRs will include:

1. Adding a MaxFileCount to loadtest
2. Richer output for humans in the tool (Highlighted fields, progress bar, etc.)
3. More diagnostic features to the tool (activity, state of loading the compact map, etc. as specified in https://github.com/ydb-platform/nbs/issues/6208). The tool/command shall be renamed to diagnosefs (or similar) when new features are added.


cloud/filestore/tests/python/lib/client.py
```diff
@@ -300,6 +300,23 @@ def forced_compaction(self, fs):
 
         return common.execute(cmd, env=self.__env, check_exit_code=self.__check_exit_code).stdout
 
+    def diagnose_shards(self, fs, top=None, sort_by="load"):
+        cmd = [
+            self.__binary_path, "diagnose-shards",
```
qkrorlqr: does this work? seems that the command is called "diagnoseshards", not "diagnose-shards"

cloud/filestore/tests/python/lib/daemon_config.py
```diff
@@ -237,6 +237,12 @@ def __generate_storage_txt(self, disableLocalService=False):
     def __generate_diag_txt(self):
         config = TDiagnosticsConfig()
 
+        def fill_performance_profile(profile):
+            profile.Read.RPS = 10
```
qkrorlqr: let's pass `profile` from the calling code instead of hardcoding it here

cloud/filestore/tests/client_sharded/test.py
```diff
@@ -368,3 +427,95 @@ def _l(path, symlink):
 
     ret = common.canonical_file(results_path, local=True)
     return ret
+
+
+def test_diagnose_shards_with_loadtest():
```
qkrorlqr: this test is already much more complex than a normal client test - let's create a separate build target for it - e.g. `tests/diagnostic_test`

cloud/filestore/apps/client/lib/diagnose_filesystem.cpp
```diff
@@ -0,0 +1,181 @@
...
+            return true;
+        }
+
+        Cout << "Filesystem: " << FileSystemId << Endl;
+        Cout << "Shard count: " << rows.size() << Endl;
+        Cout << "Top loaded shards:" << Endl;
+
+        for (size_t i = 0; i < limit; ++i) {
+            const auto& row = rows[i];
+            Cout << i + 1 << ". "
```
qkrorlqr, line 160: let's generate a richer output for humans - most of the time the results will be used not by scripts but by humans so having things like colored output, highlighting of outlier values, etc would be nice

you can do it in another PR

also in the future you'll add collection of per-file (for highly loaded files) stats and then adding a progress bar will be great, etc, etc

cloud/filestore/tests/client_sharded/test.py
```diff
@@ -368,3 +427,95 @@ def _l(path, symlink):
...
+                break
+
+            if not workload.running:
+                workload.wait(check_exit_code=False)
+                break
+
+        if workload is not None and workload.running:
+            workload.wait(check_exit_code=False)
+
+        assert target_shards is not None and target_shards.issubset(seen_active_shards), (
```
qkrorlqr: I suggest to replace explicit assertions with canonical test results (with all unstable fields removed / stabilized somehow) - this way we'll be able to see the output (coarsened output) of the diagnoseshards tool command in the repository and probably it will also be easier to modify the tool in the future (will require less modifications to the test code)

cloud/filestore/tests/python/lib/client.py
szumal: It works because `nbs/cloud/filestore/apps/client/lib/app.cpp` normalizes the names. Regardless, I'll rename the usage to exactly match the command name.

cloud/filestore/apps/client/lib/diagnose_shards.cpp
```diff
@@ -0,0 +1,181 @@
...
+    {
+        TString ShardId;
+        ui64 CurrentLoad = 0;
+        ui32 Suffer = 0;
+        ui64 UsedBlocksCount = 0;
+        ui64 TotalBlocksCount = 0;
+        ui64 UsedNodesCount = 0;
+    };
+
+    ui32 Top = 10;
```
qkrorlqr: better to set these defaults via Opts (`.DefaultValue(xxx)`) - in this case they'll be seen in `--help` output

cloud/filestore/apps/client/lib/diagnose_shards.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#include <google/protobuf/util/json_util.h>
+
+namespace NCloud::NFileStore::NClient {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TDiagnoseShardsCommand final
```
qkrorlqr: let's rename this command to `DiagnoseFilesystem`

qkrorlqr (commented):
looks fine to me overall

qkrorlqr (approved):

agalibin (approved):

PR #6496, opened 2026-07-13
Fix missing return statement in HandleChangeVolumeBindingRequest

efimov-ivan-yteam:
Fix missing return statement in HandleChangeVolumeBindingRequest in no-action-required check.
Could potentially cause redundant remounts


cloud/blockstore/libs/storage/service/volume_session_actor_binding.cpp
```diff
@@ -69,6 +69,7 @@ void TVolumeSessionActor::HandleChangeVolumeBindingRequest(
         if (VolumeInfo->BindingType == NProto::BINDING_LOCAL) {
             replyError(
                 MakeError(S_ALREADY, "Volume is already running at host"));
+            return;
```
SvartMetal, line 72: We need test

Issue #6518, opened 2026-07-14
Fix `"address already in use"` errors in nemesis tests caused by port collisions


SvartMetal:
The port collision occurs due to the following race condition:

1. A recipe acquires a port using `PortManager`, which holds a file lock for that port during recipe startup.
2. `unstable-process` is launched, along with its child process that binds to the acquired port.
3. Recipe startup completes, and the file lock is released.
4. `unstable-process` terminates its child process.
5. Another recipe starts and, by chance, acquires the same port because nothing is holding it at that moment: the child process is down, the socket is free, and the file lock has already been released.
6. The second recipe launches its own `unstable-process` using the same port. At this point, two `unstable-process` instances are competing for the same port. They do not fail because `unstable-process` is designed to tolerate `"address already in use"` errors.
7. As a result, the test may exhibit very confusing behavior: child processes from different recipes can alternately own the same port, and requests from one test suite may be routed to a completely different test suite's service. This makes the logs extremely difficult to interpret.

A similar race can also occur when the second recipe launches a regular process (for example, a daemon) instead of `unstable-process`. In that case, the outcome is more straightforward: the daemon simply fails with an `"address already in use"` error.

PR #6519, opened 2026-07-14
issue-6518: fix "address already in use" errors in nemesis tests caused by port collisions

SvartMetal:
### Problem

Nemesis tests (daemons wrapped in the `unstable-process` launcher) sporadically fail with
"address already in use": after a restart the daemon cannot re-bind its own ports because a
concurrently running test suite has grabbed them.

The root cause is a lifetime mismatch. yatest `PortManager` reserves a port by holding a
flock on `<PORT_SYNC_PATH>/<port>`, but that flock only lives as long as the *allocating*
process (e.g. a recipe), while the port itself is used for the whole lifetime of the
launcher. Once the allocator exits, every restart gap - the moment between the child's
sockets being freed and the new instance binding them - leaves the port both unlocked and
unbound. A concurrent suite's `PortManager` (which takes the flock first and bind-checks
second) can then legitimately capture it.

### Fix

Keep the flocks held for the launcher's whole lifetime by handing them over from the
allocator to the launcher:

* **`port_reservation.py` - `PortManager`**: a wrapper over yatest `PortManager` that
  memorizes allocated ports (`reserve_port()` / `list_reserved_ports()`), so runners can
  tell the launcher which ports to protect.
* **`port_reservation.py` - `PortReservation`**: used inside `unstable-process`; re-takes
  the same flocks (same path convention, same `library.python.filelock.FileLock`) and holds
  them until the launcher exits. Requires `PORT_SYNC_PATH` (asserts otherwise).
* **`unstable-process --reserve-port`** (repeatable): before every terminate/kill — i.e.
  before the child's sockets are freed - the launcher calls `reserve_ports()`, so the port
  is never observable as both unlocked and unbound.
* **Integration**: `disk_agent_runner.py` (blockstore) and `daemon_config.py` (filestore)
  allocate ports through the wrapper and pass the reserved set to the launcher.

Acquisition is **non-blocking and best-effort with a short grace period** (retry in 1 ms
steps, up to 2 s per call). A single non-blocking attempt is not enough: a concurrent
`PortManager` holds the flock for a moment while it bind-checks the port (the check fails
while our child is alive, so the flock comes right back), and an attempt landing in that
window would otherwise free the sockets unprotected. A holder that survives the whole grace
period is not transient - it is the allocating process itself, which keeps the port
protected for as long as it lives, so the launcher proceeds and retries before the next
restart. At any moment at least one of the three holds the port: the allocator's flock, the
child's bound socket, or the launcher's flock.

Additionally, the kill path now reaps the killed child (`process.wait()`) before starting
the next instance: its sockets are only guaranteed to be freed once it is gone, and the new
instance binds the same ports.

### Testing

* `cloud/storage/core/tools/common/python/ut`: unit tests for the wrapper and for
  `PortReservation` acquire/release semantics.
* `cloud/storage/core/tools/testing/unstable-process/tests`: integration tests running the
  real launcher with a dummy daemon under attack from `PortStealer` threads that mimic a
  concurrent suite (flock, then bind, keep both on success):
  * `test_cannot_steal_reserved_port` - with `--reserve-port` the port survives multiple
    restart gaps under contention;
  * `test_steals_port_without_reservation` - without the option the stealer reproduces the
    original race (daemon dies with EADDRINUSE, launcher exits with code 3), proving the
    test setup actually exercises the bug.

### Issue
#6518

### DIFF
--- a/cloud/blockstore/tests/python/lib/disk_agent_runner.py
+++ b/cloud/blockstore/tests/python/lib/disk_agent_runner.py
@@ -12,6 +12,7 @@ from cloud.blockstore.config.storage_pb2 import TStorageServiceConfig
 from cloud.blockstore.config.server_pb2 import TServerAppConfig, TKikimrServiceConfig, TServerConfig, TLocation
 from cloud.storage.core.config.features_pb2 import TFeaturesConfig
 from cloud.storage.core.tools.common.python.core_pattern import core_pattern
+from cloud.storage.core.tools.common.python.port_reservation import PortManager
 from cloud.storage.core.tools.testing.access_service.lib import AccessService
 from contrib.ydb.core.protos.auth_pb2 import TAuthConfig
 from contrib.ydb.core.protos.config_pb2 import TActorSystemConfig
@@ -61,11 +62,11 @@ class LocalDiskAgent(Daemon):
                 dict(name="dynamic_storage_pool:2", kind="ssd")]
 
         self.__service_type = LocalDiskAgent.__get_service_type(server_app_config)
-        self.__port_manager = yatest_common.PortManager()
+        self.__port_manager = PortManager()
         self.__grpc_port = grpc_port
         self.__grpc_trace = grpc_trace
-        self.__ic_port = self.__port_manager.get_port()
-        self.__mon_port = self.__port_manager.get_port()
+        self.__ic_port = self.__port_manager.reserve_port()
+        self.__mon_port = self.__port_manager.reserve_port()
         self.__temporary_agent = temporary_agent
         if disk_agent_binary_path is not None:
             self.__binary_path = disk_agent_binary_path
@@ -131,8 +132,8 @@ class LocalDiskAgent(Daemon):
         self.__access_service = None
         if enable_access_service:
             host = "localhost"
-            port = self.__port_manager.get_port()
-            control_server_port = self.__port_manager.get_port()
+            port = self.__port_manager.reserve_port()
+            control_server_port = self.__port_manager.reserve_port()
             self.__access_service = AccessService(host, port, control_server_port)
             self.__proto_configs["auth.txt"] = self.__generate_auth_txt(port)
 
@@ -557,6 +558,17 @@ ModifyScheme {
             command = [
                 launcher_path,
             ] + self.__unstable_process_args
+
+            # All ports allocated by this agent (ic, mon, access-service,
+            # ...) are reserved by PortManager (flocked under PORT_SYNC_PATH)
+            # only while the allocating process lives.
+            #
+            # Pass the reserved ports to the launcher so it re-takes and holds
+            # those flocks across every restart.
+            #
+            # See issue #6518 for details
+            for reserved_port in self.__port_manager.list_reserved_ports():
+                command += ["--reserve-port", str(reserved_port)]
 
             for cmd in commands:
                 command += [
--- a/cloud/filestore/tests/python/lib/daemon_config.py
+++ b/cloud/filestore/tests/python/lib/daemon_config.py
@@ -21,6 +21,7 @@ from cloud.filestore.config.storage_pb2 import TStorageConfig
 
 from cloud.storage.core.protos.media_pb2 import STORAGE_MEDIA_HDD
 from cloud.storage.core.tools.testing.access_service.lib import AccessService
+from cloud.storage.core.tools.common.python.port_reservation import PortManager
 
 logger = logging.getLogger(__name__)
 
@@ -81,17 +82,17 @@ class FilestoreDaemonConfigGenerator:
         self.__restart_interval = restart_interval
         self.__restart_flag = restart_flag
 
-        self._port_manager = yatest_common.PortManager()
-        self.__port = self._port_manager.get_port()
-        self.__mon_port = self._port_manager.get_port()
-        self.__ic_port = self._port_manager.get_port() if ic_port is None else ic_port
+        self._port_manager = PortManager()
+        self.__port = self._port_manager.reserve_port()
+        self.__mon_port = self._port_manager.reserve_port()
+        self.__ic_port = self._port_manager.reserve_port() if ic_port is None else ic_port
         self.__access_service_type = access_service_type
         self.__access_service_port = access_service_port
 
         self.__use_secure_registration = use_secure_registration
 
         if secure:
-            self.__secure_port = self._port_manager.get_port()
+            self.__secure_port = self._port_manager.reserve_port()
             self.__app_config.ServerConfig.SecurePort = self.__secure_port
 
         with open(self.__app_config_file_path, "w") as config_file:
@@ -366,7 +367,21 @@ class FilestoreDaemonConfigGenerator:
                 str(self.__restart_interval),
                 "--cmdline",
                 " ".join(command),
-            ] + (["--allow-restart-flag", self.__restart_flag] if self.__restart_flag else [])
+            ]
+
+            if self.__restart_flag:
+                command += ["--allow-restart-flag", self.__restart_flag]
+
+            # All ports allocated by the PortManager (server, mon, ic, secure,
+            # local-service, ...) are reserved (flocked under PORT_SYNC_PATH)
+            # only until the recipe process exits.
+            #
+            # Pass the full set to the launcher so it re-takes and holds those
+            # flocks across every restart.
+            #
+            # See issue #6518 for details
+            for reserved_port in self._port_manager.list_reserved_ports():
+                command += ["--reserve-port", str(reserved_port)]
 
         return command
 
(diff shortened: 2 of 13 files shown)


cloud/storage/core/tools/common/python/port_reservation.py
```diff
@@ -0,0 +1,113 @@
...
+            if time.monotonic() > deadline:
+                logging.error(
+                    'could not reserve ports %s within %d seconds; another '
+                    'process may already hold them - possible cross-suite '
+                    'port collision; restarting without reservation',
+                    sorted(pending), self._grace_seconds)
+                break
+            time.sleep(1)
+
+        self._reserve_done = True
```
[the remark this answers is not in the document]
SvartMetal: Fixed

cloud/storage/core/tools/common/python/port_reservation.py
```diff
@@ -0,0 +1,113 @@
...
+    reserve_ports() is called synchronously right before every terminate/kill,
+    it does all its work once, before the first restart, so the sockets are
+    never freed while the ports are unreserved.
+
+    Introduced to fix a race with unreserved ports (see issue #6518 for details)
+    """
+
+    def __init__(self, ports, grace_seconds=60):
+        self._sync_dir = os.environ.get('PORT_SYNC_PATH')
+        assert len(self._sync_dir) != 0
```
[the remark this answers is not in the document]
SvartMetal: I think it is better to assert `sync_dir` to keep the code less surprising

cloud/filestore/tests/python/lib/daemon_config.py
```diff
@@ -366,7 +367,21 @@ def generate_command(self):
...
+            # All ports allocated by the PortManager (server, mon, ic, secure,
+            # local-service, ...) are reserved (flocked under PORT_SYNC_PATH)
+            # only until the recipe process exits.
+            #
+            # Pass the full set to the launcher so it re-takes and holds those
+            # flocks across every restart.
+            #
+            # See issue #6518 for details
+            for reserved_port in self._port_manager.list_reserved_ports():
+                command += ["--reserve-port", str(reserved_port)]
```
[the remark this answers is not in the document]
SvartMetal: It is OK, this port is bound by FilestoreServer which is not wrapped by unstable-process https://github.com/ydb-platform/nbs/blob/dcaff5861b096ceed68131566f41244a735dc69d/cloud/filestore/tests/recipes/service-kikimr/__main__.py#L169

cloud/blockstore/tests/python/lib/disk_agent_runner.py
```diff
@@ -558,6 +559,17 @@ def append_conf_file_arg(command, config_path, option_name, conf_file):
...
+            # All ports allocated by this agent (ic, mon, access-service,
+            # ...) are reserved by PortManager (flocked under PORT_SYNC_PATH)
+            # only while the allocating process lives.
+            #
+            # Pass the reserved ports to the launcher so it re-takes and holds
+            # those flocks across every restart.
+            #
+            # See issue #6518 for details
+            for reserved_port in self.__port_manager.list_reserved_ports():
+                command += ["--reserve-port", str(reserved_port)]
```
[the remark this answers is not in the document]
SvartMetal: Fixed

cloud/storage/core/tools/common/python/ut/port_reservation_ut.py
```diff
@@ -0,0 +1,68 @@
...
+    pm = PortManager()
+    port = pm.reserve_port()
+    pm.release_port(port)
+
+    reservation = PortReservation([port])
+    reservation.reserve_ports()
+    assert_lock_failure(sync_dir, port)
+
+    reservation.release()
+    assert_lock_success(sync_dir, port)
```
librarian, line 68:
```
P2: the lock handoff is still best-effort, not atomic. [reserve_ports()](https://github.com/ydb-platform/nbs/blob/95989135018778397186d38d9c77352309e8b4f2/cloud/storage/core/tools/common/python/port_reservation.py#L89-L125) gives up after two seconds with ports still pending, after which [unstable-process unconditionally kills the child](https://github.com/ydb-platform/nbs/blob/95989135018778397186d38d9c77352309e8b4f2/cloud/storage/core/tools/testing/unstable-process/__main__.py#L129-L147). If the existing lock owner releases between that timeout and child termination, the unlocked-and-unbound window returns. The positive test only covers releasing the allocator before the first reservation attempt. A real ownership handshake, or making restart conditional on acquiring every lock, would provide the claimed guarantee.
```

I asked codex to reproduce issue and it produced this test:

```
def test_allocator_release_after_reservation_timeout():
    sync_dir = setup_sync_dir()
    allocator = PortManager()
    port = allocator.reserve_port()

    listener = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
    listener.bind(("::", port))
    listener.listen(1)

    reservation = PortReservation(
        [port],
        grace_seconds=0.01,
        retry_step_seconds=0.001,
    )

    try:
        # The launcher cannot acquire the flock while the allocator still
        # owns it, so reserve_ports() returns after its grace period.
        reservation.reserve_ports()

        # Reproduce the vulnerable ordering: the allocator disappears after
        # the grace period, immediately before the launcher kills the daemon.
        allocator.release_port(port)
        listener.close()
        listener = None

        # A completed handoff must keep the port locked independently of the
        # original allocator. The current best-effort implementation does not.
        assert_lock_failure(sync_dir, port)
    finally:
        if listener is not None:
            listener.close()
        allocator.release_port(port)
        reservation.release()
```

cloud/storage/core/tools/common/python/port_reservation.py
```diff
@@ -0,0 +1,130 @@
...
+
+        assert ports
+        self._ports = sorted(set(ports))
+
+        self._locks = {}
+        self._grace_seconds = grace_seconds
+        self._retry_step_seconds = retry_step_seconds
+        self._reserve_done = False
+
+    def reserve_ports(self):
```
librarian, line 89: It should return boolean value because otherwise we wouldn't know if process of reservation was successful.

And in case of failed reservation we should stop.

cloud/blockstore/tests/python/lib/disk_agent_runner.py
librarian, line 571: We have double ownership here because LocalDiskAgent owns the ports in a normal way (and doesn't need new PortManager per-se) and if we were to return status of port reservation unstable process would get false because it doesn't own ports.

For filestore it is different, because we reserve port AND exits after we launch unstable process.

cloud/storage/core/tools/testing/unstable-process/__main__.py
```diff
@@ -97,28 +111,43 @@ def main():
...
                 while args.allow_restart_flag is not None and not os.path.exists(args.allow_restart_flag):
                     logging.debug("waiting for the allow restart flag")
                     time.sleep(1)
                     continue
 
+                # Hold the port flocks *before* the child's sockets are freed,
+                # so the ports are never observable as both unlocked and unbound.
+                #
+                # Needed to fix a race with unreserved ports (see issue #6518 for details)
+                if reservation:
```
librarian, line 133: And this part should look like this

```
if reservation and not reservation.reserve_ports():
    # Reservation was unsuccessful. Do not perform the restart.
    process.kill()
    process.wait()
    return 4
```

librarian (changes requested):

cloud/blockstore/tests/python/lib/disk_agent_runner.py
SvartMetal: Yes, that is why `PortReservation` has a grace period. It is currently two seconds, and for `LocalDiskAgent`, this delay will be added before every restart. I think this is a reasonable trade-off.

Alternatively, I can disable `PortReservation` for `LocalDiskAgent` completely

cloud/storage/core/tools/common/python/port_reservation.py
SvartMetal: We could use exceptions in this case, but I would prefer to keep this function best effort, meaning it becomes a no-op if port reservation fails. 

That way, it does not introduce failures on its own. If a port collision occurs during testing, we can investigate the root cause using the logs

cloud/storage/core/tools/testing/unstable-process/__main__.py
SvartMetal: I want to use a best-effort strategy here, explained in https://github.com/ydb-platform/nbs/pull/6519#discussion_r3656457914

cloud/storage/core/tools/common/python/ut/port_reservation_ut.py
SvartMetal: Yes, the current solution is not ideal, but I hope it will be good enough for real-world scenarios

librarian (approved):

komarevtsev-d (approved):

PR #6532, opened 2026-07-15
[NBS] Move read blob info logic to separate file

vladstepanyuk:
### Notes
Move read blob info logic to separate file
### Issue
#6279

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
@@ -1,6 +1,7 @@
 #include "part_actor.h"
 
 #include "part_compaction_logic.h"
+#include "part_readblobinfo_logic.h"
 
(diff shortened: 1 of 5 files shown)


cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1986,47 +2006,32 @@ bool TPartitionActor::PrepareCompaction(
...
-            continue;
-        }
-
-        for (auto& rangeCompaction: args.RangeCompactions) {
-            if (auto* ab = rangeCompaction.AffectedBlobs.FindPtr(blobId)) {
-                ab->BlobMeta = *blobMeta;
-            }
-        }
-    }
+    FillRangeCompactionInfos(
```
qkrorlqr: do we need to fill compaction infos if `ready == false`?

qkrorlqr (approved):

komarevtsev-d (approved):

Issue #5790, opened 2026-04-21
Investigate performance degradation with IoDepth > RecvQueueSize


tpashkin:
When iodepth is greater than RECV queue size (server or client, doesn't matter) we see very bad latencies up to ~4-5 seconds per request

tpashkin:
So, the answer lies in our default configuration

`rnr_cnt=7` is a special value, meaning infinite RNR retries. In other words individual requests will be retried indefinitely in case the peer doesn't have outstanding RECVs – precisely our case. Another part of the puzzle is the default timeout value – `min_rnr_timer=0` which is ~600ms

Good news are that issue can be pretty easily resolved by decreasing RNR timeout
```
tpashkin@node-1315:~$ ./rdma-test --test Initiator --storage Rdma --host node-1316 --iodepth 512 --send-queue-size 32 --recv-queue-size 256 --verbose --wait BusyWait
...
IOPS: 721, BW(r): 1.38 MiB/s, BW(w): 1.43 MiB/s, BW(rw): 2.81 MiB/s
...
```
`min_rnr_timer=13` gives us about 1ms timeout
```
tpashkin@node-1315:~$ ./rdma-test --test Initiator --storage Rdma --host node-1316 --iodepth 512 --send-queue-size 32 --recv-queue-size 256 --verbose --wait BusyWait --qp-min-rnr-timer 13
...
IOPS: 203347, BW(r): 397.64 MiB/s, BW(w): 396.68 MiB/s, BW(rw): 794.32 MiB/s
...
```

Proper fix would be simply propagating this error back to the client, introducing some backpressure

PR #6542, opened 2026-07-16
use memory windows to invalidate request buffers; resolves issue-5729

### LINKED ISSUE #5729: Invalidate request buffers
Recently we caught a scary behaviour of our RDMA.

Timeline:
1. Client sends request `R₁` 
   → Server receives TRequestMessage 
   → Server has no free SendWr → queues RDMA READ for `R₁`

2. `MaxResponseDelay` pass. Client timeout fires.
   → AbortRequest(`R₁`) → FreeRequest → InBuffer returned to pool

3. Client sends new request `R₂` 
   → AllocateRequest → pool returns the **same** buffer
   → Client PostSend → server receives new TRequestMessage

4. Server finally gets a free SendWr
   → Dequeues `R₁`'s RDMA READ and executes it
   → But it's `InBuffer` now contains `R₂`'s data!
   → Server parses `R₂` data as if it's `R₁`

5. Server also processes `R₂` normally (from step 3)

Result: `R₂`'s `VolumeRequestId` hits the RDMA tracker TWICE:
  - Once from the stale `R₁` processing (step 4)
  - Once from the real `R₂` processing (step 5)

The `R₂` hitting disk agent backend the second time triggers this critical event:
```
Apr  1 09:03:29 <fqdn> NBS_DISK_AGENT[2223481]: CRITICAL_EVENT:DiskAgentCriticalEvents/UnexpectedIdentifierRepetition:DeviceUUID="b987831b17464759ff370b11fa39ba9d"; requestId="4:49244506"; range=[21119808..21119871]
```

### Proposed solution
We should avoid using memory regions with a "static" rkey for its lifetime. Instead, we could use memory windows through `ibv_wr_bind_mw()` and `ibv_wr_local_inv()`. These would allow us to revoke the `rkey` from a memory window on request abort and assign a new one when allocating.

Other option would to be simply drop the connection on the request timeout – this will invalidate the current PD along with all keys. Not a good option for request cancellation though, since it will basically cancel all requests

tpashkin:
use memory windows to invalidate request buffers
#5729

### DIFF
--- a/cloud/storage/core/libs/rdma/impl/client.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client.cpp
@@ -38,6 +38,8 @@
 #include <util/system/mutex.h>
 #include <util/system/thread.h>
 
+#include <deque>
+
 namespace NCloud::NStorage::NRdma {
 
 using namespace NMonitoring;
@@ -151,8 +153,8 @@ struct TRequest
     ui64 ClientReqId = 0;
 
     // set upon receiving response
-    ui32 Status;
-    ui32 ResponseBytes;
+    ui32 Status = 0;
+    ui32 ResponseBytes = 0;
 
     TPooledBuffer InBuffer{};
     TPooledBuffer OutBuffer{};
@@ -623,6 +625,7 @@ private:
     TEventHandle DisconnectEvent;
 
     TSimpleList<TRequest> QueuedRequests;
+    std::deque<ui32> InvalidationQueue;
     TActiveRequests ActiveRequests;
 
     std::atomic<ui64> ReqIdPool{0};
@@ -692,6 +695,7 @@ public:
 private:
     // called from CQ thread
     void HandleQueuedRequests() noexcept;
+    bool HandleInvalidations() noexcept;
     void StartRequest(TRequestPtr req, TSendWr* send) noexcept;
     void SendRequest(TRequest* req, TSendWr* send) noexcept;
     void SendRequestCompleted(TSendWr* send) noexcept;
@@ -701,11 +705,11 @@ private:
     void BindCompleted(TSendWr* send) noexcept;
     void RecvResponse(TRecvWr* recv) noexcept;
     void RecvResponseCompleted(TRecvWr* recv) noexcept;
-    void InvalidateInBuffer(TRequestPtr req, TSendWr* send) noexcept;
+    void InvalidateInBuffer(TRequest* req, TSendWr* send) noexcept;
     void InvalidateOutBuffer(TRequest* req, TSendWr* send) noexcept;
     void InvalidateCompleted(TSendWr* send) noexcept;
-    void InvalidateBuffers(TRequestPtr req) noexcept;
-    void CompleteRequest(TRequestPtr req) noexcept;
+    void InvalidateBuffers(TRequest* req) noexcept;
+    void CompleteRequest(ui32 reqId) noexcept;
     void AbortRequest(TRequestPtr req, ui32 err, const TString& msg) noexcept;
     void FreeRequest(TRequest* creq) noexcept;
     bool PostSend(TRequest* req, TSendWr* send, ibv_send_wr* wr) noexcept;
@@ -1055,18 +1059,25 @@ bool TClientEndpoint::HandleInputRequests() noexcept
         RequestEvent.Clear();
     }
 
-    auto requests = InputRequests.DequeueAll();
-    if (!requests) {
-        return false;
+    bool hasWork = false;
+
+    if (auto requests = InputRequests.DequeueAll()) {
+        QueuedRequests.Append(std::move(requests));
+        HandleQueuedRequests();
+        hasWork = true;
     }
 
-    QueuedRequests.Append(std::move(requests));
-    HandleQueuedRequests();
-    return true;
+    if (HandleInvalidations()) {
+        hasWork = true;
+    }
+
+    return hasWork;
 }
 
 void TClientEndpoint::HandleQueuedRequests() noexcept
 {
+    // in Poll mode events can get through in any state, so it's important to
+    // cancel requests here, otherwise HandleInputEvent may get lost
     if (!CheckState(EEndpointState::Connected)) {
         while (QueuedRequests) {
             auto req = QueuedRequests.Dequeue();
@@ -1085,20 +1096,43 @@ void TClientEndpoint::HandleQueuedRequests() noexcept
 
         auto req = QueuedRequests.Dequeue();
         Y_ABORT_UNLESS(req);
-
         Counters->RequestDequeued();
 
-        switch (req->State) {
-            case ERequestState::Enqueued:
-                StartRequest(std::move(req), send);
-                break;
+        if (req->State != ERequestState::Enqueued) {
+            RDMA_ERROR(
+                "request " << req->ReqId << " has unexpected state "
+                            << req->State);
 
-            case ERequestState::InvalidateBuffers:
-                Counters->InvalidationDequeued();
-                InvalidateInBuffer(std::move(req), send);
-                break;
+            QueuedRequests.Enqueue(std::move(req));
+            Counters->RequestEnqueued();
 
-            default:
+            SendQueue.Push(send);
+            Counters->Error();
+            Disconnect();
+            return;
+        }
+
+        StartRequest(std::move(req), send);
+    }
+}
+
+bool TClientEndpoint::HandleInvalidations() noexcept
+{
+    bool hasWork = InvalidationQueue.size();
+
+    while (InvalidationQueue.size()) {
+        auto* send = SendQueue.Pop();
+        if (!send) {
+            // no more WRs available
+            break;
+        }
+
+        auto reqId = InvalidationQueue.front();
+        InvalidationQueue.pop_front();
+        Counters->InvalidationDequeued();
+
+        if (auto* req = ActiveRequests.Get(reqId)) {
+            if (req->State != ERequestState::InvalidateBuffers) {
                 RDMA_ERROR(
                     "request " << req->ReqId << " has unexpected state "
                                << req->State);
@@ -1106,9 +1140,16 @@ void TClientEndpoint::HandleQueuedRequests() noexcept
                 SendQueue.Push(send);
                 Counters->Error();
                 Disconnect();
-                return;
+                return true;
+            }
+
+            Counters->InvalidationStarted();
+            InvalidateInBuffer(req, send);
         }
+        // request has been cancelled or timed out
     }
+
+    return hasWork;
 }
 
 void TClientEndpoint::CancelRequest(ui64 clientRequestId) noexcept
@@ -1212,6 +1253,9 @@ void TClientEndpoint::AbortRequests() noexcept
         RDMA_DEBUG("abort request " << req->ReqId);
         AbortRequest(std::move(req), E_RDMA_UNAVAILABLE, "endpoint is unavailable");
     }
+
+    // drop all invalidations
+    InvalidationQueue.clear();
 }
 
 void TClientEndpoint::AbortRequest(
@@ -1669,7 +1713,7 @@ void TClientEndpoint::RecvResponseCompleted(TRecvWr* recv) noexcept
     RecvResponse(recv);
 
     // request has been aborted
-    auto req = ActiveRequests.Pop(reqId);
+    auto* req = ActiveRequests.Get(reqId);
     if (!req) {
         return;
     }
@@ -1683,13 +1727,13 @@ void TClientEndpoint::RecvResponseCompleted(TRecvWr* recv) noexcept
         req->CallContext->RequestId);
 
     if (Config.UseMemoryWindows) {
-        InvalidateBuffers(std::move(req));
+        InvalidateBuffers(req);
     } else {
-        CompleteRequest(std::move(req));
+        CompleteRequest(req->ReqId);
     }
 }
 
-void TClientEndpoint::InvalidateBuffers(TRequestPtr req) noexcept
+void TClientEndpoint::InvalidateBuffers(TRequest* req) noexcept
 {
     req->State = ERequestState::InvalidateBuffers;
 
@@ -1700,8 +1744,7 @@ void TClientEndpoint::InvalidateBuffers(TRequestPtr req) noexcept
 
     RDMA_TRACE("invalidate request " << req->ReqId << " buffers");
 
-    InputRequests.Enqueue(std::move(req));
-    Counters->RequestEnqueued();
+    InvalidationQueue.push_back(req->ReqId);
     Counters->InvalidationEnqueued();
 
     if (WaitMode == EWaitMode::Poll) {
@@ -1709,10 +1752,12 @@ void TClientEndpoint::InvalidateBuffers(TRequestPtr req) noexcept
     }
 }
 
-void TClientEndpoint::CompleteRequest(TRequestPtr req) noexcept
+void TClientEndpoint::CompleteRequest(ui32 reqId) noexcept
 {
-    RDMA_TRACE("request " << req->ReqId << " completed");
+    auto req = ActiveRequests.Pop(reqId);
+    Y_ABORT_UNLESS(req);
 
+    RDMA_TRACE("request " << reqId << " completed");
     Counters->RequestCompleted();
 
     auto* handler = req->Handler.get();
@@ -1722,16 +1767,11 @@ void TClientEndpoint::CompleteRequest(TRequestPtr req) noexcept
     handler->HandleResponse(std::move(req), status, responseBytes);
 }
 
-void TClientEndpoint::InvalidateInBuffer(
-    TRequestPtr request,
-    TSendWr* send) noexcept
+void TClientEndpoint::InvalidateInBuffer(TRequest* req, TSendWr* send) noexcept
 {
-    auto* req = request.get();
-    req->State = ERequestState::InvalidateInBuffer;
-    Counters->InvalidationStarted();
-
-    ActiveRequests.Push(std::move(request));
     send->context = reinterpret_cast<void*>(static_cast<uintptr_t>(req->ReqId));
+
+    req->State = ERequestState::InvalidateInBuffer;
 
     ibv_send_wr wr = {
         .wr_id = send->wr.wr_id,
@@ -1805,7 +1845,7 @@ void TClientEndpoint::InvalidateCompleted(TSendWr* send) noexcept
             MemoryWindows.Release(std::move(req->OutMemoryWindow));
             Counters->ReleaseMemoryWindow();
             Counters->InvalidationCompleted();
-            CompleteRequest(ActiveRequests.Pop(reqId));
+            CompleteRequest(req->ReqId);
             break;
 
         default:
--- a/cloud/storage/core/libs/rdma/impl/client.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client.cpp
@@ -75,6 +75,32 @@ using TCompletionPollerPtr = std::unique_ptr<TCompletionPoller>;
 
 ////////////////////////////////////////////////////////////////////////////////
 
+enum class ERequestState
+{
+    BindInBuffer,
+    BindOutBuffer,
+    SendRequest,
+    RecvResponse,
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+inline IOutputStream& operator<<(IOutputStream& out, ERequestState state)
+{
+    static const char* string[] = {
+        "BindInBuffer",
+        "BindOutBuffer",
+        "SendRequest",
+        "RecvResponse",
+    };
+    if (static_cast<size_t>(state) < Y_ARRAY_SIZE(string)) {
+        return out << string[static_cast<size_t>(state)];
+    }
+    return out << "Undefined";
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
 struct TRequest
     : TClientRequest
     , TListNode<TRequest>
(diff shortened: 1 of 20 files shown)


tpashkin:
Current version works, but is extremely slow. Next step would be to set up the cache for memory windows

@EvgeniyKozev @yegorskii I'll be on vacation for the next 2 weeks, so feel free to pick this up

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1164,29 +1206,165 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+    req->State = ERequestState::BindInBuffer;
+
+    ibv_send_wr wr = {
+        .wr_id = send->wr.wr_id,
+        .opcode = IBV_WR_BIND_MW,
+        .bind_mw = {
+            .mw = req->InMemoryWindow.get(),
+            .rkey = req->InMemoryWindow->rkey,
+            .bind_info = {
+                .mr = req->InBuffer.GetMemoryRegion(),
```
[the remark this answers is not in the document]
tpashkin: Interestingly, mlx5 provider doesn't really enforce this

tpashkin:
> Current version works, but is extremely slow. Next step would be to set up the cache for memory windows
> 
> @EvgeniyKozev @yegorskii I'll be on vacation for the next 2 weeks, so feel free to pick this up

Implemented MW cache. Doesn't seem to be any degradation – https://gist.github.com/tpashkin/e4de6d87469f5e497c922de1b80ba262

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,29 +1322,164 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+    auto* msg = send->Message();
+    Zero(*msg);
+
+    InitMessageHeader(msg, NegotiatedProtocolVersion);
+    msg->ReqId = req->ReqId;
+    msg->In = req->InBuffer;
+    msg->Out = req->OutBuffer;
+
+    if (Config.UseMemoryWindows) {
+        req->InMemoryWindow = MemoryWindows.Acquire();
```
tpashkin: In theory it's possible for `In/OutBuffer` to have 0 length, but no client does this right now. For the sake of simplicity I'm going to leave it out of this PR and make a separate issue

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,29 +1322,164 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+        return;
+    }
+
+    LWTRACK(
+        BindInBufferStarted,
+        req->CallContext->LWOrbit,
+        req->CallContext->RequestId);
+}
+
+void TClientEndpoint::BindOutBuffer(TRequest* req, TSendWr* send) noexcept
```
budevg, line 1574: looks same as BindInBuffer maybe make single function with direction parameter (in/out)

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1269,31 +1550,167 @@ void TClientEndpoint::RecvResponseCompleted(TRecvWr* recv) noexcept
...
+        return;
+    }
+
+    LWTRACK(
+        InvalidateInBufferStarted,
+        req->CallContext->LWOrbit,
+        req->CallContext->RequestId);
+}
+
+void TClientEndpoint::InvalidateOutBuffer(
```
budevg, line 1794: same as InvalidateInBuffer let's consolidate into function with buffer type param (in/out)

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,29 +1322,164 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+void TClientEndpoint::BindInBuffer(TRequest* req, TSendWr* send) noexcept
+{
+    req->State = ERequestState::BindInBuffer;
+
+    ibv_send_wr wr = {
+        .wr_id = send->wr.wr_id,
+        .opcode = IBV_WR_BIND_MW,
+        .bind_mw = {
+            .mw = req->InMemoryWindow.get(),
+            .rkey = req->InMemoryWindow->rkey,
```
budevg: shouldn't you rotate rkey using `ibv_inc_rkey` to avoid reusing it ?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1782,16 +2203,22 @@ class TCompletionPoller final
...
-                    TStringBuilder() << "request " << reqId << " timed out "
-                                     << "[peer=" << endpoint->Host << ":"
-                                     << endpoint->Port << "]");
+                    TStringBuilder()
+                        << "request " << req->ReqId << " timed out "
+                        << "[peer=" << endpoint->Host << ":" << endpoint->Port
+                        << "]");
+
+                if (Config->UseMemoryWindows) {
+                    endpoint->InvalidateBuffers(std::move(req));
```
budevg: This moves the req into QueuedRequests but no cancel request is registered or cancel event signaled. So there is possibility that the request will be stuck without completing it

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -980,31 +1092,54 @@ bool TClientEndpoint::HandleCancelRequests() noexcept
...
     cancelled.Append(QueuedRequests.DequeueIf(wasCancelled));
+
     while (auto req = cancelled.Dequeue()) {
         RDMA_TRACE("request " << req->ReqId << " cancelled");
         Counters->RequestDequeued();
-        AbortRequest(std::move(req), E_CANCELLED, "request was cancelled");
+        req->Response = MakeError(E_CANCELLED, "request was cancelled");
+
+        if (Config.UseMemoryWindows) {
+            InvalidateBuffers(std::move(req));
```
budevg: is it possible that request is canceled before it's memory window initialized and in/out regions bounded? In this case InvalidateBuffers will lead to segmentation fault or undefined behaviour

cloud/storage/core/libs/rdma/impl/memory_window.cpp
```diff
@@ -0,0 +1,34 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+void TMemoryWindowsPool::Init(
+    NVerbs::IVerbsPtr verbs,
+    ibv_pd* protectionDomain,
+    size_t capacity)
+{
+    Verbs = verbs;
+    ProtectionDomain = protectionDomain;
+    Capacity = capacity;
```
budevg, line 16: you should clean the pool as well since in reconnect it contains objects with old `pd`

budevg (changes requested):

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: I can do this, but quite sure the code will lose clarity

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Probably yes, but not here. Normally we invalidate buffers after we've already received a response, so this only matters for cancelled and timed out requests

The original version actually just destroyed memory windows in such cases, which is much cleaner way to solve the problem. But since we share workers between different endpoints, this may slow down other endpoints as well

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Yeah, already fixed. Going to add in the next iteration

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Indeed it is. I'm actually leaning towards abandoning this quest trying to invalidate aborted requests and simply destroying tainted memory windows

cloud/storage/core/libs/rdma/impl/memory_window.cpp
tpashkin: In reality PD doesn't really change, but you are right, this behavior is not guaranteed – will do

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: > I'm actually leaning towards abandoning this quest trying to invalidate aborted requests and simply destroying tainted memory windows

Did exactly that

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: > The original version actually just destroyed memory windows in such cases, which is much cleaner way to solve the problem

Returned to that version

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Okay, I've done what I could – https://github.com/ydb-platform/nbs/pull/6542/changes/ed8299d90826382a8303f6480890abee45ad5612

> maybe make single function with direction parameter (in/out)

I think this specific change would make them harder to reason about. I've made a few wrappers instead

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,30 +1312,97 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
     }
 }
 
-void TClientEndpoint::SendRequest(TRequestPtr req, TSendWr* send) noexcept
+void TClientEndpoint::StartRequest(TRequestPtr request, TSendWr* send) noexcept
```
budevg, line 1481: TClientEndpoint::StartRequest marked as noexcept but it's actually can throw now following `emoryWindows.Acquire()` call

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -684,15 +766,24 @@ void TClientEndpoint::CreateQP()
...
+            << "SendQueueSize * 2 = " << Config.MemoryWindowsPoolSize);
+    }
+
+    MemoryWindows.Init(Verbs, Connection->pd, Config.MemoryWindowsPoolSize);
+
     SendBuffers.Init(
         Verbs,
         Connection->pd,
-        IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_READ);
+        IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_READ | IBV_ACCESS_MW_BIND);
```
budevg: should be added only when `UseMemoryWindows` enabled here and below

cloud/storage/core/libs/rdma/impl/client_ut.cpp
```diff
@@ -1401,4 +1401,171 @@ TEST(TRdmaClientTest, ShouldDisconnectOnUnknownProtocolVersionInRejectMessage)
...
+                        testContext->RecvEvents.pop_front();
+                        testContext->ProcessedRecvEvents.push_back(recv);
+                        testContext->CompletionHandle.Set();
+                        break;
+                    }
+                }
+            }
+        };
+
+        // normal completion will invalidate it's buffers
```
budevg: it's buffers -> its buffers

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,30 +1312,97 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
     }
 }
 
-void TClientEndpoint::SendRequest(TRequestPtr req, TSendWr* send) noexcept
+void TClientEndpoint::StartRequest(TRequestPtr request, TSendWr* send) noexcept
 {
+    // hand request over to simplify error handling
+    auto* req = request.get();
     req->ReqId = ActiveRequests.CreateId();
+    ActiveRequests.Push(std::move(request));
```
budevg, line 1486: it's now possible that active request is aborted before calling SendRequestStarted() which increases the `ActiveRequests` counter. Then the abort will decrease the counter without increasing it in first place.

Same thing if disconnect happens during in flight invalidate. abort request will be called without decreasing the counter

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1269,31 +1517,151 @@ void TClientEndpoint::RecvResponseCompleted(TRecvWr* recv) noexcept
...
+{
+    req->State = ERequestState::InvalidateBuffers;
+
+    LWTRACK(
+        RequestQueuedForInvalidation,
+        req->CallContext->LWOrbit,
+        req->CallContext->RequestId);
+
+    InputRequests.Enqueue(std::move(req));
+    Counters->RequestEnqueued();
```
budevg: consider adding counters for at least MW allocation failures, created/reused windows, pool occupancy, and requests pending invalidation

cloud/storage/core/libs/rdma/impl/client.cpp
SvartMetal: @tpashkin @yegorskii let's avoid exceptions in production code

SvartMetal:
@tpashkin AI reviews:
* https://claude.ai/share/324b4700-118b-4a41-8557-29b271bae161
* https://chatgpt.com/share/6a8b31e6-e35c-83eb-8521-9bcae9920de2

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: What difference does it make?

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Fair point, I'll review the counters

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: > TClientEndpoint::StartRequest marked as noexcept but it's actually can throw now following `MemoryWindows.Acquire` call

Good point, I'll fix that

> let's avoid exceptions in production code

They are not supposed to leak, that's why `noexcept` is here

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Actually, scratch that. Buffers have to be at least `RDMA_PROTO_HEADER_SIZE`
https://github.com/ydb-platform/nbs/blob/7b9fa821caed9693ee2a45b1c147dfa3bf988967/cloud/storage/core/libs/rdma/iface/protobuf.cpp#L112-L116
https://github.com/ydb-platform/nbs/blob/7b9fa821caed9693ee2a45b1c147dfa3bf988967/cloud/storage/core/libs/rdma/iface/protobuf.cpp#L23-L26

cloud/storage/core/libs/rdma/impl/client.cpp
tpashkin: Initially I've been on a fence about adding the counters for memory windows – historically they've been quite useless. But then again, in light of #5790, we don't really know the root cause and anything could help

> MW allocation failures

These are nothing special, they fall into the generic error category

> pool occupancy

Doesn't seem to be very useful. We can have a stable 1 iops load which will guarantee this to be 0 or we can go over the pool capacity and still have zero

Metrics usefulness is still very limited by the fact they are aggregated across all endpoints. Probably should create an issue for that

tpashkin:
So regarding the AI reviews. First claude – https://claude.ai/share/324b4700-118b-4a41-8557-29b271bae161

> 1. The rkey is never rotated, which undermines the whole point of the feature

This is wrong. Happy path is guaranteed to invalidate rkey's and timeout / cancellation will drop the memory window altogether, which will also invalidate the rkey. Furthermore `ibv_inc_rkey` has a working range of only 256 values, since it just rotates 8 LSB of the rkey

Nevertheless this API can be utilized for asynchronous invalidation. By rotating rkey with `ibv_inc_rkey` we can schedule invalidation and return MW to the pool right away. This will ensure that the next 255 requests using this MW can't really access the original buffer. However, it doesn't solve the problem of server writing to the response buffer at the same time user handler is working, so we would still have to drop or invalidate memory window upon timeout or cancellation

> 2. Null-pointer dereference for zero-sized buffers

Buffers have to be at least `RDMA_PROTO_HEADER_SIZE`
https://github.com/ydb-platform/nbs/blob/7b9fa821caed9693ee2a45b1c147dfa3bf988967/cloud/storage/core/libs/rdma/iface/protobuf.cpp#L112-L116
https://github.com/ydb-platform/nbs/blob/7b9fa821caed9693ee2a45b1c147dfa3bf988967/cloud/storage/core/libs/rdma/iface/protobuf.cpp#L23-L26

> 3. Successful responses get converted into errors

I don't really see it as a problem, just an implementation detail. Same thing applies to the requests, that have been successfully executed, but server has been unable to send the response in time

> 4. MW destruction races with in-flight WRs

The spec is not very clear about this. I'd assume any bind on destroyed MW will fail, but this needs to checked. Realistically though, our timeout values are measured in minutes and local completions take microseconds, so it's unlikely to ever become a problem. Either way I think this should be a separate issue

> 5. `MemoryWindows.Acquire()` exception

Fixed

> 6. Counter/gauge corruption

Added new counters and fixed the existing ones

> 7. Retry-on-transient-failure regression

This is intentional, all requests should be aborted on disconnect

> 8. Failed bind/invalidate completions aren't handled per-request

This is wrong. All errors except flush (which itself means we are disconnecting) lead to disconnect

> 9. HandleQueuedRequests default branch drops the request silently

This is true, but the default branch should be unreachable. Aborting request might be reasonable, need to think about it

> RecvResponseCompleted now silently returns on an unknown reqId, dropping the old RDMA_WARN + error counter — a genuinely bogus/corrupt response is now indistinguishable from the legit "request was aborted" case

This happens frequently enough to be a nuisance in the logs. We've actually removed the code that stores recently completed/aborted requests to distinguish these cases

> TMemoryWindowsPool has uninitialized ProtectionDomain/Capacity members, and the default capacity (SendQueueSize * 2) can be exceeded by in-flight windows (requests hold windows through the response wait, and ActiveRequests isn't bounded by SendQueueSize), so under load you get constant ibv_alloc_mw/dealloc_mw churn while Release silently destroys the overflow.

This is intentional. Proper fix would be an in-flight limit

> TRequest::Status/ResponseBytes are uninitialized members — currently always set before use, but a foot-gun.

Fair enough, I'll add the initialization

> Test nits: ASSERT_GE(6, bound) only asserts the upper bound although the comment promises 4–6; the cancellation subtest's assertions depend on timing of when the cancel lands.

Not true, previous ASSERT ensures the counter to be 4

> The ibv_wr_opcode name table in verbs.h is position-dependent on the libibverbs enum and lives in the global namespace; fragile if the enum grows.

`operator<<` must be in the global namespace. As for the enum growth – yes, but it's extremely unlikely for upstream to change the enum values

tpashkin:
Now chatgpt – https://chatgpt.com/share/6a8b31e6-e35c-83eb-8521-9bcae9920de2

> 1. P1: Reusing the same rkey defeats the entire invalidation scheme

Wrong. Already covered this in the previous comment

> 2. P1: ibv_alloc_mw() failure terminates the process

Already fixed

> 3. P1: Zero-length request or response buffers cause a null dereference

Wrong. Covered in the previous comment

> 4. P2: Request IDs become reusable before the request has actually completed

Fair point, will think about it

> 5. P2: Request and send counters are no longer paired with state transitions

Already fixed

> 6. P2, provider-dependent: The feature changes MR registration even when disabled

Already fixed

tpashkin:
Follow-up to AI review

> HandleQueuedRequests default branch drops the request silently

I've update the code to push request back into the `QueuedRequests`. It will be handled by the normal disconnect flow later

> TRequest::Status/ResponseBytes are uninitialized members — currently always set before use, but a foot-gun.

Initialized to 0

> P2: Request IDs become reusable before the request has actually completed

Changed the flow to use a separate invalidation queue without popping request from the `ActiveRequests` map

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,30 +1432,124 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+        BindInBuffer(req, send);
+    }
+    catch (const TServiceError& e) {
         RDMA_ERROR(send << " " << e.what());
+
+        if (req->InMemoryWindow) {
+            req->InMemoryWindow.reset();
+            Counters->ReleaseMemoryWindow();
+        }
+
```
budevg, line 1546: The following failure sequence seems to be possible which leads to unbalanced counters
```
  Failure sequence:

  1. HandleQueuedRequests() decrements QueuedRequests.
  2. StartRequest() inserts the request into ActiveRequests and increments the ActiveRequests gauge.
  3. MemoryWindows.Acquire() throws.
  4. BindBuffers() releases the partially acquired window, returns the send WR, and starts disconnecting.
  5. The request still has state Enqueued.
  6. AbortRequests() eventually calls AbortRequest(), which treats it as queued:
      - Decrements QueuedRequests again.
      - Does not decrement ActiveRequests.
      - Does not increment AbortedRequests.
```

maybe fix this like this
```
       catch (const TServiceError& e) {                                                                                       
           RDMA_ERROR(send << " " << e.what());                                                                               
           if (req->InMemoryWindow) {                                                                                         
               req->InMemoryWindow.reset();                                                                                   
               Counters->ReleaseMemoryWindow();                                                                               
           }                                                                                                                  
           SendQueue.Push(send);                                                                                              
           Counters->Error();                                                                                                 
           Counters->RequestAborted();                                                                                        
           AbortRequest(ActiveRequests.Pop(req->ReqId), E_RDMA_UNAVAILABLE, e.what());                                        
           Disconnect();                                                                                                      
       }                                                                                                                      
```

budevg (changes requested):

budevg (approved):

komarevtsev-d (approved):

yegorskii (approved):

PR #6544, opened 2026-07-16
[Filestore] Inject index tablet database failures

tdiff:
### Notes
This change allows randomly to restart transactions in tests using TTestEnv. Random page faults are enabled for a subset of unit and integration tests.

Currently there is some gap in testing transactions restarted due to "page faults" in YDB, which reschedule and restart read transactions after some delay. As we did not find a way to trigger page faults in YDB, I extended previously existing failure injection to all Read* methods of TInMemoryIndexState. 

This PR is a rework of https://github.com/ydb-platform/nbs/pull/6423.

### Issue
#6467

### DIFF
--- a/cloud/filestore/config/storage.proto
+++ b/cloud/filestore/config/storage.proto
@@ -848,4 +848,8 @@ message TStorageConfig
 
     // Pass data buffer to the tablet as a message payload
     optional bool ExternalWriteDataPayloadEnabled = 526;
+
+    // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
+    optional bool FakeTxPageFaultsEnabled = 527;
+    optional double FakeTxPageFaultsProbabilityPercentage = 528;
 }
--- a/cloud/filestore/config/storage.proto
+++ b/cloud/filestore/config/storage.proto
@@ -851,5 +851,5 @@ message TStorageConfig
 
     // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
     optional bool FakeTxPageFaultsEnabled = 527;
-    optional double FakeTxPageFaultsProbabilityPercentage = 528;
+    optional double FakeTxPageFaultsProbability = 528;
 }
--- a/cloud/filestore/config/storage.proto
+++ b/cloud/filestore/config/storage.proto
@@ -850,6 +850,5 @@ message TStorageConfig
     optional bool ExternalWriteDataPayloadEnabled = 526;
 
     // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
-    optional bool FakeTxPageFaultsEnabled = 527;
-    optional double FakeTxPageFaultsProbability = 528;
+    optional double FakeTxPageFaultsProbability = 527;
 }
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -384,7 +384,6 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(SoftBackpressureMaxReadIops,                   ui32,    100'000       )\
                                                                                \
     xxx(ExternalWriteDataPayloadEnabled,               bool,    false         )\
-    xxx(FakeTxPageFaultsEnabled,                       bool,    false         )\
     xxx(FakeTxPageFaultsProbability,                   double,   0            )\
 // FILESTORE_STORAGE_CONFIG
 
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -385,7 +385,7 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
                                                                                \
     xxx(ExternalWriteDataPayloadEnabled,               bool,    false         )\
     xxx(FakeTxPageFaultsEnabled,                       bool,    false         )\
-    xxx(FakeTxPageFaultsProbabilityPercentage,         double,   0            )\
+    xxx(FakeTxPageFaultsProbability,                   double,   0            )\
 // FILESTORE_STORAGE_CONFIG
 
 #define FILESTORE_STORAGE_CONFIG_REF(xxx)                                      \
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -384,6 +384,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(SoftBackpressureMaxReadIops,                   ui32,    100'000       )\
                                                                                \
     xxx(ExternalWriteDataPayloadEnabled,               bool,    false         )\
+    xxx(FakeTxPageFaultsEnabled,                       bool,    false         )\
+    xxx(FakeTxPageFaultsProbabilityPercentage,         double,   0            )\
 // FILESTORE_STORAGE_CONFIG
 
 #define FILESTORE_STORAGE_CONFIG_REF(xxx)                                      \
--- a/cloud/filestore/libs/storage/core/config.h
+++ b/cloud/filestore/libs/storage/core/config.h
@@ -441,7 +441,6 @@ public:
 
     [[nodiscard]] bool GetExternalWriteDataPayloadEnabled() const;
 
-    [[nodiscard]] bool GetFakeTxPageFaultsEnabled() const;
     [[nodiscard]] double GetFakeTxPageFaultsProbability() const;
 };
 
--- a/cloud/filestore/libs/storage/core/config.h
+++ b/cloud/filestore/libs/storage/core/config.h
@@ -440,6 +440,9 @@ public:
     ui32 GetSoftBackpressureMaxReadIops() const;
 
     [[nodiscard]] bool GetExternalWriteDataPayloadEnabled() const;
+
+    [[nodiscard]] bool GetFakeTxPageFaultsEnabled() const;
+    [[nodiscard]] double GetFakeTxPageFaultsProbabilityPercentage() const;
 };
 
 }   // namespace NCloud::NFileStore::NStorage
--- a/cloud/filestore/libs/storage/core/config.h
+++ b/cloud/filestore/libs/storage/core/config.h
@@ -442,7 +442,7 @@ public:
     [[nodiscard]] bool GetExternalWriteDataPayloadEnabled() const;
 
     [[nodiscard]] bool GetFakeTxPageFaultsEnabled() const;
-    [[nodiscard]] double GetFakeTxPageFaultsProbabilityPercentage() const;
+    [[nodiscard]] double GetFakeTxPageFaultsProbability() const;
 };
 
 }   // namespace NCloud::NFileStore::NStorage
--- a/cloud/filestore/libs/storage/core/tablet.h
+++ b/cloud/filestore/libs/storage/core/tablet.h
@@ -143,26 +143,6 @@ protected:
             Generation = tx.Generation;
             Step = tx.Step;
 
-            // Reschedule (restart) the transaction from PrepareTx when the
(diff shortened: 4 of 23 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -5,6 +5,7 @@
 #include <cloud/filestore/libs/diagnostics/critical_events.h>
 #include <cloud/filestore/libs/diagnostics/metrics/registry.h>
 #include <cloud/filestore/libs/storage/tablet/model/throttler_logger.h>
+#include "cloud/filestore/libs/storage/tablet/tablet_database_failure_injection.h"
```
qkrorlqr: <>

cloud/filestore/libs/storage/init/actorsystem.cpp
```diff
@@ -246,7 +246,11 @@ class TCustomLocalServiceInitializer final
                     systemCounters,
                     metricsRegistry,
                     fastShardServer,
-                    CreateNoOpTxRescheduler());
+                    config->GetFakeTxPageFaultsEnabled() ?
+                        CreateRescheduler({
+                            .ProbabilityPercentage = config->GetFakeTxPageFaultsProbabilityPercentage(),
```
qkrorlqr:
```
$ echo -n '                            .ProbabilityPercentage = config->GetFakeTxPageFaultsProbabilityPercentage(),'|wc -c
104
```

cloud/filestore/libs/storage/core/tablet_tx_rescheduler.cpp
```diff
@@ -1,28 +1,73 @@
...
-class TNoOpTxRescheduler final
+// With a given probability decides if a transaction should be rescheduled.
+// Used in tests to exercise behaviour in case of transaction restarts/reorderings.
+// Note that it does not reschedule anything by itself and only serves for communication
+// between TIndexActorDatabase and Tx::Execute().
+class TRandomTxRescheduler final
     : public ITxRescheduler
 {
+private:
+    const float Probability = 0;
```
qkrorlqr: let's use `double` for such things (because why not)

cloud/filestore/libs/storage/core/tablet.h
```diff
@@ -176,6 +156,23 @@ class TTabletBase
 
                 Args.Clear();
                 Args.OnRestart();
+
+                // Reschedules the transaction from PrepareTx assuming the rescheduler
+                // was triggered during PrepareTx() call.
+                if (Y_UNLIKELY(Self->TxRescheduler && Self->TxRescheduler->IsTriggered())) {
```
qkrorlqr:
```
$ echo -n '                if (Y_UNLIKELY(Self->TxRescheduler && Self->TxRescheduler->IsTriggered())) {'|wc -c
92
```

cloud/filestore/libs/storage/core/tablet.h
```diff
@@ -176,6 +156,23 @@ class TTabletBase
...
+                    Self->TxRescheduler->Reset();
+                    LOG_INFO(
```
qkrorlqr: probably `LOG_INFO` is too much - IMO `LOG_DEBUG` is enough for such things

cloud/filestore/libs/storage/core/config.cpp
```diff
@@ -384,6 +384,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(SoftBackpressureMaxReadIops,                   ui32,    100'000       )\
                                                                                \
     xxx(ExternalWriteDataPayloadEnabled,               bool,    false         )\
+    xxx(FakeTxPageFaultsEnabled,                       bool,    false         )\
+    xxx(FakeTxPageFaultsProbabilityPercentage,         float,   0             )\
```
qkrorlqr: let's get rid of the `Enabled` flag and just treat prob==0 as "disabled"

cloud/filestore/config/storage.proto
```diff
@@ -848,4 +848,8 @@ message TStorageConfig
 
     // Pass data buffer to the tablet as a message payload
     optional bool ExternalWriteDataPayloadEnabled = 526;
+
+    // Enables injecting of errors in TIndexTabletDatabase. Must be only used in tests!
+    optional bool FakeTxPageFaultsEnabled = 527;
+    optional float FakeTxPageFaultsProbabilityPercentage = 528;
```
qkrorlqr: nit: IMO `TxPageFaultInjectionProbability` sounds better

cloud/filestore/libs/storage/core/config.h
```diff
@@ -440,6 +440,9 @@ class TStorageConfig
     ui32 GetSoftBackpressureMaxReadIops() const;
 
     [[nodiscard]] bool GetExternalWriteDataPayloadEnabled() const;
+
+    [[nodiscard]] bool GetFakeTxPageFaultsEnabled() const;
+    [[nodiscard]] float GetFakeTxPageFaultsProbabilityPercentage() const;
```
qkrorlqr: `double`

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1821,15 +1822,29 @@ std::unique_ptr<IIndexTabletDatabase>
 TIndexTabletActor::CreateIndexTabletDatabase(
     NKikimr::NTable::TDatabase& database)
 {
-    return std::make_unique<TIndexTabletDatabase>(database);
+    std::unique_ptr<IIndexTabletDatabase> db =
+        std::make_unique<TIndexTabletDatabase>(database);
+    if (TxRescheduler) {
+        db = std::make_unique<TIndexTabletDatabaseFailureInjection>(
```
qkrorlqr: nit: `TIndexTabletDatabaseFailureInjection` sounds like the injection event itself, let's rename it to something like `TTestIndexTabletDatabase`?

cloud/filestore/libs/storage/testlib/test_env.h
```diff
@@ -76,6 +76,11 @@ struct TTestEnvConfig
     NActors::NLog::EPriority LogPriority_NFS = NActors::NLog::PRI_TRACE;
     NActors::NLog::EPriority LogPriority_KiKiMR = NActors::NLog::PRI_WARN;
     NActors::NLog::EPriority LogPriority_Others = NActors::NLog::PRI_WARN;
+
+    // This controls probability that read will be restarted
+    bool FakePageFaultsEnabled = true;
```
qkrorlqr: let's keep it disabled by default

cloud/filestore/tests/common_configs/nfs-storage-newfeatures-patch.txt
```diff
@@ -33,3 +33,5 @@ GidPropagationEnabled: true
 NodeRefsNoAutoPrecharge: true
 UseCustomReadDataResponseParser: true
 MaxShardManagementRequestsInFlight: 3
+FakeTxPageFaultsEnabled: true
+FakeTxPageFaultsProbabilityPercentage: 0.1
```
qkrorlqr: I would suggest to make a separate test suite where this would be enabled and keep the current integration test suites as they are now

qkrorlqr (commented):
and the most important question - IIRC an assertion inside the localdb code used to be triggered when a transaction returned `false` from the prepare stage when actually all the pages that it wanted to read were already present in the page cache - have you encountered this issue? or is it not true?

cloud/filestore/config/storage.proto
qkrorlqr: // and in this case it won't be a percentage - we'll be able to delete the division by 100 from the code

cloud/filestore/tests/common_configs/nfs-storage-newfeatures-patch.txt
qkrorlqr: ok, decided to keep it enabled for all integration tests by default - otherwise we'll either need to significantly increase the number of tests or the coverage won't be good enough

qkrorlqr:
> and the most important question - IIRC an assertion inside the localdb code used to be triggered when a transaction returned `false` from the prepare stage when actually all the pages that it wanted to read were already present in the page cache - have you encountered this issue? or is it not true?

ah, it works because we explicitly call `tx.Reschedule();`, understood

cloud/filestore/config/storage.proto
tdiff: Why I used Percentage here - there was another parameter named like %Percentage. I think for uniformity its reasonable to follow single style at least in the same config, no?

cloud/filestore/libs/storage/core/config.cpp
tdiff: I can change that, but in storage.proto there are other examples of configs with separate Enable flag, e.g. one could use InMemoryIndexCacheNodesCapacity=0 as implicit way to disable InMemoryCaching
```
    // Enables in-memory index cache for all index-related requests
    optional bool InMemoryIndexCacheEnabled = 371;
    // Capacity of in-memory index cache, in number of entries per each table
    optional uint64 InMemoryIndexCacheNodesCapacity = 372;
```

cloud/filestore/config/storage.proto
qkrorlqr: those are different parameters - they're not about the probability of doing some action but about various thresholds or buffer sizes

and those are also integers - there's no need for those parameters to be configured with precision higher than 1%

here `ProbabilityPercentage` looks "heavy" and might sound misleading to some people, IMO it's better to have just `Probability` (as a value between 0 and 1) - especially considering the fact that you need to be able to configure values smaller than 0.01 (smaller than 1%)

cloud/filestore/libs/storage/core/config.cpp
qkrorlqr: > I can change that, but in storage.proto there are other examples of configs with separate Enable flag, e.g. one could use InMemoryIndexCacheNodesCapacity=0 as implicit way to disable InMemoryCaching

In those cases there's a separate `Enabled` flag because:
* there're multiple settings and we want to be able to enable/disable all of them together via a single flag
* most of the time the defaults are good enough and it's just a question of enabling/disabling - setting the thresholds every time is inconvenient

here we have only one parameter and the natural default is 0 so I don't see any reason to have a separate `Enabled` flag

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: Test* feels too abstract.

What about TIndexTabletDatabaseWithFailureInjection?

cloud/filestore/libs/storage/init/actorsystem.cpp
tdiff: now its only 85

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: agreed

qkrorlqr (dismissed):

qkrorlqr (approved):

cloud/filestore/libs/storage/core/tablet.h
```diff
@@ -176,6 +156,27 @@ class TTabletBase
...
                 Args.OnRestart();
+
+                // Reschedules the transaction from PrepareTx assuming the
+                // rescheduler was triggered during PrepareTx() call.
+                if (Y_UNLIKELY(
+                        Self->TxRescheduler &&
+                        Self->TxRescheduler->IsTriggered()))
+                {
+                    Self->TxRescheduler->Reset();
+                    LOG_DEBUG(
```
drbasic, line 167: LOG_DEBUG?

tdiff: https://github.com/ydb-platform/nbs/pull/6544#discussion_r3597809082

drbasic (approved):

PR #6552, opened 2026-07-17
[Disk Manager] Add ability to report external storage limits/usage

jkuradobery:
### Notes
We would like to be able to add additional snapshot metrics reporting (e.g. ydb storage usage from ydb-controlplane, S3 storage usage via api).
We want disk-manager to report it's storage metrics without us having to rely on external metrics providers.


cloud/disk_manager/cmd/disk-manager/main.go
```diff
@@ -17,5 +18,6 @@ func main() {
 			// TODO: should use real authorizer, not stub.
 			return auth.NewStubAuthorizer(), nil
 		},
+		snapshot.NewStubSnapshotStorageQuotaReporter,
```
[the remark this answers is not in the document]
jkuradobery: I think that if we use authorizer stub, we can use this stub as well.

cloud/disk_manager/pkg/snapshot/quota.go
```diff
@@ -0,0 +1,67 @@
...
+	_ *Config,
+	_ auth.Credentials,
+) (SnapshotStorageQuotaReporter, error) {
+
+	return &emptySnapshotStorageQuotaReporter{}, nil
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+type stubSnapshotStorageQuotaReporter struct {
```
SvartMetal: It is not needed in production code, let's define it in mocks/disk-manager/main.go

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
```diff
@@ -1094,12 +1094,76 @@ func GetCounters(
 	return result
 }
 
+func GetGauges(
+	t *testing.T,
+	ports []string,
+	name string,
+	labels map[string]string,
+) []float64 {
```
SvartMetal, line 1102: \n

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
```diff
@@ -1094,12 +1094,76 @@ func GetCounters(
...
+	)
+	return result
+}
+
 func GetCounter(
 	t *testing.T,
 	port string,
 	name string,
 	labels map[string]string,
 ) (float64, bool) {
```
SvartMetal, line 1125: \n

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
```diff
@@ -1094,12 +1094,76 @@ func GetCounters(
...
+		},
+	)
+}
+
+func GetGauge(
+	t *testing.T,
+	port string,
+	name string,
+	labels map[string]string,
+) (float64, bool) {
```
SvartMetal, line 1143: \n

cloud/disk_manager/test/mocks/disk-manager/ya.make
```diff
@@ -0,0 +1,7 @@
+GO_PROGRAM(disk-manager-test)
```
SvartMetal: disk-manager-mock

cloud/disk_manager/internal/pkg/dataplane/collect_snapshot_metrics_task.go
```diff
@@ -8,14 +8,17 @@ import (
...
 	"github.com/ydb-platform/nbs/cloud/tasks"
+	"github.com/ydb-platform/nbs/cloud/tasks/logging"
 )
 
 ////////////////////////////////////////////////////////////////////////////////
 
 type collectSnapshotMetricsTask struct {
 	registry                  metrics.Registry
 	storage                   storage.Storage
+	quotaReporter             snapshot.SnapshotStorageQuotaReporter
```
SvartMetal: storageQuotaReporter

cloud/disk_manager/internal/pkg/dataplane/collect_snapshot_metrics_task.go
SvartMetal, line 97: It looks like we need to clear the quota metrics here, along with other metrics such as `snapshots/count`

cloud/disk_manager/pkg/snapshot/quota.go
```diff
@@ -0,0 +1,67 @@
...
+	registry metrics.Registry
+}
+
+func (r *stubSnapshotStorageQuotaReporter) Report(_ context.Context) error {
+	r.registry.Gauge("snapshots/quotas/used_bytes").Set(0)
```
SvartMetal: As far as I know, we use camelCase for metric names, so it should be `usedBytes`

cloud/disk_manager/pkg/snapshot/quota.go
```diff
@@ -0,0 +1,67 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+type Config = snapshot_config.SnapshotConfig
+
+type SnapshotStorageQuotaReporter interface {
+	Report(ctx context.Context) error
+}
+
+type NewSnapshotStorageQuotaReporterFunc = func(
```
SvartMetal, line 23: Let's define the contract, including the expected sensors and their semantics, at least in a comment

cloud/disk_manager/pkg/snapshot/quota.go
```diff
@@ -0,0 +1,67 @@
+package snapshot
+
+import (
+	"context"
+
+	snapshot_config "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/dataplane/snapshot/config"
```
SvartMetal, line 6: We should not reference internal packages here

cloud/disk_manager/internal/pkg/facade/disk_service_test/disk_service_metrics_test.go
```diff
@@ -97,6 +97,21 @@ func TestNbsClientReportsMetrics(t *testing.T) {
...
+	require.Equal(t, []float64{0}, testcommon.GetGaugesDataplane(
+		t,
+		"snapshots_quotas_used_bytes",
+		snapshotStorageQuotaLabels,
+	))
+	require.Equal(t, []float64{1000}, testcommon.GetGaugesDataplane(
+		t,
+		"snapshots_quotas_limit_bytes",
+		snapshotStorageQuotaLabels,
+	))
```
SvartMetal, line 116: Are we sure this test is not flaky?

How do we know that the snapshot collection metrics iteration has completed by this point?

SvartMetal (changes requested):

cloud/disk_manager/pkg/snapshot/quota.go
jkuradobery: See auth in the same package

cloud/disk_manager/internal/pkg/dataplane/collect_snapshot_metrics_task.go
jkuradobery: I do not think we need to delete quota metrics here. We can just do max(usage) / min(limit) and call it a day. 
We also do not unregister other metrics in clearMetrics, such as totalStorageSize, so I do not see the reason to delete the metric.

SvartMetal: > We also do not unregister other metrics in clearMetrics

Looks like we missed that, so it is a bug

SvartMetal: > We can just do max(usage) / min(limit) and call it a day.

Looks very kostylno and does not work correctly as well

jkuradobery: Yes this is not a correct behavior. We can actually fail within the task for any reason, yet the disk-manager would keep on running, thus reporting metrics, which may lead to stale metrics.

SvartMetal (approved):

gy2411 (approved):

PR #6558, opened 2026-07-20
Report endpoint restore failures

EvgeniyKozev:
## What changed

- Track endpoint restore errors in `TEndpointManager` with an atomic counter.
- Mark endpoint restoration as failed when any restore step fails.
- Return `E_FAIL` from `ListEndpoints` after failed restore, including the number of failed endpoints.
- Extend `ShouldRestoreEndpointWithNbdDevice` to verify the new `ListEndpoints` error behavior.

## Why

Previously restore errors were reported to logs/counters, but the final restore stage still became `Completed`. As a result, clients listing endpoints could observe `EndpointsWereRestored=true` even when some endpoints failed to restore.

## Validation

- `git diff --check`
- `./ya make -tt cloud/blockstore/libs/endpoints/ut -F TEndpointManagerTest::ShouldRestoreEndpointWithNbdDevice`
- `./ya make -tt cloud/blockstore/libs/endpoints/ut`

### DIFF
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -672,6 +672,24 @@ private:
         return false;
     }
 
+    NProto::EEndpointRestoringState GetEndpointRestoringState()
+    {
+        switch (AtomicGet(RestoringStage)) {
+            case WaitingForRestoring:
+                return NProto::ENDPOINT_RESTORING_STATE_WAITING_FOR_RESTORING;
+            case ReadingStorage:
+                return NProto::ENDPOINT_RESTORING_STATE_READING_STORAGE;
+            case StartingEndpoints:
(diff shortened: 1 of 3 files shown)


cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -1243,6 +1251,14 @@ NProto::TListEndpointsResponse TEndpointManager::DoListEndpoints(
     Y_UNUSED(ctx);
     Y_UNUSED(request);
 
+    if (AtomicGet(RestoringStage) == Failed) {
```
yegorskii: May be it will be better to extend TListEndpointsResponse with fiels representing RestoringStage(something like EndpointsStorageState) value instead of failing request with E_FAIL? If the request could not be completed because of some underlying subsystem failure, E_INVALID_STATE seems to be more appropriate error code.

PR #6560, opened 2026-07-20
[Filestore] Do not report DescribeFileStoreError critical event from TCreateSessionActor if the request came from a source we can not control (gRPC request / filestore-client)

debnatkh:
### Notes
A create session request can come from the filestore-client/GRPC request. For those requests, we do not want to report a critical event about non-existent filesystems with these errors: `E_NOT_FOUND ss error: SEVERITY_ERROR | FACILITY_SCHEMESHARD | 2 Path not found`. 

To do it, I add the source of the request distinguishing between ones we can and cannot control

### Issue

### DIFF
--- a/cloud/filestore/public/api/protos/headers.proto
+++ b/cloud/filestore/public/api/protos/headers.proto
@@ -59,10 +59,10 @@ message THeaders
             REQUEST_ORIGIN_UNSPECIFIED = 0;
 
             // Internal caller; arguments are picked by us
-            REQUEST_ORIGIN_CONTROLLED = 1;
+            REQUEST_ORIGIN_INTERNAL = 1;
 
             // Came over the public gRPC API; arguments are client-supplied
-            REQUEST_ORIGIN_UNCONTROLLED = 2;
+            REQUEST_ORIGIN_EXTERNAL = 2;
         }
 
         ERequestOrigin RequestOrigin = 6;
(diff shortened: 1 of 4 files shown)


cloud/filestore/public/api/protos/headers.proto
```diff
@@ -53,6 +53,19 @@ message THeaders
 
         // Peer address of the request sender.
         string Peer = 5;
+
+        enum ERequestOrigin
+        {
+            REQUEST_ORIGIN_UNSPECIFIED = 0;
+
+            // Internal caller; arguments are picked by us
+            REQUEST_ORIGIN_CONTROLLED = 1;
```
qkrorlqr: REQUEST_ORIGIN_{INTERNAL / EXTERNAL}?

qkrorlqr (approved):

cloud/filestore/public/api/protos/headers.proto
```diff
@@ -53,6 +53,19 @@ message THeaders
 
         // Peer address of the request sender.
         string Peer = 5;
+
+        enum ERequestOrigin
+        {
+            REQUEST_ORIGIN_UNSPECIFIED = 0;
+
+            // Internal caller; arguments are picked by us
+            REQUEST_ORIGIN_INTERNAL = 1;
```
agalibin, line 62: Is it added for future using? I do not see this value in current changes.

agalibin (approved):

cloud/filestore/public/api/protos/headers.proto
debnatkh: Yep. For the fullness and potential following usage

PR #6562, opened 2026-07-20
issue-1541: drain HandleOpsQueue when AsyncDestroy is disabled

WilyTiger:
### Notes
HandleOpsQueue is now restored and drained when AsyncDestroy is disabled if a persistent queue from a previous session exists.

New Release requests continue to use the synchronous path while previously queued DestroyHandle requests are processed in the background.

### Issue
#1541

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
@@ -3456,36 +3456,21 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
 
     Y_UNIT_TEST(ShouldDrainHandleOpsQueueAfterAsyncDestroyDisabled)
     {
-        const TString sessionId = CreateGuidAsString();
         const ui64 queuedNodeId = 10;
         const ui64 queuedHandle = 2;
         const ui64 directNodeId = 11;
         const ui64 directHandle = 3;
 
         auto createBootstrap =
-            [&](bool asyncDestroyHandleEnabled, ISchedulerPtr scheduler)
+            [](bool asyncDestroyHandleEnabled, ISchedulerPtr scheduler)
         {
             NProto::TFileStoreFeatures features;
             features.SetAsyncDestroyHandleEnabled(asyncDestroyHandleEnabled);
 
-            TBootstrap bootstrap(
+            return TBootstrap(
                 CreateWallClockTimer(),
                 std::move(scheduler),
                 features);
-
-            bootstrap.Service->CreateSessionHandler =
-                [features, &sessionId](auto, auto)
-            {
-                NProto::TCreateSessionResponse result;
-                result.MutableSession()->SetSessionId(sessionId);
-                result.MutableFileStore()->SetBlockSize(4096);
-                result.MutableFileStore()->MutableFeatures()->CopyFrom(
-                    features);
-                result.MutableFileStore()->SetFileSystemId(FileSystemId);
-                return MakeFuture(result);
-            };
-
-            return bootstrap;
         };
 
         {
--- a/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
+++ b/cloud/filestore/libs/vfs_fuse/fs_ut.cpp
@@ -259,12 +259,10 @@ struct TBootstrap
         proto.SetDebug(true);
         proto.SetSocketPath(SocketPath);
         proto.SetFileSystemId(FileSystemId);
-        if (featuresConfig.GetAsyncDestroyHandleEnabled() ||
-            featuresConfig.GetAsyncDestroyReadOnlyHandleEnabled())
-        {
-            proto.SetHandleOpsQueuePath(TempDir.Path() / "HandleOpsQueue");
-            proto.SetHandleOpsQueueSize(handleOpsQueueSize);
-        }
+        // Keep the path configured to allow restoring
+        // an existing HandleOpsQueue.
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -992,14 +992,14 @@ class TFileSystemLoop final
...
-                        FileSystemConfig->GetFileSystemId() /
-                        SessionId;
-
+            if (Config->GetHandleOpsQueuePath()) {
+                const auto path = TFsPath(Config->GetHandleOpsQueuePath()) /
+                                  FileSystemConfig->GetFileSystemId() /
+                                  SessionId;
+                if (path.Exists() ||
+                    FileSystemConfig->GetAsyncDestroyHandleEnabled() ||
+                    FileSystemConfig->GetAsyncDestroyReadOnlyHandleEnabled())
```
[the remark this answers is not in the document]
WilyTiger: This is intentional. HandleOpsQueueInitialized is set because the existing queue is opened and owned by the mount.
An explicit mount stop runs DestroySession before this cleanup. The tablet cleans up the session state, so any remaining DestroyHandle requests are no longer needed and the queue file can be removed.

A filestore-vhost restart uses SuspendAsync instead of StopAsync. That path does not run DestroySession or delete HandleOpsQueue, so pending requests are preserved across service restarts.

SvartMetal: Let's extract this into a variable and reuse it on lines 1024-1025 as well?

SvartMetal (dismissed):

SvartMetal (dismissed):

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -992,14 +992,14 @@ class TFileSystemLoop final
...
-                        SessionId;
-
+            const bool asyncDestroyEnabled =
+                FileSystemConfig->GetAsyncDestroyHandleEnabled() ||
+                FileSystemConfig->GetAsyncDestroyReadOnlyHandleEnabled();
+            if (Config->GetHandleOpsQueuePath()) {
+                const auto path = TFsPath(Config->GetHandleOpsQueuePath()) /
+                                  FileSystemConfig->GetFileSystemId() /
+                                  SessionId;
+                if (path.Exists() || asyncDestroyEnabled) {
```
neihar, line 1002: low: should we delete file after drain is finished?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1019,13 +1019,13 @@ class TFileSystemLoop final
...
                     HandleOpsQueueInitialized = true;
-                } else {
-                    ReportHandleOpsQueueCreatingOrDeletingError(Sprintf(
-                        "[f:%s][c:%s] Error initializing HandleOpsQueue: "
-                        "HandleOpsQueuePath is not set",
-                        Config->GetFileSystemId().Quote().c_str(),
-                        Config->GetClientId().Quote().c_str()));
                 }
+            } else if (asyncDestroyEnabled) {
+                ReportHandleOpsQueueCreatingOrDeletingError(Sprintf(
```
neihar, line 1024: just curious, will we in that case proceed with nullptr HandleOpsQueue and decided whether we need to use HandlexOpsQueue based only on flags and therefore have a crash?

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -3456,6 +3454,130 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+        {
+            NProto::TFileStoreFeatures features;
+            features.SetAsyncDestroyHandleEnabled(asyncDestroyHandleEnabled);
+
+            TBootstrap bootstrap(
+                CreateWallClockTimer(),
+                std::move(scheduler),
+                features);
+
+            bootstrap.Service->CreateSessionHandler =
```
neihar: nit: do we need custom handler?

WilyTiger: No, the default TBootstrap handler already uses the supplied features and the shared SessionId. I'll remove the custom handler

cloud/filestore/libs/vfs_fuse/loop.cpp
WilyTiger: Yes, you are right. This behavior already existed before this PR. I will also check that HandleOpsQueue is initialized and fall back to synchronous processing otherwise:

```
const bool processAsynchronously =
    HandleOpsQueue &&
    (Config->GetAsyncDestroyHandleEnabled() ||
     (Config->GetAsyncDestroyReadOnlyHandleEnabled() && readOnly));
```

cloud/filestore/libs/vfs_fuse/loop.cpp
WilyTiger: I don't think we should delete it immediately after the queue becomes empty. THandleOpsQueue and its file lock are owned for the whole mount lifetime, and queue-processing tasks may still be scheduled.

Deleting it after drain would require a separate lifecycle for stopping the processor, waiting for in-flight callbacks, releasing the queue and its lock, and only then removing the file. Currently, as with WriteBackCache, drain only empties the persistent storage; the file is removed on an explicit StopEndpoint

neihar (approved):

agalibin (approved):

PR #6575, opened 2026-07-21
added --bs-file option to filestore-server and vhost

tdiff:
### Notes
This PR allows to specify blob storage configuration on command line, alike to other parameters. It was required in context of BSProxy failure injection, which is enabled in blob_storage_config.

Additionally it allows to enable BS failure injection in test config generator.

### Issue
https://github.com/ydb-platform/nbs/issues/6467


SvartMetal (dismissed):

cloud/storage/core/libs/kikimr/config_initializer.cpp
```diff
@@ -160,7 +160,13 @@ void TConfigInitializerYdbBase::InitKikimrConfig()
...
 
     auto& bsConfig = *KikimrConfig->MutableBlobStorageConfig();
-    bsConfig.MutableServiceSet()->AddAvailabilityDomains(1);
+    if (Options->BlobStorageConfig) {
+        ParseProtoTextFromFileRobust(
+            Options->BlobStorageConfig,
+            bsConfig);
+    } else {
+        bsConfig.MutableServiceSet()->AddAvailabilityDomains(1);
+    }
```
SvartMetal, line 169: Let's add support to https://github.com/ydb-platform/nbs/tree/main/cloud/storage/core/tools/common/go/configurator as well

SvartMetal: We decided to skip it because we don't need to set `bs.txt` for any cluster, and the `TBlobStorageConfig` protobuf is not as small as we initially thought

cloud/storage/core/libs/kikimr/options.cpp
```diff
@@ -99,6 +99,10 @@ TOptionsYdbBase::TOptionsYdbBase()
     Opts.AddLongOption("actor-system-available-cpu-cores-percentage")
         .OptionalArgument("NUM")
         .StoreResult(&ActorSystemAvailableCpuCoresPercentage);
+
+    Opts.AddLongOption("bs-file")
+        .OptionalArgument("FILE")
+        .StoreResult(&BlobStorageConfig);
```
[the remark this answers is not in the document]
tdiff: fixed

cloud/filestore/tests/python/lib/daemon_config.py
```diff
@@ -277,6 +289,7 @@ def __profile_file_path(self, name):
 
     def __write_configs(self):
         for name, proto in self.__proto_configs.items():
+            if not proto: continue
```
SvartMetal: Why is this needed?

cloud/filestore/tests/config_dispatcher/test.py
```diff
@@ -147,3 +150,61 @@ def test_server():
 def test_vhost():
     filestore_binary_path = yatest_common.binary_path("cloud/filestore/apps/vhost/filestore-vhost")
     assert setup_and_run_test(filestore_binary_path, FilestoreVhostConfigGenerator, wait_for_filestore_vhost)
+
+
+def test_blob_storage_config():
```
SvartMetal: It does not look like a config dispatcher test. It looks more like a config generator test

cloud/filestore/tests/python/lib/daemon_config.py
tdiff: Just to keep the style of `def generate_configs()` below and avoid writing unneeded files. Without this logic I'd have to conditionally update __proto_configs, which looks a bit ugly to me. __generate_bs_txt() returns none if it should be disabled, hence the idea is to avoid writing file that should not be used.

We cannot just pass empty bs.txt to fileserver processes, as some default configuration is applied when --bs-file option is not provided.

cloud/filestore/tests/config_dispatcher/test.py
tdiff: fair enough

cloud/filestore/tests/python/lib/daemon_config.py
SvartMetal: I think it is less surprising to simply not add the `bs.txt` mapping to the map if the failure probability is zero (or is None)

tdiff: ok

cloud/filestore/tests/python/lib/ut/ya.make
SvartMetal, line 1: The file is not reachable right now.

* Let's add a `ut` folder under `cloud/filestore/tests/python/lib`
* Add a `RECURSE_TESTS(ut)` directive to `cloud/filestore/tests/python/lib/ya.make`
* Move `daemon_config_test.py` there

SvartMetal: UPDATE: it is actually reachable from cloud/filestore/tests/python/lib/ya.make

tdiff: done

SvartMetal (approved):

nisqatsi (approved):

Issue #6486, opened 2026-07-12
[Filestore] non-blocking open: add durable async-open confirm/retry queue


budevg:
This is part of  #5432 and is based on the two-stage redo-log idea from #1541.

The tablet can apply `CreateHandle` to memory and reply before the local DB commit. filestore-vhost then needs its own durable record of acknowledged opens that may not yet be persisted by the tablet.

## Goal

Add a durable filestore-vhost queue for acked async opens and use it to confirm or retry the operation until the tablet has persisted the handle or reports a final failure.

## Proposed behavior

Fast open flow:

1. Send `CreateHandle` to the tablet with the async/early-reply mode enabled.
2. Tablet validates the request, creates the handle in memory, and replies before waiting for DB commit.
3. filestore-vhost writes the acknowledged open to a durable local queue.
4. filestore-vhost replies to the guest.
5. filestore-vhost asynchronously confirms/completes the open with the tablet.
6. Remove the queue entry only after confirmation/completion succeeds.

Recovery flow:

1. On filestore-vhost restart, reload pending async-open records.
2. Retry/confirm each pending open.
3. Keep retrying transient failures.
4. Report final failures via metric/critical event.

Tablet restart flow:

1. If an operation using a pending handle gets `E_FS_BADHANDLE`, `EBADF`, or the tablet's equivalent invalid-handle error, check the async-open queue.
2. If the handle is pending, retry/confirm the original `CreateHandle`.
3. Retry the original operation if it is safe to retry.

## Scope

* Use `HandleOpsQueue` or a similar tmpfs-backed persistent queue.
* Store enough data to retry or confirm the open:
  * filesystem id;
  * session id;
  * node id;
  * guest-visible handle id;
  * open flags and mode information needed by the tablet;
  * request id or dup-cache data if required.
* Make queue replay idempotent.
* Remove entries only after the tablet-side handle is known to be durable enough.

## Metrics

Add counters/gauges for:

* async-open queue depth;
* oldest async-open queue entry age;
* confirm/retry attempts;
* confirm/retry failures by error code;
* recovered invalid-handle operations;
* unrecoverable pending opens.

Issue #5432, opened 2026-03-09
[Filestore] non-blocking open: basic implementation, for read-only mode without O_CREAT flag


SvartMetal:
Based on ideas described in #1541.

The Open (CreateHandle) call currently performs several expensive operations:
* Handle ID generation and persistence in the tablet local DB https://github.com/ydb-platform/nbs/blob/06a5a359855e1ce9625a72281f41624ef71b0fd3/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp#L411
* Node creation when O_CREAT flag is set, which also requires a local DB transaction https://github.com/ydb-platform/nbs/blob/06a5a359855e1ce9625a72281f41624ef71b0fd3/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp#L349

The second item (node creation) will be addressed in a separate [issue](https://github.com/ydb-platform/nbs/issues/5443). Therefore, the scope of this issue can be simplified to read-only handles without the O_CREAT flag.

### Proposed approach
* Add CreateHandleRequest to a redo log (implemented via HandleOpsQueue).
* Send the request to the tablet with a flag indicating that it should perform only basic validation and not wait for local DB transactions to complete. This is very fast when the node is in the in-memory cache. This [issue](https://github.com/ydb-platform/nbs/issues/5388) is also expected to speed up client–tablet communication
* Immediately acknowledge CreateHandleResponse to the guest.
* Perform the remaining work asynchronously (persisting the handle in the tablet).
* Remove the entry from HandleOpsQueue once the operation completes.

See original comment: https://github.com/ydb-platform/nbs/issues/1541#issuecomment-2214905104

### Handle ID generation requirements
* CreateHandleResponse must return a unique handle ID.
* Currently the ID is generated by the tablet and stored in the local DB as part of TSessionHandle:
https://github.com/ydb-platform/nbs/blob/06a5a359855e1ce9625a72281f41624ef71b0fd3/cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp#L702
This handle prevents node deletion while the handle exists.

### Problems need to be addressed:
1. Handle IDs should be idempotently allocated from a range owned by the current client (guest) when a CreateHandleRequest is added to the HandleOpsQueue. These IDs must not conflict with the IDs allocated by the tablet
2. Race with unlink. Consider the following scenario:
* filestore-vhost acknowledges CreateHandle before the local DB transaction completes.
[...]

PR #6589, opened 2026-07-23
[Filestore] issue-5432: add basic async open support (tablet)

budevg:
#5432
#6486

### DIFF
--- a/cloud/filestore/libs/service/request.h
+++ b/cloud/filestore/libs/service/request.h
@@ -72,6 +72,7 @@ namespace NCloud::NFileStore {
     xxx(ReadData,                           __VA_ARGS__)                       \
     xxx(WriteData,                          __VA_ARGS__)                       \
     xxx(AllocateData,                       __VA_ARGS__)                       \
+    xxx(ConfirmCreateHandle,                __VA_ARGS__)                       \
 // FILESTORE_DATA_METHODS
 
 #define FILESTORE_LOCAL_DATA_METHODS(xxx, ...)                                 \
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.h
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.h
@@ -182,6 +182,8 @@ private:
 
     TSet<NActors::TActorId> WorkerActors;
     TIntrusiveList<TRequestInfo> ActiveRequests;
+    // Handles waiting for their registration tx to commit.
+    THashSet<ui64> PendingCreateHandleCommits;
 
     TInstant ReassignRequestSentTs;
 
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.h
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.h
@@ -182,8 +182,6 @@ private:
 
     TSet<NActors::TActorId> WorkerActors;
     TIntrusiveList<TRequestInfo> ActiveRequests;
-    // Handles waiting for their registration tx to commit.
-    THashSet<ui64> PendingCreateHandleCommits;
 
     TInstant ReassignRequestSentTs;
 
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
@@ -560,7 +560,7 @@ void TIndexTabletActor::ExecuteTx_CreateHandle(
 
     if (safeAsync || unsafeAsync) {
         args.AsyncHandle = args.Response.GetHandle();
-        PendingCreateHandleCommits.insert(args.AsyncHandle);
+        StartCreateHandleCommit(args.AsyncHandle);
 
         LOG_DEBUG(
             ctx,
@@ -654,7 +654,7 @@ void TIndexTabletActor::CompleteTx_CreateHandle(
     TTxIndexTablet::TCreateHandle& args)
 {
     if (args.Completed) {
-        PendingCreateHandleCommits.erase(args.AsyncHandle);
+        EndCreateHandleCommit(args.AsyncHandle);
         return;
     }
 
@@ -683,7 +683,7 @@ void TIndexTabletActor::HandleConfirmCreateHandle(
     if (handle
             && handle->Session == session
             && handle->GetNodeId() == request.GetNodeId()
-            && !PendingCreateHandleCommits.contains(request.GetHandle()))
+            && !HasPendingCreateHandleCommit(request.GetHandle()))
     {
         auto response =
             std::make_unique<TEvService::TEvConfirmCreateHandleResponse>();
@@ -750,7 +750,7 @@ bool TIndexTabletActor::PrepareTx_ConfirmCreateHandle(
     }
 
     if (args.Node->Attrs.GetType() != NProto::E_REGULAR_NODE) {
-        args.Error = ErrorIsDirectory(args.NodeId);
+        args.Error = ErrorInvalidNodeType(args.NodeId);
         return true;
     }
 
@@ -796,9 +796,15 @@ void TIndexTabletActor::ExecuteTx_ConfirmCreateHandle(
     }
 
     if (!handleExists) {
-        // Keep retries off the fast path until this tx commits.
+        // RegisterHandle makes the new handle visible in memory before this
+        // transaction is durably committed. Mark the commit as pending so a
+        // concurrent confirmation skips the fast path and runs as a tablet
+        // transaction. Its response will then be ordered after this handle
+        // registration commits. Otherwise the client could remove its durable
+        // confirmation request, and a tablet restart could leave it with a
+        // handle that was never persisted.
         args.HandleRegistered = true;
-        PendingCreateHandleCommits.insert(args.Handle);
+        StartCreateHandleCommit(args.Handle);
     }
 
     if (args.CreateRequestId && !session->LookupDupEntry(args.CreateRequestId))
@@ -826,7 +832,7 @@ void TIndexTabletActor::CompleteTx_ConfirmCreateHandle(
     RemoveInFlightRequest(*args.RequestInfo);
 
     if (args.HandleRegistered) {
-        PendingCreateHandleCommits.erase(args.Handle);
+        EndCreateHandleCommit(args.Handle);
     }
 
     auto response =
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
@@ -18,7 +18,8 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-NProto::TError ValidateRequest(const NProto::TCreateHandleRequest& request)
+NProto::TError ValidateCreateHandleRequest(
+    const NProto::TCreateHandleRequest& request)
 {
     if (request.GetNodeId() == InvalidNodeId) {
         return ErrorInvalidArgument();
@@ -41,6 +42,16 @@ NProto::TError ValidateRequest(const NProto::TCreateHandleRequest& request)
     return {};
 }
 
+NProto::TError ValidateConfirmCreateHandleRequest(
+    const NProto::TConfirmCreateHandleRequest& request)
+{
+    if (request.GetNodeId() == InvalidNodeId || !request.GetHandle()) {
+        return ErrorInvalidArgument();
+    }
+
+    return {};
+}
+
 }   // namespace
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -51,8 +62,10 @@ void TIndexTabletActor::HandleCreateHandle(
 {
     using TResponse = TEvService::TEvCreateHandleResponse;
 
-    auto* session =
-        AcceptRequest<TEvService::TCreateHandleMethod>(ev, ctx, ValidateRequest);
+    auto* session = AcceptRequest<TEvService::TCreateHandleMethod>(
+        ev,
+        ctx,
+        ValidateCreateHandleRequest);
     if (!session) {
         return;
     }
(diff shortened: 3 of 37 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -629,4 +657,168 @@ void TIndexTabletActor::CompleteTx_CreateHandle(
...
+    if (!session) {
+        auto message = ReportSessionNotFoundInTx(TStringBuilder()
+            << "ConfirmCreateHandle: " << args.Request.ShortDebugString());
+        args.Error = MakeError(E_INVALID_STATE, std::move(message));
+        return;
+    }
+
+    TIndexTabletDatabase db(tx.DB);
+
+    if (auto* handle = FindHandle(args.Handle)) {
```
qkrorlqr: do we even need to start a transaction if the requested handle has already been created? seems that we can check it and reply directly from `HandleConfirmCreateHandle`

yes, we'll still need to doublecheck here (because there might be a concurrent request attempting to create the same handle), but seems that we can make the happy path much faster

cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
```diff
@@ -109,6 +109,636 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Handles)
...
+
+        ui32 nodeIdx = env.AddDynamicNode();
+        ui64 tabletId = env.BootIndexTablet(nodeIdx);
+
+        TIndexTabletClient tablet(env.GetRuntime(), nodeIdx, tabletId);
+        tablet.InitSession("client", "session");
+
+        auto id = CreateNode(tablet, TCreateNodeArgs::File(RootNodeId, "test"));
+
+        // Hold the create tx commit so the async response is acknowledged
```
qkrorlqr: I understand the purpose of the test but I don't understand this comment - it says that something's acknowledged before SessionHandle becomes durable but in this test the feature is disabled so nothing is acknowledged before becoming durable

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -629,4 +657,168 @@ void TIndexTabletActor::CompleteTx_CreateHandle(
...
+            auto message = TStringBuilder()
+                << "ConfirmCreateHandle collision: "
+                << args.Request.ShortDebugString()
+                << " existing session: " << handle->GetSessionId()
+                << " existing node: " << handle->GetNodeId();
+            args.Error = MakeError(E_INVALID_STATE, std::move(message));
+            return;
+        }
+    } else {
+        handle = UnsafeCreateHandle(
```
qkrorlqr: now the comment and the name of this method are misleading: https://github.com/ydb-platform/nbs/blob/db72a9e4bcae9b6ed3c30ca494425c1ebce60f48/cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp#L791

> // This method is unsafe because it accepts a handleId that may not follow the
> // conventions enforced by TIndexTabletState::GenerateHandle. We need this
> // method for testing purposes, such as verifying the response to handles with
> // an incorrect shard number.

I suggest to add something like `NProto::TError RegisterHandle(db, session, handleId, nodeId, commitId, flags)` instead which would take args.NodeId as input and would do the NodeId check inside it instead. This way:
* `Unsafe` state methods will still be used only by `Unsafe` private api handlers
* the other methods would be safe in the sense that they would do the necessary checks

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
```diff
@@ -209,28 +400,105 @@ void TFileSystem::Open(
...
+    fuse_req_t req,
+    fuse_ino_t ino,
+    const NProto::TCreateHandleRequest& createRequest,
+    const NProto::TCreateHandleResponse& response)
+{
+    const auto requestId = createRequest.GetHeaders().GetRequestId()
+        ? createRequest.GetHeaders().GetRequestId()
+        : callContext->RequestId;
+
+    THandleOpsQueue::EResult result;
```
neihar, line 422: just curious, do we really want to handle destroy/create from the single queue?

cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
budevg: sorry it's leftover from previous refactoring I modified the comment to match the test

cloud/filestore/libs/service/request.h
```diff
@@ -72,6 +72,7 @@ namespace NCloud::NFileStore {
     xxx(ReadData,                           __VA_ARGS__)                       \
     xxx(WriteData,                          __VA_ARGS__)                       \
     xxx(AllocateData,                       __VA_ARGS__)                       \
+    xxx(ConfirmCreateHandle,                __VA_ARGS__)                       \
```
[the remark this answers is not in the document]
budevg: this is not relevant after rebase

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -513,6 +526,29 @@ void TIndexTabletActor::ExecuteTx_CreateHandle(
...
+         | ProtoFlag(NProto::TCreateHandleRequest::E_TRUNCATE));
+
+    const bool isReadOnly = (args.Flags & wflags) == 0;
+
+    const bool hasLocalHandle = args.Response.GetHandle() != 0;
+
+    const bool safeAsync = hasLocalHandle &&
+                           isReadOnly &&
+                           Config->GetAsyncCreateHandleEnabled() &&
+                           args.Request.GetAllowAsyncCreateHandle();
```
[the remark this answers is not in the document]
budevg: this is not relevant after the rebase

cloud/filestore/libs/vfs_fuse/fs_impl_data.cpp
budevg: For now, a single queue is the simplest design and preserves the required ordering between `ConfirmCreateHandle` and `DestroyHandle.` We also plan to introduce batching to make queue processing faster. If the single processing  path becomes a bottleneck in the future, we can consider splitting the queues or throttling asynchronous requests.

cloud/filestore/libs/storage/tablet/tablet_actor.h
```diff
@@ -182,6 +182,8 @@ class TIndexTabletActor final
 
     TSet<NActors::TActorId> WorkerActors;
     TIntrusiveList<TRequestInfo> ActiveRequests;
+    // Handles waiting for their registration tx to commit.
+    THashSet<ui64> PendingCreateHandleCommits;
```
qkrorlqr: This boundary is not stated explicitly anywhere but generally we try to keep only things like pending event queues, active actor sets, state flags, etc in the `TIndexTabletActor` fields and store the filesystem-specific things inside `TIndexTabletState`. Here it's a set of handles which are filesystem-specific and not actorsystem-specific so I would suggest to move this field into `TIndexTabletState`.

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -635,13 +667,187 @@ void TIndexTabletActor::HandleConfirmCreateHandle(
...
+        return false;
+    }
+
+    if (!args.Node) {
+        args.Error = ErrorInvalidTarget(args.NodeId);
+        return true;
+    }
+
+    if (args.Node->Attrs.GetType() != NProto::E_REGULAR_NODE) {
+        args.Error = ErrorIsDirectory(args.NodeId);
```
qkrorlqr: hmm, not necessarily a directory actually

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -635,13 +667,187 @@ void TIndexTabletActor::HandleConfirmCreateHandle(
...
+            TFileStoreComponents::TABLET,
+            "%s ConfirmCreateHandle failed: %s, %s",
+            LogTag.c_str(),
+            args.Request.ShortDebugString().c_str(),
+            FormatError(args.Error).Quote().c_str());
+        return;
+    }
+
+    if (!handleExists) {
+        // Keep retries off the fast path until this tx commits.
```
qkrorlqr: could you please expand this comment a bit - right now it's not entirely clear

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
```diff
@@ -788,6 +787,56 @@ TSessionHandle* TIndexTabletState::CreateHandle(
...
+
+NProto::TError TIndexTabletState::RegisterHandle(
+    IIndexTabletDatabase& db,
+    TSession* session,
+    ui64 handleId,
+    ui64 nodeId,
+    ui64 commitId,
+    ui32 flags)
+{
+    if (const auto* handle = FindHandle(handleId)) {
```
qkrorlqr, line 838: `FindHandle(handleId)` looks a bit strange because handles are stored per-session in the tablet LSM

mb refactoring the current code a bit so that we would store handles per session in the in-memory state would be reasonable - maybe in a preliminary PR (before this one) or as a follow-up (after this one)

qkrorlqr (commented):
generally looks good, I don't see any significant issues

cloud/filestore/libs/storage/tablet/tablet_state_sessions.cpp
budevg: I guess refactoring will make code more readable. There is even very old comment with `TODO: move to TSession`
https://github.com/ydb-platform/nbs/blob/41b79897cfd35407a4d0e4018ca62bdcf630273d/cloud/filestore/libs/storage/tablet/tablet_state_impl.h#L53

But I guess this will require to add cross session collision rejection logic to preserve invariant that exist now and it doesn't look like something worth doing in this pr. I will open an issue for this

qkrorlqr (dismissed):

yegorskii (dismissed):

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -635,13 +667,193 @@ void TIndexTabletActor::HandleConfirmCreateHandle(
...
+        // concurrent confirmation skips the fast path and runs as a tablet
+        // transaction. Its response will then be ordered after this handle
+        // registration commits. Otherwise the client could remove its durable
+        // confirmation request, and a tablet restart could leave it with a
+        // handle that was never persisted.
+        args.HandleRegistered = true;
+        StartCreateHandleCommit(args.Handle);
+    }
+
+    if (args.CreateRequestId && !session->LookupDupEntry(args.CreateRequestId))
```
SvartMetal: Should we validate CreateRequestId and return an error beforehand?

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -635,13 +667,193 @@ void TIndexTabletActor::HandleConfirmCreateHandle(
...
+
+    if (!handleExists) {
+        // RegisterHandle makes the new handle visible in memory before this
+        // transaction is durably committed. Mark the commit as pending so a
+        // concurrent confirmation skips the fast path and runs as a tablet
+        // transaction. Its response will then be ordered after this handle
+        // registration commits. Otherwise the client could remove its durable
+        // confirmation request, and a tablet restart could leave it with a
+        // handle that was never persisted.
+        args.HandleRegistered = true;
```
SvartMetal: `CreateHandleCommitStarted = true`?

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
```diff
@@ -635,13 +667,193 @@ void TIndexTabletActor::HandleConfirmCreateHandle(
...
+            << "ConfirmCreateHandle: " << args.Request.ShortDebugString());
+        args.Error = MakeError(E_INVALID_STATE, std::move(message));
+        return;
+    }
+
+    TIndexTabletDatabase db(tx.DB);
+
+    const bool handleExists = FindHandle(args.Handle) != nullptr;
+
+    args.Error = RegisterHandle(
```
SvartMetal, line 781: I think ConfirmCreateHandle looks clearer - it's the main side effect of the whole transaction

SvartMetal: Or CommitCreateHandle, or CommitHandleCreation

SvartMetal (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor_createhandle.cpp
budevg: I think RegisterHandle is simple and explains the intention well, it was suggested here https://github.com/ydb-platform/nbs/pull/6589#discussion_r3644465596

CommitCreateHandle/ConfirmCreateHandle ties a generic state method to one particular request so I prefer to keep it as is.

budevg:
calude report https://claude.ai/share/3d80308f-3872-4aca-905e-8e4de7d7a93b
fixed shard bits validation (issue 1)
The rest are not valid
2. Flags aren't restricted to read-only -  not valid for the supported flow. Async confirmation is generated only for eligible read-only handles and copies the original flags.
3. Confirm-after-destroy resurrects handles -  not valid. Confirm and async destroy share a HandleOps queue, so destroy runs only after confirmation completes.

SvartMetal (approved):

yegorskii (approved):

PR #6590, opened 2026-07-23
[Filestore] issue-5432: add basic async open support (integration tests)

budevg:
#5432
#6486


cloud/filestore/tests/async_open_test/script.py
```diff
@@ -0,0 +1,81 @@
...
+        count,
+        ready_path,
+        release_path,
+        verify_path=None,
+        verified_path=None,
+        unlink_path=None,
+        unlinked_path=None):
+    os.sync()
+    with open("/proc/sys/vm/drop_caches", "w") as f:
+        f.write("3\n")
```
qkrorlqr: let's put this "3" into a properly named constant

cloud/filestore/tests/async_open_test/test.py
```diff
@@ -0,0 +1,228 @@
...
+        f"sudo rm -f {READY_PATH} {RELEASE_PATH} {VERIFY_PATH} "
+        f"{VERIFIED_PATH} {UNLINK_PATH} {UNLINKED_PATH} {LOG_PATH}")
+    ssh(
+        f"sudo bash -c 'cd {mount_dir} && "
+        f"python3 {script} prepare {OPEN_HANDLE_COUNT}'")
+
+    with client.CreateClient(
+            f"localhost:{server_port}", log=logger) as filestore_client:
+        file_system_ids = [FILE_SYSTEM_ID]
+        if SHARD_COUNT:
```
qkrorlqr: but we have recipes for filesystem creation - why do we need extra checks here? this should be handled by the recipe

cloud/filestore/tests/common_configs/nfs-storage-newfeatures-patch.txt
```diff
@@ -17,6 +17,7 @@ UseMixedBlocksInsteadOfAliveBlocksInCompaction: true
 MixedBlocksOffloadedRangesCapacity: 1024
 CalculateCleanupScoreBasedOnUsedBlocksCount: true
 AsyncDestroyHandleEnabled: true
+AsyncCreateHandleEnabled: true
```
qkrorlqr: let's not put this into common_configs/nfs-storage-newfeatures-patch.txt for now - let's make a separate config patch for this specific test

cloud/filestore/tests/async_open_test/test.py
budevg: The recipe configures the shards but test needs the ids to check handle counts and restart every shard. The check also waits until all shards are visible, since that may happen asynchronously.

qkrorlqr: the shards should be visible immediately

if they're not visible, then I assume that this test actually waits for 15s (stats collection interval) here which is bad - if this practice spreads, the tests will start working much slower overall

qkrorlqr (approved):

cloud/filestore/tests/async_open_noshards_test/test.py
```diff
@@ -0,0 +1,7 @@
+import runpy
+
+import yatest.common as common
+
+
+test = runpy.run_path(common.source_path(
+    "cloud/filestore/tests/async_open_test/test.py"))["test"]
```
qkrorlqr: this looks ugly, let's just move this function into a lib and import it properly

qkrorlqr (changes requested):

qkrorlqr (dismissed):

yegorskii (dismissed):

SvartMetal (approved):

yegorskii (approved):

budevg:
failed test cloud/filestore/libs/storage/service/ut/TStorageServiceTest.ShouldReassignTablet doesn't look like related to this pr

PR #6594, opened 2026-07-23
Add certificate expiration metric to the static TLS certificate provider

sharpeye:
Add the ExpireTs metric with the certificate expiration timestamp to the static TLS certificate provider.

### DIFF
--- a/cloud/storage/core/libs/grpc/tls_certificate_provider.cpp
+++ b/cloud/storage/core/libs/grpc/tls_certificate_provider.cpp
@@ -2,8 +2,11 @@
 
 #include "tls_utils.h"
 
+#include <cloud/storage/core/libs/common/error.h>
+
 #include <library/cpp/monlib/dynamic_counters/counters.h>
 
+#include <util/folder/dirut.h>
 #include <util/generic/yexception.h>
 #include <util/stream/file.h>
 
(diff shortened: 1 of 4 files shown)


yegorskii (dismissed):

yegorskii (approved):

cloud/storage/core/libs/grpc/tls_certificate_provider.cpp
```diff
@@ -54,28 +67,63 @@ class TStaticCertificateProvider final
...
+                "cert",
+                GetBaseName(cert.Files.CertChainPath));
+
+            auto expireTs = certMetrics->GetCounter("ExpireTs", false);
+
+            auto [seconds, error] =
+                NTlsUtils::GetCertificateNotAfterTimestampSec(cert.CertChain);
+            if (HasError(error)) {
+                STORAGE_WARN(
+                    "Unable to parse certificate notAfter date for "
```
SvartMetal, line 119: Critical event?

sharpeye: If an error occurs here, `ExpireTs` will be set to zero, which will trigger the expired certificate alerts (the alert should treat `0` as `NODATA`). I think that should be enough to draw attention to the issue.

SvartMetal (approved):

PR #6595, opened 2026-07-23
[NBS] Fix NBS shutdown hang when Device Provider socket is missing

sharpeye:
Fix shutdown hanging when the Device Provider socket is unavailable.

The client now cancels active gRPC requests, rejects new requests after shutdown starts, and uses request deadlines.

### DIFF
--- a/cloud/blockstore/libs/local_nvme/generic_inventory_service_client.h
+++ b/cloud/blockstore/libs/local_nvme/generic_inventory_service_client.h
@@ -6,6 +6,7 @@
 #include <cloud/storage/core/libs/common/thread.h>
 #include <cloud/storage/core/libs/diagnostics/logging.h>
 #include <cloud/storage/core/libs/grpc/init.h>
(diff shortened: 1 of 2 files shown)


yegorskii (dismissed):

cloud/blockstore/libs/local_nvme/generic_inventory_service_client.h
```diff
@@ -58,25 +63,26 @@ class TGenericInventoryServiceClient
...
                 ClientContext,
                 Request,
                 cq);
             Reader->Finish(&Response, &Status, this);
-
-            return future;
         }
 
-        void Complete() override
+        void Complete(bool ok) final
```
SvartMetal, line 82: Why do we need `bool ok`?

sharpeye: Forgot to add proper handling for `!ok`.

SvartMetal (approved):

yegorskii (approved):

Issue #6390, opened 2026-07-01
[NBS] Introduce blocks filter for mixed index


vladstepanyuk:
Read blocks transaction takes a lot of cpu time

<img width="1788" height="561" alt="Image" src="https://github.com/user-attachments/assets/147665e4-88be-4257-a95f-e8ed823a1698" />

And if we look at flame graphs, we will see that mixed index reading takes a lot of CPU, but, on the other hand, mixed index is very sparse. We have approximately only 6% blocks in the mixed index, so most reads don't read any useful information. 

<img width="2430" height="774" alt="Image" src="https://github.com/user-attachments/assets/e8a975e0-aa6c-4dc7-b38c-ab3ab0a0ce80" />

We can add bloom or compressed bit mask filter for mixed index.

I wrote prototype of this optimization and with randrw 4k load it shows such results:

partition tablet consumes less CPU time  by 28%

<img width="1788" height="443" alt="Image" src="https://github.com/user-attachments/assets/50558faa-124c-4bdd-a02d-616e05f04e65" />

ReadBlocks Transaction ExecuteCPUTime and ReadBytes metrics

<img width="2940" height="443" alt="Image" src="https://github.com/user-attachments/assets/442b5fa8-63cf-4e27-b0a3-79d3b4531ea8" />

and compressed bitmap version consumes only ~10mb of RAM for 512 GB disk. So it is very cheap.

PR #6601, opened 2026-07-27
[NBS] corrected leaky bucket Register method return value with non-standard rate

vladstepanyuk:
### Notes
I expect this code to work, because we use the return value as the time we need to postpone the request before retrying it:
```
TLeakyBucket bucket(0.5, 0.5, 0);
auto now = TInstant::Now();
auto delay = bucket.Register(now, 42.0);
UNIT_ASSERT_VALUES_EQUAL(0, bucket.Register(now + SecondsToDuration(delay), 42.0));
```
Before the fix this code did not work: enough budget to register the update is accumulated only after `(update - State.Budget) / (Rate * 1e6)` seconds, but we were returning `(update - State.Budget)`.

Now this is not a big deal, but I want to use leaky bucket in implementing mixed blocks filter feature. And in that case pre-fix bucket behavior causing bugs.

### Issue
#6390

### DIFF
--- a/cloud/storage/core/libs/throttling/leaky_bucket.h
+++ b/cloud/storage/core/libs/throttling/leaky_bucket.h
@@ -151,6 +151,9 @@ public:
             Min(diff1, Min(Burst, Standard.TimePassed() / 1e6) * Beta);
 
         const auto boostDelaySeconds = Boost.Register(ts, maxBoostableDiff1);
+
+        Y_DEBUG_ABORT_UNLESS(
+            boostDelaySeconds == 0 || maxBoostableDiff1 > Boost.Budget());
         const auto diff2 =
             boostDelaySeconds ? maxBoostableDiff1 - Boost.Budget() : 0;
         // the remaining part of the update
--- a/cloud/storage/core/libs/throttling/leaky_bucket.h
+++ b/cloud/storage/core/libs/throttling/leaky_bucket.h
@@ -36,6 +36,7 @@ public:
     }
 
 public:
+    // returns time to wait in seconds to be able to spend the update
     double Register(TInstant ts, double update)
     {
         if (Y_LIKELY(State.LastUpdateTs.GetValue())) {
@@ -54,7 +55,7 @@ public:
             return 0;
         }
 
-        return update - State.Budget;
+        return (update - State.Budget) / (Rate * 1e6);
     }
 
     void Flush()
@@ -149,7 +150,9 @@ public:
         auto maxBoostableDiff1 =
             Min(diff1, Min(Burst, Standard.TimePassed() / 1e6) * Beta);
 
-        auto diff2 = Boost.Register(ts, maxBoostableDiff1);
+        const auto boostDelay = Boost.Register(ts, maxBoostableDiff1);
+        const auto diff2 =
+            boostDelay ? maxBoostableDiff1 - Boost.Budget() : 0;
         // the remaining part of the update
         auto fullDiff = diff2 + diff1 - maxBoostableDiff1;
         if (fullDiff == 0) {
--- a/cloud/storage/core/libs/throttling/leaky_bucket.h
+++ b/cloud/storage/core/libs/throttling/leaky_bucket.h
@@ -36,7 +36,7 @@ public:
     }
 
 public:
-    // returns time to wait in seconds to be able to spend the update
+    // returns time to wait in seconds to be able to register the update
     double Register(TInstant ts, double update)
     {
         if (Y_LIKELY(State.LastUpdateTs.GetValue())) {
@@ -150,9 +150,9 @@ public:
         auto maxBoostableDiff1 =
             Min(diff1, Min(Burst, Standard.TimePassed() / 1e6) * Beta);
 
-        const auto boostDelay = Boost.Register(ts, maxBoostableDiff1);
+        const auto boostDelaySeconds = Boost.Register(ts, maxBoostableDiff1);
         const auto diff2 =
-            boostDelay ? maxBoostableDiff1 - Boost.Budget() : 0;
+            boostDelaySeconds ? maxBoostableDiff1 - Boost.Budget() : 0;
         // the remaining part of the update
         auto fullDiff = diff2 + diff1 - maxBoostableDiff1;
         if (fullDiff == 0) {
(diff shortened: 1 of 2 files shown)


komarevtsev-d (dismissed):

cloud/storage/core/libs/throttling/leaky_bucket_ut.cpp
```diff
@@ -34,62 +34,62 @@ Y_UNIT_TEST_SUITE(TLeakyBucketTest)
...
-        REG_AND_CHECK(0, 20'010'000, 200);      // 199. + 0.01 * 100 - 200 = 0
+        REG_AND_CHECK(0, 100'000, 110);                 // 200 - 110 = 90
+        REG_AND_CHECK(0, 500'000, 70);                  // 90 + 0.4 * 100 - 70 = 60
+        REG_AND_CHECK(0, 600'000, 40);                  // 60 + 0.1 * 100 - 40 = 30
+        REG_AND_CHECK(0.1, 800'000, 60);                // 30 + 0.2 * 100 = 50
+        REG_AND_CHECK(0, 900'000, 60);                  // 50 + 0.1 * 100 - 60 = 0
+        REG_AND_CHECK(0, 10'000'000, 500);              // min(max(0, 200 - 500), 0 + 9.1 * 100 - 500) = 0
+        REG_AND_CHECK(0, 20'000'000, 1);                // 0 + min(max(0, 200 - 1), 10 * 100) = 199
+        REG_AND_CHECK(0.01, 20'000'000, 200);           // 199 + 0 = 199
+        REG_AND_CHECK(0, 20'010'000, 200);              // 199. + 0.01 * 100 - 200 = 0
```
SvartMetal: Why is it necessary to move the comment block to the right?

cloud/storage/core/libs/throttling/leaky_bucket.h
```diff
@@ -36,6 +36,7 @@ class TLeakyBucket
     }
 
 public:
+    // returns time to wait in seconds to be able to spend the update
```
SvartMetal: I'm having trouble understanding what **"spend the update"** means

cloud/storage/core/libs/throttling/leaky_bucket.h
```diff
@@ -149,7 +150,9 @@ class TBoostedTimeBucket
         auto maxBoostableDiff1 =
             Min(diff1, Min(Burst, Standard.TimePassed() / 1e6) * Beta);
 
-        auto diff2 = Boost.Register(ts, maxBoostableDiff1);
+        const auto boostDelay = Boost.Register(ts, maxBoostableDiff1);
```
SvartMetal: boostDelaySeconds

cloud/storage/core/libs/throttling/leaky_bucket.h
```diff
@@ -149,7 +150,9 @@ class TBoostedTimeBucket
...
+        const auto diff2 =
+            boostDelay ? maxBoostableDiff1 - Boost.Budget() : 0;
```
SvartMetal: Do we need `maxBoostableDiff1` to be greater than `Boost.Budget()`?

SvartMetal:
```
In our code, we interpret the return value of the Register method as the delay we need to wait before we can register the update.

But now it returns only the budget difference, although the budget will be replenished enough after the budgetDiff / rate seconds.
```

I don't fully understand the description. Could you explain in more detail what the problem was and how it was fixed?

> But now it returns only the budget difference

By "now," do you mean after the fix? Again, I don't understand why this is needed

cloud/storage/core/libs/throttling/leaky_bucket.h
vladstepanyuk: boostDelay != 0 means that  maxBoostableDiff1 greater than Boost.Budget()

vladstepanyuk:
> ```
> In our code, we interpret the return value of the Register method as the delay we need to wait before we can register the update.
> 
> But now it returns only the budget difference, although the budget will be replenished enough after the budgetDiff / rate seconds.
> ```
> 
> I don't fully understand the description. Could you explain in more detail what the problem was and how it was fixed?
> 
> > But now it returns only the budget difference
> 
> By "now," do you mean after the fix? Again, I don't understand why this is needed

@SvartMetal 

"Now" is the pre-fix behavior.

I expect this code to work, because we use the return value as the time we need to postpone the request before retrying it:
```
TLeakyBucket bucket(0.5, 0.5, 0);
auto now = TInstant::Now();
auto delay = bucket.Register(now, 42.0);
UNIT_ASSERT_VALUES_EQUAL(0, bucket.Register(now + SecondsToDuration(delay), 42.0));
```
Before the fix this code did not work: enough budget to register the update is accumulated only after `(update - State.Budget) / (Rate * 1e6)` seconds, but we were returning `(update - State.Budget)`.

Now this is not a big deal, but I want to use leaky bucket in implementing mixed blocks filter feature(#6390). And in that case pre-fix bucket behavior causing bugs.

cloud/storage/core/libs/throttling/leaky_bucket.h
SvartMetal: Could you please add a DEBUG_ABORT_UNLESS? 

As this is not obvious to me

SvartMetal (approved):

cloud/storage/core/libs/throttling/leaky_bucket.h
```diff
@@ -149,7 +150,12 @@ class TBoostedTimeBucket
         auto maxBoostableDiff1 =
             Min(diff1, Min(Burst, Standard.TimePassed() / 1e6) * Beta);
 
-        auto diff2 = Boost.Register(ts, maxBoostableDiff1);
+        const auto boostDelaySeconds = Boost.Register(ts, maxBoostableDiff1);
```
qkrorlqr, line 153: nit: `auto` -> `double`

cloud/storage/core/libs/throttling/leaky_bucket.h
```diff
@@ -149,7 +150,12 @@ class TBoostedTimeBucket
...
+
+        Y_DEBUG_ABORT_UNLESS(
+            boostDelaySeconds == 0 || maxBoostableDiff1 > Boost.Budget());
+        const auto diff2 =
```
qkrorlqr, line 157: nit: `auto` -> `double`

qkrorlqr (approved):

komarevtsev-d (approved):

Issue #5895, opened 2026-05-05
[Filestore] Low-latency file-only filesystem shard implementation


qkrorlqr:
This issue will be used to group all the activities related to the implementation of a specialized file-only shard which is supposed to be much more efficient than the current index tablet. This new implementation is going to use our own storage specialized for filesystem workloads - i.e. **not** YDB BlobStorage.

We already have a success story for a similar approach in network block devices - our mirrored disks. Implementing a filesystem shard is more complex, for example:
* we'll actually need to implement some data structures on top of the storage layer: inode table, radix tree for block indexing, bitmaps for block range allocation
* we also won't be able to get away without implementing journalling on storage nodes - it's needed to be able to combine the modification of multiple data structures into a single atomic op and also to recover upon shard restart (not only upon storage node restart)

Latency reduction for small requests is the main goal. We want to achieve at least 200us avg latency (and preferably 100us) for the following requests:
* `ReadData` and `WriteData` calls with small (ten-ish) number of blocks
* `CreateHandle` for an existing file if inode id is known, `DestroyHandle`

So basically we want open + read/write + close to be 300us (and if we do close asynchronously, then 200us). virtio-fs overhead should be measured separately - it might be almost 100us total for 3 calls (won't be optimized in this issue most probably).

Specialized storage will also allow us to serve large read/write requests with lower latencies and higher throughput. So our secondary goal is to be able to reach line rate on modern 400Gbps NICs, which means 40-50GB/s for reads and multi-hop writes and 15GB/s for direct mirrored filestore-vhost <-> storage-node writes.

qkrorlqr:
Preliminary work:
* https://github.com/ydb-platform/nbs/issues/5643
* https://github.com/ydb-platform/nbs/issues/5644
* https://github.com/ydb-platform/nbs/issues/5894 - prototype (PoC)

Also needed to achieve the latency goals:
* https://github.com/ydb-platform/nbs/issues/5388

Design:
* https://github.com/ydb-platform/nbs/tree/main/doc/filestore/design/storage/fastshard

Production version implementation plan - a rough oversimplified version looks like this:
* support journalling in blockstore-disk-agent devices
* support a separate device pool with journalled devices in DiskRegistry, support device allocation from this pool in the filesystem create/resize logic
* implement quorum and recovery on top of a group of journalled devices
* implement block allocator
* implement inode table
* implement radix tree for block indexing
[...]

Issue #5894, opened 2026-05-05
[Filestore] Filestore file shard prototype implementation on top of blockstore-disk-agent devices


qkrorlqr:
Here https://github.com/ydb-platform/nbs/issues/5643 `IFileSystemShard` interface and `TIndexTabletActor`'s adapter mode were implemented

Here https://github.com/ydb-platform/nbs/issues/5644 a way to have separate file-shards and separate non-file shards was implemented

In this issue we want to make an implementation of `IFileSystemShard` that would store data persistently on the devices belonging to multiple `blockstore-disk-agent`s. This implementation is not supposed to perfectly recover from crashes - it's main purpose is to show that we can achieve the latency numbers that we want to achieve on top of a full-stack implementation consisting of filestore-vhost <-RTT-> filestore-server <-RTT-> multiple blockstore-disk-agents <-libaio/uring-> multiple NVMe SSD devices

The main goal is to achieve at least 200us avg latency for single-block `ReadData` and `WriteData` requests. Preferably - 100us (not taking virtio-fs overhead into account). 200us can be achieved even if we use YDB Interconnect as our transport library. To achieve 100us we'll probably need to use something more efficient - for example, our rdma lib, which is going to be integrated into filestore here: https://github.com/ydb-platform/nbs/issues/5388

PR #6606, opened 2026-07-27
issue-5894, issue-5895: NaiveMirroredFileSystemShard implementation [NOT FOR MERGING]

qkrorlqr:
### Notes
Prototype implementation of a mirrored filesystem shard on top of journalled devices. This PR won't be merged as is - I'll split it into smaller PRs. This one is just to show the end result.

```
     shard
       |
    page store
       |
[storage groups] <--- the prototype uses only one group
```

#### Shard
The shard implements `IFileSystemShard` and consists of:
* inode table - stores the `NodeId -> NodeAttr` mapping, responsible for `NodeId` allocation
* name table - stores the `Name -> NodeId` mapping
* handle table - stores the `Handle -> NodeId` mapping, responsible for `Handle` allocation
* page index - stores the `<NodeId, PageNo> -> PhysicalPageNo` mapping
* page allocator - stores the `PhysicalPageNo -> free/allocated` bitmap, responsible for `PhysicalPage` allocation

#### PageStore
The shard accesses storage via a "page store" layer which is responsible for:
* forwarding read/write requests to storage groups
* maintaining in-memory page cache
* tracking dirty pages

#### Differences from the planned production implementation
* We'll need to add explicit storage group numbers into the `IPageStore` interface to let the shard co-locate the data and metadata of a single inode inside the same storage group.
* We'll need cross-group 2PC support - for the files that don't fit into a single storage group

#### StorageGroup
Each storage group consists on N storage devices and is responsible for:
* data redundancy
* device locking [missing in prototype]
* recovery [missing in prototype]
* proper retries and ordering - ensuring that LSN moves forward monotonically and without gaps [missing in prototype]

#### StorageNode
_The interface was merged and then a naive implementation was created in other PRs_
Responsible for:
* applying all writes belonging to a single write-log-record atomically (via a journal) [the prototype doesn't physically have a journal]
* rejecting the writes whose LSNs are not properly ordered [missing in prototype]
* reading (obvious)
* rejecting old writers (whose leases have been invalidated by another writer with a newer `Generation`) [missing in prototype]

#### NodeTable, NameTable, HandleTable
All tables are implemented using `TPersistentHashTable` which is an open-addressed hash table with linear probing and with batch tombstone cleanup upon element deletion

#### PageIndex
Page index is also implemented using `TPersistentHashTable` - with `<NodeId, PageClusterNo>` pairs as its key and `StoragePageClusterNo` as its value. Each page cluster is just a group of 8 consecutive pages.

#### PageAllocator
The bitmap used by page allocator is implemented using `TPersistentBitmap` which stores the data in page-sized chunks (so for PageSize==4KiB each chunk stores 2^15 bits) and uses an in-memory stack to track the chunks that have zero bits. Each bit corresponds to a page cluster.

#### Differences from the planned production implementation
* Production implementation will most probably use per-node radix trees to index file data (instead of a hash table). So PageIndex will be implemented very differently.
* We'll probably also have a more sophisticated page allocator implementation because always allocating in multiples of 32KiB (8 pages) would probably waste too much space. 
* We'll need to be able to address storage pages from other groups (for large files).

### Layout
Per-group layout:
```
// offsets here are in bytes but in the implementation they are rounded up to PageSize
Offset 0: --------------------------------------------- Node Table - 100 bytes per slot, N slots
Offset 100 * N: --------------------------------------- Name Table - 32 bytes per slot, N slots
Offset 100 * N + 32 * N: ------------------------------ Handle Table - 16 bytes per slot, 10 * N slots
Offset 100 * N + 32 * N + 160 * N: -------------------- Page Index - 24 bytes per slot, M slots
Offset 100 * N + 32 * N + 160 * N + 24 * M ------------ Page Allocator Bitmap - M bits, M / 2^15 pages
Offset 100 * N + 32 * N + 160 * N + 24 * M + M / 8 ---- Data Pages
```

For N == 100'000 (100k files per storage group) and M == 3'000'000 (3m page clusters per storage group - i.e. 3m x 32KiB = 91.5GiB of space) and hashtable loadfactor of 0.5 the total space allocated for the metadata is thus:
```
(100000 * 100 + 100000 * 32 + 10 * 100000 * 16 + 3000000 * 24 + 3000000 / 8) / 1024 / 1024 / 0.5 = 194MiB
```

### Issue


qkrorlqr:
This PR was split into the following PRs all of which have been merged:
* https://github.com/ydb-platform/nbs/pull/6607
* https://github.com/ydb-platform/nbs/pull/6615
* https://github.com/ydb-platform/nbs/pull/6614
* https://github.com/ydb-platform/nbs/pull/6617
* https://github.com/ydb-platform/nbs/pull/6630

Issue #6610, opened 2026-07-27
[Filestore] TSAN reports a race when reading from and writing to FUSE headers mapped from guest memory


SvartMetal:
First (not working) attempt to fix the issue https://github.com/ydb-platform/nbs/pull/4600

```
==================
WARNING: ThreadSanitizer: data race (pid=246293)
  Read of size 8 at 0x7f3b00e0f728 by thread T153:
    #0 __tsan_memcpy /actions-runner/_work/nbs/nbs/contrib/libs/clang16-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:3165:3 (filestore-vhost+0x13709a00) (BuildId: a18247a9b02226f48c4137dba3368e013f1b21ad)
    #1 iov_iter_to_buf /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:228:9 (filestore-vhost+0x2f0c61cd) (BuildId: a18247a9b02226f48c4137dba3368e013f1b21ad)
    #2 process_request /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:278:20 (filestore-vhost+0x2f0c61cd)
    #3 virtio_session_loop /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:544:19 (filestore-vhost+0x2f0c61cd)
    #4 fuse_session_loop /actions-runner/_work/nbs/nbs/cloud/contrib/virtiofsd/fuse_lowlevel.c:2859:12 (libvirtiofsd.so+0xbafe) (BuildId: b656ddf5b581b810ed9c809f242f20a24dc2176e)
    #5 NCloud::NFileStore::NFuse::(anonymous namespace)::TFuseLoopThread::ThreadProc() /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/loop.cpp:612:9 (filestore-vhost+0x2ef8fc7b) (BuildId: a18247a9b02226f48c4137dba3368e013f1b21ad)
    #6 void* (anonymous namespace)::ThreadProcWrapper<ISimpleThread>(void*) /actions-runner/_work/nbs/nbs/util/system/thread.cpp:383:45 (filestore-vhost+0x13911182) (BuildId: a18247a9b02226f48c4137dba3368e013f1b21ad)
[...]

PR #4600, opened 2025-11-03
[Filestore] mute tsan data race between iov_copy_to_iov and iov_iter_to_buf in vfs_fuse/vhost/fuse_virtio.c

SvartMetal:
```
sanitizer found errors in /home/github/.ya/build/build_root/3xuw/002f10/cloud/filestore/tests/fio_index/qemu-kikimr-multishard-test/test-results/py3test/chunk2/testing_out_stuff/filestore-vhost.err:
: data race (pid=392811)
  Read of size 8 at 0x7fe574c88728 by thread T146:
    #0 __tsan_memcpy /actions-runner/_work/nbs/nbs/contrib/libs/clang16-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:3165:3 (filestore-vhost+0xf9f06e0) (BuildId: 835a0e8006e8b5af5d7f83246796d057ca0346b8)
    #1 iov_iter_to_buf /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:190:9 (filestore-vhost+0x2797175a) (BuildId: 835a0e8006e8b5af5d7f83246796d057ca0346b8)
    #2 process_request /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:253:5 (filestore-vhost+0x2797175a)
    #3 virtio_session_loop /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:494:19 (filestore-vhost+0x2797175a)
    #4 fuse_session_loop /actions-runner/_work/nbs/nbs/contrib/libs/virtiofsd/fuse_lowlevel.c:2694:12 (libvirtiofsd.so+0xb67e) (BuildId: d669c68e288fc6da51483d7bfbbc6b4af5a8ba09)
    #5 NCloud::NFileStore::NFuse::(anonymous namespace)::TFuseLoopThread::ThreadProc() /actions-runner/_work/nbs/nbs/cloud/filestore/libs/vfs_fuse/loop.cpp:531:9 (filestore-vhost+0x278adf31) (BuildId: 835a0e8006e8b5af5d7f83246796d057ca0346b8)
[...]

PR #6611, opened 2026-07-27
[Filestore] mute tsan data race between iov_copy_to_iov and iov_iter_to_buf in vfs_fuse/vhost/fuse_virtio.c

SvartMetal:
### Notes

The annotations added in #4600 did not work. The nightly tsan build reports the same pair again, at the same two memcpy calls, with the annotations compiled in: a read in `iov_iter_to_buf` against a previous write in `iov_copy_to_iov`, both landing in guest memory backed by `memfd:memory-backend-memfd`.

The reason is that `__tsan_acquire` and `__tsan_release` key the sync object on the exact address. `MetaMap::GetSync` looks it up with `s->addr == addr`, so a release and an acquire only meet when both name the same byte. #4600 released on the destination chunk start in `iov_copy_to_iov` and acquired on the source chunk start in `iov_iter_to_buf`, and those two addresses are almost never equal. The virtio-fs driver lays a request out as one allocation holding the in header and args followed by the out header and args, so when the guest recycles those pages the buffer the response was written at and the buffer the next request is read from are offset from each other. No happens-before edge gets created and the race is reported as soon as the recycled ranges overlap.

This PR replaces the per-buffer annotation with a single sync token for the whole guest memory. Since the token no longer depends on the buffer address the two sides always meet, and the annotation can sit directly on the memcpy calls rather than around the enclosing loop, which keeps the window where two threads are inside the token short. Both directions are annotated now, so neither a write after a read nor a read after a write is reported.

Report from the nightly tsan run, trimmed to the top frames of both stacks:

```
data race (pid=246293)
  Read of size 8 at 0x7f3b00e0f728 by thread T153:
    #0 __tsan_memcpy contrib/libs/clang16-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:3165:3
    #1 iov_iter_to_buf cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:228:9
    #2 process_request cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:278:20
    #3 virtio_session_loop cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:544:19
    #4 fuse_session_loop cloud/contrib/virtiofsd/fuse_lowlevel.c:2859:12
    #5 NCloud::NFileStore::NFuse::(anonymous namespace)::TFuseLoopThread::ThreadProc() cloud/filestore/libs/vfs_fuse/loop.cpp:612:9

  Previous write of size 8 at 0x7f3b00e0f728 by thread T88:
    #0 __tsan_memcpy contrib/libs/clang16-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:3165:3
    #1 iov_copy_to_iov cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:118:17
    #2 virtio_send_msg cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c:574:9
    #3 fuse_send_msg cloud/contrib/virtiofsd/fuse_lowlevel.c:234:16
    #4 fuse_send_reply_iov_nofree cloud/contrib/virtiofsd/fuse_lowlevel.c:258:12
    #5 send_reply_iov cloud/contrib/virtiofsd/fuse_lowlevel.c:266:11
    #6 send_reply cloud/contrib/virtiofsd/fuse_lowlevel.c:281:12
    #7 send_reply_ok cloud/contrib/virtiofsd/fuse_lowlevel.c:353:12
    #8 fuse_reply_write cloud/contrib/virtiofsd/fuse_lowlevel.c:520:12
    #9 NCloud::NFileStore::NFuse::ReplyWrite(...) cloud/filestore/libs/vfs_fuse/fs.cpp:254:15
    ...
    #29 NCloud::NFileStore::NFuse::TWriteBackCache::TImpl::ExecuteRequestUnderBarrier<...> cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp:600:25

  Location is global '??' at 0x7f3a40000000 (memfd:memory-backend-memfd (deleted)+0xc0e0f728)
```

Full log: https://github-actions-s3.storage.eu-north2.nebius.cloud/ydb-platform/nbs/Nightly-build-(tsan)/29546143985/1/nebius-x86-64-tsan/summary/1/ya-test.html

Failing suite: `cloud/filestore/tests/fio_index/qemu-kikimr-multishard-test`,
chunk6, `SANITIZER_TYPE=thread`.

### Issue

#6610

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c
+++ b/cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c
@@ -21,12 +21,36 @@
 #define Y_UNUSED(x) (void)x;
 #endif
 
+// Each fuse request comes with an iovec that is split into in and out iovecs.
+// The in iovec is read in iov_iter_to_buf to copy its content to an allocated
+// buffer. The out iovec is written in iov_copy_to_iov to send the data
+// response. Both point into guest memory.
+//
+// Tsan cannot see the synchronization that actually orders these accesses,
+// because it happens in the guest via the virtqueue. The guest reuses the
+// pages of an already completed request as the in iovec of the next one, so
+// tsan shadow memory still thinks the memory is read while it is written.
+//
+// Per-buffer __tsan_acquire/__tsan_release do not help here: the address the
+// response was written at and the address the next request is read from are
+// almost never the same, and these annotations only match when the addresses
+// are equal. Use a single sync token for the whole guest memory instead, taken
+// and released around each copy, and annotate both directions, so that neither
+// ordering is reported.
+//
+// See https://github.com/ydb-platform/nbs/pull/6611
 #if __has_feature(thread_sanitizer)
-#define TSAN_ACQUIRE(x) __tsan_acquire(x)
-#define TSAN_RELEASE(x) __tsan_release(x)
+
+static char guest_memory_sync_token;
+
+void __tsan_acquire(void* addr);
+void __tsan_release(void* addr);
+
+#define TSAN_ACQUIRE_GUEST_MEMORY() __tsan_acquire(&guest_memory_sync_token)
+#define TSAN_RELEASE_GUEST_MEMORY() __tsan_release(&guest_memory_sync_token)
 #else
-#define TSAN_ACQUIRE(x)
-#define TSAN_RELEASE(x)
+#define TSAN_ACQUIRE_GUEST_MEMORY()
+#define TSAN_RELEASE_GUEST_MEMORY()
 #endif
 
 struct fuse_virtio_dev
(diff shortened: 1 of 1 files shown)


SvartMetal:
image

Not relevant to this change

debnatkh (approved):

cloud/filestore/libs/vfs_fuse/vhost/fuse_virtio.c
```diff
@@ -21,12 +21,36 @@
...
+// tsan shadow memory still thinks the memory is read while it is written.
+//
+// Per-buffer __tsan_acquire/__tsan_release do not help here: the address the
+// response was written at and the address the next request is read from are
+// almost never the same, and these annotations only match when the addresses
+// are equal. Use a single sync token for the whole guest memory instead, taken
+// and released around each copy, and annotate both directions, so that neither
+// ordering is reported.
+//
+// See https://github.com/ydb-platform/nbs/pull/6611
```
debnatkh, line 41: References to the same PR seem a bit redundant, since blame does the same job. Maybe reference the issue instead of the PR

SvartMetal: I would prefer not to rely on `blame`, as it can be difficult to search through when a large number of commits are affected

komarevtsev-d (approved):

Issue #1559, opened 2024-07-06
[Disk Manager] Implement filesystem backups


qkrorlqr:
This issue is mostly about the Disk Manager part. Filestore checkpointing is out of scope here - in this issue we need to implement filesystem backups by simply reading the current state of the filesystem. Yes, it will not be consistent, but:
1. it's much better than nothing
2. after we add checkpoints we will simply switch the implementation that is going to appear in this issue from reading the current state to making a checkpoint and reading the state from that checkpoint

qkrorlqr:
There is actually an alternative approach for backups: create a VM, attach Filestore to it and do a simple rsync from Filestore to a S3-based FS.

jkuradobery:
The decided approach have slightly change from parallelized layer-by-layer approach towards recursive traversal approach. 

Also, the bigger concern now is losing underlying blob-storage group for a single shard with directories in shards enabled.
In that case we would loose huge parts of filesystem while having lost a single shard. 
To handle this, we would like to:
1) Backup data as usual saving shard node name and shard id.
2) Implement a task RestoreShardFromSnapshot, which for each node in given shard would call UnsafeCreateNode, and for each child of each directory in this shard would call UnsafeCreateNodeRef. (We need to iterate over inodes children, but not deeper). 

<img width="961" height="631" alt="Image" src="https://github.com/user-attachments/assets/306d30f5-fc80-42ba-a4b3-2200d332a3ec" />

PR #6620, opened 2026-07-28
issue-1559: [Disk Manager] Wire filesystem snapshot deletion

jkuradobery:
### Issue
#1559


cloud/disk_manager/internal/pkg/services/filesystem_snapshot/delete_filesystem_snapshot_task_test.go
SvartMetal, line 1: Let's also add facade integration tests that cover filesystem deletion and cancellation

cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
```diff
@@ -233,6 +295,69 @@ func TestCreateFilesystemSnapshot(t *testing.T) {
...
+	nfsClient := testcommon.NewNfsTestingClient(t, ctx, "zone-a")
+	defer nfsClient.Close()
+
+	filesystemID := t.Name() + "_filesystem"
+	snapshotID := t.Name() + "_snapshot"
+
+	createFilesystem(t, ctx, client, filesystemID)
+	defer func() {
+		err := nfsClient.Delete(ctx, filesystemID, true)
+		require.NoError(t, err)
```
SvartMetal, line 321: Why do you need to delete filesystem manually?

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
```diff
@@ -947,6 +980,32 @@ func GetSnapshotMeta(
...
+	ctx context.Context,
+	snapshotID string,
+) (*resources.FilesystemSnapshotMeta, error) {
+
+	storage, err := newResourceStorage(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return storage.GetFilesystemSnapshotMeta(ctx, snapshotID)
```
SvartMetal: It looks like the YDB client is leaked in these methods

cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
```diff
@@ -49,6 +50,67 @@ func createFilesystem(
...
+
+	require.Eventually(t, func() bool {
+		meta, err := testcommon.GetFilesystemSnapshotMeta(ctx, snapshotID)
+		if err != nil || meta == nil {
+			return false
+		}
+
+		dataplaneMeta, err :=
+			testcommon.GetDataplaneFilesystemSnapshotMeta(ctx, snapshotID)
+		return err == nil && dataplaneMeta != nil
```
SvartMetal: Maybe we should log the errors?

I know the errors should be logged in the persistence package, but we could also add a helpful message here indicating that we are still waiting for the operation to complete

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
jkuradobery: I made it consistent with surrounding methods.

cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
jkuradobery: Because we have sessions open. I think we can remove this unit test and check snapshot deletion during fs-from-snapshot creation instead.

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
SvartMetal: I don't think it would be hard to fix this for the existing methods as well

cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
jkuradobery: I mean delete TestDeleteFilesystemDuringSnapshotCreationDeletesSnapshot, because in the current test we do not test race with deletion but we need to delete afterwards. And if we do not delete with force, there is a chance of failure of the deletion in nemesis tests.

SvartMetal (approved):

gy2411 (approved):

Issue #6624, opened 2026-07-28
[Filestore] ResetSession tx split


neihar:
Currently we can have huge amount of open handles which will 
1) take a lot of time to prepare TX, sometimes with a lot of restarts 
2) [tx can be large than size limit for tx](https://github.com/ydb-platform/nbs/blob/1ef2d8df021fe62ce5f174c339aedbab9e15ba2d/contrib/ydb/core/tablet_flat/flat_executor.cpp#L139) . 

For example, we faced this error:
`TABLET_EXECUTOR CRIT: Leader{72075186227744157:1026:2} Tx{5, NCloud::NFileStore::NStorage::TTabletBase<NCloud::NFileStore::NStorage::TIndexTabletActor>::TTransaction<NCloud::NFileStore::NStorage::TIndexTabletActor::TResetSession>} fatal commit failure: Redo commit of 5454462681 bytes is more than the allowed limit`

So, the proposed solution is to split tx into some batches and proceed iteratively 
The split will be applied both for reset session and destroy session

PR #6625, opened 2026-07-28
Filestore: split reset session into parts #6624

neihar:
### Notes
Splited handles cleanup to iterative parts 
Added ai-assisted tests

### Issue
#6624

### DIFF
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -35,6 +35,7 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
                                                                                \
     xxx(EstablishSessionTimeout,       TDuration, TDuration::Seconds(30)      )\
     xxx(IdleSessionTimeout,            TDuration, TDuration::Hours(1)         )\
+    xxx(MaxResetSessionHandlesPerTx,   ui32,      1'000'000                   )\
                                                                                \
     xxx(WriteBlobThreshold,            ui32,      128_KB                      )\
                                                                                \
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
@@ -62,6 +62,8 @@ void TIndexTabletActor::HandleResetSession(
         std::move(requestInfo),
         sessionId,
         seqNo,
+        Config->GetMaxResetSessionHandlesPerTx(),
+        false /* continuation */,
         std::move(msg->Record));
 }
 
@@ -94,8 +96,12 @@ bool TIndexTabletActor::PrepareTx_ResetSession(
 
     bool ready = true;
     auto commitId = GetCurrentCommitId();
-    args.Nodes.reserve(session->Handles.Size());
+    ui32 handleCount = 0;
     for (const auto& handle: session->Handles) {
+        if (args.MaxHandlesPerTx && ++handleCount > args.MaxHandlesPerTx) {
+            break;
+        }
+
         if (args.Nodes.contains(handle.GetNodeId())) {
             continue;
         }
@@ -124,21 +130,39 @@ void TIndexTabletActor::ExecuteTx_ResetSession(
 
     auto* session = FindSession(args.SessionId);
     if (!session) {
+        if (args.Continuation) {
+            args.Error = MakeError(
+                E_REJECTED,
+                "session reset interrupted: session destroyed");
+        }
+        args.Completed = true;
         return;
     }
 
     if (!CheckSessionForDestroy(session, args.SessionSeqNo)) {
+        if (args.Continuation) {
+            args.Error = MakeError(
+                E_REJECTED,
+                "session reset interrupted: session recovered");
+        }
+        args.Completed = true;
         return;
     }
 
     auto commitId = GenerateCommitId();
     if (commitId == InvalidCommitId) {
         args.OnCommitIdOverflow();
+        args.Completed = true;
         return;
     }
 
+    ui32 handleCount = 0;
     auto handle = session->Handles.begin();
     while (handle != session->Handles.end()) {
+        if (args.MaxHandlesPerTx && ++handleCount > args.MaxHandlesPerTx) {
+            break;
+        }
+
         auto nodeId = handle->GetNodeId();
         DestroyHandle(*db, &*(handle++));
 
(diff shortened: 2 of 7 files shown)


cloud/filestore/libs/storage/tablet/tablet_ut_sessions.cpp
```diff
@@ -402,6 +402,313 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_Sessions)
...
+        tablet.UnlinkNode(RootNodeId, "yyy", false);
+
+        auto statsBefore = GetStorageStats(tablet);
+        UNIT_ASSERT_VALUES_EQUAL(5, statsBefore.GetUsedHandlesCount());
+
+        ui32 commitCount = 0;
+        env.GetRuntime().SetObserverFunc(
+            [&](TAutoPtr<NActors::IEventHandle>& event)
+            {
+                if (event->GetTypeRewrite() == NKikimr::TEvTablet::EvCommit &&
```
qkrorlqr, line 441: this looks fragile - I would suggest to track the number of tx-es via our own counters

cloud/filestore/libs/storage/tablet/tablet_tx.h
```diff
@@ -542,18 +542,25 @@ struct TTxIndexTablet
         /* const */ TRequestInfoPtr RequestInfo;
         const TString SessionId;
         const ui64 SessionSeqNo;
+        const ui32 MaxHandlesPerTx;
+        const bool Continuation;
```
qkrorlqr: nit: IsContinuation?

qkrorlqr (commented):
overall LGTM

cloud/filestore/libs/storage/tablet/tablet_ut_sessions.cpp
qkrorlqr: but that's not critical

cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
```diff
@@ -124,21 +130,39 @@ void TIndexTabletActor::ExecuteTx_ResetSession(
 
     auto* session = FindSession(args.SessionId);
     if (!session) {
+        if (args.Continuation) {
+            args.Error = MakeError(
+                E_REJECTED,
```
SvartMetal, line 137: It looks like the durable client will keep retrying `ResetSession` and continue receiving `E_REJECTED` in this case.

Is that expected?

A critical event might also be helpful

cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
```diff
@@ -124,21 +130,39 @@ void TIndexTabletActor::ExecuteTx_ResetSession(
...
+                "session reset interrupted: session destroyed");
+        }
+        args.Completed = true;
         return;
     }
 
     if (!CheckSessionForDestroy(session, args.SessionSeqNo)) {
+        if (args.Continuation) {
+            args.Error = MakeError(
+                E_REJECTED,
```
SvartMetal, line 151: Same comment as above.

It looks odd that we can leave the session in a partially destroyed state

cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
qkrorlqr: no, on the next retry `args.Continuation` will be `false`

qkrorlqr: but a critical event can be useful, yes

neihar: It will be rejected only mid-chain. After retry we will not have "Continuation" flag at first tx and exit as before.

cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
neihar: I agree that this looks odd, but from correctness perspective it is okay to leave some handles uncleaned. Before it could happen for none or for all, but I don't see a big harm in partial state as well. Moreover, I don't really see a case where it can fire at the moment

cloud/filestore/libs/storage/tablet/tablet_ut_sessions.cpp
neihar: I used this approach many time, so far, I do not recall problems, if it. fires, I will consider to use different approach

cloud/filestore/libs/storage/tablet/tablet_tx.h
neihar: changed

cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
SvartMetal: I think, critical event might be helpful in this case also

neihar: Added critical even for both cases

cloud/filestore/libs/storage/core/config.cpp
```diff
@@ -35,6 +35,7 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
                                                                                \
     xxx(EstablishSessionTimeout,       TDuration, TDuration::Seconds(30)      )\
     xxx(IdleSessionTimeout,            TDuration, TDuration::Hours(1)         )\
+    xxx(MaxResetSessionHandlesPerTx,   ui32,      1'000'000                   )\
```
SvartMetal, line 38: Do we have integration tests with more than 1 million handles?

Maybe we should set this parameter to a lower value (for example, 10-100) in the `new-features` configuration?

neihar: I don't see tests which will execute resetsession with open handles in py tests. We can consider adding openhandles to migration test

SvartMetal (approved):

yegorskii (approved):

cloud/filestore/libs/storage/core/config.cpp
SvartMetal: I think it is reasonable to add a test case here https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/tests/async_close_test

PR #6634, opened 2026-07-29
Filestore: Split destroy session tx 

neihar:
### Notes
Execute DestorySession tx iteratively instead of single tx
Added AI asissited tests

### Issue
#6624

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -238,11 +238,11 @@ bool TIndexTabletActor::CheckSessionForDestroy(const TSession* session, ui64 seq
         session->GetSessionRwSeqNo() == seqNo;
 }
 
-bool TIndexTabletActor::ReadNodesForSessionHandles(
+bool TIndexTabletActor::ReadNodesToRemoveForSessionHandles(
     IIndexTabletDatabase& db,
     const TSession& session,
     ui32 maxHandlesPerTx,
-    TNodeSet& nodes)
+    TNodeSet& nodesToRemove)
 {
     bool ready = true;
     auto commitId = GetCurrentCommitId();
@@ -252,7 +252,7 @@ bool TIndexTabletActor::ReadNodesForSessionHandles(
             break;
         }
 
-        if (nodes.contains(handle.GetNodeId())) {
+        if (nodesToRemove.contains(handle.GetNodeId())) {
             continue;
         }
 
@@ -263,7 +263,7 @@ bool TIndexTabletActor::ReadNodesForSessionHandles(
             TABLET_VERIFY(node);
             if (node->Attrs.GetLinks() == 0) {
                 // candidate to be removed
-                nodes.insert(*node);
+                nodesToRemove.insert(*node);
             }
         }
     }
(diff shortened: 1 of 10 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -238,6 +238,106 @@ bool TIndexTabletActor::CheckSessionForDestroy(const TSession* session, ui64 seq
         session->GetSessionRwSeqNo() == seqNo;
 }
 
+bool TIndexTabletActor::ReadNodesForSessionHandles(
```
SvartMetal: ReadNodesToRemoveForSessionHandles?

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -238,6 +238,106 @@ bool TIndexTabletActor::CheckSessionForDestroy(const TSession* session, ui64 seq
...
+}
+
+void TIndexTabletActor::DestroySessionHandles(
+    IIndexTabletDatabase& db,
+    const TActorContext& ctx,
+    TSession* session,
+    ui64 commitId,
+    ui32 maxHandlesPerTx,
+    bool isContinuation,
+    const TNodeSet& nodes,
```
SvartMetal: nodesToRemove

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -238,6 +238,106 @@ bool TIndexTabletActor::CheckSessionForDestroy(const TSession* session, ui64 seq
...
+                // candidate to be removed
+                nodes.insert(*node);
+            }
+        }
+    }
+
+    return ready;
+}
+
+void TIndexTabletActor::DestroySessionHandles(
```
SvartMetal: DestroySessionHandleAndRemoveNodes?

SvartMetal (changes requested):

SvartMetal:
LGTM overall

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
neihar: Ok

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
neihar: ok

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
neihar: I don't know, this one looks strange for me, but if you insist I am okay with it

neihar: discussed offline, agree that ToRemove make sense here

SvartMetal: ReadNodesForSessionHandles also looks strange because it also checks `GetLinks() == 0`

cloud/filestore/libs/storage/tablet/tablet_actor_resetsession.cpp
SvartMetal, line 119: Can this critical event occur during normal operation because of concurrent `ResetSession` requests?

For example:

* `tx0`: `ResetSession` is executed and creates continuation `tx1`
* `ResetSession` is retried by a durable client, and the request is queued as `tx2`, which is not a continuation
* `tx1` is executed and creates another continuation, `tx3`
* `tx2` is executed and destroys all remaining handles and the session
* `tx3` is executed, finds that the session has already been destroyed, and fires the critical event

neihar: At first glance it is possible, theoretically, but shouldn't be a problem. In real cases, looks like we didn't have cases for more than 1 tx anyway.

neihar: let's observe those critical events and act reactively instead of proactively in this case, as I don't see harmful side effects

SvartMetal (approved):

yegorskii (approved):

PR #6638, opened 2026-07-29
[Filestore] issue-4547: fake DescribeData should report metrics

SvartMetal:
### Notes
Follow-up for #4930

`HandleFakeDescribeData` replied without touching `Metrics.DescribeData`, so fake DescribeData responses were invisible in the tablet request metrics. A load test running with `FakeDescribeDataEnabled` reported zero DescribeData RPS and zero throughput, which makes the fake path hard to use for performance modeling because its numbers cannot be compared against the real path.

### Issue
#4547

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
@@ -1194,6 +1194,12 @@ void TIndexTabletActor::HandleFakeDescribeData(
 
     ReportFakeDescribeDataHappened();
 
+    const TByteRange byteRange(
+        msg->Record.GetOffset(),
+        msg->Record.GetLength(),
+        GetBlockSize()
+    );
+
     auto requestInfo = CreateRequestInfo(
         ev->Sender,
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
```diff
@@ -1254,7 +1255,17 @@ void TIndexTabletActor::HandleFakeDescribeData(
     // Spin until deadline
     while (GetCycleCount() < deadlineCycles) {}
 
+    CompleteResponse<TEvIndexTablet::TDescribeDataMethod>(
+        response->Record,
+        requestInfo->CallContext,
+        ctx);
```
SvartMetal, line 1261: To be consistent with https://github.com/ydb-platform/nbs/blob/24ed3c4a3f73c340702de5285f515a8c5ae5735c/cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp#L1051

qkrorlqr (approved):

agalibin (approved):

PR #6642, opened 2026-07-30
[DM] Add TTL mechanism for idle base disks in pools

agalibin:
### Notes
Add automatic cleanup of idle base disks in pools via the existing `OptimizeBaseDisks` periodic task. Base disks with zero active overlay disks for longer than a configurable TTL are retired and their pool capacity is reduced to prevent the scheduler from recreating them.

### DIFF
--- a/cloud/disk_manager/internal/pkg/services/pools/config/config.proto
+++ b/cloud/disk_manager/internal/pkg/services/pools/config/config.proto
@@ -45,4 +45,10 @@ message PoolsConfig {
     //  * Reducing this number also increases the total size of the base disks
     //    that need to be allocated.
     optional uint32 BaseDiskOverSubscription = 24 [default = 2];
+    // TTL for idle base disks in pools. Base disks with zero overlay disks
+    // for longer than this duration will be retired (ejected from pool and deleted).
+    // Set to "0s" to disable.
+    optional string BaseDiskIdleTTL = 25 [default = "0s"];
+    // Maximum number of idle base disks to process in a single cleanup run.
+    optional uint64 CleanupIdleBaseDisksLimit = 26 [default = 100];
 }
--- a/cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
+++ b/cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
@@ -22,39 +22,39 @@ type optimizeBaseDisksTask struct {
 	convertToImageSizedBaseDisksThreshold   uint64
 	convertToDefaultSizedBaseDisksThreshold uint64
 	minPoolAge                              time.Duration
+	baseDiskIdleTTL                         time.Duration
+	cleanupIdleBaseDisksLimit               uint64
 	state                                   *protos.OptimizeBaseDisksTaskState
 }
 
-func (t *optimizeBaseDisksTask) Save() ([]byte, error) {
-	return proto.Marshal(t.state)
-}
-
-func (t *optimizeBaseDisksTask) Load(_, state []byte) error {
-	t.state = &protos.OptimizeBaseDisksTaskState{}
-	return proto.Unmarshal(state, t.state)
-}
-
-func (t *optimizeBaseDisksTask) Run(
+func (t *optimizeBaseDisksTask) ensureResourcesToOptimizeCollected(
 	ctx context.Context,
 	execCtx tasks.ExecutionContext,
+	sizeOptimizationEnabled bool,
+	idleCleanupEnabled bool,
 ) error {
 
-	t1 := t.convertToDefaultSizedBaseDisksThreshold
-	t2 := t.convertToImageSizedBaseDisksThreshold
-	common.Assert(t1 > t2, `ConvertToDefaultSizedBaseDisksThreshold should be greater 
-		than convertToImageSizedBaseDisksThreshold`)
-
-	if t1 == 0 && t2 == 0 {
+	if t.state.CollectedResourcesToOptimize {
 		return nil
 	}
 
-	if len(t.state.ConfigurePoolRequests) == 0 {
-		poolInfos, err := t.storage.GetReadyPoolInfos(ctx)
-		if err != nil {
-			return err
-		}
+	poolInfos, err := t.storage.GetReadyPoolInfos(ctx)
+	if err != nil {
+		return err
+	}
 
-		now := time.Now()
+	now := time.Now()
+
+	type poolKey struct {
+		imageID string
+		zoneID  string
+	}
+
+	optimizedPools := make(map[poolKey]bool)
+
+	if sizeOptimizationEnabled {
+		t1 := t.convertToDefaultSizedBaseDisksThreshold
+		t2 := t.convertToImageSizedBaseDisksThreshold
 
 		for _, poolInfo := range poolInfos {
 			dateThreshold := poolInfo.CreatedAt.Add(t.minPoolAge)
@@ -75,8 +75,8 @@ func (t *optimizeBaseDisksTask) Run(
 				continue
 			}
 
-			t.state.ConfigurePoolRequests = append(
-				t.state.ConfigurePoolRequests,
+			t.state.ConfigurePoolForImageSizeRequests = append(
+				t.state.ConfigurePoolForImageSizeRequests,
 				&protos.ConfigurePoolRequest{
 					ZoneId:       poolInfo.ZoneID,
 					ImageId:      poolInfo.ImageID,
@@ -84,17 +84,115 @@ func (t *optimizeBaseDisksTask) Run(
 					UseImageSize: newUseImageSize,
 				},
 			)
-		}
 
-		err = execCtx.SaveState(ctx)
-		if err != nil {
-			return err
+			// Pools selected for size optimization will have all their base disks retired,
+			// including idle ones, so there is no need to run idle cleanup for them.
+			optimizedPools[poolKey{poolInfo.ImageID, poolInfo.ZoneID}] = true
 		}
 	}
 
+	if idleCleanupEnabled {
+		for _, poolInfo := range poolInfos {
+			key := poolKey{poolInfo.ImageID, poolInfo.ZoneID}
+			if optimizedPools[key] {
+				continue
+			}
+
+			idleDisks, err := t.storage.GetIdleBaseDisks(
+				ctx,
+				poolInfo.ImageID,
+				poolInfo.ZoneID,
+				t.baseDiskIdleTTL,
+				t.cleanupIdleBaseDisksLimit,
+			)
+			if err != nil {
+				return err
+			}
+
+			if len(idleDisks) == 0 {
+				continue
+			}
+
+			var reduction uint32
+			for _, disk := range idleDisks {
+				reduction += uint32(disk.FreeSlots)
+				t.state.IdleBaseDiskIds = append(
+					t.state.IdleBaseDiskIds,
+					disk.ID,
+				)
+			}
+
+			// TODO(https://github.com/ydb-platform/nbs/issues/6684):
+			// setting capacity to 0 may cause a race. If a slot is acquired
+			// on an idle disk between ConfigurePool and RetireBaseDisk, the
+			// retire triggers a rebase that creates a new base disk with srcDisk=nil.
+			// The scheduler (takeBaseDisksToSchedule) never picks it up: no SrcDisk
+			// for the global path, and capacity=0 excludes the pool from the per-pool
+			// path. The rebase task retries indefinitely, hanging the optimization.
+			newCapacity := uint32(0)
+			if poolInfo.Capacity > reduction {
+				newCapacity = poolInfo.Capacity - reduction
+			}
+
+			t.state.PoolReductionRequests = append(
+				t.state.PoolReductionRequests,
+				&protos.ConfigurePoolRequest{
+					ZoneId:       key.zoneID,
+					ImageId:      key.imageID,
+					Capacity:     newCapacity,
+					UseImageSize: poolInfo.ImageSize > 0,
+				},
+			)
+		}
+	}
+
+	t.state.CollectedResourcesToOptimize = true
+	return execCtx.SaveState(ctx)
+}
+
+func (t *optimizeBaseDisksTask) Save() ([]byte, error) {
+	return proto.Marshal(t.state)
+}
+
+func (t *optimizeBaseDisksTask) Load(_, state []byte) error {
+	t.state = &protos.OptimizeBaseDisksTaskState{}
+	return proto.Unmarshal(state, t.state)
+}
+
+func (t *optimizeBaseDisksTask) Run(
+	ctx context.Context,
+	execCtx tasks.ExecutionContext,
+) error {
+
+	t1 := t.convertToDefaultSizedBaseDisksThreshold
+	t2 := t.convertToImageSizedBaseDisksThreshold
+
+	sizeOptimizationEnabled := t1 > 0 || t2 > 0
+	if sizeOptimizationEnabled {
+		common.Assert(t1 > t2, `ConvertToDefaultSizedBaseDisksThreshold should be greater
+			than convertToImageSizedBaseDisksThreshold`)
+	}
+
+	idleCleanupEnabled := t.baseDiskIdleTTL > 0
+
+	if !sizeOptimizationEnabled && !idleCleanupEnabled {
+		return nil
+	}
+
+	err := t.ensureResourcesToOptimizeCollected(
+		ctx,
+		execCtx,
+		sizeOptimizationEnabled,
+		idleCleanupEnabled,
+	)
+	if err != nil {
+		return err
+	}
+
+	// Execute optimization: ConfigurePool + RetireBaseDisks for each pool.
 	var configurePoolTaskIDs []string
 
-	for _, request := range t.state.ConfigurePoolRequests {
+	for _, request := range t.state.ConfigurePoolForImageSizeRequests {
 		taskID, err := t.scheduler.ScheduleTask(
 			headers.SetIncomingIdempotencyKey(
 				ctx,
@@ -124,7 +222,7 @@ func (t *optimizeBaseDisksTask) Run(
 			return err
 		}
 
-		request := t.state.ConfigurePoolRequests[i]
+		request := t.state.ConfigurePoolForImageSizeRequests[i]
 
 		taskID, err = t.scheduler.ScheduleTask(
 			headers.SetIncomingIdempotencyKey(
@@ -138,6 +236,57 @@ func (t *optimizeBaseDisksTask) Run(
 				ZoneId:           request.GetZoneId(),
 				ImageId:          request.GetImageId(),
 				UseBaseDiskAsSrc: true,
+			},
+		)
+		if err != nil {
+			return err
+		}
+
+		_, err = t.scheduler.WaitTask(ctx, execCtx, taskID)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Execute idle cleanup: reduce pool capacities, then retire idle base disks.
+	for _, request := range t.state.PoolReductionRequests {
+		taskID, err := t.scheduler.ScheduleTask(
+			headers.SetIncomingIdempotencyKey(
+				ctx,
+				"idle_configure_pool_"+execCtx.GetTaskID()+
+					":"+request.GetZoneId()+":"+request.GetImageId(),
+			),
+			"pools.ConfigurePool",
+			"",
+			request,
+		)
+		if err != nil {
+			return err
+		}
+
+		_, err = t.scheduler.WaitTask(ctx, execCtx, taskID)
+		if err != nil {
+			if !errors.CanRetry(err) {
+				// Ignore non-retriable (fatal) error, because image might be already
+				// deleted.
+				continue
+			}
+
+			return err
+		}
+	}
+
+	for _, baseDiskID := range t.state.IdleBaseDiskIds {
+		taskID, err := t.scheduler.ScheduleTask(
+			headers.SetIncomingIdempotencyKey(
+				ctx,
+				"retire_idle_base_disk_"+execCtx.GetTaskID()+
+					":"+baseDiskID,
+			),
+			"pools.RetireBaseDisk",
+			"",
+			&protos.RetireBaseDiskRequest{
+				BaseDiskId: baseDiskID,
 			},
 		)
 		if err != nil {
--- a/cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
+++ b/cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
@@ -9,5 +9,8 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 ////////////////////////////////////////////////////////////////////////////////
 
 message OptimizeBaseDisksTaskState {
-    repeated ConfigurePoolRequest ConfigurePoolRequests = 1;
+    repeated ConfigurePoolRequest ConfigurePoolForImageSizeRequests = 1;
+    repeated ConfigurePoolRequest PoolReductionRequests = 2;
+    repeated string IdleBaseDiskIds = 3;
+    bool CollectedResourcesToOptimize = 4;
 }
--- a/cloud/disk_manager/internal/pkg/services/pools/register.go
+++ b/cloud/disk_manager/internal/pkg/services/pools/register.go
@@ -79,6 +79,13 @@ func RegisterForExecution(
 		return err
 	}
 
+	baseDiskIdleTTL, err := time.ParseDuration(
+		config.GetBaseDiskIdleTTL(),
+	)
+	if err != nil {
+		return err
+	}
+
 	err = taskRegistry.RegisterForExecution("pools.AcquireBaseDisk", func() tasks.Task {
 		return &acquireBaseDiskTask{
 			scheduler: taskScheduler,
@@ -189,6 +196,8 @@ func RegisterForExecution(
 			convertToImageSizedBaseDisksThreshold:   config.GetConvertToImageSizedBaseDiskThreshold(),
 			convertToDefaultSizedBaseDisksThreshold: config.GetConvertToDefaultSizedBaseDiskThreshold(),
 			minPoolAge:                              minOptimizedPoolAge,
+			baseDiskIdleTTL:                         baseDiskIdleTTL,
+			cleanupIdleBaseDisksLimit:               config.GetCleanupIdleBaseDisksLimit(),
 		}
 	})
 	if err != nil {
--- a/cloud/disk_manager/internal/pkg/services/pools/storage/common.go
+++ b/cloud/disk_manager/internal/pkg/services/pools/storage/common.go
@@ -86,6 +86,15 @@ func baseDiskStatusToString(status baseDiskStatus) string {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// It checks whether a timestamp is effectively zero.
+// Go's time.Time{} is 0001-01-01, but after a YDB round-trip it becomes
+// time.Unix(0, 0) (1970-01-01), so t.IsZero() alone is insufficient.
+func isTimestampUnset(t time.Time) bool {
+	return !t.After(time.Unix(0, 0))
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
 type baseDisk struct {
 	id                  string
 	imageID             string
@@ -105,6 +114,7 @@ type baseDisk struct {
 	fromPool       bool
 	retiring       bool
 	deletedAt      time.Time
+	idleSince      time.Time // zero if disk has at least one active unit
 	status         baseDiskStatus
 }
 
@@ -128,6 +138,7 @@ func (d *baseDisk) toBaseDisk() BaseDisk {
 		CreateTaskID:        d.createTaskID,
 		Size:                d.size,
 		Units:               d.units,
+		FreeSlots:           d.freeSlots(),
 		Ready:               d.status == baseDiskStatusReady,
 	}
 }
@@ -152,6 +163,13 @@ func (d *baseDisk) applyInvariants() {
 		if !d.isDoomed() && !d.fromPool {
 			d.status = baseDiskStatusDeleting
 		}
+
+		if isTimestampUnset(d.idleSince) {
+			d.idleSince = time.Now()
+		}
+	} else {
+		// Disk is no longer idle.
+		d.idleSince = time.Time{}
 	}
 }
 
@@ -205,6 +223,7 @@ func (d *baseDisk) structValue() persistence.Value {
 		persistence.StructFieldValue("from_pool", persistence.BoolValue(d.fromPool)),
 		persistence.StructFieldValue("retiring", persistence.BoolValue(d.retiring)),
 		persistence.StructFieldValue("deleted_at", persistence.TimestampValue(d.deletedAt)),
+		persistence.StructFieldValue("idle_since", persistence.TimestampValue(d.idleSince)),
 		persistence.StructFieldValue("status", persistence.Int64Value(int64(d.status))),
 	)
 }
@@ -229,6 +248,7 @@ func baseDiskStructTypeString() string {
 		from_pool: Bool,
 		retiring: Bool,
 		deleted_at: Timestamp,
+		idle_since: Timestamp,
 		status: Int64>`
 }
 
@@ -252,6 +272,7 @@ func baseDisksTableDescription() persistence.CreateTableDescription {
 		persistence.WithColumn("from_pool", persistence.Optional(persistence.TypeBool)),
 		persistence.WithColumn("retiring", persistence.Optional(persistence.TypeBool)),
 		persistence.WithColumn("deleted_at", persistence.Optional(persistence.TypeTimestamp)),
+		persistence.WithColumn("idle_since", persistence.Optional(persistence.TypeTimestamp)),
 		persistence.WithColumn("status", persistence.Optional(persistence.TypeInt64)),
 
 		persistence.WithPrimaryKeyColumn("id"),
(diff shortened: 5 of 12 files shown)


agalibin:
@codex review

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -41,57 +44,130 @@ func (t *optimizeBaseDisksTask) Run(
...
-					UseImageSize: newUseImageSize,
-				},
-			)
+			for key, count := range idleCounts {
+				info, ok := poolInfoMap[key]
+				if !ok {
+					continue
+				}
+
+				reduction := count * t.maxActiveSlots
```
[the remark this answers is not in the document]
agalibin: agree, need to use min(disk.Units, s.maxActiveSlots)

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3602,76 @@ func (s *storageYDB) unlockPool(
...
+		declare $idle_before as Timestamp;
+		declare $status as Int64;
+		declare $limit as Uint64;
+
+		select *
+		from base_disks
+		where from_pool = true
+			and active_units = 0
+			and status = $status
+			and last_unused_at > cast(0 as Timestamp)
```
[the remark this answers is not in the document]
agalibin: agree, set the "lastUnusedAt" when disk becomes ready

cloud/disk_manager/internal/pkg/services/pools/register.go
```diff
@@ -237,7 +247,7 @@ func RegisterForExecution(
 		},
 	)
 
-	if config.GetRegularBaseDiskOptimizationEnabled() {
+	if config.GetRegularBaseDiskOptimizationEnabled() || baseDiskIdleTTL > 0 {
```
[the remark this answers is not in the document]
agalibin: send RegularBaseDiskOptimizationEnabled as constructor parameter of the optimizeBaseDisksTask

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3602,76 @@ func (s *storageYDB) unlockPool(
...
+		pragma TablePathPrefix = "%v";
+		declare $idle_before as Timestamp;
+		declare $status as Int64;
+		declare $limit as Uint64;
+
+		select *
+		from base_disks
+		where from_pool = true
+			and active_units = 0
+			and status = $status
```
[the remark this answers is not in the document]
agalibin: not matter now

agalibin:
@codex review

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -150,6 +226,59 @@ func (t *optimizeBaseDisksTask) Run(
...
+		if err != nil {
+			if !errors.CanRetry(err) {
+				continue
+			}
+
+			return err
+		}
+	}
+
+	for _, baseDiskID := range t.state.IdleBaseDiskIds {
```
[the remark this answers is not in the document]
agalibin: case is valid but the probability is extremely low (same case we already have in OptimizeBaseDisks)

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3603,76 @@ func (s *storageYDB) unlockPool(
...
+		--!syntax_v1
+		pragma TablePathPrefix = "%v";
+		declare $idle_before as Timestamp;
+		declare $status as Int64;
+		declare $limit as Uint64;
+
+		select *
+		from base_disks
+		where from_pool = true
+			and active_units = 0
```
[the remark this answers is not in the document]
agalibin: I checked at the DM DB and did not find any base disks by filter "active_slots != 0 AND from_pool = TRUE AND active_units = 0" so I think this comment is not actual.

agalibin:
@codex review

cloud/disk_manager/internal/pkg/services/pools/config/config.proto
```diff
@@ -45,4 +45,9 @@ message PoolsConfig {
     //  * Reducing this number also increases the total size of the base disks
     //    that need to be allocated.
     optional uint32 BaseDiskOverSubscription = 24 [default = 2];
+    // TTL for idle base disks in pools. Base disks with zero overlay disks
+    // for longer than this duration will be ejected from pool and deleted.
```
SvartMetal: > will be ejected from pool and deleted

will be retired (ejected from pool and deleted)

cloud/disk_manager/internal/pkg/services/pools/config/config.proto
```diff
@@ -45,4 +45,9 @@ message PoolsConfig {
...
+    // Set to "0s" to disable.
+    optional string BaseDiskIdleTTL = 25 [default = "0s"];
+    optional uint64 CleanupIdleBaseDisksLimit = 26 [default = 100];
```
SvartMetal, line 53: This needs a comment explaining the semantics

cloud/disk_manager/internal/pkg/services/pools/storage/common.go
```diff
@@ -105,6 +111,7 @@ type baseDisk struct {
 	fromPool       bool
 	retiring       bool
 	deletedAt      time.Time
+	lastUnusedAt   time.Time // zero if disk has at least one active unit
```
SvartMetal: idleSince?

To be more consistent with the other names in this PR

cloud/disk_manager/internal/pkg/services/pools/storage/common.go
```diff
@@ -800,6 +811,8 @@ func acquireUnitsAndSlots(
 	slot.allottedUnits = computeAllottedUnits(slot)
 	disk.activeUnits += slot.allottedUnits
 
+	disk.lastUnusedAt = time.Time{}
```
SvartMetal: Let's add a comment explaining that the disk is no longer idle (unused)

cloud/disk_manager/internal/pkg/services/pools/storage/common.go
```diff
@@ -829,6 +842,8 @@ func acquireTargetUnitsAndSlots(
 	slot.targetAllottedUnits = computeAllottedUnits(slot)
 	disk.activeUnits += slot.targetAllottedUnits
 
+	disk.lastUnusedAt = time.Time{}
```
SvartMetal: Let's add a comment explaining that the disk is no longer idle (unused)

cloud/disk_manager/internal/pkg/services/pools/storage/common_test.go
```diff
@@ -283,3 +283,96 @@ func TestCommonAcquireAndReleaseUnitsAndSlots(t *testing.T) {
...
+	ctx := newContext()
+
+	disk := &baseDisk{
+		maxActiveSlots: 4,
+		units:          6,
+		fromPool:       true,
+		status:         baseDiskStatusReady,
+	}
+
+	require.True(t, disk.lastUnusedAt.IsZero())
```
SvartMetal: Why can't we use `lastUnusedAt.IsZero()` instead of `isTimestampUnset` here?

cloud/disk_manager/internal/pkg/services/pools/storage/common.go
```diff
@@ -829,6 +842,8 @@ func acquireTargetUnitsAndSlots(
...
+
```
SvartMetal: applyInvariants looks like a better place for that

https://github.com/ydb-platform/nbs/blob/d54dcf5289f6d787cde2d60150b5049511d19b5f/cloud/disk_manager/internal/pkg/services/pools/storage/common.go#L143

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3605,76 @@ func (s *storageYDB) unlockPool(
...
+
+	res, err := session.ExecuteRO(ctx, fmt.Sprintf(`
+		--!syntax_v1
+		pragma TablePathPrefix = "%v";
+		declare $idle_before as Timestamp;
+		declare $status as Int64;
+		declare $limit as Uint64;
+
+		select *
+		from base_disks
```
SvartMetal: We do not need to scan every base disk

We can scan only the "free" table because it contains "ready" base disks with at least one free slot

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3605,76 @@ func (s *storageYDB) unlockPool(
...
+		where from_pool = true
+			and active_units = 0
+			and status = $status
+			and last_unused_at > cast(0 as Timestamp)
```
SvartMetal: `last_unused_at != 0`?

SvartMetal: And do we need to check for NULL?

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3605,76 @@ func (s *storageYDB) unlockPool(
...
+		pragma TablePathPrefix = "%v";
+		declare $now as Timestamp;
+		declare $status as Int64;
+
+		update base_disks
+		set last_unused_at = $now
+		where from_pool = true
+			and active_units = 0
+			and status = $status
+			and (last_unused_at = cast(0 as Timestamp) OR last_unused_at IS NULL)
```
SvartMetal: Let's keep this condition consistent with the one on line 3630

cloud/disk_manager/internal/pkg/services/pools/register.go
```diff
@@ -275,5 +286,12 @@ func RegisterForExecution(
 		},
 	)
 
+	if baseDiskIdleTTL > 0 {
+		err = storage.InitializeIdleTimestamps(ctx)
+		if err != nil {
+			return err
+		}
+	}
```
SvartMetal: We do not want this to run when the service starts

Let's add a private command for it instead, or find another place in the code

cloud/disk_manager/internal/pkg/services/pools/register.go
```diff
@@ -237,7 +248,7 @@ func RegisterForExecution(
 		},
 	)
 
-	if config.GetRegularBaseDiskOptimizationEnabled() {
+	if config.GetRegularBaseDiskOptimizationEnabled() || baseDiskIdleTTL > 0 {
```
SvartMetal: I don't think this is needed because the "idle TTL" feature is part of the base disk optimization

cloud/disk_manager/internal/pkg/services/pools/register.go
```diff
@@ -186,9 +193,13 @@ func RegisterForExecution(
 		return &optimizeBaseDisksTask{
 			scheduler:                               taskScheduler,
 			storage:                                 storage,
+			optimizationEnabled:                     config.GetRegularBaseDiskOptimizationEnabled(),
```
SvartMetal: Why do we need this?

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -41,57 +45,130 @@ func (t *optimizeBaseDisksTask) Run(
...
+			than convertToImageSizedBaseDisksThreshold`)
+	}
+
+	idleCleanupEnabled := t.baseDiskIdleTTL > 0
+
+	if !optimizationEnabled && !idleCleanupEnabled {
 		return nil
 	}
 
-	if len(t.state.ConfigurePoolRequests) == 0 {
```
SvartMetal, line 51: Why do we need to change this branch?

If the "optimize base disks" task runs "retire base disks" for an image, it will retire all base disks, including idle ones, so there is no need to run idle base disk cleanup

We can visit every pool not processed by "configure pool" or "retire base disks", collect the idle base disks, and clean them up separately

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
```diff
@@ -8,6 +8,13 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 
 ////////////////////////////////////////////////////////////////////////////////
 
+message IdlePoolCleanupRequest {
+    ConfigurePoolRequest ConfigurePoolRequest = 1;
+    repeated string BaseDiskIds = 2;
+}
```
SvartMetal: I don't think this is needed. You can use the existing `RetireBaseDisk` task to clean up each idle base disk

SvartMetal (changes requested):

cloud/disk_manager/internal/pkg/services/pools/storage/common_test.go
agalibin: We already use IsZero here. If your question about "common.go" file then there is a difference between how go-code and YDB-code interpret zero time.
When we write Go's zero time.Time{} to YDB via TimestampValue(), it becomes 0 microseconds. When read back, TimestampToTime(0) returns time.Unix(0, 0) (1970-01-01), not Go's zero time (0001-01-01). So after a round-trip through YDB, IsZero() returns false while isTimestampUnset correctly handles both representations. In tests where values don't pass through YDB, IsZero() works fine.

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
agalibin: I can add IS NOT NULL but it is not needed in general.

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
agalibin: These conditions are opposite: 'getIdleBaseDisks' selects disks with a set 'idle_since', while 'initializeIdleTimestamps' selects disks without a 'idle_since'. Making them "consistent" would break one of the two queries.

cloud/disk_manager/internal/pkg/services/pools/register.go
agalibin: I'll remove it from RegisterForExecution and add 'todo' in my list for future.

cloud/disk_manager/internal/pkg/services/pools/register.go
agalibin: It was needed to separate enabling optimize and cleanup features, no needed now.

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
agalibin: Previously, the function would early-return if optimization was disabled. We added idle cleanup and we need the function to continue when optimization is off — so we wrapped the optimization logic in an 'if optimizationEnabled' block. The early return now only happens when both optimization and idle cleanup are disabled 'if !optimizationEnabled && !idleCleanupEnabled'. The optimization logic itself is unchanged, it's just guarded by a condition.

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
agalibin: We need the 'IdlePoolCleanupRequest' to couple 'ConfigurePool' (capacity reduction) with the corresponding disk retirements. Without reducing 'config.capacity' first, 'RetireBaseDisk' alone would decrease 'pool.size', but the scheduler would see 'pool.size' < 'config.capacity' and recreate the disk — defeating the purpose of idle cleanup. The grouping also ensures crash-safety: if we restart mid-execution, we know which disks belong to which pool and can skip retires for pools whose ConfigurePool failed.

cloud/disk_manager/internal/pkg/services/pools/storage/common_test.go
SvartMetal: I mean it seems clearer to use the same approach for detecting a zero or unset timestamp. `IsZero()` also seems much clearer than the more exotic `!t.After(time.Unix(0, 0))`

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
SvartMetal: By "consistent", I mean they should use the same approach. If one uses `NULL`, the other should use it as well, and vice versa

The conditions may still differ, for example, one may negate the other, while remaining consistent

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
SvartMetal: I don't think these actions need to be coupled in a single transaction. You can run `ConfigurePool` and retire idle base disks as separate actions

cloud/disk_manager/internal/pkg/services/pools/storage/common.go
```diff
@@ -152,6 +159,9 @@ func (d *baseDisk) applyInvariants() {
 		if !d.isDoomed() && !d.fromPool {
 			d.status = baseDiskStatusDeleting
 		}
+	} else if d.activeUnits > 0 {
```
SvartMetal: Do we really need this branch? We know that `activeUnits` is a `ui64` and is guaranteed to be nonzero

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
SvartMetal: I don't think we should run this task when optimization is disabled, and optimization is not disabled in production

Therefore, the newly introduced `optimizationEnabled` variable is not needed

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -41,57 +44,130 @@ func (t *optimizeBaseDisksTask) Run(
...
+				)
+
+				optimizedPools[poolKey{poolInfo.ImageID, poolInfo.ZoneID}] = true
 			}
+		}
 
-			useImageSize := poolInfo.ImageSize > 0
-			newUseImageSize := useImageSize
+		if idleCleanupEnabled {
+			idleDisks, err := t.storage.GetIdleBaseDisks(
```
SvartMetal, line 101: I think we should use a separate query for each pool, identified by image ID and zone ID. A single transaction for all pools might be too heavy

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -150,6 +226,61 @@ func (t *optimizeBaseDisksTask) Run(
...
+			"",
+			request,
+		)
+		if err != nil {
+			return err
+		}
+
+		_, err = t.scheduler.WaitTask(ctx, execCtx, taskID)
+		if err != nil {
+			if !errors.CanRetry(err) {
```
SvartMetal, line 269: This needs the same comment as above

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -150,6 +226,61 @@ func (t *optimizeBaseDisksTask) Run(
...
+					BaseDiskId: baseDiskID,
+				},
+			)
+			if err != nil {
+				return err
+			}
+
+			_, err = t.scheduler.WaitTask(ctx, execCtx, taskID)
+			if err != nil {
+				if !errors.CanRetry(err) {
```
SvartMetal: I don't think this is needed because the `RetireBaseDisk` task should be idempotent with respect to base disk deletion and complete successfully if the base disk has been deleted

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -1843,6 +1843,9 @@ func (s *storageYDB) baseDiskCreated(
 
 	updated := found
 	updated.status = baseDiskStatusReady
+	if found.activeUnits == 0 {
+		updated.idleSince = time.Now()
+	}
```
SvartMetal: Is this really needed?

`applyInvariants` should be called for every base disk update

cloud/disk_manager/internal/pkg/services/pools/storage/common.go
```diff
@@ -939,6 +957,10 @@ func releaseTargetUnitsAndSlots(
 	}
 	disk.activeUnits -= slot.targetAllottedUnits
 
+	if disk.activeUnits == 0 && isTimestampUnset(disk.idleSince) {
+		disk.idleSince = time.Now()
+	}
```
SvartMetal: I don't think this is needed. The code in `baseDisk.applyInvariants` should be enough

cloud/disk_manager/internal/pkg/services/pools/storage/common_test.go
agalibin: I have already explained why we cannot use IsZero everywhere.

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
agalibin: Why should «CleanupTask» depend on «convertToDefaultSizedBaseDisksThreshold» and «convertToImageSizedBaseDisksThreshold»? These are parameters of 'optimisation' not a 'cleanup'.

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
agalibin: moved the idle timestamp logic into applyInvariants and removed it from baseDiskCreated, releaseUnitsAndSlots, and releaseTargetUnitsAndSlots.

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
agalibin: They are not in a single transaction — ConfigurePool and each RetireBaseDisk are separate scheduled tasks. The proto grouping only ensures that on crash-recovery we know which disks belong to which pool, so we can skip retires for pools whose ConfigurePool failed with a non-retriable error. Without this association, a crash between the two phases would leave us retiring disks whose capacity was never reduced.

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
agalibin: renamed variable as discussed

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -41,57 +44,124 @@ func (t *optimizeBaseDisksTask) Run(
...
+					t.state.ConfigurePoolRequests,
+					&protos.ConfigurePoolRequest{
+						ZoneId:       poolInfo.ZoneID,
+						ImageId:      poolInfo.ImageID,
+						Capacity:     poolInfo.Capacity,
+						UseImageSize: newUseImageSize,
+					},
+				)
+
+				optimizedPools[poolKey{poolInfo.ImageID, poolInfo.ZoneID}] = true
```
SvartMetal: Let's add a comment explaining that we optimize only one thing at a time for a given pool

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
```diff
@@ -10,4 +10,7 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 
 message OptimizeBaseDisksTaskState {
     repeated ConfigurePoolRequest ConfigurePoolRequests = 1;
+    repeated ConfigurePoolRequest IdleCleanupRequests = 2;
```
SvartMetal: PoolReductionRequests

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
```diff
@@ -10,4 +10,7 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 
 message OptimizeBaseDisksTaskState {
     repeated ConfigurePoolRequest ConfigurePoolRequests = 1;
```
SvartMetal: ConfigurePoolForImageSizeRequests

cloud/disk_manager/internal/pkg/services/pools/protos/optimize_base_disks_task.proto
```diff
@@ -10,4 +10,7 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
...
+    repeated ConfigurePoolRequest IdleCleanupRequests = 2;
+    repeated string IdleBaseDiskIds = 3;
+    bool Collected = 4;
```
SvartMetal: CollectedResourcesToOptimize

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -41,57 +44,124 @@ func (t *optimizeBaseDisksTask) Run(
...
+					},
+				)
+			}
 		}
 
+		t.state.Collected = true
 		err = execCtx.SaveState(ctx)
 		if err != nil {
 			return err
 		}
```
SvartMetal: The function is becoming too complex. Let's extract an `ensureResourcesToOptimizeCollected` method

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -150,6 +220,63 @@ func (t *optimizeBaseDisksTask) Run(
...
 	}
 
+	// Execute idle cleanup: reduce pool capacities, then retire idle disks.
```
SvartMetal: retire idle base disks

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -150,6 +220,63 @@ func (t *optimizeBaseDisksTask) Run(
...
+		if err != nil {
+			return err
+		}
+
+		_, err = t.scheduler.WaitTask(ctx, execCtx, taskID)
+		if err != nil {
+			if !errors.CanRetry(err) {
+				// Ignore non-retriable (fatal) error, because image might be already
+				// deleted.
+				continue
```
SvartMetal: I mentioned earlier that this is not needed because `RetireBaseDisk` should be idempotent with respect to base disk deletion

agalibin: oh, I reverted my previous changes and forgot about this, sorry.

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -3602,3 +3602,88 @@ func (s *storageYDB) unlockPool(
...
+
+	var result []BaseDisk
+	for _, d := range baseDisks {
+		result = append(result, d.toBaseDisk())
+	}
+
+	return result, nil
+}
+
+func (s *storageYDB) initializeIdleTimestamps(
```
SvartMetal, line 3665: Do we use this?

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -22,39 +22,40 @@ type optimizeBaseDisksTask struct {
 	convertToImageSizedBaseDisksThreshold   uint64
 	convertToDefaultSizedBaseDisksThreshold uint64
 	minPoolAge                              time.Duration
+	baseDiskIdleTTL                         time.Duration
+	cleanupIdleBaseDisksLimit               uint64
+	maxActiveSlots                          uint32
```
SvartMetal: This constant should not be passed to the task during construction because it may differ for each base disk

I think we should expose `FreeSlots` on `BaseDisk` https://github.com/ydb-platform/nbs/blob/1c16d8b469ce9c0215ed74ffef51fad251ce880a/cloud/disk_manager/internal/pkg/services/pools/storage/storage.go#L13

FreeSlots computed here https://github.com/ydb-platform/nbs/blob/1c16d8b469ce9c0215ed74ffef51fad251ce880a/cloud/disk_manager/internal/pkg/services/pools/storage/common.go#L158

FreeSlots correspond to pool size here https://github.com/ydb-platform/nbs/blob/1c16d8b469ce9c0215ed74ffef51fad251ce880a/cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go#L2150

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -75,26 +76,117 @@ func (t *optimizeBaseDisksTask) Run(
...
 		}
 	}
 
+	t.state.CollectedResourcesToOptimize = true
+	err = execCtx.SaveState(ctx)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
SvartMetal: `return execCtx.SaveState(ctx)`

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -75,26 +76,117 @@ func (t *optimizeBaseDisksTask) Run(
...
+			than convertToImageSizedBaseDisksThreshold`)
+	}
+
+	idleCleanupEnabled := t.baseDiskIdleTTL > 0
+
+	if !sizeOptimizationEnabled && !idleCleanupEnabled {
+		return nil
+	}
+
+	err := t.ensureResourcesToOptimizeCollected(ctx, execCtx, sizeOptimizationEnabled, idleCleanupEnabled)
```
SvartMetal: 80 symbols

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -75,26 +76,117 @@ func (t *optimizeBaseDisksTask) Run(
...
+			var reduction uint32
+			for _, disk := range idleDisks {
+				reduction += min(uint32(disk.Units), t.maxActiveSlots)
+				t.state.IdleBaseDiskIds = append(
+					t.state.IdleBaseDiskIds,
+					disk.ID,
+				)
+			}
+
+			newCapacity := uint32(0)
```
SvartMetal, line 132: Could this code cause a race condition? https://github.com/ydb-platform/nbs/blob/1c16d8b469ce9c0215ed74ffef51fad251ce880a/cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go#L1024

Is it possible that, in some cases, we overwrite a capacity of `1` with `0`, causing overlay disk creation to hang?

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
agalibin: Not now, but I want to use it later

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
agalibin: I could not find the way to do so.

SvartMetal: Discussed that we have another hang (during overlay disk rebase)

agalibin: created issue https://github.com/ydb-platform/nbs/issues/6684

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb.go
```diff
@@ -476,6 +476,36 @@ func (s *storageYDB) CheckConsistency(ctx context.Context) error {
...
+	limit uint64,
+) ([]BaseDisk, error) {
+
+	var baseDisks []BaseDisk
+
+	err := s.db.Execute(
+		ctx,
+		func(ctx context.Context, session *persistence.Session) error {
+			var err error
+			baseDisks, err = s.getIdleBaseDisks(ctx, imageID, zoneID, session, idleDuration, limit)
```
SvartMetal: minor: 80 symbols

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_test.go
```diff
@@ -1906,7 +1932,7 @@ func TestStorageYDBRetireBaseDisks(t *testing.T) {
 
 		for i := 0; i < len(toSchedule); i++ {
 			toSchedule[i].CreateTaskID = "create"
-			require.NotContains(t, oldBaseDisks, toSchedule[i])
+			require.NotContains(t, normalizeBaseDisks(oldBaseDisks), normalizeBaseDisk(toSchedule[i]))
```
SvartMetal: 80 symbols

and other places

SvartMetal (dismissed):

SvartMetal (approved):

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
```diff
@@ -75,26 +75,124 @@ func (t *optimizeBaseDisksTask) Run(
...
 				&protos.ConfigurePoolRequest{
 					ZoneId:       poolInfo.ZoneID,
 					ImageId:      poolInfo.ImageID,
 					Capacity:     poolInfo.Capacity,
 					UseImageSize: newUseImageSize,
 				},
 			)
+
+			// Pools selected for size optimization will have all their base disks retired,
+			// including idle ones, so there is no need to run idle cleanup for them.
```
gy2411, line 89: nit: It is not obvious how this line related to idle disks. Maybe move this comment to line 97?

gy2411 (approved):

cloud/disk_manager/internal/pkg/services/pools/optimize_base_disks_task.go
agalibin: I decided to write it here to explain why it need to set "true" in optimizedPools array. If it is not convenient we could move it in the next PR.

Issue #3781, opened 2025-06-26
TFileRingBuffer: make more stable in the case of file corruption


e673:
Currently, only CRC validation is performed.
There is lack of checks against corruption of headers.
For example:
* `SkipSlackSpace` may read from the wrong place if `Header()->ReadPos` is invalid.
It may also happen even without corruption when the slack space is less than the entry header size.
* After visiting the entries, `Header()->WritePos` is not checked for equality to the expected position of a new entry.
* There is no way to know from the outside that the headers were corrupted

PR #6654, opened 2026-07-30
issue-3781: TFileRingBuffer: introduce TFileRingBufferAccessor

e673:
### Notes
Extracted from https://github.com/ydb-platform/nbs/pull/6563

Introduce `TFileRingBufferAccessor` to access memory mapped file from `TFileRingBuffer`:
- Extract common code that will be used in https://github.com/ydb-platform/nbs/pull/6563;
- Strengthen data validation and give verbose messages;
- Replace `Y_ABORT_UNLESS` and throwing exceptions with `NProto::TError`.

Also:
- Fixed the test `TFileRingBufferTest::ShouldMigrate` when migrating to V6 (test was passed because validation was not complete).

### Issue
https://github.com/ydb-platform/nbs/issues/3781

### DIFF
--- a/cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
+++ b/cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
@@ -1,4 +1,5 @@
 #include "file_ring_buffer.h"
+#include "file_ring_buffer_accessor.h"
 #include "file_ring_buffer_format.h"
 
 #include <cloud/storage/core/libs/diagnostics/critical_events.h>
@@ -127,10 +128,7 @@ class TFileRingBuffer::TImpl
 {
 private:
     const TFileRingBufferArgs Args;
-    TFileMap Map;
-
-    std::unique_ptr<IFileRingBufferDataProcessor> Data;
-    TFileRingBufferCapabilities Capabilities;
+    TFileMapFileRingBufferAccessor Accessor;
     bool Corrupted = false;
 
     TEntryInfo CurrentAllocation = TEntryInfo::CreateInvalid();
@@ -142,12 +140,27 @@ private:
 private:
     THeader* Header()
     {
-        return reinterpret_cast<THeader*>(Map.Ptr());
+        return Accessor.GetHeader();
     }
 
     const THeader* Header() const
     {
-        return reinterpret_cast<THeader*>(Map.Ptr());
+        return Accessor.GetHeader();
+    }
+
+    IFileRingBufferDataProcessor* Data()
+    {
+        return Accessor.GetDataProcessor();
+    }
+
+    const IFileRingBufferDataProcessor* Data() const
+    {
+        return Accessor.GetDataProcessor();
+    }
+
+    const TFileRingBufferCapabilities& Capabilities() const
+    {
+        return Accessor.GetCapabilities();
     }
 
     TEntryInfo GetEntry(ui64 pos) const
@@ -162,15 +175,15 @@ private:
                 return TEntryInfo::CreateInvalid();
             }
 
-            const auto eh = Data->ReadEntryHeader(pos);
+            const auto eh = Data()->ReadEntryHeader(pos);
             if (eh.DataSize != 0) {
-                const auto* data = Data->GetEntryDataPtr(pos, eh.DataSize);
+                const auto* data = Data()->GetEntryDataPtr(pos, eh.DataSize);
                 if (data == nullptr) {
                     return TEntryInfo::CreateInvalid();
                 }
 
                 if (eh.FreeFlag && eh.DataChecksum != 0 &&
-                    Capabilities.EntryHeaderIsCoveredByChecksum)
+                    Capabilities().EntryHeaderIsCoveredByChecksum)
                 {
                     return TEntryInfo::CreateInvalid();
                 }
@@ -192,22 +205,22 @@ private:
             return TEntryInfo::CreateInvalid();
         }
 
-        const auto eh = Data->ReadEntryHeader(pos);
+        const auto eh = Data()->ReadEntryHeader(pos);
         if (eh.DataSize == 0) {
             return TEntryInfo::CreateInvalid();
         }
 
-        const auto* data = Data->GetEntryDataPtr(pos, eh.DataSize);
+        const auto* data = Data()->GetEntryDataPtr(pos, eh.DataSize);
         if (data == nullptr) {
             return TEntryInfo::CreateInvalid();
         }
 
-        if (pos + Data->GetEntrySize(eh.DataSize) > Header()->WritePos) {
+        if (pos + Data()->GetEntrySize(eh.DataSize) > Header()->WritePos) {
             return TEntryInfo::CreateInvalid();
         }
 
         if (eh.FreeFlag && eh.DataChecksum != 0 &&
-            Capabilities.EntryHeaderIsCoveredByChecksum)
+            Capabilities().EntryHeaderIsCoveredByChecksum)
         {
             return TEntryInfo::CreateInvalid();
         }
@@ -223,111 +236,29 @@ private:
     TEntryInfo GetNextEntry(const TEntryInfo& e) const
     {
         return e.HasValue()
-            ? GetEntry(e.ActualPos + Data->GetEntrySize(e.Header.DataSize))
+            ? GetEntry(e.ActualPos + Data()->GetEntrySize(e.Header.DataSize))
             : TEntryInfo::CreateInvalid();
-    }
-
-    void CreateDataProcessor(EVersion version)
-    {
-        Data = CreateFileRingBufferDataProcessor(
-            version,
-            GetMappedData(Header()->DataOffset, Header()->DataCapacity));
-
-        Capabilities = Data->GetCapabilities(true);
-    }
-
-    void ValidateStructure()
-    {
-        const ui64 mapLength = static_cast<ui64>(Map.Length());
-        const auto& h = *Header();
-
-        Y_ABORT_UNLESS(sizeof(THeader) == h.HeaderSize);
-        Y_ABORT_UNLESS(h.HeaderSize <= h.MetadataOffset);
-        Y_ABORT_UNLESS(h.MetadataOffset <= h.DataOffset);
-        Y_ABORT_UNLESS(h.MetadataCapacity <= h.DataOffset - h.MetadataOffset);
-        Y_ABORT_UNLESS(h.DataOffset <= mapLength);
-        Y_ABORT_UNLESS(h.DataCapacity <= mapLength - h.DataOffset);
-    }
-
-    void ValidateDataStructure()
-    {
-        TEntryInfo cur = GetFrontEntry();
-
-        if (cur.IsInvalid()) {
-            SetCorrupted();
-            return;
-        }
-
-        if (cur.ActualPos != Header()->ReadPos) {
-            if (cur.ActualPos == 0 && Header()->WritePos == 0) {
-                // Valid situation when Alloc was interrupted for empty buffer
-                Header()->ReadPos = 0;
-            } else {
-                SetCorrupted();
-                return;
-            }
-        }
-
-        while (cur.HasValue()) {
-            cur = GetNextEntry(cur);
-        }
-
-        if (cur.IsInvalid() || cur.ActualPos != Header()->WritePos) {
-            SetCorrupted();
-        }
-    }
-
-    void ValidateMetadata()
-    {
-        auto data =
-            GetMappedData(Header()->MetadataOffset, Header()->MetadataCapacity);
-
-        if (Header()->MetadataSize > data.size() ||
-            Crc32c(data.data(), Header()->MetadataSize) !=
-                Header()->MetadataChecksum)
-        {
-            SetCorrupted();
-        }
-    }
-
-    void ValidateEntriesChecksums()
-    {
-        Visit(
-            [&](ui32 checksum, ui32 tag, TStringBuf entry)
-            {
-                Y_UNUSED(tag);
-                const ui32 actualChecksum = Crc32c(entry.data(), entry.size());
-                if (actualChecksum != checksum) {
-                    SetCorrupted();
-                }
-            });
-    }
-
-    void ResizeAndRemap(ui64 fileSize)
-    {
-        Map.ResizeAndRemap(0, fileSize);
-        // File can be mapped to a different address - we should invalidate
-        // the data processor working with the old address range
-        Data.reset();
-    }
-
-    std::span<char> GetMappedData(ui64 offset, ui64 size) const
-    {
-        const ui64 mapLength = static_cast<ui64>(Map.Length());
-        Y_ABORT_UNLESS(offset <= mapLength);
-        Y_ABORT_UNLESS(size <= mapLength - offset);
-        return {static_cast<char*>(Map.Ptr()) + offset, size};
     }
 
     void CopyMappedData(ui64 destPos, ui64 srcPos, ui64 size)
     {
-        auto src = GetMappedData(srcPos, size);
-        auto dst = GetMappedData(destPos, size);
+        auto src = Accessor.GetRawData(srcPos, size);
+        auto dst = Accessor.GetRawData(destPos, size);
 
         // Copied data regions cannot overlap
         Y_ABORT_UNLESS(destPos + size <= srcPos || srcPos + size <= destPos);
 
         MemCopy(dst.data(), src.data(), size);
+    }
+
+    bool ResizeAndRemap(size_t newSize)
+    {
+        auto status = Accessor.ResizeAndRemap(newSize);
+        if (HasError(status)) {
+            SetCorrupted(FormatError(status));
+            return false;
+        }
+        return true;
     }
 
     bool IsMigrationNeeded() const
@@ -343,13 +274,13 @@ private:
 
         // Migration to any version can be performed when the buffer is empty
         if (Empty()) {
+            SetReadAndWritePosToZeroForEmptyBuffer();
             Header()->Version = Args.Version;
-            CreateDataProcessor(Args.Version);
-            return;
+            Validate();
         }
     }
 
-    void ResizeMetadata(ui64 desiredMetadataCapacity)
+    bool ResizeMetadata(ui64 desiredMetadataCapacity)
     {
         Header()->MetadataCapacity =
             Min(Header()->MetadataCapacity,
@@ -371,7 +302,9 @@ private:
             const ui64 tempDataOffset =
                 Max(newFileSize, Header()->DataOffset + Header()->DataCapacity);
 
-            ResizeAndRemap(tempDataOffset + Header()->DataCapacity);
+            if (!ResizeAndRemap(tempDataOffset) || !Validate()) {
+                return false;
+            }
 
             CopyMappedData(
                 tempDataOffset,
@@ -391,9 +324,13 @@ private:
             Header()->DataOffset = newDataOffset;
         }
 
-        ResizeAndRemap(newFileSize);
+        if (!ResizeAndRemap(newFileSize) || !Validate()) {
+            return false;
+        }
 
         Header()->MetadataCapacity = newMetadataCapacity;
+
+        return Validate();
     }
 
     void VisitEntries(auto&& visitor)
@@ -406,7 +343,7 @@ private:
         }
 
         if (e.IsInvalid()) {
-            SetCorrupted();
+            SetCorrupted("Invalid entry detected at VisitEntries");
         }
     }
 
@@ -418,7 +355,7 @@ private:
         }
 
         if (front.IsInvalid()) {
-            SetCorrupted();
+            SetCorrupted("Invalid front entry");
         } else {
             Header()->ReadPos = front.ActualPos;
         }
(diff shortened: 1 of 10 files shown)


cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -142,12 +140,27 @@ class TFileRingBuffer::TImpl
 private:
     THeader* Header()
     {
-        return reinterpret_cast<THeader*>(Map.Ptr());
+        return Accessor.GetHeader();
     }
```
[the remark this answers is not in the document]
e673: fixed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -438,38 +424,59 @@ class TFileRingBuffer::TImpl
...
-        ValidateStructure();
+            if (!Accessor.IsInitialized()) {
+                SetCorrupted("Failed to initialize after creating header");
+                return;
+            }
+        }
 
         if (Header()->MetadataCapacity != Args.MetadataCapacity) {
             ResizeMetadata(Args.MetadataCapacity);
         }
```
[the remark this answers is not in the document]
e673: fixed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -835,13 +834,14 @@ class TFileRingBuffer::TImpl
...
+        return Capabilities().MaxAllocationByteCount;
     }
 
     bool ValidateMetadata() const
     {
-        auto data =
-            GetMappedData(Header()->MetadataOffset, Header()->MetadataCapacity);
+        auto data = Accessor.GetRawData().subspan(
+            Header()->MetadataOffset,
+            Header()->MetadataCapacity);
```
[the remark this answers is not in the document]
e673: fixed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_accessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+            if (eh.DataChecksum != 0) {
+                return MakeError(Sprintf(
+                    "Invalid file ring buffer entry header at position %lu "
+                    "(free flag is set and data checksum is non-zero)",
+                    pos));
+            }
+        } else {
+            auto actualCrc = Crc32c(payload, eh.DataSize);
+            if (actualCrc != eh.DataChecksum) {
+                return MakeError(Sprintf(
```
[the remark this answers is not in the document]
e673: I want to enforce checksum validation on startup.
The purpose of `Config.EnableChecksumValidation` should be validating the checksums at flush.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -438,38 +424,59 @@ class TFileRingBuffer::TImpl
...
-        Y_ABORT_UNLESS(
-            IsSupportedFileRingBufferVersion(
-                static_cast<EVersion>(Header()->Version)),
-            "Unsupported current FileRingBuffer version - %u, file: %s",
-            static_cast<ui32>(Header()->Version),
-            args.FilePath.c_str());
+        auto validateResult = Accessor.ValidateAndInitialize();
+        if (HasError(validateResult)) {
+            SetCorrupted(FormatError(validateResult));
+            return;
```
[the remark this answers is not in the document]
e673: fixed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -328,14 +279,18 @@ class TFileRingBuffer::TImpl
         // Transition V4 -> V5 can be made without emptying the buffer.
         if (Header()->Version == EVersion::V4 && Args.Version > EVersion::V4) {
             // Parse entry headers using newer version
-            CreateDataProcessor(EVersion::V5);
+            Header()->Version = EVersion::V5;
```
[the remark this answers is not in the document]
e673: Not needed. V4 is no more used and this code may be safely deleted (in a separate PR).

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_accessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+        if (eh.Tag != 0) {
+            return MakeError(Sprintf(
+                "Invalid file ring buffer entry header at position %lu "
+                "(data size is zero and tag is non-zero)",
+                pos));
+        }
+
+        if (eh.DataChecksum != 0) {
+            return MakeError(Sprintf(
```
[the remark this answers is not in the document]
e673: Not needed. WriteBackCache already uses V6 everywhere. HandleOpsQueue will switch to V6 soon.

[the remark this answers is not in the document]
e673: added

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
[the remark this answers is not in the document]
e673: V4 is removed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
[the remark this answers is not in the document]
e673: Implemented differently

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -345,7 +302,9 @@ class TFileRingBuffer::TImpl
             const ui64 tempDataOffset =
                 Max(newFileSize, Header()->DataOffset + Header()->DataCapacity);
 
-            ResizeAndRemap(tempDataOffset + Header()->DataCapacity);
+            if (!ResizeAndRemap(tempDataOffset) || !Validate()) {
```
neihar: how does it work in new code? will below CopyMappedData be safe?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_accessor.cpp
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    if (rawDataSize < header.DataOffset) {
+        return MakeError(Sprintf(
+            "Invalid file ring buffer data offset %lu (expected <= %lu)",
+            header.DataOffset,
+            rawDataSize));
+    }
+
+    if (header.DataOffset % sizeof(ui64) != 0) {
```
neihar, line 80: will it work with tempDataOffset?  can we check that with test?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_accessor.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+void TFileRingBufferAccessor::UpdateRawData(std::span<char> rawData)
+{
+    RawData = rawData;
+    ResetValidationState();
+}
+
+std::span<char> TFileRingBufferAccessor::GetRawData(ui64 offset, ui64 byteCount)
+{
+    return RawData.subspan(offset, byteCount);
```
neihar, line 325: I liked abort on wrong params here more, than without it

e673: Previously, we worked with raw pointers so the check were necessary.
Now the same validation is performed inside `subspan` so it would be just a duplication of the same functional.
I can return the checks back if needed.

neihar: does this validation inside subspan exists in release build?

neihar: is see that it is UB https://en.cppreference.com/cpp/container/span/subspan
I am not a big fun of Aborts, but just curious why did you remove it

e673: Yes, indeed.
Fixed the code.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_accessor.cpp
e673: I've changed `len` in `ShouldResizeMetadata` test to a non-round value `19`.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: No, it didn't work because I make a mistake in a file size - and tests passed because of UB in subspan.
Fixed now.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
neihar, line 325: is tempDataOffset here aligned and will pass alignment check if we crash midway?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
neihar: should we check f2 here? looks like this covers problem

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: I've added alignment here

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
e673: Fixed

neihar (dismissed):

e673:
AI review

Reviewed at head f18102a9 (Introduce TFileRingBufferAccessor, extracted from #6563). The refactoring itself is sound — the accessor cleanly owns mapping + validation, and the ResizeMetadata AlignUp fix and the migration position reset are both correct and necessary for V6. Findings below, roughly by severity.

**1. Missing compiler fence before the Version store in TryMigrate — crash during migration can permanently brick the buffer**

`file_ring_buffer.cpp:277-279`:

```cpp
SetReadAndWritePosToZeroForEmptyBuffer();
Header()->Version = Args.Version;
Validate();
```

`SetReadAndWritePosToZeroForEmptyBuffer` (lines 376-390) ends with the `ReadPos = 0` store and no trailing fence. Nothing prevents the compiler from sinking `ReadPos = 0` below `Header()->Version = Args.Version` — both are plain stores to distinct members, and the last `atomic_signal_fence` is between the WritePos and ReadPos stores. The reachable in-memory state is then `WritePos = 0, Version = V6, ReadPos = <old unaligned V5 position>`. If the page is flushed in that window and the process dies, restart validation takes the `writePos == 0` branch in `ValidateData` (accessor.cpp:192) and fails the V6 ReadPos alignment check — status Failed, buffer marked corrupted forever. This is exactly the class of reordering the existing fences guard against; add a `std::atomic_signal_fence(std::memory_order_seq_cst)` between the position reset and the Version store (or give `SetReadAndWritePosToZeroForEmptyBuffer` a trailing fence, which also hardens the call site at line 541).

> Added a barrier

**2. Every ValidateAndInitialize is now a full CRC pass over all live data, and the constructor can do it up to 4 times**

`ValidateData` walks every entry and computes crc32c over every payload, plus the metadata CRC. That runs: once in the constructor, then in `ResizeMetadata` at line 308, line 330, and line 336 (`return Validate()`) whenever `MetadataCapacity` differs from the requested value — i.e., every startup after a config change does up to 4 full passes over up to DataCapacity bytes. The two trailing calls at 330/336 are back-to-back with only `MetadataCapacity = newMetadataCapacity` between them; the last one exists only to rebuild `RawMetadata`. Consider a lightweight "re-derive spans/processor from an already-validated header" path so ResizeMetadata doesn't repeatedly re-checksum the whole buffer. Not correctness, but for a multi-GB write-back cache file this is real mount latency.

> Metdata resize is not a hot path. This is acceptable.

**3. Behavior change: entry checksum validation is now unconditional and any single bad entry bricks the whole buffer**

Previously `EnableChecksumValidation` defaulted to false, so production WBC never verified entry CRCs at startup (only `ValidateMetadata` was unconditional). Now one flipped bit in any entry payload marks the entire buffer corrupted at construction, and since this is a write-back cache holding dirty unflushed data, that converts localized corruption into total loss of all pending writes. That may well be the intended trade (serving a silently corrupted entry is arguably worse), but it deserves an explicit call-out in the PR description, and it's worth considering a salvage mode later: entries preceding the first CRC failure are verifiably intact, so truncating at the first bad entry would preserve them. The Debug validation mode exists but nothing wires it up for recovery yet. Also confirm nothing else in the tree still references `EnableChecksumValidation` or plumbs a filestore config flag into it — if there was a config knob feeding this field, it's now dead config (I couldn't complete a repo-wide search, GitHub API rate limit).

> Unconiditional checksum validation is expected and mentioned here: https://github.com/ydb-platform/nbs/pull/6665

**4. SetMetadata crash window now unconditionally fatal**

`file_ring_buffer.cpp:848-850` writes `MetadataSize` and `MetadataChecksum` before copying the data, with no fences at all. A crash (or page flush) between the header update and the data copy yields a checksum mismatch, which the constructor now treats as full corruption. The ordering is pre-existing, but this PR makes the consequence unconditional. Minimum improvement: copy data first, signal fence, then size + checksum — that at least removes the compiler-reordering exposure and shrinks the window to the size/checksum pair. True crash atomicity would need versioned double-buffered metadata, which is probably out of scope, but the current order is strictly worse than necessary.

> It will be corruption in either case so I see to reason to change anything.

**5. Files smaller than the header are now permanently Failed instead of re-initialized**

Old code: `Map.Length() < sizeof(THeader)` → initialize fresh. New code (accessor.cpp:354): any file of 1..71 bytes → Failed → corrupted, including an all-zero one. `ShouldNotValidateTooSmallFile` shows this is deliberate, but note two things: it contradicts the doc comment on `NotInitialized` in the header ("Header will be accessible if file length >= header size" implies small zeroed files were meant to land in NotInitialized), and it changes recovery behavior for externally truncated files. Probably fine — silent re-init of a truncated file is data loss too — but worth an explicit decision, and the enum doc should match the code either way.

> Small files cannot be produced by `TFileRingBuffer` logic so it is a corruption even if the file is zeroed.

**6. Condition order in TryMigrate**

`file_ring_buffer.cpp:271`: `if (!IsMigrationNeeded() || IsCorrupted())` — `IsMigrationNeeded()` dereferences `Header()` before the corruption check. Today all callers reach this only through paths where corruption implies early return before `Header()` can be null, but that's an invariant maintained by call-site discipline, not by this function. Swapping the operands makes it robust for free.

> Changed.

**7. Minor accessor issues**

- `ProcessMap` (accessor.cpp:480): when `rawData.size() != fileSize` it returns E_IO but leaves `RawData` pointing at the partial mapping. Callers mark corrupted so it's benign, but `UpdateRawData({})` on that path would match the catch blocks.

> Done. Accessing partially mapped file has no sense.

- `TFileMapFileRingBufferAccessor` inherits publicly, so callers can invoke `UpdateRawData`/`ValidateAndInitialize` directly and desync `RawData` from the actual map. Protected inheritance or composition would close that.

> Made `UpdateRawData` protected but `ValidateAndInitialize` should be public since if doesn't desync the mapping.

- `UpdateRawData` accepts arbitrary spans and `DoValidateAndInitialize` does `reinterpret_cast<TFileRingBufferHeader*>(RawData.data())` with no alignment check. `TData<ui64, true>` asserts alignment for V6 data, but the header cast is unchecked — a test feeding a `TVector<char>` buffer at an odd offset gets UB. Cheap to assert in `UpdateRawData`.

> Added check and test.

- `LastValidationError` is set to "File is not initialized" for the NotInitialized status, which is not a failure; `TImpl::Validate` then reports it via `SetCorrupted` if that status ever occurs post-construction. Fine in practice, but the message reads oddly as a corruption report.

> File is expected to be initialized by that moment. So non initialzed = error.

- Every validation error is E_FAIL through the local `MakeError` shadow; distinguishing E_IO (map failures already do this) from E_INVALID_STATE would help triage.

> Not needed here.

- Typos: "initialzed" in the header enum doc; "metadata offset aligned is enforced" in accessor.cpp.

> Fixed.

**What's correct that I explicitly verified:** the `ValidateData` wrap-walk handles implicit slack markers (entry chain landing within `sizeof(EntryHeader)` of capacity) via the default-constructed header from an out-of-bounds `ReadEntryHeader`; the `writePos == 0` interrupted-Alloc case is preserved; the crash windows inside `SetReadAndWritePosToZeroForEmptyBuffer` itself are all restart-valid in the *source* order (the problem in finding 1 is purely the missing fence after it); the temp-offset `AlignUp` fix is required because the never-validated `DataOffset = tempDataOffset` intermediate state must pass V6 header validation after a crash-restart; and `EntryMap` pointer stability across `Validate()` holds because revalidation never remaps.

Findings 1 and 4 are the ones I'd block on; 2 and 3 need at least an explicit acknowledgment in the PR description since they change production startup cost and failure semantics.

neihar (dismissed):

e673:
GitHub's API is rate-limited from this container, so I'll fetch the PR branch directly with git instead.Now let me look at the format header/cpp (which this PR touches) and the diff to the ring buffer to compare against the base version.Let me check how `TFileRingBuffer::Validate()` is used elsewhere in the tree (it's now a full CRC re-scan), and skim the new tests.Let me check the base (main) constructor and validation to correctly attribute what's new vs pre-existing.I pulled the current head of the branch (`69a74ca`) and reviewed the actual code rather than relying on the review threads (the PR has already absorbed several rounds of bot/human feedback — fences added in `TryMigrate`/`SetReadAndWritePosToZeroForEmptyBuffer`, `UpdateRawData` made protected, the buffer-alignment check + test, the `IsMigrationNeeded`/corruption ordering, typo fixes). Here's what I'd still flag on the current state:

**1. Inconsistent capability flags guarding the same invariant (introduced by this PR).**
The rule "a freed/slack entry must have `DataChecksum == 0`" is enforced in two places with *different* flags:
- `file_ring_buffer_accessor.cpp`, `ReadAndValidateEntry`: gated on `capabilities.EntryHeaderIsProcessedAtomically` (the new flag)
- `file_ring_buffer.cpp`, `GetEntry` (lines ~185 and ~222): gated on `Capabilities().EntryHeaderIsCoveredByChecksum` (the old flag)

They coincide for V5 (false/false) and V6 (true/true), so there's no functional bug today — but they encode different concepts, and a future version where they diverge (e.g., checksum covers the header but writes aren't atomic) would silently make startup validation and runtime entry reading disagree about what's corrupt. The atomicity flag is the semantically correct one here: V5 must tolerate stale checksums on freed entries precisely because its two-word header write is non-atomic. I'd unify `GetEntry` on `EntryHeaderIsProcessedAtomically`.

> I've merged two flags into one (EntryHeaderIsProcessedAtomically) that is more semantically correct

**2. Crash during first-time initialization now permanently bricks the file.**
In the constructor's `NotInitialized` branch, the fresh header is written as a plain struct assignment with no store-ordering guarantee and no fence. A crash mid-write leaves a non-zero file with an invalid header (e.g., `Version == 0` plus partial fields), which the new strict validation permanently rejects — and `ShouldNotValidateFileWithZeroVersionAndNonZeroContents` deliberately locks that in. The old code was arguably worse (`Y_ABORT_UNLESS` on the version → process abort loop, or silent re-init for sub-header files), so this is a changed failure mode rather than a regression — but since a freshly-initializing file provably holds no user data, ending up in "corrupted, needs manual deletion" is a needlessly harsh outcome. Cheap mitigation: write all fields, `atomic_signal_fence`, then write `Version` last — and decide explicitly whether `Version == 0` with `ReadPos == WritePos == 0` should be recoverable. At minimum this trade-off deserves a sentence in the PR description.

> This is not a problem at all and should not be implemented here

**3. `ResizeMetadata` never converges when shrinking below `MetadataSize`, and each pass is now expensive.**
If the requested capacity is smaller than the stored `MetadataSize`, the final capacity is clamped to `MetadataSize`, so `Header()->MetadataCapacity != Args.MetadataCapacity` remains true forever and the constructor re-enters `ResizeMetadata` on **every** open. That was true before, but each pass was cheap; now every pass does a same-size `ResizeAndRemap` plus **two full `Validate()` runs** — i.e., two complete CRC walks over all live entries per open, forever. The existing test even exercises this configuration (`ShouldResizeMetadata` constructs with capacity 1 after writing 9 bytes of metadata) without noticing the cost. An early return when `newMetadataCapacity == Header()->MetadataCapacity && newDataOffset == Header()->DataOffset` would fix it.

> Changed the logic

**4. Smaller items.**
- The comment above the new alignment check in `DoValidateAndInitialize` says "the raw data **size** should be aligned to 8 bytes" while the code checks the buffer *address*; the note about page-size mapping is also about the pointer, not the size.

> Fixed the comment

- `ResizeAndRemap` in the accessor reuses the error text "Failed to map file" for resize failures — worth distinguishing since these get surfaced as corruption reasons.

> Fixed the message 

- `TFileRingBufferAccessor` is a public, non-final base with a non-virtual destructor. Nothing deletes through the base today, but marking `TFileMapFileRingBufferAccessor` `final` (or the base dtor protected) closes the trap.

> Added `final`

- `TFileRingBuffer::Validate()` is now a full pass — all-zeros scan, every entry CRC, metadata CRC — on every call. Fine at startup, but any caller using it as a cheap periodic invariant check will regress badly on a multi-GB write-back cache file; worth a doc note on the public method.

> Added a comment that it validates data checksum

- Test coverage gaps in the new accessor suite: the interrupted-Alloc state (`WritePos == 0`, `ReadPos` → slack marker) is never validated at the accessor level, the "Last entry ends at position …" overshoot branch and the V6 unaligned `ReadPos`/`WritePos` branches in `ValidateData` are untested, and Debug-mode behavior for `NotInitialized` isn't covered.

> Added tests

For balance: the things I specifically checked and found correct include the wrap-walk in `ValidateData` (both explicit and implicit slack markers terminate it properly, and `ShouldValidateWrappedFile` happens to exercise the explicit marker on V5 and the implicit one on V6), the fence discipline in `TryMigrate`/`SetReadAndWritePosToZeroForEmptyBuffer` after the earlier review fix, the overflow-safe ordering of the header range checks, the `AlignUp` on `tempDataOffset` (genuinely required now that intermediate `Validate()` calls create a V6 processor on the temp layout), header re-fetch through the accessor after every remap, and the `ShouldMigrate` test fix (it really was asserting against the wrong file, `f1` instead of `f2`). Items 1–3 are the ones I'd actually raise on the PR; the rest are take-it-or-leave-it polish.

neihar (approved):

yegorskii (approved):

Issue #5456, opened 2026-03-11
[NBS] Asynchronous handler can access memory freed on reconnect


komarevtsev-d:
```c++
Thread 9751 (LWP 3605349):
#0 raise () from /lib/x86_64-linux-gnu/libc.so.6
#1 abort () from /lib/x86_64-linux-gnu/libc.so.6
#2 tcmalloc::tcmalloc_internal::Crash (filename=<optimized out>, line=<optimized out>, msg=0x7f1a83df3120 "3605349 /mount_path/contrib/libs/tcmalloc/tcmalloc/tcmalloc.cc:692] CHECK in ReportDoubleFree: Possible double free detected of 0x505a8ea00000\n", msglen=187, oom=<optimized out>) at /mount_path/contrib/libs/tcmalloc/tcmalloc/internal/logging.cc +176
#3 tcmalloc::tcmalloc_internal::CheckFailed (file=<optimized out>, line=<optimized out>, msg=0x2 <error: Cannot access memory at address 0x2>, msglen=-2002448373) at /mount_path/contrib/libs/tcmalloc/tcmalloc/internal/logging.cc +181
#4 tcmalloc::tcmalloc_internal::CheckFailed<void*> (func=0x55e4d4cc7436 "ReportDoubleFree", file=<optimized out>, line=692, format=..., args=@0x7f1a83df3360: 0x505a8ea00000) at /mount_path/contrib/libs/tcmalloc/tcmalloc/internal/logging.h +231
#5 tcmalloc::tcmalloc_internal::(anonymous namespace)::ReportDoubleFree (ptr=0x505a8ea00000) at /mount_path/contrib/libs/tcmalloc/tcmalloc/tcmalloc.cc +692
#6 tcmalloc::tcmalloc_internal::(anonymous namespace)::InvokeHooksAndFreePages (ptr=0x505a8ea00000, size=...) at /mount_path/contrib/libs/tcmalloc/tcmalloc/tcmalloc.cc +711
#7 tcmalloc::tcmalloc_internal::(anonymous namespace)::do_free (ptr=0x2) at /mount_path/contrib/libs/tcmalloc/tcmalloc/tcmalloc.cc +777
#8 TCMallocInternalFree (ptr=0x2) at /mount_path/contrib/libs/tcmalloc/tcmalloc/tcmalloc.cc +1392
[...]

PR #6661, opened 2026-07-31
Cached rdma memory windows

yegorskii:
### Notes
**Summary**
This PR adds optional RDMA Type 2 Memory Window support to reduce the cost of exposing request buffers and improve rkey lifecycle management.
Instead of exposing long-lived Memory Region keys directly, the client binds reusable Memory Windows to individual request buffers. These windows are invalidated after the request completes and reused only when invalidation has been confirmed.
The feature is controlled by the UseMemoryWindows client setting and the --memory-windows option in the RDMA test tool.
**How it works**
**Connection setup**
When Memory Windows are enabled, the client:
registers buffer pools with IBV_ACCESS_MW_BIND;
preallocates an endpoint-local pool of Type 2 Memory Windows;
increases the QP send queue capacity to account for BIND_MW, SEND, and LOCAL_INV work requests;
negotiates SEND_WITH_INV support with the server during connection establishment.
The server advertises SEND_WITH_INV support in the accept message.
**Request submission**
For each input and output buffer, the client allocates or reuses a Memory Window and generates a new rkey.
The request is posted as a chained sequence:
BIND_MW(input) -> BIND_MW(output) -> SEND
The terminal SEND is fenced so that the request cannot reach the server before the Memory Window bindings become visible.
The request message contains the temporary Memory Window rkeys instead of the underlying Memory Region keys.
**Memory Window invalidation**
The input window is invalidated locally after request completion.
For the output window, the client selects one of two modes:
If the server supports SEND_WITH_INV, the server invalidates the output rkey while sending the response.
Otherwise, the client posts a local invalidation for the output window as well.
Remote invalidation completions are validated against the expected rkey. A missing or unexpected invalidation prevents the affected window from being recycled.
Locally invalidated windows are returned to the endpoint cache only after the terminal LOCAL_INV completion. This guarantees that the old rkey is no longer usable before the window and its buffer are reused.
**Resource ownership**
Request resources are exclusively owned by TRequest. They include:
input and output buffers;
input and output Memory Windows;
invalidation mode and expected rkey;
deferred completion or abort state;
buffer-pool generation.
Invalidation requests have two states:
PendingInvalidationRequests: invalidation has not yet been posted.
PostedInvalidationRequests: invalidation is visible to hardware and may still reference request resources.
A posted invalidation retains the entire request until its terminal CQE is received or the QP is destroyed. This avoids separate user-side and hardware-side ownership and keeps the lifetime rules explicit.
**Cancellation and error handling**
Requests cannot be destroyed while a posted send chain may still reference their Memory Windows.
Cancellation, timeout, and disconnect handling therefore distinguish between:
requests that have not been submitted and can be completed immediately;
requests waiting for their terminal SEND completion;
requests with pending invalidations;
requests whose work chain was only partially accepted by ibv_post_send().
Errors reported by SEND completions now abort the corresponding request even when the CQE is the first place where the error is observed. This prevents requests from remaining indefinitely in ActiveRequests with SendRequest state during disconnect.
ibv_post_send() now exposes bad_wr, allowing the client to distinguish between:
a completely rejected chain, whose resources can be released immediately;
a partially posted chain, whose resources must remain alive until QP destruction.
Memory Windows associated with cancelled, timed-out, failed, or partially posted requests are dropped instead of returned to the reuse pool.
**Disconnect and teardown**
Pending invalidations that were never posted can be completed immediately.
Posted invalidations retain their requests until either:
their terminal CQE arrives; or
the QP is destroyed.
QP destruction is used as the final lifetime boundary for WQEs that may still reference Memory Windows. The teardown order ensures that Memory Windows are not deallocated while the QP may still access them:
stop polling/posting
    -> destroy QP
    -> finalize retained requests
    -> deallocate Memory Windows
    -> release buffers and remaining transport resources
Send slots owned by a QP that is committed to destruction are tracked separately, allowing disconnect flushing to finish without waiting indefinitely for invalidation completions that may never arrive.
A buffer-pool generation counter also prevents resources retained across reconnect from being returned to a newly created buffer pool.
**Additional changes**
Added Memory Window allocation support to the verbs abstraction and test implementation.
Added opcode and work-request formatting for diagnostics.
Added tracing probes for input and output Memory Window binding.
Added configuration and monitoring output for UseMemoryWindows.
Updated buffer descriptors to expose their underlying Memory Region.
Extended the RDMA protocol with request and accept capability flags.
Added comprehensive tests for:Memory Window allocation and reuse;
local and remote invalidation;
capability negotiation;
cancellation and timeout ordering;
disconnect and QP teardown;
failed and flushed completions;
partially posted send and invalidation chains;
retained resource lifetime across reconnect.

### Issue

### DIFF
--- a/cloud/storage/core/libs/rdma/impl/client.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client.cpp
@@ -461,6 +461,11 @@ struct TRequestResources
     ui64 BufferPoolGeneration = 0;
     NProto::TError Error;
 
+    // Cleanup is performed only after the user releases TRequest and every
+    // posted WR chain has either completed or been destroyed with its QP.
+    bool UserReleased = false;
+    ui32 HardwareOwners = 0;
+
     ui32 Status = RDMA_PROTO_FAIL;
     ui32 ResponseBytes = 0;
 };
@@ -565,12 +570,18 @@ private:
     TEventHandle DisconnectEvent;
 
     TSimpleList<TRequest> QueuedRequests;
+    TSimpleList<TRequest> PartiallyPostedRequests;
+    // Send slots whose terminal CQE is no longer required because the current
+    // QP is committed to destruction.
+    std::atomic<size_t> QpOwnedSendSlots = 0;
     TIntrusiveListWithAutoDelete<TInvalidationRequest, TInvalidationRequestDelete>
         PendingInvalidationRequests;
     TIntrusiveListWithAutoDelete<TInvalidationRequest, TInvalidationRequestDelete>
         InflightInvalidationRequests;
     TIntrusiveListWithAutoDelete<TInvalidationRequest, TInvalidationRequestDelete>
         RetiredInvalidationRequests;
+    TIntrusiveListWithAutoDelete<TInvalidationRequest, TInvalidationRequestDelete>
+        PartiallyPostedInvalidationRequests;
     TActiveRequests ActiveRequests;
 
     std::atomic<ui64> ReqIdPool{0};
@@ -664,6 +675,7 @@ private:
         TRequestPtr& req,
         NProto::TError error) noexcept;
     void AbortRequest(ui32 reqId) noexcept;
+    void AbortRequest(ui32 reqId, NProto::TError error) noexcept;
     void AbortRequest(TRequestPtr req, const NProto::TError& error) noexcept;
     void AbortRequest(
         TRequestPtr req,
@@ -690,6 +702,9 @@ private:
 
     // Resource lifetime and allocation helpers.
     void ReleaseRequestResources(TRequestResourcesPtr resources) noexcept;
+    void AcquireHardwareResources(const TRequestResourcesPtr& resources) noexcept;
+    void ReleaseHardwareResources(const TRequestResourcesPtr& resources) noexcept;
+    void CleanupRequestResourcesLocked(TRequestResources& resources) noexcept;
     void FreeRequest(TRequest* creq) noexcept;
     NVerbs::TMemoryWindowPtr AcquireMemoryWindowLocked();
     void RecycleMemoryWindowLocked(NVerbs::TMemoryWindowPtr& memoryWindow);
@@ -912,13 +927,8 @@ void TClientEndpoint::SetupQP()
 
 void TClientEndpoint::DestroyQP() noexcept
 {
-    RetiredInvalidationRequests.Clear();
-
-    with_lock (AllocationLock) {
-        MaxFreeMemoryWindows = 0;
-        FreeMemoryWindows.clear();
-    }
-
+    // QP destruction is the lifetime boundary for every MW which may have
+    // been referenced by a posted WQE.
     if (Connection && Connection->qp) {
         if (Config.VerbsQP) {
             Verbs->DestroyQP(Connection->qp);
@@ -927,6 +937,34 @@ void TClientEndpoint::DestroyQP() noexcept
             Verbs->RdmaDestroyQP(Connection.get());
             Connection->qp = nullptr;
         }
+    }
+
+    while (RetiredInvalidationRequests) {
+        auto* req = RetiredInvalidationRequests.PopFront();
+        Y_ABORT_UNLESS(req);
+        ReleaseHardwareResources(req->Resources);
+        delete req;
+    }
+
+    while (auto req = PartiallyPostedRequests.Dequeue()) {
+        AbortRequest(
+            std::move(req),
+            EAbortCounter::Aborted,
+            MakeError(E_RDMA_UNAVAILABLE, "failed to post complete send chain"));
+    }
+
+    while (PartiallyPostedInvalidationRequests) {
+        auto* req = PartiallyPostedInvalidationRequests.PopFront();
+        Y_ABORT_UNLESS(req);
+        ReleaseHardwareResources(req->Resources);
+        FinalizeInvalidationRequest(req, false);
+    }
+    QpOwnedSendSlots = 0;
+
+    // No WQE can reference windows from the endpoint-local pool anymore.
+    with_lock (AllocationLock) {
+        MaxFreeMemoryWindows = 0;
+        FreeMemoryWindows.clear();
     }
 
     CompletionQueue.reset();
@@ -1283,15 +1321,19 @@ int TClientEndpoint::ValidateCompletion(ibv_wc* wc) noexcept
     if (id.Magic == SendMagic && id.Index < SendWrs.size()) {
         auto* send = &SendWrs[id.Index];
         auto handleLocalInvalidationCompletion = [&] {
+            if (!send->context) {
+                RDMA_TRACE(send << " local invalidate completion already handled");
+                return;
+            }
             send->context = nullptr;
             SendQueue.Push(send);
         };
-        auto handleRequestSendCompletion = [&] {
+        auto handleRequestSendCompletion = [&] (NProto::TError error) {
             const ui32 reqId =
                 SafeCast<ui32>(reinterpret_cast<uintptr_t>(send->context));
             Counters->SendRequestCompleted();
             SendQueue.Push(send);
-            AbortRequest(reqId);
+            AbortRequest(reqId, std::move(error));
         };
 
         switch (wc->status) {
@@ -1302,7 +1344,11 @@ int TClientEndpoint::ValidateCompletion(ibv_wc* wc) noexcept
                         handleLocalInvalidationCompletion();
                         break;
                     case IBV_WC_SEND:
-                        handleRequestSendCompletion();
+                        handleRequestSendCompletion(MakeError(
+                            E_RDMA_UNAVAILABLE,
+                            TStringBuilder()
+                                << "send request flushed: "
+                                << NVerbs::GetStatusString(wc->status)));
                         break;
                     default:
                         break;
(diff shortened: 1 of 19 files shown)


yegorskii:
@codex review

yegorskii:
@codex review

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -761,15 +906,39 @@ void TClientEndpoint::SetupQP()
...
         if (Config.VerbsQP) {
             Verbs->DestroyQP(Connection->qp);
             Connection->qp = nullptr;
         } else {
             Verbs->RdmaDestroyQP(Connection.get());
+            Connection->qp = nullptr;
         }
     }
 
+    while (auto req = PartiallyPostedRequests.Dequeue()) {
```
tpashkin, line 921: `DestroyQP` gets called from `CM` thread, so you either need synchronization or do this from `CQ` thread

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -489,11 +552,24 @@ class TClientEndpoint final
     TEventHandle DisconnectEvent;
 
     TSimpleList<TRequest> QueuedRequests;
+    TSimpleList<TRequest> PartiallyPostedRequests;
```
tpashkin, line 555: Why not put them back into the `QueuedRequests`?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -823,14 +992,22 @@ TResultOrError<TClientRequestPtr> TClientEndpoint::AllocateRequest(
         shared_from_this(),
         std::move(handler),
         std::move(context));
+    req->Resources.BufferPoolGeneration = BufferPoolGeneration.load();
```
tpashkin, line 995: We actually have an issue for this – #5456. Let's move it to a separate PR

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1026,26 +1221,39 @@ void TClientEndpoint::AbortRequests() noexcept
...
 
 void TClientEndpoint::AbortRequest(
     TRequestPtr req,
-    ui32 err, const
-    TString& msg) noexcept
+    const NProto::TError& error) noexcept
 {
+    // For timed out/cancelled/aborted requests we cannot guarantee that remote
+    // peer invalidated keys, so force MW deallocation instead of recycling.
+    req->Resources.RecyclePolicy = TRequestResources::ERecyclePolicy::Drop;
```
tpashkin, line 1251: `ibv_dealloc_mw` automatically unbinds the window, you don't have to do it manually – https://man7.org/linux/man-pages/man3/ibv_alloc_mw.3.html

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,42 +1447,277 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+    if (Config.UseMemoryWindows) {
+        if (req->Resources.InMemoryWindow) {
+            if (req->CallContext) {
+                LWTRACK(
+                    BindInBufferStarted,
+                    req->CallContext->LWOrbit,
+                    req->CallContext->RequestId);
+            }
+
+            const ui32 inRKey = GetNextMemoryWindowRKey(
```
tpashkin, line 1496: Why do you need to do that?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,42 +1447,277 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+                    req->CallContext->LWOrbit,
+                    req->CallContext->RequestId);
+            }
 
+            const ui32 outRKey = GetNextMemoryWindowRKey(
+                req->Resources.OutMemoryWindow.get());
+            msg->Out.RKey = outRKey;
+            ConfigureOutInvalidation(req->Resources, msg, outRKey);
+
+            bindOut.wr_id = send->wr.wr_id;
```
tpashkin, line 1526: Can we actually use the same `wr_id` for all 3 send's?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,42 +1447,277 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+                head = &bindOut;
+            }
+            tail = &bindOut;
+        }
+    }
+
+    if (tail) {
+        // BIND_MW is asynchronous on HCA, SEND must be fenced to ensure
+        // memory window bindings become visible before request delivery.
+        send->wr.send_flags |= IBV_SEND_FENCE;
```
tpashkin, line 1547: [man](https://www.man7.org/linux/man-pages/man3/ibv_bind_mw.3.html) is actually very explicit about the ordering
> The user may safely send the R_key using a send request on the same QP, (based on QP ordering rules: a send after a bind request on the same QP are always ordered)

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,42 +1447,277 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
+        tail->next = &send->wr;
```
tpashkin, line 1548: in my opinion, chaining wrs is an interesting idea, but we shouldn't do it, at least not right away

I expect latency gains to be negligible, but error handling and reasoning about request ownership gets way more complicated

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1173,42 +1447,277 @@ void TClientEndpoint::HandleCompletionEvent(ibv_wc* wc) noexcept
...
-    ActiveRequests.Push(std::move(req));
+void TClientEndpoint::ConfigureOutInvalidation(
+    TRequestResources& resources,
+    TRequestMessage* msg,
+    ui32 outRKey) noexcept
+{
+    if (PeerSupportsSendWithInvalidate) {
+        msg->Unused |= RDMA_REQUEST_FLAG_USE_MEMORY_WINDOWS;
+        resources.OutMode =
+            TRequestResources::EOutMode::RemoteInvalidate;
```
tpashkin, line 1597: looks like an unnecessary complication. i think we should stick with the local invalidation for the first version

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1388,14 +1940,247 @@ bool TClientEndpoint::FlushHanging() const
...
+        memoryWindow = NVerbs::NullPtr;
+    }
+}
+
+ui32 TClientEndpoint::GetNextMemoryWindowRKey(ibv_mw* memoryWindow)
+{
+    Y_ABORT_UNLESS(memoryWindow);
+
+    const ui32 current = memoryWindow->rkey;
+    const ui32 next = (current & 0xFFFFFF00u) | ((current + 1u) & 0x000000FFu);
```
komarevtsev-d, line 2179: This magic needs a comment

cloud/storage/core/libs/rdma/impl/client.cpp
komarevtsev-d: Yea, @tpashkin is probably right. Let's get a working version first. Then we can optimize it.

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -656,9 +783,12 @@ void TClientEndpoint::CreateQP()
         ResetConfig = false;
     }
 
+    const ui32 sendWrPerRequest = Config.UseMemoryWindows ? 3 : 1;
```
komarevtsev-d, line 786: Do we really need this code? If we enable "UseMemoryWindows" in the config, we can simultaneously increase the size of the send queue if needed.

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1388,14 +1940,247 @@ bool TClientEndpoint::FlushHanging() const
...
+NVerbs::TMemoryWindowPtr TClientEndpoint::AcquireMemoryWindowLocked()
+{
+    if (!FreeMemoryWindows.empty()) {
+        auto memoryWindow = std::move(FreeMemoryWindows.back());
+        FreeMemoryWindows.pop_back();
+        return memoryWindow;
+    }
+
+    // Slow path for request spikes over the preallocated pool size.
+    return Verbs->CreateMemoryWindow(Connection->pd);
```
komarevtsev-d, line 2157: Should we have counters for getting MW from cache and creating it?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -684,15 +814,28 @@ void TClientEndpoint::CreateQP()
...
-        Verbs,
-        Connection->pd,
-        IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE);
+    if (Config.UseMemoryWindows) {
+        // Endpoint-local pool: windows are reused only inside this endpoint/QP
+        // flow (Type 2A semantics at application level).
+        MaxFreeMemoryWindows = static_cast<size_t>(Config.SendQueueSize) * 2;
+        FreeMemoryWindows.reserve(MaxFreeMemoryWindows);
+        for (size_t i = 0; i < MaxFreeMemoryWindows; ++i) {
+            FreeMemoryWindows.push_back(Verbs->CreateMemoryWindow(Connection->pd));
```
komarevtsev-d, line 833: If this throws, then `TClient::BeginConnect` will call `Disconnect()` which does nothing in "Connecting" state. Probably should fix this...

cloud/storage/core/libs/rdma/impl/client.cpp
komarevtsev-d: Note: i think, there is a minor catch to current implementation. If the endpoint reconnects between `AllocateRequest` and `SendRequest`, the request is still accepted. 
We should probably reject stale-generation requests, and/or retain each generation's PD/pool until its allocated requests are released.

PR #6664, opened 2026-07-31
issues-6645 Fix crash in silk library on arm with page_size 64k

proller:
### Notes

### Issue


qkrorlqr:
Fixed here https://github.com/ydb-platform/nbs/pull/6711

PR #6667, opened 2026-07-31
[Disk Manager] cleanup testcommon/common.go: use require instead of returning error for YDB-related constructors

SvartMetal:
### Notes
Refactor the storage helpers in `cloud/disk_manager/internal/pkg/facade/testcommon` so they take `*testing.T` and assert with `require` instead of returning an error.

Affected helpers: `newYDB`, `newResourceStorage`, `newFilesystemSnapshotStorage`, `newPoolStorage`, `newCellStorage`, `newSnapshotStorage`, `NewTaskStorage`, `newScheduler` and `storageWithDB.close`. All of them now return only the value.

These helpers construct YDB clients and storages against the test recipe. A failure there is a broken test environment, not a condition any caller can handle, so every call site did the same `require.NoError(t, err)` immediately after.

Also replaced `time.ParseDuration("30m")` with `30 * time.Minute` in `newResourceStorage`

### DIFF
--- a/cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
+++ b/cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
@@ -58,7 +58,7 @@ func waitForFilesystemSnapshotCreationStarted(
 ) {
 
 	require.EventuallyWithT(t, func(collect *assert.CollectT) {
-		meta, err := testcommon.GetFilesystemSnapshotMeta(ctx, snapshotID)
+		meta, err := testcommon.GetFilesystemSnapshotMeta(t, ctx, snapshotID)
 		assert.NoError(collect, err)
 		assert.NotNil(collect, meta)
 		if err != nil || meta == nil {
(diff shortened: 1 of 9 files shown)


cloud/disk_manager/internal/pkg/facade/filesystem_snapshot_service_test/filesystem_snapshot_service_test.go
```diff
@@ -58,15 +58,15 @@ func waitForFilesystemSnapshotCreationStarted(
 ) {
 
 	require.EventuallyWithT(t, func(collect *assert.CollectT) {
-		meta, err := testcommon.GetFilesystemSnapshotMeta(ctx, snapshotID)
+		meta, err := testcommon.GetFilesystemSnapshotMeta(t, ctx, snapshotID)
```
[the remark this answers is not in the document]
SvartMetal: Thanks, fixed

cloud/disk_manager/internal/pkg/facade/testcommon/common.go
```diff
@@ -667,20 +658,15 @@ func newResourceStorage(
 		db,
 		endedMigrationExpirationTimeout,
 	)
-	if err != nil {
-		_ = db.Close(ctx)
-		return nil, err
```
jkuradobery, line 672: Why not require.fail here?

SvartMetal: Because we did not fail on close before your recent change, and because close is a best-effort operation, its main value is preventing resource leaks

jkuradobery (approved):

gy2411 (approved):

PR #6613, opened 2026-07-28
issue-6208: [Filestore] Track most accessed nodes

szumal:
### Notes
This PR covers this part of the issue:
> Per-file: most loaded files per shard -> output for the whole filesystem

EDIT:
This PR will only introduce client and proto changes, actual tracking will be introduced in: https://github.com/ydb-platform/nbs/pull/6670

### Issue
https://github.com/ydb-platform/nbs/issues/6208

PR #6670, opened 2026-08-03
issue-6208: [Filestore] Add node access tracking

szumal:
### Notes
This PR is a follow up to: https://github.com/ydb-platform/nbs/pull/6613
Here, I am adding tracking the node access stats in the tablet code.

Both PRs cover this part of the issue:
> Per-file: most loaded files per shard -> output for the whole filesystem

### Issue


cloud/filestore/private/api/protos/tablet.proto
szumal, line 179: not removing this yet, will do when updating the client to use the new request/response

cloud/filestore/libs/storage/core/request_info.h
```diff
@@ -38,6 +38,8 @@ struct TRequestInfo
     // Request processing start ts. Might be empty.
     TInstant StartedTs;
 
+    bool NodeDiagnosticStatsStarted = false;
```
qkrorlqr: doesn't seem to be required

we should call the code that inserts an item into those stats only once even without this flag, adding this flag and checking whether it's set looks like a hack which conceals the inability to prove that there's a place in the request processing code that's visited exactly once

cloud/filestore/libs/storage/tablet/tablet_actor_counters.cpp
```diff
@@ -919,6 +919,26 @@ void TIndexTabletActor::HandleGetStorageStats(
...
 }
 
+void TIndexTabletActor::HandleGetDiagnosticStats(
+    const TEvIndexTablet::TEvGetDiagnosticStatsRequest::TPtr& ev,
+    const TActorContext& ctx)
+{
+    auto response =
+        std::make_unique<TEvIndexTablet::TEvGetDiagnosticStatsResponse>();
+
+    for (const auto& accessStats: GetNodeAccessStats(ctx.Now(), ev->Get()->Record.GetLimit())) {
```
qkrorlqr:
```
$ echo -n '    for (const auto& accessStats: GetNodeAccessStats(ctx.Now(), ev->Get()->Record.GetLimit())) {'|wc -c
96
```

cloud/filestore/libs/storage/tablet/tablet_state.cpp
```diff
@@ -16,6 +16,9 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+constexpr size_t MaxNodeDiagnosticEntries = 10'000;
+constexpr TDuration AccessStatsHalfLife = TDuration::Minutes(10);
```
qkrorlqr: let's put this value into `TStorageConfig`

cloud/filestore/libs/storage/tablet/tablet_state.cpp
```diff
@@ -16,6 +16,9 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+constexpr size_t MaxNodeDiagnosticEntries = 10'000;
```
qkrorlqr: let's put this value into `TStorageConfig`

cloud/filestore/libs/storage/tablet/tablet_state.cpp
```diff
@@ -70,6 +73,7 @@ ui64 CalculateInMemoryIndexCacheCapacity(
 
 TIndexTabletState::TIndexTabletState()
     : Impl(new TImpl(AllocatorRegistry))
+    , NodeAccessStatsTracker(MaxNodeDiagnosticEntries, AccessStatsHalfLife)
```
qkrorlqr: this should live inside `TImpl`

qkrorlqr: `NodeToSessionCounters` and some other state fields should live there as well - let's move the other ones there in a separate PR

cloud/filestore/config/storage.proto
```diff
@@ -870,4 +870,8 @@ message TStorageConfig
 
     // Async processing of read-only create handle requests.
     optional bool AsyncCreateHandleEnabled = 532;
+
+    optional uint32 MaxNodeDiagnosticEntries = 533;
+
+    optional uint32 AccessStatsHalfLife = 534; // in minutes
```
szumal: P1 — AccessStatsHalfLife interprets minutes as microseconds
TStorageConfig converts uint32 duration fields directly to TDuration, whose numeric value is microseconds. Thus setting AccessStatsHalfLife = 10 produces a 10µs half-life, despite the field documenting minutes. This makes access scores decay almost immediately.

cloud/filestore/libs/storage/tablet/tablet_state.cpp
szumal, line 287: P2 — Runtime config changes do not update the tracker
AccessTracker is initialized only in LoadState(). UpdateConfig() does not recreate or reconfigure it, so changes to MaxNodeDiagnosticEntries or AccessStatsHalfLife have no effect until the tablet restarts.

szumal (commented):
CODEX REVIEW

cloud/filestore/libs/storage/tablet/tablet_state.cpp
szumal: I assume we only really care about setting the tracker params at tablet startup

cloud/filestore/config/storage.proto
```diff
@@ -873,4 +873,8 @@ message TStorageConfig
 
     // Allow to create load actor.
     optional bool EnableLoadActor = 533;
+
+    optional uint32 MaxNodeDiagnosticEntries = 534;
```
qkrorlqr, line 878: add comments plz

cloud/filestore/libs/storage/core/config.cpp
```diff
@@ -393,6 +393,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
     xxx(FanoutStatsCollectionInShardsDisabled,         bool,    false         )\
                                                                                \
     xxx(EnableLoadActor,                               bool,    false         )\
+    xxx(MaxNodeDiagnosticEntries,                      ui32,    10'000        )\
```
qkrorlqr: let's set defaults to 0, then we'll gradually enable this, then change the defaults

cloud/filestore/config/storage.proto
```diff
@@ -873,4 +873,8 @@ message TStorageConfig
...
+
+    optional uint32 AccessStatsHalfLife = 535;
```
qkrorlqr: let's use a more detailed name, this name doesn't explain what kind of AccessStats are implied

cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
```diff
@@ -1060,6 +1074,13 @@ void TIndexTabletActor::CompleteTx_ReadData(
             args.OriginByteRange.Length,
             ctx.Now() - args.RequestInfo->StartedTs);
 
+        if (args.Node->Attrs.GetType() == NProto::ENodeType::E_REGULAR_NODE) {
```
qkrorlqr: reads can only happen for `E_REGULAR_NODE`s, this check looks redundant

cloud/filestore/libs/storage/tablet/tablet_actor_writedata.cpp
```diff
@@ -452,6 +456,13 @@ void TIndexTabletActor::CompleteTx_WriteData(
             args.ByteRange.Length,
             ctx.Now() - args.RequestInfo->StartedTs);
 
+        if (args.Node->Attrs.GetType() == NProto::ENodeType::E_REGULAR_NODE) {
```
qkrorlqr: same

cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
```diff
@@ -1144,6 +1165,13 @@ void TIndexTabletActor::CompleteTx_ReadData(
             MakeError(S_OK),
             ProfileLog);
 
+        if (args.Node->Attrs.GetType() == NProto::ENodeType::E_REGULAR_NODE) {
```
qkrorlqr: same

cloud/filestore/libs/storage/tablet/tablet_actor_readdata.cpp
```diff
@@ -1060,6 +1074,13 @@ void TIndexTabletActor::CompleteTx_ReadData(
             args.OriginByteRange.Length,
             ctx.Now() - args.RequestInfo->StartedTs);
 
+        if (args.Node->Attrs.GetType() == NProto::ENodeType::E_REGULAR_NODE) {
+            if (!UpdateAccessStats(args.NodeId, ctx.Now())) {
+                ReportDiagnosticStatsInsertFailed(
+                    "Failed to insert access statistics into ranking");
```
qkrorlqr: this message seems to be the same across all calls, it doesn't give any extra info, I suggest to discard it and just call `ReportDiagnosticStatsInsertFailed()`

cloud/filestore/libs/storage/tablet/tablet_state.cpp
```diff
@@ -16,6 +16,7 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+
```
qkrorlqr: x

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -28,6 +28,7 @@
 #include <cloud/filestore/libs/storage/tablet/model/verify.h>
 #include <cloud/filestore/libs/storage/tablet/protos/tablet.pb.h>
 #include <cloud/filestore/private/api/protos/tablet.pb.h>
+#include <cloud/filestore/libs/storage/tablet/model/node_access_stats.h>
```
qkrorlqr: includes not ordered correctly

cloud/filestore/libs/storage/tablet/tablet_state_impl.h
```diff
@@ -61,6 +61,8 @@ struct TIndexTabletState::TImpl
 
     THandlesStats HandlesStats;
 
+    std::optional<TNodeAccessStatsTracker> AccessTracker;
```
qkrorlqr: let's just have `TNodeAccessStatsTracker AccessTracker` and let's move its setup from ctor to a separate `Reset()` func (since it requires some params which aren't available at the time when `TImpl` is constructed)

cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
```diff
@@ -32,6 +33,14 @@ bool TNodeAccessComparator::operator()(
 
 ////////////////////////////////////////////////////////////////////////////////
 
+TNodeAccessStatsTracker::TNodeAccessStatsTracker()
+    : Ranking(
+          0,
+          TNodeAccessComparator{TDuration::Minutes(0)},
+          TNodeAccessKeyExtractor{})
+    , HalfLife(TDuration::Minutes(0))
```
qkrorlqr, line 41: nit: `TDuration` is zero by default, no need to do explicit initialization in this case

qkrorlqr (approved):

yegorskii (approved):

PR #6671, opened 2026-08-03
qemu should notice the block device resize

faucct:
### Notes
Makes the attached disks notice size changes without need to reattach.
Referenced the unmerged vhost changes until they are merged.
Can also make endpoints refreshing happen in parallel if needed.

### DIFF
--- a/cloud/blockstore/libs/endpoints/service_endpoint.cpp
+++ b/cloud/blockstore/libs/endpoints/service_endpoint.cpp
@@ -3,13 +3,10 @@
 #include "endpoint_manager.h"
 
 #include <cloud/blockstore/libs/service/context.h>
-#include <cloud/blockstore/libs/service/service_method.h>
 #include <cloud/blockstore/libs/service/service.h>
 #include <cloud/storage/core/libs/common/error.h>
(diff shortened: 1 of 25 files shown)


cloud/blockstore/libs/endpoints/service_endpoint.cpp
```diff
@@ -83,6 +89,49 @@ class TEndpointService final
 #undef ENDPOINT_IMPLEMENT_METHOD
 
 private:
+    template <typename TMethod>
+    TFuture<typename TMethod::TResponse> ExecuteStorageRequest(
+        TCallContextPtr ctx,
+        std::shared_ptr<typename TMethod::TRequest> request)
+    {
+        if constexpr (std::is_same_v<TMethod, TBlockStoreResizeVolumeMethod>) {
```
qkrorlqr: this looks like a hack and it doesn't seem to be correct - this will refresh only the endpoints local to this blockstore-server instance whereas the `ResizeVolume` request can be received by any blockstore-server instance, not necessarily the one which hosts the endpoints that need to be refreshed

PR #6675, opened 2026-08-03
making TBusyIdleTimeCalculator actually thread-safe

qkrorlqr:
### Notes
_It's a follow-up PR for https://github.com/ydb-platform/nbs/pull/6673_

`TBusyIdleTimeCalculator` pretends to be thread-safe but actually it's not thread-safe. Example race:
1. T1 - `OnRequestStarted()`
2. T1 - starts `OnRequestCompleted()`, decrements `InflightCount`, sees a zero value, moves on to `FinishState()`, calls `MicroSeconds()`, gets preempted
3. T2 - `OnRequestStarted()` - sees a value of one, updates `State[state]` with a later value of `MicroSeconds()`
4. T1 - continues, swaps-out the new value and subtracts it from its (stale) `MicroSeconds()` result obtained at step 2 - gets a huge value because of unsigned integer overflow

We have seen this multiple times in our blockstore metrics but IIRC we have never found the bug. This seems to be the bug.

Fixing it by reimplementing `TBusyIdleTimeCalculator` via double CAS. Also injected `ITimer` into it and used it to write unit tests (this component didn't have unit tests). 

### Issue

### DIFF
--- a/cloud/storage/core/libs/diagnostics/busy_idle_calculator_ut.cpp
+++ b/cloud/storage/core/libs/diagnostics/busy_idle_calculator_ut.cpp
@@ -1,0 +1,135 @@
+#include "busy_idle_calculator.h"
+
+#include <library/cpp/testing/unittest/registar.h>
+
+#include <util/system/event.h>
+#include <util/thread/factory.h>
+
+namespace NCloud {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TTestTimer: ITimer
+{
+    std::atomic<ui64> T;
+
+    TInstant Now() override
+    {
+        Cdbg << "T=" << T.load() << Endl;
+        return TInstant::MicroSeconds(1 + T.fetch_add(1));
+    }
+
+    void Sleep(TDuration duration) override
+    {
+        Y_UNUSED(duration);
+
+        UNIT_ASSERT(false);
+    }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TFixture
+{
+    std::shared_ptr<TTestTimer> Timer = std::make_shared<TTestTimer>();
+    TBusyIdleTimeCalculatorAtomics Calc{Timer};
+
+    std::atomic<i64> Busy;
+    std::atomic<i64> Idle;
+
+    TFixture()
+    {
+        Calc.Register(&Busy, &Idle);
+    }
+
+};
+
+}   // namespace
+
+////////////////////////////////////////////////////////////////////////////////
+
+Y_UNIT_TEST_SUITE(TBusyIdleTimeCalculatorTest)
+{
+    Y_UNIT_TEST(BasicOps)
+    {
+        TFixture fx;
+        fx.Calc.OnRequestStarted();
+        fx.Calc.OnRequestCompleted();
+
+        UNIT_ASSERT_VALUES_EQUAL(1, fx.Busy.load());
+        UNIT_ASSERT_VALUES_EQUAL(1, fx.Idle.load());
+
+        fx.Calc.OnRequestStarted();
+
+        UNIT_ASSERT_VALUES_EQUAL(1, fx.Busy.load());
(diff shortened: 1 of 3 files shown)


cloud/storage/core/libs/diagnostics/busy_idle_calculator_ut.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TTestTimer: ITimer
+{
+    std::atomic<ui64> T;
+
+    TInstant Now() override
+    {
+        Cdbg << "T=" << T.load() << Endl;
```
debnatkh, line 20: Is this debug output intended to be there?

debnatkh (dismissed):

cloud/storage/core/libs/diagnostics/busy_idle_calculator_ut.cpp
qkrorlqr: yes

it won't be printed if `DBGOUT` env var is not set

debnatkh (approved):

agalibin (approved):

qkrorlqr:
Claude is satisfied with the diff:

I read through the full diff (3 files, +241/−30). Overall verdict: good fix, well-motivated, I'd approve with a few minor comments.

What the PR does: TBusyIdleTimeCalculator pretended to be thread-safe but had a real race: one thread completing a request could get preempted between reading the clock and swapping state, another thread could then start a request and store a later timestamp, and the first thread would subtract the newer value from its stale clock reading — producing a huge value via unsigned underflow. This matched anomalies seen repeatedly in blockstore metrics. The fix reimplements the calculator via a double-width CAS, injects ITimer, and adds unit tests (the component previously had none). 
[github](https://github.com/ydb-platform/nbs/pull/6675/changes)

Why the approach is sound. Packing {Gen, Inflight, Started} into one 16-byte std::atomic<TFields> and mutating it with a compare_exchange_weak loop makes each transition (idle→busy, busy→idle, stats flush) atomic as a whole, so the "stale clock vs. swapped state" race is structurally impossible. The intervals also telescope correctly: every successful transition sets Started = now and credits now - prevStarted, so no time is lost or double-counted even when a Now() call is wasted on a failed CAS. The Gen counter neatly closes the ABA window where Inflight and Started could cycle back to identical values between a load and the CAS. The three static_asserts (size 16, trivially copyable, is_always_lock_free) are exactly the right guardrails, since 16-byte lock-free atomics are platform/flag-dependent (cmpxchg16b on x86-64); a build that can't do it fails loudly instead of silently taking a lock. Memory ordering (acquire on load/failure, release on success) is appropriate.

Minor comments I'd leave on the PR:

Constructor: Fields.load() before the first store relies on std::atomic<T>'s default constructor value-initializing T, which is only guaranteed since C++20. If any consumer builds this header at C++17, that's a read of indeterminate Gen/Inflight. Simpler and unambiguous: Fields.store(TFields{.Started = Timer->Now().MicroSeconds()});
UpdateProgress doesn't bump Gen while the other two mutators do. It's safe today because Started always moves forward, but bumping Gen there too would make the invariant ("every successful CAS increments Gen") uniform and future-proof.
Test timer: the Cdbg << "T=" ... line in TTestTimer::Now() looks like leftover debug output — it fires on every clock read (I see a review thread was already resolved around it; if it's staying intentionally, fine).
Test coverage: the ThreadSafety test only runs two symmetric start/complete threads. A third thread hammering OnUpdateStats() concurrently would exercise the UpdateProgress CAS path under contention, which is where the subtlest interleavings live. Also note the final assertion (T - 1 == Busy + Idle) implicitly depends on the telescoping property absorbing wasted ticks from failed CASes — it holds, but a short comment explaining why would help future readers.
Perf note (non-blocking): the common non-transition case (Inflight > 0) went from a single atomic increment to a 16-byte load + CAS loop. For a metrics component this is almost certainly fine, but worth knowing it's on the request hot path.

None of these block merging (and indeed it's already merged). The core reasoning is correct and the bug analysis in the description is convincing.

Issue #6608, opened 2026-07-27
[Filestore] Capacity-based directory quotas


debnatkh:
The goal is to allow a filesystem user to create per-directory capacity quotas. Such quotas could allow users to select paths that will have an additional limit enforced on the number of files and/or the total bytes that can be stored under that path. The quota is enforced recursively, so all files and directories under the quota root are counted against the limit.

An example of a quota could be: "the directory `/home/user1` is limited to 100 GiB, while the filesystem itself has 1 TiB of total capacity". In this case, any write operation that would cause the total size of `/home/user1` to exceed 100 GiB would fail with an `ENOSPC` error, just as it would if the filesystem itself ran out of space.

debnatkh:
### User interface

1. Having a special file in the root of the filesystem (e.g. `/.quotas`) that can be read/written to manage quotas. This has one potential problem:

   1. It would require all vhosts to either be aware of this special file node_id and handle it specially, or the shard that handles this special file would have to be aware of its semantics and handle IO to it accordingly.

2. Using extended attributes (xattr) on the quota root directory to manage quotas. This will be problematic in two ways:

   1. Sometimes we disable xattr support for performance reasons. This happens by passing ENOTSUPP to the kernel when xattr is requested. This would make all the following xattr requests not leave the kernel and instead return ENOTSUPP.
   2. It would make listing quotas much harder since it would require either implementing a special xattr on a pre-defined directory (e.g. root), or scanning the entire filesystem for directories with a quota_id set

3. Having the ability to pass arbitrary text commands to filestore-vhost through ioctl. This would require a client to manage the quotas through sending ioctl commands to the vhost. We can proceed with a separate lightweight binary wrapper for the ioctl commands that is aware of the proper quota management interface that can be interpreted by the vhost.
4. Having full disk-manager-level support with quota management passed from disk-manager to the filestore

I believe that the **_third option_** works for the first release, and is the simplest to implement

### Usage scenario

Since there is no user-friendly option to call ioctl with arbitrary arguments from bash, a simple binary wrapper will be introduced to send the right ioctl requests and parse the responses. The following example shows a quota being created and attached to a directory:

```markdown
$ filestore-quotas create --path /mnt/nfs --max-bytes 10GiB
Created quota 42

$ filestore-quotas attach --path /mnt/nfs/projects/team-a --quota-id 42
Attached /mnt/nfs/projects/team-a to quota 42

$ filestore-quotas get --quota-id 42
QuotaId:   42
[...]

PR #6697, opened 2026-08-05
issue-6608: Add QuotaId attribute to TNode and implement setting it via SetNodeAttr

debnatkh:
### Notes
First tablet-side changes necessary for quotas: introducing the ability to mark qirectory's quota

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
@@ -125,8 +125,8 @@ bool TIndexTabletActor::PrepareTx_SetNodeAttr(
     }
 
     if (HasFlag(flags, NProto::TSetNodeAttrRequest::F_SET_ATTR_QUOTA_ID)) {
-        const auto currentQuotaId = args.Node->Attrs.GetQuotaId();
-        const auto newQuotaId = update.GetQuotaId();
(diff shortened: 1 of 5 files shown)


qkrorlqr (approved):

cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
```diff
@@ -108,14 +115,49 @@ bool TIndexTabletActor::PrepareTx_SetNodeAttr(
...
+        } else if (currentQuotaId != 0) {
+            // already attached to a *different* quota - no nested/
+            // multi-domain quotas
+            args.Error = ErrorInvalidArgument(
+                TStringBuilder()
+                << "node " << args.NodeId
+                << " is already attached to a different quota: "
+                << currentQuotaId);
+            return true;
+        } else if (args.Node->Attrs.GetType() != NProto::E_DIRECTORY_NODE) {
```
agalibin, line 143: here and below: extra 'else' because you have 'return' above

agalibin (approved):

cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
debnatkh: There is an initial `if (currentQuotaId == newQuotaId) {`guard, so removing else here will require keeping this condition enclosing all cases, which, I believe, hurts readability

PR #6698, opened 2026-08-05
[Blockstore] Advertise read-only feature for vhost devices

antonmyagkov:
issue: #6767


tpashkin (approved):

tpashkin:
Should probably do the same for NBD in the future

nisqatsi (approved):

PR #6721, opened 2026-08-07
issue-6608: add ability to set QuotaId for node using filestore-client

debnatkh:
### Notes
Also adding QuotaId to the TNodeAttr proto from `cloud/filestore/public/api/protos/node.proto` 

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
@@ -2172,6 +2172,60 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
         UNIT_ASSERT_VALUES_EQUAL(target, readLinkResponse->Record.GetSymLink());
     }
 
+    TABLET_TEST_4K_ONLY(ShouldSetAttrsViaUnsafeCreateNode)
+    {
+        TTestEnv env(testEnvConfig);
+
+        ui32 nodeIdx = env.AddDynamicNode();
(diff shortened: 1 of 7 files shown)


cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -2172,6 +2172,60 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
         UNIT_ASSERT_VALUES_EQUAL(target, readLinkResponse->Record.GetSymLink());
     }
 
+    TABLET_TEST_4K_ONLY(ShouldSetAttrsViaUnsafeCreateNode)
```
antonmyagkov, line 2175: Some  tests are missing like trying to modify the quota or adding quota to non-empty directory, etc.

antonmyagkov (changes requested):

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
debnatkh: This tests checks this change: https://github.com/ydb-platform/nbs/blob/users/debnatkh/issue-6608-client-set-quota/cloud/filestore/libs/storage/tablet/helpers.cpp#L156

Since ConverAttrToNode is only used in Unsafe* methods

And the tests for logic for enforcement of quotas on non-empty or non-directories were already introduced on the tablet level in this PR: https://github.com/ydb-platform/nbs/pull/6697

antonmyagkov (approved):

agalibin (approved):

PR #6722, opened 2026-08-07
fix ShouldNotFreezeSocketReadingDueToLimiter by implementing connection negotiation

librarian:
### What happened
Invalid NBS connection disconnected active client
```
2026-08-07T08:33:43.024097Z :BLOCKSTORE_NBD DEBUG: cloud/blockstore/libs/nbd/client_handler.cpp:233: [d:TestDiskId]  SEND Request type:6 handle:1093403632696175061 from:53248 length:4096
2026-08-07T08:33:43.257456Z :BLOCKSTORE_NBD DEBUG: cloud/blockstore/libs/nbd/server.cpp:367: new connection from [::1]:14851
2026-08-07T08:33:43.257727Z :BLOCKSTORE_NBD INFO: cloud/blockstore/libs/nbd/server.cpp:183: lost connection with client, failed to receive: (yexception) cloud/blockstore/libs/nbd/server_handler.cpp:313: Condition violated: `client.Flags == (NBD_FLAG_C_FIXED_NEWSTYLE | NBD_FLAG_C_NO_ZEROES)'
...
[[bad]]assertion failed at cloud/blockstore/libs/nbd/server_ut.cpp:1104, virtual void NCloud::NBlockStore::NBD::NTestSuiteTServerTest::TTestCaseShouldNotFreezeSocketReadingDueToLimiter::Execute_(NUnitTest::TTestContext &): (!HasError(future3.GetValue(TDuration::Seconds(3)))) [[rst]]
[[alt1]]NUnitTest::NPrivate::RaiseError(char const*, TBasicString<char, std::__y1::char_traits<char>> const&, bool)+136 (0x10C5398)
NCloud::NBlockStore::NBD::NTestSuiteTServerTest::TTestCaseShouldNotFreezeSocketReadingDueToLimiter::Execute_(NUnitTest::TTestContext&)+8138 (0xF596CA)
NCloud::NBlockStore::NBD::NTestSuiteTServerTest::TCurrentTest::Execute()::'lambda'()::operator()() const+71 (0xF5C567)
NUnitTest::TTestBase::Run(std::__y1::function<void ()>, TBasicString<char, std::__y1::char_traits<char>> const&, char const*, bool)+127 (0x10C734F)
NCloud::NBlockStore::NBD::NTestSuiteTServerTest::TCurrentTest::Execute()+429 (0xF5BF1D)
NUnitTest::TTestFactory::Execute()+780 (0x10C7AAC)
NUnitTest::RunMain(int, char**)+3005 (0x10DAA7D)
??+0 (0x75283DE29D90)
__libc_start_main+128 (0x75283DE29E40)
??+0 (0xF0E029)
[[rst]]
```
###  Issue
The server replaced its active connection immediately after accepting a new socket, before NBD negotiation completed. A malformed, incomplete, or unrelated connection could therefore terminate a valid client and cancel its in-flight requests with server is unavailable.
This caused ShouldNotFreezeSocketReadingDueToLimiter to fail when unexpected IPv6 connections reached the test endpoint. The limiter itself operated correctly; the active client had already been displaced before receiving its response.

### Fix

New sockets are now treated as candidate connections:

- The active connection remains usable during negotiation.
- Invalid candidates are discarded without affecting it.
- A candidate replaces the active connection only after successful NBD negotiation.
- Shutdown handles both active and candidate connections.
- Connection logs now report the peer address.

### DIFF
--- a/cloud/blockstore/libs/nbd/server.cpp
+++ b/cloud/blockstore/libs/nbd/server.cpp
@@ -72,6 +72,7 @@ private:
     TContLockFreeQueue<TServerResponsePtr> ResponseQueue;
 
     size_t InFlightBytes = 0;
+    std::atomic_bool ReportErrors = false;
     std::atomic_flag ShuttingDown = false;
 
 public:
@@ -105,6 +106,7 @@ public:
 
     void Stop() override
     {
+        ReportErrors.store(false, std::memory_order_release);
         ShutDown();
     }
 
@@ -185,7 +187,7 @@ private:
             if (!IsShuttingDown() && !c->Cancelled()) {
                 STORAGE_INFO("lost connection with client, failed to receive: "
                     << CurrentExceptionMessage());
-                Handler->ProcessException(std::current_exception());
+                ReportException(std::current_exception());
             }
         }
 
@@ -199,6 +201,7 @@ private:
         if (Handler->NegotiateClient(io, io) &&
             ConnectionNegotiatedHandler(this))
         {
+            ReportErrors.store(true, std::memory_order_release);
             Handler->ProcessRequests(this, io, io, c);
         }
     }
@@ -211,7 +214,7 @@ private:
         while (ResponseQueue.Dequeue(&response)) {
             if (!response) {
                 // stop signal received
-                Handler->ProcessException(
+                ReportException(
                     std::make_exception_ptr(TSystemError(-ESHUTDOWN)));
                 break;
             }
@@ -221,7 +224,7 @@ private:
             } catch (...) {
                 STORAGE_INFO("lost connection with client, failed to send: "
                     << CurrentExceptionMessage());
-                Handler->ProcessException(std::current_exception());
+                ReportException(std::current_exception());
             }
 
             ReleaseRequest(response->RequestBytes);
@@ -257,6 +260,15 @@ private:
         }
 
         return false;
+    }
+
+    void ReportException(std::exception_ptr e)
+    {
+        if (ReportErrors.load(std::memory_order_acquire) &&
+            !IsShuttingDown())
+        {
+            Handler->ProcessException(std::move(e));
+        }
     }
 
     void ShutDown()
--- a/cloud/blockstore/libs/nbd/server.cpp
+++ b/cloud/blockstore/libs/nbd/server.cpp
@@ -23,6 +23,7 @@
 #include <util/thread/singleton.h>
 
 #include <atomic>
+#include <functional>
 
 namespace NCloud::NBlockStore::NBD {
 
@@ -65,6 +66,7 @@ private:
     TContExecutor* Executor;
     ILimiterPtr Limiter;
     IServerHandlerPtr Handler;
+    std::function<bool(TConnection*)> ConnectionNegotiatedHandler;
     TSocketHolder Socket;
 
     TContLockFreeQueue<TServerResponsePtr> ResponseQueue;
@@ -78,12 +80,14 @@ public:
             TContExecutor* e,
             ILimiterPtr limiter,
             IServerHandlerPtr handler,
+            std::function<bool(TConnection*)> connectionNegotiatedHandler,
             TSocketHolder socket)
         : AppCtx(appCtx)
         , Log(appCtx.Log)
         , Executor(e)
         , Limiter(std::move(limiter))
         , Handler(std::move(handler))
+        , ConnectionNegotiatedHandler(std::move(connectionNegotiatedHandler))
         , Socket(std::move(socket))
         , ResponseQueue(e)
     {}
@@ -192,7 +196,9 @@ private:
     {
         TContIO io(Socket, c);
 
-        if (Handler->NegotiateClient(io, io)) {
+        if (Handler->NegotiateClient(io, io) &&
+            ConnectionNegotiatedHandler(this))
+        {
             Handler->ProcessRequests(this, io, io, c);
         }
     }
@@ -284,6 +290,7 @@ private:
 
     std::unique_ptr<TContListener> Listener;
     TConnectionPtr Connection;
+    TConnectionPtr CandidateConnection;
 
 public:
     TEndpoint(
@@ -337,6 +344,10 @@ public:
                 Connection->Stop();
             };
 
+            if (CandidateConnection) {
+                CandidateConnection->Stop();
+            }
+
             if (Listener) {
                 Listener->Stop();
             }
@@ -363,25 +374,43 @@ private:
     {
         TSocketHolder socket(accept.S->Release());
 
-        auto address = NAddr::GetSockAddr(socket);
+        auto address = NAddr::GetPeerAddr(socket);
         STORAGE_DEBUG("new connection from " << PrintHostAndPort(*address));
 
         if (IsTcpAddress(*address)) {
             SetNoDelay(socket, true);
         }
 
-        if (Connection) {
-            Connection->Stop();
+        if (CandidateConnection) {
+            CandidateConnection->Stop();
         }
 
-        Connection = MakeIntrusive<TConnection>(
+        CandidateConnection = MakeIntrusive<TConnection>(
             AppCtx,
             Executor,
             Limiter,
             HandlerFactory->CreateHandler(),
+            [this](TConnection* connection) {
+                return ActivateConnection(connection);
+            },
             std::move(socket));
 
-        Connection->Start();
+        CandidateConnection->Start();
+    }
+
+    bool ActivateConnection(TConnection* connection)
+    {
+        if (CandidateConnection.Get() != connection) {
+            return false;
+        }
+
+        if (Connection) {
+            Connection->Stop();
+        }
+
+        Connection = CandidateConnection;
+        CandidateConnection.Reset();
+        return true;
     }
 
     void OnError() override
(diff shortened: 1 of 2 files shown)


librarian:
Issues from Codex:

[P1] Candidate negotiation failures still restart the endpoint — [server.cpp (line 199)](https://github.com/ydb-platform/nbs/blob/users/librarian/fix-connection-negotiation/cloud/blockstore/libs/nbd/server.cpp#L199). Failed candidates call Handler->ProcessException; production’s shared error handler schedules an endpoint restart after 10 ms, closing the active connection. The new test uses a stub handler, so it misses this. Suppress error propagation until the candidate is activated.

Addressed here https://github.com/ydb-platform/nbs/pull/6722/changes/2edecfb3acb9d11506b4bef19eb18c193f7eb096

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -205,7 +214,7 @@ class TConnection final
         while (ResponseQueue.Dequeue(&response)) {
             if (!response) {
                 // stop signal received
-                Handler->ProcessException(
+                ReportException(
                     std::make_exception_ptr(TSystemError(-ESHUTDOWN)));
```
tpashkin, line 218: This is actually important to call the handler (which will stop/start nbd device) here, when client shuts down the socket

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -363,25 +386,43 @@ class TEndpoint final
...
         }
 
-        if (Connection) {
-            Connection->Stop();
+        if (CandidateConnection) {
+            CandidateConnection->Stop();
         }
 
-        Connection = MakeIntrusive<TConnection>(
+        CandidateConnection = MakeIntrusive<TConnection>(
```
tpashkin, line 400: what if multiple connections get through? wouldn't it be more appropriate to use a future (instead of a callback), which will be set by connection once negotiation has been completed and `WaitFor` it here, in listener's coroutine

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -65,11 +66,13 @@ class TConnection final
     TContExecutor* Executor;
     ILimiterPtr Limiter;
     IServerHandlerPtr Handler;
+    std::function<bool(TConnection*)> ConnectionNegotiatedHandler;
     TSocketHolder Socket;
 
     TContLockFreeQueue<TServerResponsePtr> ResponseQueue;
 
     size_t InFlightBytes = 0;
+    std::atomic_bool ReportErrors = false;
```
tpashkin, line 75: could you please explain the reasoning behind this flag. why wouldn't we call an exception handler, if connection is shutting down?

cloud/blockstore/libs/nbd/server.cpp
librarian: Agreed. The intent was to suppress failures only for a candidate that has not replaced the active connection, because ProcessException triggers an endpoint restart. An activated connection closing its socket must still call the handler. The ReportErrors/IsShuttingDown combination was meant to preserve that distinction, but it makes this path too subtle. I will split negotiation from request processing and keep error reporting for the active connection.

cloud/blockstore/libs/nbd/server.cpp
librarian: The callback version guards overlapping candidates by stopping the previous CandidateConnection and checking its identity again in ActivateConnection, so a stale negotiation cannot replace the active connection. Still, I agree that the lifetime and error-handling logic is unnecessarily complicated. Waiting for a negotiation future in the listener coroutine naturally serializes accepted connections and removes the stale-callback case, so I will rework it that way.

cloud/blockstore/libs/nbd/server.cpp
librarian: The flag distinguishes an unactivated candidate from the active connection. ProcessException is wired to the production error handler, which restarts the endpoint; reporting a candidate negotiation failure would therefore close the still-valid active connection. It is enabled only after promotion and cleared for an intentional Stop. A remote shutdown of an activated connection is still reported because the send loop observes the stop signal before shutting the connection down. That said, the state is subtle; separating negotiation from the request/send loops removes the need for this flag.

librarian: removed it

cloud/blockstore/libs/nbd/server.cpp
librarian: remade using futures

cloud/blockstore/libs/nbd/server.cpp
librarian: fixed

librarian:
Currently it should work like this (ai generated):

```mermaid
sequenceDiagram
    participant A as Active client A
    participant E as Endpoint
    participant B as Candidate B
    participant C as Candidate C

    A->>E: Backend request is in flight
    B->>E: Connect
    E->>B: Start independent negotiation
    B->>E: Valid NBD_OPT_GO
    E->>E: Select B as candidate
    E->>A: Stop reading and begin draining

    C->>E: Connect while A is draining
    E->>C: Start independent negotiation
    C->>E: Valid NBD_OPT_GO
    E->>E: Replace candidate B with C
    E->>B: Cancel and disconnect

    A-->>E: Last active request completes
    E->>C: Allow negotiation to finish
    C-->>E: Final GO response succeeds
    E->>E: Promote C to active connection
    E->>C: Start request processing
```

There is a catch though: if client A has some IO going on we will not be able to start B until this io finishes, and there is no timeout here or some other way to force connection/cancel running requests.

It was the same before, though.

I am not sure adding it would be good idea.

librarian:
# AI review report: PR #6722

Reviewed revision: `a03ff673f56f44be57c9b632440a0fef45ecbae1`

## Outcome

No blocking correctness issues found in the current revision. The change fixes the original connection-replacement race and addresses the previously raised concerns around invalid candidates, active-client error reporting, concurrent candidates, and endpoint shutdown.

## What changed

Previously, accepting any new socket stopped the active connection before the new peer completed NBD negotiation. A malformed or stalled peer could therefore disconnect a healthy client and cancel its in-flight requests.

The new implementation separates connection setup into these stages:

1. Every accepted socket negotiates in an independent coroutine, leaving the accept loop available.
2. A connection becomes a candidate only after reaching a valid `NBD_OPT_GO` request.
3. The latest valid candidate has exclusive ownership; older negotiating connections and candidates are cancelled.
4. The active connection stops accepting requests and drains its existing backend operations.
5. After draining, the candidate is allowed to finish its GO response.
6. The candidate is promoted to the active connection only if negotiation finishes successfully and it still owns candidate status.

The ownership checks before and after draining prevent stale setup callbacks from activating an obsolete connection.

## Correctness observations

- Invalid or incomplete negotiations do not disturb the active connection and are not reported as active-endpoint failures.
- Unexpected failure of an active connection still reaches `ProcessException`, preserving the existing endpoint restart behavior.
- A failure while writing the final GO response rejects the candidate instead of promoting it or restarting the endpoint.
- Endpoint shutdown cancels all negotiating connections and prevents their later promotion.
- Several clients may connect and negotiate concurrently, but only one connection can become active.
- Socket shutdown is idempotent and best-effort, so an already disconnected candidate cannot abort replacement.

## Follow-up concern

### Bound negotiation lifetime or pending connection count

Every accepted peer is retained in `NegotiatingConnections` and owns setup/negotiation coroutines until it disconnects, reaches `NBD_OPT_GO`, or the endpoint stops. If many peers connect and stall before GO, socket, coroutine, and vector usage can grow without a bound.

This does not invalidate the handoff fix, and a later valid candidate cancels the stalled peers. Still, a negotiation deadline or a maximum pending-connection count would make the endpoint robust against broken or hostile clients. Whether this is blocking depends on whether the TCP listener can be reached by untrusted peers.

## Known inherited limitation

Handoff and `StopEndpoint` still wait indefinitely if the active connection has a backend I/O future that never completes. The PR preserves the behavior from `main`: it stops new requests, waits for `ActiveRequests` to reach zero, and has no timeout or backend cancellation mechanism.

`ShouldKeepHandoffAndEndpointShutdownPendingWhileDraining` explicitly characterizes this behavior. A timeout/cancellation policy should be designed separately because forcibly activating the next client while old I/O remains in flight can violate ordering and exclusivity.

A candidate that disconnects while the active connection drains is normally detected when draining completes and the final GO response fails. A newer valid candidate or endpoint shutdown cancels it immediately.

## Test coverage

The focused suite covers:

- a peer stalled before GO while a later client connects;
- an invalid peer leaving the active connection intact;
- candidate disconnect during active-connection drain;
- shutdown while waiting for negotiation readiness and completion;
- failure while writing the final GO response;
- multiple candidates waiting for the same drain, with only the latest activating;
- preservation of active-client error reporting;
- the inherited indefinite-drain behavior.

Local verification at the reviewed revision:

```text
cloud/blockstore/libs/nbd: 35/35 tests passed
git diff --check: passed
```

The tests-only revision was red before the implementation fix: the stalled-before-GO and multiple-candidate tests failed. They pass after the implementation commit.

At review time, GitHub pre-commit checks had passed and the larger component test jobs were still running. `approval-check` was failing because the PR still required approval, not because of a build or test failure.

## Recommendation

Approve the current change, with negotiation resource bounds tracked as a follow-up if the listener is exposed beyond trusted local clients.

Issue #6628, opened 2026-07-29
[NBS] Fix flushing the same block twice


vladstepanyuk:
Now it is possible that we will try to flush the same fresh block `(BlockIndex, CommitId)` twice.

 I.e. during the first flush we write a record to the local database
`(BlockIndex, CommitId, BlobId1)`

and during the second flush we write a new line with a new BlobId under the same key `(BlockIndex, CommitId)`
`(BlockIndex, CommitId, BlobId2)`

by doing so, we rewrite the line from the first blob, but we don't mark it in the block mask or anywhere else that this block from the blob is no longer in the index, which leads to a leak because this block in the blob will never be rewritten.

 It's also possible that all the lines in the index for the blob will be rewritten, resulting in a completely leaked blob that is missing from the mixed index.

The main scenario for a double flush block is if the block was flushed, but the trim queue did not reach it:

 1. Flush blocks
 2. The tablet restarts, and there is no trim after the flush
 3. After the restart, flush again
 4. The first block is completely rewritten and leaked from the mixed index

PR #6739, opened 2026-08-10
[NBS] Persist flush commit ID in partition metadata

vladstepanyuk:
### Notes
Persist flush commit id in partition metadata to avoid flushing the same block twice and blob leakage 
### Issue
#6628

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
@@ -170,7 +170,7 @@ public:
         if (UseFlushCommitIdAsTrimFreshLogToCommitId &&
             Args.Mode == ADD_FLUSH_RESULT)
         {
-            const auto trimFreshLogToCommitId =
+            const ui64 trimFreshLogToCommitId =
                 Max(State.GetTrimFreshLogToCommitId(), Args.CommitId);
             State.AccessMeta().SetTrimFreshLogToCommitId(
                 trimFreshLogToCommitId);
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -82,13 +83,16 @@ class TAddBlobsExecutor
             TString diskId,
             ui64 deletionCommitId,
             ui32 maxBlocksInBlob,
+            const bool useFlushCommitIdAsTrimFreshLogToCommitId,
```
qkrorlqr: nit: `const` looks a bit weird here next to the other primitive arguments which don't have `const`, let's drop this `const`

cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -162,7 +166,17 @@ class TAddBlobsExecutor
...
 
         UpdateCompactionMap(db);
-        State.UpdateTrimFreshLogToCommitIdInMeta();
+
+        if (UseFlushCommitIdAsTrimFreshLogToCommitId &&
+            Args.Mode == ADD_FLUSH_RESULT)
+        {
+            const ui64 trimFreshLogToCommitId =
+                Max(State.GetTrimFreshLogToCommitId(), Args.CommitId);
+            State.AccessMeta().SetTrimFreshLogToCommitId(
```
qkrorlqr, line 181: looks pretty non-straightforward

I would expect waiting for inflight fresh write completion before flush to naturally cause trimming at least up to flush commit id because there would be no trim-fresh-log-barriers with commit-ids smaller than flush commit-id

so I don't understand why we might need to make any changes to this code

qkrorlqr (changes requested):

qkrorlqr (approved):

komarevtsev-d (approved):

PR #6741, opened 2026-08-10
[NBS] take ownership of source partition during disk migration, add UTs

agalibin:
### Notes
Two scenarios:
1. during disk migration, the old leader volume could terminate while its source partition actor remained alive because the follower wrapper did not own it. This could leave a stale partition serving I/O after cutover;
2. during disk migration shutdown, TVolumeAsPartitionActor could wait indefinitely for an in-flight follower request whose response never arrived. This prevents PoisonTaken propagation and blocks graceful shutdown of the leader volume.

The change in commit 1 transfers source partition ownership to the follower wrapper, ensuring the entire actor tree is stopped.
The change in commit 2 bounds request draining by DestroyVolumeTimeout, cancels stuck requests, and allows the migration actor tree to terminate.

### DIFF
--- a/cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
+++ b/cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
@@ -556,7 +556,7 @@ void TVolumeAsPartitionActor::CancelRequests(const TActorContext& ctx)
 
         const auto& requestCtx = request->Value;
         auto error = MakeError(
-            E_CANCELLED,
+            E_REJECTED,
             "Follower request timed out during shutdown");
 
         switch (requestCtx.RequestType) {
--- a/cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
+++ b/cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
@@ -13,6 +13,8 @@
 #include <contrib/ydb/library/actors/core/hfunc.h>
 #include <contrib/ydb/library/actors/core/log.h>
 
+#include <util/generic/vector.h>
+
 namespace NCloud::NBlockStore::NStorage {
 
 using namespace NActors;
@@ -22,10 +24,12 @@ using namespace NActors;
 TVolumeAsPartitionActor::TVolumeAsPartitionActor(
         TChildLogTitle logTitle,
         ui32 originalBlockSize,
-        TString diskId)
+        TString diskId,
+        TDuration shutdownTimeout)
     : LogTitle(std::move(logTitle))
     , OriginalBlockSize(originalBlockSize)
     , DiskId(std::move(diskId))
+    , ShutdownTimeout(shutdownTimeout)
 {}
 
 TVolumeAsPartitionActor::~TVolumeAsPartitionActor() = default;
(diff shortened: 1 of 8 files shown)


cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
```diff
@@ -517,9 +524,88 @@ void TVolumeAsPartitionActor::HandlePoisonPill(
...
+
+    for (const ui64 requestId: requestIds) {
+        auto request = RequestsInProgress.ExtractRequest(requestId);
+        Y_DEBUG_ABORT_UNLESS(request);
+        if (!request) {
+            continue;
+        }
+
+        const auto& requestCtx = request->Value;
+        auto error = MakeError(
```
drbasic, line 558: разве мы должны отдавать E_CANCELLED? Клиент же живой и ждет выполнения запроса. Это партишен по какой-то причине кувыркнулся, значит нужно E_REJECTED.

agalibin: согласен

cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
```diff
@@ -517,9 +524,88 @@ void TVolumeAsPartitionActor::HandlePoisonPill(
...
+            "Follower request timed out during shutdown");
+
+        switch (requestCtx.RequestType) {
+            case ERequestType::WriteBlocks:
+                ctx.Send(
+                    requestCtx.OriginalSender,
+                    std::make_unique<TEvService::TEvWriteBlocksResponse>(
+                        std::move(error)),
+                    0,   // flags
+                    requestCtx.OriginalCookie);
```
SvartMetal, line 569: Can we use NCloud::Reply here?

And maybe we should change TRequestCtx to TRequestInfo with OriginalSender -> Sender and OriginalCookie renamed to Sender and Cookie

cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
```diff
@@ -517,9 +524,88 @@ void TVolumeAsPartitionActor::HandlePoisonPill(
...
+    for (const ui64 requestId: requestIds) {
+        auto request = RequestsInProgress.ExtractRequest(requestId);
+        Y_DEBUG_ABORT_UNLESS(request);
+        if (!request) {
+            continue;
+        }
+
+        const auto& requestCtx = request->Value;
+        auto error = MakeError(
+            E_REJECTED,
```
SvartMetal, line 559: Can we add a comment that it is important to have retriable error code here?

SvartMetal (approved):

drbasic (approved):

cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
agalibin: will added in next PR

cloud/blockstore/libs/storage/volume/actors/volume_as_partition_actor.cpp
agalibin: next PR

PR #6674, opened 2026-08-03
issue-6208: [Filestore] Track latency in the filesystem (tablet changes)

szumal:
### Notes
This PR adds latency tracking to the filesystem. It is a follow up to https://github.com/ydb-platform/nbs/pull/6691, which introduced client + proto changes. Here, I am introducing collecting stats in the tablet.

### Issue
https://github.com/ydb-platform/nbs/issues/6208

>Per-file: most loaded files per shard and slowest (needs to be collected in tablet code) -> output for the whole filesystem

PR #6749, opened 2026-08-10
issue-6208: [Filestore] Track latency in the filesystem (datastructures model)

szumal:
### Notes
This is a split from https://github.com/ydb-platform/nbs/pull/6674, applying only the model changes

### Issue

### DIFF
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -102,8 +102,7 @@ namespace NCloud::NFileStore{
     xxx(ListNodesInternalFailedToAddNodeRef)                                   \
     xxx(InMemoryIndexStateNotInitialized)                                      \
     xxx(WriteDataRequestWithBufferAndPayload)                                  \
-    xxx(DiagnosticStatsInsertFailed)                                           \
-    // FILESTORE_IMPOSSIBLE_EVENTS
+// FILESTORE_IMPOSSIBLE_EVENTS
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -102,7 +102,8 @@ namespace NCloud::NFileStore{
     xxx(ListNodesInternalFailedToAddNodeRef)                                   \
     xxx(InMemoryIndexStateNotInitialized)                                      \
     xxx(WriteDataRequestWithBufferAndPayload)                                  \
-// FILESTORE_IMPOSSIBLE_EVENTS
+    xxx(DiagnosticStatsInsertFailed)                                           \
+    // FILESTORE_IMPOSSIBLE_EVENTS
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
@@ -1,4 +1,5 @@
 #include "node_access_stats.h"
+#include <cloud/filestore/libs/diagnostics/critical_events.h>
 
 namespace NCloud::NFileStore::NStorage {
 
@@ -56,8 +57,9 @@ void TNodeAccessStatsTracker::RequestStarted(ui64 nodeId, TInstant now)
     ++stats.RequestCount;
     stats.LastAccessed = now;
 
-    Ranking.InsertOrUpdate(std::move(stats));
-}
+    if(!Ranking.InsertOrUpdate(std::move(stats))) {
+        ReportDiagnosticStatsInsertFailed("Failed to insert access statistics into ranking");
+    };}
 
 TVector<TNodeAccessStats> TNodeAccessStatsTracker::GetStats(
     TInstant now,
--- a/cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
@@ -1,6 +1,4 @@
 #include "node_access_stats.h"
-#include <cloud/filestore/libs/diagnostics/critical_events.h>
-
 namespace NCloud::NFileStore::NStorage {
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -44,7 +42,7 @@ TNodeAccessStatsTracker::TNodeAccessStatsTracker(
     , HalfLife(halfLife)
 {}
 
-void TNodeAccessStatsTracker::RequestStarted(ui64 nodeId, TInstant now)
+bool TNodeAccessStatsTracker::UpdateAccessStats(ui64 nodeId, TInstant now)
 {
     TNodeAccessStats stats;
 
@@ -57,9 +55,8 @@ void TNodeAccessStatsTracker::RequestStarted(ui64 nodeId, TInstant now)
     ++stats.RequestCount;
     stats.LastAccessed = now;
 
-    if(!Ranking.InsertOrUpdate(std::move(stats))) {
-        ReportDiagnosticStatsInsertFailed("Failed to insert access statistics into ranking");
-    };}
+    return Ranking.InsertOrUpdate(std::move(stats));
+}
 
 TVector<TNodeAccessStats> TNodeAccessStatsTracker::GetStats(
     TInstant now,
--- a/cloud/filestore/libs/storage/tablet/model/node_access_stats_ut.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/node_access_stats_ut.cpp
@@ -14,7 +14,7 @@ Y_UNIT_TEST_SUITE(TNodeAccessStatsTrackerTest)
 
         const auto start = TInstant::Hours(1);
 
-        tracker.UpdateAccessStats(1, start);
+        UNIT_ASSERT(tracker.UpdateAccessStats(1, start));
 
         const auto stats = tracker.GetStats(start, 3);
 
@@ -31,8 +31,8 @@ Y_UNIT_TEST_SUITE(TNodeAccessStatsTrackerTest)
 
         const auto start = TInstant::Hours(1);
 
-        tracker.UpdateAccessStats(1, start);
-        tracker.UpdateAccessStats(1, start);
+        UNIT_ASSERT(tracker.UpdateAccessStats(1, start));
+        UNIT_ASSERT(tracker.UpdateAccessStats(1, start));
 
         const auto stats = tracker.GetStats(start, 3);
 
@@ -56,10 +56,10 @@ Y_UNIT_TEST_SUITE(TNodeAccessStatsTrackerTest)
         const auto old = start - TDuration::Minutes(10);
 
         for (ui32 i = 0; i < 10; ++i) {
-            tracker.UpdateAccessStats(1, old);
+            UNIT_ASSERT(tracker.UpdateAccessStats(1, old));
         }
         for (ui32 i = 0; i < 9; ++i) {
-            tracker.UpdateAccessStats(2, start);
+            UNIT_ASSERT(tracker.UpdateAccessStats(2, start));
         }
 
         const auto stats = tracker.GetStats(start, 5);
@@ -76,8 +76,8 @@ Y_UNIT_TEST_SUITE(TNodeAccessStatsTrackerTest)
 
         const auto start = TInstant::Hours(1);
 
-        tracker.UpdateAccessStats(1, start);
-        tracker.UpdateAccessStats(2, start);
+        UNIT_ASSERT(tracker.UpdateAccessStats(1, start));
+        UNIT_ASSERT(tracker.UpdateAccessStats(2, start));
 
         const auto stats = tracker.GetStats(start, 5);
 
@@ -93,11 +93,12 @@ Y_UNIT_TEST_SUITE(TNodeAccessStatsTrackerTest)
 
         const auto start = TInstant::Hours(1);
 
-        tracker.UpdateAccessStats(1, start - TDuration::Minutes(10));
-        tracker.UpdateAccessStats(2, start);
-        tracker.UpdateAccessStats(3, start);
-        tracker.UpdateAccessStats(4, start);
-        tracker.UpdateAccessStats(4, start);
+        UNIT_ASSERT(
+            tracker.UpdateAccessStats(1, start - TDuration::Minutes(10)));
+        UNIT_ASSERT(tracker.UpdateAccessStats(2, start));
+        UNIT_ASSERT(tracker.UpdateAccessStats(3, start));
+        UNIT_ASSERT(tracker.UpdateAccessStats(4, start));
+        UNIT_ASSERT(tracker.UpdateAccessStats(4, start));
 
         const auto stats = tracker.GetStats(start, 5);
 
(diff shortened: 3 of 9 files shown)


cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,59 @@
+#include <cloud/filestore/libs/storage/tablet/model/node_latency_stats.h>
+
+#include <util/datetime/base.h>
+#include <util/generic/hash_set.h>
+#include <util/generic/vector.h>
+
+namespace NCloud::NFileStore::NStorage {
+
```
qkrorlqr, line 7: nit: /////... before declarations/definitions

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,59 @@
...
+double TNodeLatencyStatsTracker::CalculateLatencyDecay(
+    const TNodeLatencyStats& stats,
+    TInstant now)
+{
+    const auto elapsed = now >= stats.LastAccessed ? now - stats.LastAccessed
+                                                   : TDuration::Zero();
+    return stats.AverageLatencyDecayedMs *
+           exp(-log(2) * elapsed.MilliSeconds() /
+               TDuration::Minutes(10).MilliSeconds());
```
qkrorlqr: things like `TDuration::Minutes(10)` should be configurable from the outside, in this case - configurable via ctor params

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
...
+            const double lhsScore = CalculateLatencyDecay(lhs, comparisonTime);
+            const double rhsScore = CalculateLatencyDecay(rhs, comparisonTime);
+
+            // AverageLatencyDecayedMs ASC, NodeId ASC
+            return std::tie(lhsScore, lhs.NodeId) <
+                   std::tie(rhsScore, rhs.NodeId);
+        }
+    };
+
+    using LatencyKey = std::pair<ui64, EFileStoreRequest>;
```
qkrorlqr: nit: TLatencyKey

and it's better to avoid std::pair / std::tuple wherever possible in favor of explicit data structures with properly named fields - it's very hard to read code with heavy `first` / `second` usage

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
...
+    size_t MaxEntries = 0;
+    using LatencyRanking = TSet<TNodeLatencyStats, TNodeLatencyStatsComparator>;
```
qkrorlqr: nit: TLatencyRanking

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
...
+    THashMap<LatencyKey, LatencyRanking::iterator> IdAndRequest2Stats;
```
qkrorlqr: nit: `Key2Stats` / `Lat2Stats`?

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
...
+        EFileStoreRequest requestType,
+        TInstant now,
+        TDuration latency);
+    static double CalculateLatencyDecay(
+        const TNodeLatencyStats& stats,
+        TInstant now);
+
+    TVector<TNodeLatencyStats> GetLatencyStats(TInstant now) const
+    {
+        TVector<TNodeLatencyStats> result;
```
qkrorlqr: doesn't have to be defined in the header

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
...
+    void UpdateLatencyStats(
+        ui64 nodeId,
+        EFileStoreRequest requestType,
+        TInstant now,
+        TDuration latency);
+    static double CalculateLatencyDecay(
+        const TNodeLatencyStats& stats,
+        TInstant now);
+
+    TVector<TNodeLatencyStats> GetLatencyStats(TInstant now) const
```
qkrorlqr: probably we don't need the whole list in the response - I assume that you'd only need a small top from each shard, e.g. 10 elements or so

so adding a parameter like `ui32 n` and returning just top-n items seems reasonable

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,59 @@
+#include <cloud/filestore/libs/storage/tablet/model/node_latency_stats.h>
+
+#include <util/datetime/base.h>
+#include <util/generic/hash_set.h>
```
qkrorlqr: I don't see hashset usage anywhere

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
+#pragma once
+
+#include <cloud/filestore/libs/service/request.h>
+
+#include <util/datetime/base.h>
+#include <util/digest/multi.h>
+#include <util/generic/hash.h>
+#include <util/generic/hash_set.h>
```
qkrorlqr: I don't see hashset usage anywhere

and some of the other `include`s are unneeded as well

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,59 @@
...
+        NodeLatencyStats.erase(it->second);
+    } else {
+        stats.NodeId = nodeId;
+        stats.RequestType = requestType;
+    }
+
+    stats.AverageLatencyDecayedMs = CalculateLatencyDecay(stats, now);
+
+    ++stats.RequestCount;
+    stats.TotalLatencyMs += latency.MilliSeconds();
```
qkrorlqr: `MilliSeconds()` is too coarse - most of the operations inside the tablet take less that 1ms to complete

you should use `MicroSeconds()`

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,59 @@
...
+    stats.AverageLatencyDecayedMs =
+        static_cast<double>(stats.TotalLatencyMs) / stats.RequestCount;
+    stats.LastAccessed = now;
+
+    auto [newLatencyIterator, inserted] = NodeLatencyStats.insert(stats);
+    Y_ABORT_UNLESS(inserted);
+    IdAndRequest2Stats[key] = newLatencyIterator;
```
qkrorlqr: why not `it->second = newLatencyIterator`?

cloud/filestore/libs/storage/tablet/model/node_latency_stats_ut.cpp
```diff
@@ -0,0 +1,177 @@
...
+        UNIT_ASSERT_VALUES_EQUAL(now, stats[0].LastAccessed);
+        UNIT_ASSERT_VALUES_EQUAL(now, stats[1].LastAccessed);
+    }
+
+    Y_UNIT_TEST(DecayCalculation)
+    {
+        TNodeLatencyStatsTracker tracker;
+        tracker.Initialize(1);
+
+        const auto now = TInstant::Now();
```
qkrorlqr: please don't use wall clock timer in tests

cloud/filestore/libs/storage/tablet/model/node_latency_stats.h
```diff
@@ -0,0 +1,87 @@
...
+        bool operator()(
+            const TNodeLatencyStats& lhs,
+            const TNodeLatencyStats& rhs) const
+        {
+            const auto comparisonTime = Max(lhs.LastAccessed, rhs.LastAccessed);
+            const double lhsScore = CalculateLatencyDecay(lhs, comparisonTime);
+            const double rhsScore = CalculateLatencyDecay(rhs, comparisonTime);
+
+            // AverageLatencyDecayedMs ASC, NodeId ASC
+            return std::tie(lhsScore, lhs.NodeId) <
```
szumal: CODEX REVIEW
[P1] Ranking comparator can abort on equal scores — node_latency_stats.h:29-40

TNodeLatencyStatsComparator compares only (decayed latency, NodeId), while entries are keyed by (NodeId, RequestType). Two request types for the same node with equal latency are therefore equivalent in TSet; the second insertion fails and Y_ABORT_UNLESS(inserted) terminates the process.
Include RequestType (or another unique key component) in the comparator, and add a regression test with equal-latency request types for one node.

Existing reviewer comments also identify configurable decay, millisecond precision, wall-clock-dependent tests, and unnecessary includes. CI’s Filestore workload passed, but it does not cover this comparator collision case.

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
szumal: true, no need to rehash (edit: if we now that the key exists)

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,83 @@
...
+    TNodeLatencyStats stats;
+
+    if (const auto* oldStats = Ranking.Find(key)) {
+        stats = *oldStats;
+    }
+    stats.NodeId = nodeId;
+    stats.RequestType = requestType;
+    ++stats.RequestCount;
+    stats.TotalLatencyUs += latency.MicroSeconds();
+    stats.AverageLatencyDecayedUs =
```
szumal, line 62: [P1] Existing latency history is not decayed before incorporating a new sample.
[node_latency_stats.cpp:68](https://github.com/ydb-platform/nbs/blob/e89aa27a747251abf9ecbb625934fdbe86e44263/cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp#L68-L75) averages the lifetime TotalLatencyUs, then resets LastAccessed to now. After a long idle period, old samples therefore retain full weight instead of decaying. For example, an old 1000 µs request followed much later by a 1 µs request produces a 500.5 µs average. The update should first decay the prior aggregate to now, then incorporate the new sample (similarly to TNodeAccessStatsTracker::RequestStarted).

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+double TNodeLatencyStatsTracker::CalculateLatencyDecay(
+    const TNodeLatencyStats& stats,
+    TInstant now,
+    TDuration halfLife)
+{
+    const auto elapsed = now - stats.LastAccessed;
+
+    return stats.AverageLatencyDecayedUs *
```
szumal, line 42: [P2] Millisecond conversion loses precision for sub-millisecond operations.
[node_latency_stats.cpp:47](https://github.com/ydb-platform/nbs/blob/e89aa27a747251abf9ecbb625934fdbe86e44263/cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp#L47-L52) uses MilliSeconds(), truncating elapsed durations below 1 ms to zero. This was also raised in an existing review comment; the current head still contains it. Use microseconds consistently.

szumal (commented):
CODEX REVIEW

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
szumal: This is a design decision, where the `AverageLatencyDecayedUs` represents the avg latency decayed since the last usage. An alternative would be to rename this field to smth like `LatencyDecayedUs` and make it a cumulative result where at each usage the previous value is decayed.

cloud/filestore/libs/storage/tablet/model/node_latency_stats.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+    stats.NodeId = nodeId;
+    stats.RequestType = requestType;
+    ++stats.RequestCount;
+    stats.TotalLatencyUs += latency.MicroSeconds();
+    stats.AverageLatencyDecayedUs =
+        static_cast<double>(stats.TotalLatencyUs) / stats.RequestCount;
+    stats.LastAccessed = now;
+
+    Ranking.InsertOrUpdate(std::move(stats));
```
qkrorlqr: seems that you're ignoring return value here

and afair in the previous pr as well

cloud/filestore/libs/diagnostics/critical_events.h
```diff
@@ -102,7 +102,8 @@ namespace NCloud::NFileStore{
     xxx(ListNodesInternalFailedToAddNodeRef)                                   \
     xxx(InMemoryIndexStateNotInitialized)                                      \
     xxx(WriteDataRequestWithBufferAndPayload)                                  \
-// FILESTORE_IMPOSSIBLE_EVENTS
+    xxx(DiagnosticStatsInsertFailed)                                           \
+    // FILESTORE_IMPOSSIBLE_EVENTS
```
qkrorlqr: nit: unneeded indent (we might change this indent if we decide to reformat __the whole file__, but not now)

cloud/filestore/libs/storage/tablet/model/node_access_stats_ut.cpp
```diff
@@ -14,7 +14,7 @@ Y_UNIT_TEST_SUITE(TNodeAccessStatsTrackerTest)
 
         const auto start = TInstant::Hours(1);
 
-        tracker.RequestStarted(1, start);
+        tracker.UpdateAccessStats(1, start);
```
qkrorlqr: probably makes sense to do `UNIT_ASSERT(tracker.UpdateAccessStats(...))` now

cloud/filestore/libs/diagnostics/critical_events.h
```diff
@@ -102,7 +102,8 @@ namespace NCloud::NFileStore{
     xxx(ListNodesInternalFailedToAddNodeRef)                                   \
     xxx(InMemoryIndexStateNotInitialized)                                      \
     xxx(WriteDataRequestWithBufferAndPayload)                                  \
-// FILESTORE_IMPOSSIBLE_EVENTS
+    xxx(DiagnosticStatsInsertFailed)                                           \
```
qkrorlqr: actually this seems to be unused in this PR, let's remove it and add it back in the next PR where you'll integrate this data structure with the tablet

qkrorlqr (approved):

cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
```diff
@@ -1,5 +1,4 @@
 #include "node_access_stats.h"
-
```
agalibin, line 2: why was it deleted?

agalibin (approved):

cloud/filestore/libs/storage/tablet/model/node_access_stats.cpp
szumal: by mistake, I'll make sure the file is formatted in the following PRs

PR #6758, opened 2026-08-11
Add additional external file mappings param to the config-generator

debnatkh:
### Notes
This way configs can be st up in the following way:

before:
```yaml
    additionalFiles: [ 'nbs-names.txt', 'nbs-domains.txt' ]
    additionalFilesPath: '/extra_configs/clustername'
```

after:
```yaml
    additionalFileMappings:
      /extra_configs/clustername/names.txt: nbs-names.txt
      /extra_configs/clustername/domains.txt: nbs-domains.txt
```

### Issue
Simple change for mostly internall purposes – no issue

### DIFF
--- a/cloud/storage/core/tools/common/go/configurator/configurator.go
+++ b/cloud/storage/core/tools/common/go/configurator/configurator.go
@@ -47,6 +47,7 @@ type ClusterSpec struct {
 	ZoneCfgOverride                 map[string]CfgFileOverride `yaml:"zoneCfgOverride"`
 	ZoneCfgOverrideSeed             map[string]CfgFileOverride `yaml:"zoneCfgOverrideSeed"`
 	AdditionalFiles                 []string                   `yaml:"additionalFiles"`
+	AdditionalFileMappings          map[string]string          `yaml:"additionalFileMappings"`
 	AdditionalFilesPath             string                     `yaml:"additionalFilesPath"`
 	AdditionalFilesPathTargetPrefix string                     `yaml:"additionalFilesPathTargetPrefix"`
 	TargetForSeed                   string                     `yaml:"targetForSeed"`
@@ -307,7 +308,7 @@ func (g *ConfigGenerator) dumpConfigs(
 
 	clusterConfig := g.spec.ServiceSpec.Clusters[cluster]
 
-	for _, fileName := range clusterConfig.AdditionalFiles {
+	appendAdditionalFile := func(fileName string, resultFileName string) error {
 		var filePath string
 		if strings.HasPrefix(fileName, "/") {
 			filePath = path.Join(g.spec.ArcadiaPath, fileName)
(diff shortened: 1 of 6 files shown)


dvrazumov (dismissed):

debnatkh:
I found the diff (via the Files changed tab — GitHub blocked the raw `.diff` URL, but the rendered file view has everything). Here's the review:

## Summary
Small, focused change to `cloud/storage/core/tools/common/go/configurator/configurator.go`: adds a new `additionalFileMappings` YAML field (`map[string]string`) alongside the existing `additionalFiles` list, letting a config author map an arbitrary source file path to an arbitrary output filename, instead of being stuck with the source file's basename. Well-covered by an updated golden test and a new spec.yaml case.

## What it does correctly
- **Backward compatible**: the old `AdditionalFiles` path is refactored into the same `appendAdditionalFile(fileName, resultFileName)` closure but still called with `filepath.Base(fileName)`, so existing configs behave identically.
- **Semantics match the description**: map key = source file path, map value = desired output filename — confirmed by the test fixture (`/extra_configs/nbs-names.txt: 'nfs-names.txt'` → golden output `.../nfs-names.txt`).
- **Test coverage**: adds a real file on disk, wires it into `test_service/spec.yaml`, and updates the golden output — good practice for this kind of generator.
- CI passed, one approval already.

## Things worth raising before merge

1. **Unexplained reverse sort.**
```go
mappedFiles := maps.Keys(clusterConfig.AdditionalFileMappings)
sort.Sort(sort.Reverse(sort.StringSlice(mappedFiles)))
```
Sorting is needed since Go map iteration is unordered, but why *reverse* alphabetical rather than ascending? If it's arbitrary, ascending is more conventional/readable. If it's deliberate (e.g., to control which entry "wins" on a filename collision), that deserves a one-line comment — as written it looks like it could be an accidental leftover from debugging.

2. **No collision handling between the two mechanisms.** If an entry in `AdditionalFiles` and an entry in `AdditionalFileMappings` (or two mapping entries) resolve to the same `resultFileName`, both get appended to `resultConfigs` with no dedup/overwrite check. Worth confirming downstream consumers tolerate duplicate-named entries, or add a guard/error.

3. **Doc comment on the new struct field would help.** `AdditionalFileMappings map[string]string` isn't self-explanatory about which side is source vs. destination (it's the reverse of what you might guess from the name alone — key is the *source* path). A short comment next to the field would save the next reader some diff-archaeology.

4. **Minor**: the only mapping test case has a single entry, so the sort-order logic (reverse or not) isn't actually exercised for correctness — a second mapping in the test fixture would give real coverage of ordering behavior.

None of these are blockers — the change is small and the existing test passes — but I'd flag #1 and #2 to the author before merging.

---

1. Intentionally done for some inner-working purposes. So that our current extra configs won't get reordered 
2. fixed
3. no need in documenting
4. fixed

dvrazumov (approved):

gy2411 (approved):

PR #6759, opened 2026-08-11
refactor grpc tls settings to prepare for automatic certs update

yegorskii:
### Notes

- Add static TLS providers for Disk Manager gRPC clients and server.

- Load certificates during application startup without automatic certificate refresh.

- Report Root CA fingerprints through the existing nbs_client and nfs_client metric registries.

- Report server certificate expireTs using the earliest NotAfter value from the entire certificate chain.

- Calculate the existing certificateValidity metric using the same earliest chain expiration and refresh the gauge every 24 hours.

- Share a single TLS provider between NFS control-plane and data-plane factories.

- Add provider-aware client transport credentials shared by the NBS and Filestore Go SDKs.

- Request a TLS config snapshot from the client provider for every new connection.

- Use tls.Config.GetCertificate on the server so new handshakes read the current provider state.

- Store server certificates and expiration metadata as synchronized provider state, preparing for automatic certificate refresh in a follow-up PR.

### Issue


BarkovBG (dismissed):

yegorskii:
@codex review

yegorskii:
https://claude.ai/share/591f4776-6eb5-419a-8bdc-ba3a047ebf6d
The following review comments were intentionally not addressed:
**certificateValidity semantics changed despite the "preserved" claim (medium)**
This is intentional. We previously agreed to inspect the entire certificate chain and use the earliest NotAfter. This allows alerts to detect an expiring intermediate certificate, not only the leaf. The behavior is now explicitly documented in the PR summary.
Inconsistent error types in readServerCertificate
LoadX509KeyPair failures remain non-retriable, while certificate-chain parsing failures remain regular errors. This preserves the behavior of the old controlplane.go, where key-pair loading errors were non-retriable but x509.ParseCertificate errors were returned directly.
**The Root CA fingerprint is calculated from the raw PEM instead of parsed DER**
This matches the existing NBS/Filestore implementation:
CityHash64(rootCa)
The metric identifies the contents of the CA bundle file. Therefore, changing whitespace or certificate ordering intentionally changes the fingerprint.
**Setting both TLSProvider and certificate file fields is not rejected**
Instead, precedence is explicitly documented: TLSProvider takes priority over RootCertsFile, CertFile, and CertPrivateKeyFile. This keeps the public SDK flexible and avoids introducing additional runtime validation. Disk Manager itself no longer sets both sources simultaneously.

cloud/disk_manager/internal/pkg/common/grpc_client_tls_provider.go
```diff
@@ -0,0 +1,67 @@
...
+
+type GRPCClientTLSProvider struct {
+	mutex     sync.RWMutex
+	tlsConfig *tls.Config
+}
+
+func NewGRPCClientTLSProvider(
+	config GRPCClientTLSProviderConfig,
+	registry metrics.Registry,
+) (*GRPCClientTLSProvider, error) {
```
SvartMetal: \n

we usually add \n after multi-line function signature

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
```diff
@@ -0,0 +1,199 @@
...
+	mutex        sync.RWMutex
+	certificates []tls.Certificate
+	expirations  []certificateExpiration
+}
+
+func NewGRPCServerTLSProvider(
+	ctx context.Context,
+	certs []GRPCServerCertificateConfig,
+	registry metrics.Registry,
+) (*GRPCServerTLSProvider, error) {
```
SvartMetal: \n

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
```diff
@@ -0,0 +1,199 @@
...
+			after: expiration,
+		})
+	}
+
+	for i := range expirations {
+		expiration := &expirations[i]
+		certRegistry := registry.WithTags(
+			map[string]string{"path": expiration.path},
+		)
+		expiration.expireTsGauge = certRegistry.Gauge("ExpireTs")
```
SvartMetal: expireTs?

for consistency with other metrics in DM (like certificateValidity)

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
```diff
@@ -0,0 +1,199 @@
...
+	cfg := &tls.Config{
+		GetCertificate: p.getCertificate,
+		MinVersion:     tls.VersionTLS12,
+	}
+	return grpc_credentials.NewTLS(cfg)
+}
+
+func (p *GRPCServerTLSProvider) getCertificate(
+	info *tls.ClientHelloInfo,
+) (*tls.Certificate, error) {
```
SvartMetal, line 145: \n

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	certificate := p.certificates[0]
+	return &certificate, nil
+}
+
+func readServerCertificate(
+	cert GRPCServerCertificateConfig,
+) (tls.Certificate, time.Time, error) {
```
SvartMetal, line 168: \n

cloud/disk_manager/internal/pkg/common/grpc_tls_provider_test.go
```diff
@@ -0,0 +1,199 @@
...
+	require.Error(t, err)
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+func generateCertificate(
+	t *testing.T,
+	commonName string,
+	notAfter time.Time,
+) ([]byte, []byte) {
```
SvartMetal, line 202: \n

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg/monitoring/metrics"
+	task_errors "github.com/ydb-platform/nbs/cloud/tasks/errors"
+	grpc_credentials "google.golang.org/grpc/credentials"
+)
+
+////////////////////////////////////////////////////////////////////////////////
+
+// How often certificate validity metrics should be updated.
+const certificateValidationPeriod = 24 * time.Hour
```
SvartMetal, line 20: Maybe once an hour?

cloud/disk_manager/pkg/app/run.go
```diff
@@ -175,18 +176,53 @@ func run(
...
+
+		nbsTLSProvider, err = common.NewGRPCClientTLSProvider(
+			common.GRPCClientTLSProviderConfig{
+				RootCertsFile: nbsConfig.GetRootCertsFile(),
+			},
+			mon.NewRegistry("nbs_tls"),
+		)
+		if err != nil {
+			return err
+		}
```
SvartMetal: I think it would be cleaner to move this code to `nbs.NewFactoryWithCreds` and you can use `nbsClientMetricsRegistry` for reporting tls metrics

cloud/disk_manager/pkg/app/run.go
```diff
@@ -175,18 +176,53 @@ func run(
...
+		nfsTLSProvider, err = common.NewGRPCClientTLSProvider(
+			common.GRPCClientTLSProviderConfig{
+				RootCertsFile: nfsConfig.GetRootCertsFile(),
+			},
+			mon.NewRegistry("nfs_tls"),
+		)
+		if err != nil {
+			return err
+		}
+	}
```
SvartMetal: Same comment as for nbs sdk code above

cloud/storage/core/go/grpc/grpc_client_transport_credentials.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return &grpcClientTransportCredentials{provider: provider}, nil
+}
+
+func (c *grpcClientTransportCredentials) ClientHandshake(
+	ctx context.Context,
+	authority string,
+	rawConn net.Conn,
+) (net.Conn, credentials.AuthInfo, error) {
```
SvartMetal: \n

cloud/storage/core/go/grpc/grpc_client_transport_credentials.go
```diff
@@ -0,0 +1,118 @@
...
+		reflect.Ptr, reflect.Slice:
+		return value.IsNil()
+	default:
+		return false
+	}
+}
+
+func (c *grpcClientTransportCredentials) ServerHandshake(
+	net.Conn,
+) (net.Conn, credentials.AuthInfo, error) {
```
SvartMetal: \n

cloud/storage/core/go/grpc/grpc_client_transport_credentials.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return &grpcClientTransportCredentials{
+		provider:           c.provider,
+		serverNameOverride: serverNameOverride,
+	}
+}
+
+func (c *grpcClientTransportCredentials) OverrideServerName(
+	serverNameOverride string,
+) error {
```
SvartMetal, line 119: \n

cloud/storage/core/go/grpc/grpc_client_transport_credentials_test.go
```diff
@@ -0,0 +1,262 @@
...
+		if transportCredentials != nil {
+			t.Fatal("expected nil transport credentials")
+		}
+	}
+}
+
+func newGRPCClientTransportCredentials(
+	t *testing.T,
+	provider TLSConfigProvider,
+) credentials.TransportCredentials {
```
SvartMetal, line 165: \n

cloud/storage/core/go/grpc/grpc_client_transport_credentials_test.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return transportCredentials
+}
+
+func performTLSHandshake(
+	t *testing.T,
+	transportCredentials credentials.TransportCredentials,
+	serverCertificate tls.Certificate,
+	authority string,
+) {
```
SvartMetal, line 182: \n

cloud/storage/core/go/grpc/grpc_client_transport_credentials_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if err := <-serverResult; err != nil {
+		t.Fatalf("server TLS handshake failed: %v", err)
+	}
+}
+
+func newServerCertificate(
+	t *testing.T,
+	serialNumber int64,
+) (tls.Certificate, *tls.Config) {
```
SvartMetal, line 230: \n

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
yegorskii: I did not change behaviour, just moved existing code. After automatic certs refresh, we can discuss if we really need such a validity metric. ExpireTs will be much more useful.

cloud/disk_manager/pkg/app/run.go
yegorskii: I moved the TLS-specific branching into common.NewGrpcClientTlsProvider. The constructor returns nil when the client is insecure or uses system roots, and otherwise returns a configured provider or an error. The provider now reports its metrics through nbsClientMetricsRegistry, so run() only contains a linear constructor call and error handling.

cloud/disk_manager/pkg/app/run.go
yegorskii: Applied the same approach to Filestore: the optional constructor handles the TLS configuration conditions, and the provider reports metrics through the nfs_client registry.
I kept the provider creation outside nfs.NewFactoryWithCreds because the same provider is shared between the control-plane and data-plane factories. Creating it inside each factory would produce multiple providers and, after automatic refresh is added, multiple refresh goroutines for the same certificate files.

cloud/disk_manager/internal/pkg/common/grpc_client_tls_provider.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (p *grpcClientTlsProvider) GetTlsConfig() *tls.Config {
+	p.mutex.RLock()
+	defer p.mutex.RUnlock()
+
+	if p.tlsConfig == nil {
+		return nil
+	}
+
+	return p.tlsConfig
```
SvartMetal: `return p.tlsConfig?`

cloud/disk_manager/internal/pkg/common/grpc_server_tls_provider.go
```diff
@@ -0,0 +1,212 @@
...
+		if expiration.after.Sub(now) <= certificateValidationThreshold {
+			validity = 0
+		}
+
+		expiration.validityGauge.Set(validity)
+	}
+}
+
+func (p *GrpcServerTlsProvider) getExpirations() []certificateExpiration {
+
```
SvartMetal: \n is only needed after a multi-line function signature

One-line function signatures don't need \n

:)

SvartMetal (changes requested):
Minor comments

Overall LGTM

SvartMetal (approved):

gy2411 (approved):

PR #6765, opened 2026-08-11
issue-6608: Implement quota inheritance in CreateNode and CreateHandle operations

debnatkh:
### Notes
Creating new files should inherit the parent's quota_id (same as uid/guid in case of set_gid bits)

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut_quotas.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_quotas.cpp
@@ -20,12 +20,6 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-NProto::TStorageConfig MakeStorageConfig()
-{
-    NProto::TStorageConfig config;
-    return config;
-}
-
 NProto::TStorageConfig MakeStorageConfigWithDirectoryCreationInShards()
 {
     NProto::TStorageConfig config;
@@ -58,8 +52,7 @@ protected:
         Service = MakeHolder<TServiceClient>(Env->GetRuntime(), nodeIdx);
 
         if (sharded) {
-            auto fsInfo = CreateFileSystem(*Service, fsConfig);
-            Y_UNUSED(fsInfo);
+            CreateFileSystem(*Service, fsConfig);
             FsId = fsConfig.FsId;
         } else {
             FsId = "test";
@@ -73,7 +66,7 @@ class TUnshardedFixture: public TShardingModeFixtureBase
 public:
     void SetUp(NUnitTest::TTestContext&) override
     {
-        Init(MakeStorageConfig(), false);
+        Init(NProto::TStorageConfig{}, false);
     }
 };
 
@@ -82,7 +75,7 @@ class TShardedFixture: public TShardingModeFixtureBase
 public:
     void SetUp(NUnitTest::TTestContext&) override
     {
-        Init(MakeStorageConfig(), true);
+        Init(NProto::TStorageConfig{}, true);
     }
 };
 
--- a/cloud/filestore/libs/storage/service/service_ut_quotas.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_quotas.cpp
@@ -1,0 +1,231 @@
+#include "service.h"
+#include "service_private.h"
+#include "service_ut_helpers.h"
+#include "service_ut_sharding.h"
+
+#include <cloud/filestore/libs/storage/model/utils.h>
+#include <cloud/filestore/libs/storage/testlib/service_client.h>
+#include <cloud/filestore/libs/storage/testlib/test_env.h>
+
+#include <library/cpp/testing/unittest/registar.h>
+
+#include <util/generic/size_literals.h>
+
+namespace NCloud::NFileStore::NStorage {
+
+using namespace NActors;
+using namespace NKikimr;
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+NProto::TStorageConfig MakeStorageConfig()
+{
+    NProto::TStorageConfig config;
+    return config;
+}
+
+NProto::TStorageConfig MakeStorageConfigWithDirectoryCreationInShards()
+{
+    NProto::TStorageConfig config;
+    config.SetDirectoryCreationInShardsEnabled(true);
+    return config;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TShardingModeFixtureBase: public NUnitTest::TBaseTestCase
+{
+public:
+    THolder<TTestEnv> Env;
+    THolder<TServiceClient> Service;
+    TString FsId;
+
+protected:
+    void Init(NProto::TStorageConfig config, bool sharded)
+    {
+        TShardedFileSystemConfig fsConfig;
+        if (sharded) {
+            config.SetAutomaticShardCreationEnabled(true);
+            config.SetAutomaticallyCreatedShardSize(
+                fsConfig.ShardBlockCount * 4_KB);
+            config.SetShardAllocationUnit(fsConfig.ShardBlockCount * 4_KB);
+        }
+
+        Env = MakeHolder<TTestEnv>(TTestEnvConfig{}, config);
+        ui32 nodeIdx = Env->AddDynamicNode();
+        Service = MakeHolder<TServiceClient>(Env->GetRuntime(), nodeIdx);
+
+        if (sharded) {
+            auto fsInfo = CreateFileSystem(*Service, fsConfig);
+            Y_UNUSED(fsInfo);
+            FsId = fsConfig.FsId;
+        } else {
+            FsId = "test";
+            Service->CreateFileStore(FsId, 1'000);
+        }
+    }
+};
+
+class TUnshardedFixture: public TShardingModeFixtureBase
+{
+public:
+    void SetUp(NUnitTest::TTestContext&) override
+    {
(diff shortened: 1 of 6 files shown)


debnatkh:
Here's my review of PR #6765 ([issue-6608: Implement quota inheritance in CreateNode and CreateHandle operations](https://github.com/ydb-platform/nbs/pull/6765)).

## What it does
Adds quota-id inheritance: when a node is created under a directory that has (or has inherited) a `QuotaId`, the new node's `QuotaId` attribute is copied from the parent at creation time. Touches:
- `node.proto` — new `uint32 QuotaId = 20` field on `TCreateNodeRequest`, documented as an internal field for carrying the resolved quota id across the cross-shard hop.
- `tablet_actor_createnode.cpp` — `InitAttrs` now copies `request.GetQuotaId()` onto the new node's attrs; `PrepareTx_CreateNode` overwrites `args.Attrs`/`args.Request` QuotaId with the *real* parent's QuotaId when `!behaveAsShard`.
- `tablet_actor_createhandle.cpp` — similarly sets `args.QuotaId` from the real parent when `!BehaveAsShard`, then applies it to the locally-created node and to the `CreateNodeRequest` forwarded to a shard.
- New `service_ut_quotas.cpp` with good coverage across unsharded, sharded, and sharded-with-directory-creation fixtures (direct child, grandchild inheriting transitively, and an unrelated sibling staying unquota'd).

## Design looks sound where I could trace it
- Inheritance is always recomputed from the real parent on the owning tablet and only trusted from the request when forwarded internally (`behaveAsShard`), so a client can't spoof `QuotaId` in a create request — nice.
- The cross-shard `CreateNode`-via-`OpLogEntry` path (`RegisterCreateNodeInShardActor`) correctly threads `args.QuotaId` through `shardRequest->SetQuotaId(...)`, and the shard's own `InitAttrs` picks it up via `request.GetQuotaId()`.

## Things worth double-checking before merge

1. **Possible gap in `CreateHandle`'s own quota resolution.** `args.QuotaId` is populated only inside `if (!BehaveAsShard(...))`. If `PrepareTx_CreateHandle` ever takes the `args.ShardId.empty()` branch while `BehaveAsShard == true` (i.e., a `CreateHandle` with `Name` + `E_CREATE` executed directly against a shard tablet, not routed through the `CreateNodeInShardActor`), `args.QuotaId` stays at its default `0`, and — unlike `TCreateNodeRequest` — I don't see a `QuotaId` field added to `TCreateHandleRequest` in this diff to carry a pre-resolved value in from the caller. If that code path is actually reachable, quota inheritance would silently be lost there. Worth either confirming it's unreachable in practice (and maybe asserting/commenting why), or wiring it up the same way `CreateNode` does.
2. **Test coverage doesn't exercise that path.** The new tests call `CreateHandle` through the normal service client, which (per the existing `SelectShard` logic) routes new-file creation into shards via the `CreateNode`-forwarding path — the one that's verified to work. A test that directly forces a `behaveAsShard=true` `CreateHandle`-with-creation call would help confirm/close the gap above.
3. **Inheritance is a creation-time snapshot, not dynamic.** Changing a directory's `QuotaId` via `SetNodeAttr` (as in the test) only affects nodes created afterward — existing children keep their old value. That's probably intentional, but it'd be worth a one-line comment/doc note so it isn't surprising to whoever operates this later.
4. **Interaction with `Rename`/hardlinks isn't addressed** — moving a node into a differently-quota'd directory presumably doesn't update its `QuotaId`. Likely fine/out-of-scope for this issue, but worth a follow-up ticket if quota enforcement depends on this.
5. **Minor:** `BehaveAsShard(args.Request.GetHeaders())` is computed twice in `PrepareTx_CreateHandle` (once for the new quota check, once later for the existing shard-selection logic) — could hoist to a single local for clarity, though it's cheap so not a big deal.
6. **PR description is still the unfilled template** ("Describe your PR..." / "Put links to the related issues here") — should be filled in with the actual rationale and a link to issue #6608 before merging, for anyone reading history later.

CI (`linux-x86_64-relwithdebinfo`, filestore target) is green with all 4468 tests passing.

Overall this is a focused, well-tested change for the common path; I'd just get clarity on point 1 (the `CreateHandle`/`BehaveAsShard` gap) before approving.

debnatkh:
`ExecuteTx_CreateNode`'s shard path does `shardRequest->CopyFrom(args.Request)`, so the `args.Request.SetQuotaId()` in Prepare is what propagates it. Now the CreateHandle side:One quick look at what the follow-up "fix review issue" commit changed, then I have the full picture:I pulled the branch and read the full diff (6 files, +254/−2) plus the surrounding code in `PrepareTx_CreateNode`, `PrepareTx/ExecuteTx_CreateHandle`, `TCreateNodeInShardActor`, `BehaveAsShard`, and the `SetNodeAttr` quota rules from the earlier PR in this series. The core mechanics are sound — parent is validated non-null before the new dereferences, the `behaveAsShard` guard is correct on both ends (the shard actor clears `BehaveAsDirectoryTablet` before forwarding, so shards correctly trust the request's `QuotaId`), the shard request is persisted in the oplog so inheritance survives a leader restart, and the three-fixture test macro genuinely exercises the local, leader→shard, and dir-in-shards cross-shard chains. That said, here's what I'd raise:

**1. Rename and hardlink silently break the subtree invariant.** Inheritance only happens at creation. `RenameNode` moving an existing file into a quota'd directory leaves it with `QuotaId=0`, and moving a quota'd file *out* leaves it tagged forever. Same for `CreateNode(Link)`: the linked node keeps its original quota. The earlier PR's "attach only to an *empty* directory" rule in `tablet_actor_setnodeattr.cpp` exists precisely to guarantee subtree consistency — and a single `mv` or `ln` now bypasses it. XFS project quotas handle this by refusing cross-project renames (EXDEV); here it's just silent. If this is deliberately deferred to a follow-up, it should at least be noted in the PR/issue, ideally with a test pinning current behavior.

**2. One attach is now irreversible for an entire growing subtree.** `SetNodeAttr` rejects `QuotaId=0` ("quota id must be non-zero") and rejects changing to a different quota. Pre-PR that meant attach-once per directory; post-PR every node ever created underneath is permanently tagged, and there's no detach API. A mistaken attach on the wrong directory can never be corrected. (Perhaps `AttachQuota`/detach is planned per the `TODO(6608)` in `tablet_state_quotas.cpp`, but worth confirming.)

**3. The propagated id is never validated, and it's in the public API.** Nothing checks `FindQuota(quotaId)` anywhere in this path, so nodes can carry dangling quota ids. More importantly, `QuotaId = 20` sits in the public `TCreateNodeRequest`: on the main FS a client-supplied value is silently overwritten (fine), but on the shard path `InitAttrs` applies `request.GetQuotaId()` verbatim. That matches how Uid/Gid are trusted, but the field deserves a comment marking it as leader-populated/internal, like the one on `OriginalNodeId` — notably, the "fix review issue" commit replaced the original, more informative comment with a vaguer one.

**4. `TCreateHandle::Clear()` doesn't reset the new `QuotaId` member.** Every other mutable field in that struct (`ReadCommitId`, `ShardId`, `IsNewShardNode`, `ParentNode`, `UpdatedNodes`…) is reset in `Clear()` for tx restarts. It's benign today because Prepare unconditionally reassigns it on the same branch, but it breaks the struct's convention and invites a stale-state bug later. (`Gid` has the same pre-existing omission, which is probably where the habit came from.)

**5. `args.QuotaId` stays 0 in one `IsNewShardNode` combination.** In `PrepareTx_CreateHandle`, `shardId` can come from `args.RequestShardId` even when `behaveAsShard` is true, in which case `IsNewShardNode` is set but `args.QuotaId` was never assigned, and `shardRequest->SetQuotaId(0)` goes out. If that combination is unreachable, a comment saying so would help; if it's reachable, inheritance silently drops there.

**6. Dead work for hardlink requests.** In `PrepareTx_CreateNode` the quota block also runs for `Link` requests, where `args.Attrs` is never applied and the request isn't forwarded to a shard by the leader — harmless dead assignments, but a `!isParentNodeLinkRequest` guard or comment would make intent clearer.

**7. Test coverage is happy-path only.** Nothing pins down the rename/hardlink behavior from point 1, nothing verifies a client-supplied `QuotaId` is ignored on the main FS, and there's no case with a second, differently-quota'd directory to catch cross-contamination. Small nits in the new test file: `Y_UNUSED(fsInfo)` can go by just not capturing the return value, and `MakeStorageConfig()` is a trivial wrapper around a default-constructed config.

Also a cosmetic point: the proto comment "Id of the quota this node's parent is attached to" is slightly off — after this PR the parent may have merely *inherited* the id rather than been attached, and the field is also what gets stamped on the node itself.

Points 1–3 are the ones I'd actually want answered before merge; 4–7 are cleanups. Happy to dig into any of these further — for example, checking how the planned enforcement (MaxBytes/MaxNodes in `TQuota`) is supposed to account for sharded files that have no local node on the leader.

debnatkh:
1. Rename/hardlink invariant – intentional; hardlinks across quota domains will be rejected outright (not silently retagged or left inconsistent) once that validation lands.
2. Irreversible attach – planned (detach mechanism).
3. Unvalidated propagated ID – validation is planned for fruther PRs
4. Clear() fix – done
5. Same as 1, since RequestShardId is only set on harlinks 
6. Amount of work is insignificant, not worth of a worry
7. Nits fixed

SvartMetal (approved):

agalibin (approved):

Issue #6645, opened 2026-07-30
Adapt the code to run on the ARM architecture


proller:
`ya make`, `ya test` should work same as on x86
Binaries should work on 64k page size

- ya make broken on systems with 64k page size
- some built by yandex binaries does not work on 64k page size
- lot of binaries uploaded to sandbox have no arm variant https://github.com/ydb-platform/nbs/pull/6540
- need to make qemu arm images  https://github.com/ydb-platform/nbs/pull/6126 
- qemu lib improvements https://github.com/ydb-platform/nbs/pull/6181/changes
- lot of tests using  ydb sql fails on arm
- some other tests fail on arm (silk)

also related pr's :
https://github.com/ydb-platform/nbs/pull/6457
https://github.com/ydb-platform/nbs/pull/6445
https://github.com/ydb-platform/nbs/pull/6627

PR #6772, opened 2026-08-12
issue-6645 Fix test cloud/blockstore/tests/plugin/uds when built with forced allocator change (-DALLOCATOR=TCMALLOC_256K) 

proller:
https://github.com/ydb-platform/nbs/issues/6645

Without this patch  -DALLOCATOR=TCMALLOC_256K changes FAKE allocator for plugin with TCMALLOC_256K


librarian (approved):
afaik we do not use plugin at all

nisqatsi (approved):

PR #6773, opened 2026-08-12
issue-5894, issue-5895: fastshard server should listen on ipv6 as well as on ipv4

qkrorlqr:
### Notes
Claude was so kind to fix its own bug regarding listening only on ipv4 address. I reviewed it, looks good. Just logging the cases when we trigger fallbacks to ipv6-only or ipv4-only is okay here. This is currently an experimental component AND ipv4 is needed for tests only AND it's still going to be immediately visible if it happens.

This server is used only for tcp-side-channel which is itself experimental.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/fastshard/server/server.cpp
+++ b/cloud/filestore/libs/storage/fastshard/server/server.cpp
@@ -45,6 +45,27 @@ constexpr ui64 MaxMessageSize = 64_MB;
 constexpr int SocketBacklog = 128;
 
 ////////////////////////////////////////////////////////////////////////////////
+
+int BindAndListen(int fd, const sockaddr* addr, socklen_t addrLen)
+{
+    int one = 1;
+    ::setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one));
+
+    if (::bind(fd, addr, addrLen) < 0) {
+        SILK_ERROR("bind: %s", ::strerror(errno));
+        ::close(fd);
+        return -1;
+    }
+
+    if (::listen(fd, SocketBacklog) < 0) {
+        SILK_ERROR("listen: %s", ::strerror(errno));
+        ::close(fd);
+        return -1;
+    }
+    return fd;
+}
+
+////////////////////////////////////////////////////////////////////////////////
 // Handler bookkeeping.
 //
 // Every accepted connection is represented by a THandler which owns the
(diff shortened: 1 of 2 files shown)


qkrorlqr:
```
● Review of the diff (one file changed for the fix, one for tests):

  Finding 1 — real, though benign today: the accept path still uses sockaddr_in (server.cpp:418). AcceptFiberMain passes a 16-byte sockaddr_in buffer to accept4 on what is now an AF_INET6 listener whose peer addresses are 28-byte sockaddr_in6.
  accept4 doesn't fail on a short buffer — it truncates the address and returns a valid fd — and nothing reads addr afterwards, so behavior is correct right now. But it's a trap armed for whoever first logs the peer address: they'll read a
  truncated struct through the wrong type. Since the value is unused, the honest fix is ::accept4(lfd, nullptr, nullptr, ...); if peer logging is anticipated, sockaddr_storage. This is the one thing I'd change before merging.

  Finding 2 — comment overstates the fallback condition. The comment says v6-disabled hosts fail with EAFNOSUPPORT, but the code falls back to IPv4 on any socket(AF_INET6) errno. On, say, EMFILE the v4 attempt fails identically and we return -1
  with both errors logged — the outcome is fine, but the code doesn't match the comment's precision. Either check errno == EAFNOSUPPORT explicitly or loosen the comment. Cosmetic.

  Finding 3 — judgment call I made silently: IPV6_V6ONLY clear failure only warns. If that setsockopt ever failed on a bindv6only=1 host, the server would run v6-only — the mirror image of the original bug, with only a WARN line as evidence. I left
  it a warning because clearing the flag before bind essentially cannot fail on Linux, and a v6-only listener still serves the production (IPv6) case. If you'd rather fail loudly than degrade quietly, that line should return -1 instead.

  What holds up: ownership is clean (BindAndListen consumes the fd on every failure path, single close, no leak — including the fallback path); SO_REUSEADDR semantics unchanged; the wildcard scope didn't widen beyond adding the v6 family; error
  handling and 80-column style match the file; build is -Werror-clean and all 6 tests pass. Tests: the ::1 test is a genuine regression test (fails against the old AF_INET code), the 127.0.0.1 test pins the IPV6_V6ONLY clearing, string-literal
  lifetimes in THostParam are static so the fiber params slot is safe, and the skip guard probes the right thing.

  Known remaining gap, not part of this fix: sn/server/server.cpp:496 still has the identical AF_INET-only listener (test-only component), and the socket(AF_INET6)-failure fallback branch has no test — triggering it needs syscall fault injection,
  which isn't worth a seam in bootstrap code.

```

debnatkh (approved):

WilyTiger (approved):

PR #6757, opened 2026-08-11
issue-6208: [Filestore] Track access in the filesystem (datastructures model)

szumal:
### Notes
This is a split from https://github.com/ydb-platform/nbs/pull/6670. It contains only the datastructures.

### Issue
https://github.com/ydb-platform/nbs/issues/6208

PR #6778, opened 2026-08-13
issue-6208: [Filestore] Ranking datastructures

szumal:
### Notes
Extracting common data structures that can be reused in https://github.com/ydb-platform/nbs/pull/6757 and  https://github.com/ydb-platform/nbs/pull/6749

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/tablet/model/ranking.h
+++ b/cloud/filestore/libs/storage/tablet/model/ranking.h
@@ -1,0 +1,95 @@
+#pragma once
+
+#include <util/datetime/base.h>
+#include <util/generic/hash.h>
+#include <util/generic/set.h>
+#include <util/generic/vector.h>
+
+namespace NCloud::NFileStore::NStorage {
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <
+    typename TKey,
+    typename TValue,
+    typename TComparator,
+    typename TKeyExtractor,
+    typename TKeyHash = THash<TKey>>
+class TBoundedRanking
+{
+private:
+    using TEntries = TSet<TValue, TComparator>;
+    using TIterator = typename TEntries::iterator;
+
+    size_t MaxEntries;
+    TKeyExtractor ExtractKey;
+    TEntries Entries;
+    THashMap<TKey, TIterator, TKeyHash> KeyToRanking;
+
+private:
+    void EvictFirst()
+    {
+        while (Entries.size() > MaxEntries) {
+            auto first = Entries.begin();
+            const TKey key = ExtractKey(*first);
+
+            KeyToRanking.erase(key);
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/model/ranking.h
```diff
@@ -0,0 +1,95 @@
...
+        const auto key = ExtractKey(value);
+        auto it = KeyToRanking.find(key);
+
+        if (it != KeyToRanking.end()) {
+            Entries.erase(it->second);
+            KeyToRanking.erase(it);
+        }
+
+        auto [newIt, inserted] = Entries.insert(std::move(value));
+        Y_ABORT_UNLESS(inserted);
```
qkrorlqr: better to return `false` and abort if needed at the layer above (and probably not even abort but abort in debug builds and report a critical event in release builds - by reporting this https://github.com/ydb-platform/nbs/blob/4290008f572c13f29f19eafea55be986d1368f83/cloud/filestore/libs/diagnostics/critical_events.h#L77)

qkrorlqr: this way you'll also be able to test this behaviour

cloud/filestore/libs/storage/tablet/model/ranking.h
```diff
@@ -0,0 +1,99 @@
...
+        auto it = KeyToRanking.find(key);
+
+        if (it != KeyToRanking.end()) {
+            Entries.erase(it->second);
+            KeyToRanking.erase(it);
+        }
+
+        auto [newIt, inserted] = Entries.insert(std::move(value));
+        if (!inserted) {
+            return false;
```
szumal, line 73: P1 — Existing entries can be silently lost on a failed update. In [ranking.h:58-67](https://github.com/ydb-platform/nbs/blob/643e6aeaa93ad1d200a10a995ea38f33dbdb56e4/cloud/filestore/libs/storage/tablet/model/ranking.h#L58-L67), an existing key is erased before inserting the replacement. If the replacement compares equivalent to another entry, Entries.insert() fails, returns false, and the original entry is already gone from both containers. For example, with score-only comparison, updating key A to a score already held by key B removes A. Insert the replacement first or restore the old entry on failure; add a regression test for this collision.

szumal (commented):
CODEX REVIEW

cloud/filestore/libs/storage/tablet/model/ranking.h
szumal: This would only happen if the supplied comparator does not provide a unique tiebreaker. We will catch failures in the clients of this code.

qkrorlqr (approved):

agalibin (approved):

Issue #6780, opened 2026-08-13
[Filestore] support xattrs and locks in filestore-client


qkrorlqr:
The xattr and lock-related command sets are missing in the cli, let's add them

PR #6781, opened 2026-08-13
issue-6780: support for xattrs and locks in filestore-client

qkrorlqr:
### Notes
Claude struggled for a couple hours but eventually spat out the implementation of:
* acquirelock
* releaselock
* testlock
* getnodexattr
* listnodexattr
* removenodexattr
* setnodexattr

together with `filestore/tests/client` testcases

This cost me around 30 euros of tokens for Fable5, better be good. Especially considering the fact that it's almost copy-paste of the implementation of the other commands - exactly the AGI type of task.

UPD: the extra fixes took 15 more euros
and 5 euros for the review

total: 50 euros
AGI-level stuff

### Issue
https://github.com/ydb-platform/nbs/issues/6780

### DIFF
--- a/cloud/filestore/apps/client/lib/factory.cpp
+++ b/cloud/filestore/apps/client/lib/factory.cpp
@@ -43,6 +43,13 @@ TCommandPtr NewFindCommand();
 TCommandPtr NewDiffCommand();
 TCommandPtr NewPingCommand();
 TCommandPtr NewDiagnoseFilesystemCommand();
+TCommandPtr NewSetNodeXAttrCommand();
+TCommandPtr NewGetNodeXAttrCommand();
+TCommandPtr NewListNodeXAttrCommand();
+TCommandPtr NewRemoveNodeXAttrCommand();
+TCommandPtr NewAcquireLockCommand();
+TCommandPtr NewReleaseLockCommand();
+TCommandPtr NewTestLockCommand();
 
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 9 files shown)


debnatkh (dismissed):

yegorskii (approved):

qkrorlqr:
AI review:
```
● Second pass over the fixed code. The five original findings are properly closed — what remains is smaller:

  1. Comment style drift in locks.cpp. The three // Validate the arguments before creating a session. remarks are plain one-liners sitting on top of statement blocks, while the other inline comments in the same file (Handle access mode follows 
  POSIX..., Unlocking has no access-mode requirement...) use the project's sandwiched // style. CLAUDE.md explicitly says not to collapse to a plain one-liner above a statement block. Same file, two conventions — should be sandwiched.

  2. HandleFlags(NProto::E_SHARED) in releaselock says one thing and means another. It passes a lock-type constant to express "open for read". The comment above it rescues the intent, but the honest shape is a separate constant or just
  ProtoFlag(E_READ) inline — release has no lock type, so borrowing the shared-lock mapping is a pun.

  3. The access-mode fix has a mirror-image edge. releaselock now opens E_READ, so a write-only file (mode 0200) can no longer be unlocked via the CLI — the exact inverse of the wart I fixed for read-only files, just on a far rarer file kind. If
  the backend enforces mode on CreateHandle, the only fully general answer is trying read then falling back to write, which is more machinery than the case deserves. Worth knowing, not worth code.

  4. The new output branches are only half-tested. The python helpers hardcode --json, so the fixed text-mode paths — name=value listing lines and the enriched incompatible: owner=... conflict line — never execute in tests. The JSON branches are
  now covered including values (ValuesBase64 asserted), but text mode is exercised only by humans. The conflict branch additionally still needs two concurrent sessions to trigger at all, which the harness cannot do — that limitation is structural
  and stands from last time.

```

WilyTiger (approved):

Issue #6783, opened 2026-08-13
[Filestore] Deterministic weighted shard balancer


ruslan-gar:
Requirements:
1. Work deterministically
2. The goal is make shards to have the same UsedBytesCount.
The new balancer works only with strict filesystem size enforcement.
In this mode all the shards have the same TotalBlocksCount  which is equal to 
the that of the whole file system. 
So, free space in a shard may be evaluated as: TotalBlocksCount / ShardCount - UsedBlocksCount 
3. Should have linear complexity

Implementation:
Deterministically selects shards with frequencies proportional to their
scores. The balancer traverses score levels from lowest to highest. At each
level, it visits every shard whose score is at least that
level. Therefore, a shard with score N is selected N + 1 times during a
 complete traversal.

For example, five shards with scores {0, 2, 4, 3, 1} are visited according
to the following table. Read it row by row from top to bottom, from left to
right; an "x" marks a shard skipped at that score level:

 0 1 2 3 4
 x 1 2 3 4
 x 1 2 3 x
 x x 2 3 x
 x x 2 x x

PR #6784, opened 2026-08-13
issue-6783: [Filestore] Deterministic weighted shard balancer

ruslan-gar:
### Notes
Deterministic weighed shard balancer is implemented

### Issue
#6783

### DIFF
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -66,6 +66,7 @@ namespace NCloud::NFileStore{
     xxx(BadValueInFeatureConfig)                                               \
     xxx(ResetSessionInterrupted)                                               \
     xxx(DestroySessionInterrupted)                                             \
+    xxx(WeightedDeterministicBalancerRequiresStrictEnforcement)                \
 // FILESTORE_CRITICAL_EVENTS
 
 #define FILESTORE_CRITICAL_EVENTS_WITHOUT_LOGGING(xxx)                         \
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -66,7 +66,7 @@ namespace NCloud::NFileStore{
     xxx(BadValueInFeatureConfig)                                               \
     xxx(ResetSessionInterrupted)                                               \
     xxx(DestroySessionInterrupted)                                             \
-    xxx(WeightedDeterministicBalancerRequiresStrictEnforcement)                \
+    xxx(IncompatibleFeatures)                                                  \
 // FILESTORE_CRITICAL_EVENTS
 
 #define FILESTORE_CRITICAL_EVENTS_WITHOUT_LOGGING(xxx)                         \
--- a/cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
@@ -328,21 +328,41 @@ void TShardBalancerWeightedDeterministic::CalcNextShard()
 
     for (ui64 i = 1; i < Metas.size(); ++i) {
         const auto& meta = Metas[i];
-        // Update Next for the last shard
+        // Update Next for the last shard.
         setNextShard(lastShardIdx, i, meta.Score);
 
-        // Pop stack
+        // Pop the stack.
         while (stack && meta.Score >= stack.back()->Score) {
-            const auto& back = *stack.back();
-            setNextShard(back.ShardIdx, meta.ShardIdx, back.Score);
+            for (ui64 shardIdx = stack.back()->ShardIdx;
+                 shardIdx < meta.ShardIdx;
+                 ++shardIdx)
+            {
+                setNextShard(shardIdx, meta.ShardIdx, meta.Score);
+            }
             stack.pop_back();
         }
 
-        if (stack) {
-            setNextShard(stack.back()->ShardIdx, meta.ShardIdx, meta.Score);
+        const ui64 leftBorder = stack ? stack.back()->ShardIdx : 0;
+        for (ui64 shardIdx = leftBorder; shardIdx < meta.ShardIdx; ++shardIdx) {
+            setNextShard(shardIdx, meta.ShardIdx, meta.Score);
         }
 
         stack.push_back(&meta);
+    }
+
+    // Fill the gap on the right side of the matrix.
+    for (ui64 score = 0; score < ScoreLevelsCount &&
+                         GetNextShard(lastShardIdx, score) < Metas.size();
+         ++score)
+    {
+        const ui64 nextShard = GetNextShard(lastShardIdx, score);
+        for (ui64 shardIdx = lastShardIdx - 1;
+             shardIdx != Max<ui64>() &&
+             GetNextShard(shardIdx, score) == Metas.size();
+             --shardIdx)
+        {
+            GetNextShard(shardIdx, score) = nextShard;
+        }
     }
 }
 
@@ -379,24 +399,22 @@ NProto::TError TShardBalancerWeightedDeterministic::SelectShard(
             TStringBuilder() << "Metas.size() is zero");
     }
 
-    // In the middle but no next shard.
-    if (ShardSelector < Metas.size() - 1 &&
-        GetNextShard(ShardSelector, CurrentScore) == Metas.size())
-    {
+    // If jumping to the next shard crosses the right boundary,
+    // increment CurrentScore.
+    const ui64 nextShardIdx = GetNextShard(ShardSelector, CurrentScore);
+    if (nextShardIdx < Metas.size() && nextShardIdx <= ShardSelector) {
+        CurrentScore = (++CurrentScore) % ScoreLevelsCount;
         ShardSelector = Metas.size() - 1;
     }
 
-    // In the last column the information where to start is stored.
-    if (ShardSelector == Metas.size() - 1) {
-        CurrentScore = (++CurrentScore) % ScoreLevelsCount;
+    // We are in an undefined part of the NextShard matrix.
+    // That means that we should start from the initial state.
+    if (GetNextShard(ShardSelector, CurrentScore) == Metas.size()) {
+        CurrentScore = 0;
+        ShardSelector = Metas.size() - 1;
     }
 
     ShardSelector = GetNextShard(ShardSelector, CurrentScore);
-    if (ShardSelector == Metas.size()) {
-        CurrentScore = 0;
-        ShardSelector = 0;
-    }
-
     *shardId = Metas[ShardSelector].Stats.ShardId;
 
     return {};
--- a/cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
+++ b/cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
@@ -254,12 +254,12 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
 ////////////////////////////////////////////////////////////////////////////////
 
 TShardBalancerWeightedDeterministic::TShardBalancerWeightedDeterministic(
-    ui32 blockSize,
-    ui64 precisionBytes,
-    ui32 maxFileBlocks,
-    ui64 desiredFreeSpaceReserve,
-    ui64 minFreeSpaceReserve,
-    TVector<TString> shardIds)
+        ui32 blockSize,
+        ui64 precisionBytes,
+        ui32 maxFileBlocks,
+        ui64 desiredFreeSpaceReserve,
+        ui64 minFreeSpaceReserve,
+        TVector<TString> shardIds)
     : TShardBalancerBase(
           blockSize,
           precisionBytes,
@@ -268,8 +268,6 @@ TShardBalancerWeightedDeterministic::TShardBalancerWeightedDeterministic(
           minFreeSpaceReserve,
           std::move(shardIds))
 {
-    Y_UNUSED(precisionBytes);
-
     // Update with empty stats.
     TVector<TShardStats> shardStats(Metas.size());
     for (ui64 i = 0; i < shardStats.size(); ++i) {
@@ -278,7 +276,7 @@ TShardBalancerWeightedDeterministic::TShardBalancerWeightedDeterministic(
     Update(shardStats, {}, {});
 
     ShardSelector = Metas.size() - 1;
-    CurrentScore = MaxScore();
+    CurrentScore = MaxScore;
 }
 
 void TShardBalancerWeightedDeterministic::CalcScore(
@@ -299,7 +297,7 @@ void TShardBalancerWeightedDeterministic::CalcScore(
     Metas.reserve(stats.size());
     for (ui64 i = 0; i < stats.size(); ++i) {
         const ui64 score = Min<ui64>(
-            MaxScore(),
+            MaxScore,
             (maxBlocksCount - stats[i].UsedBlocksCount) / step);
         Metas.emplace_back(i, stats[i], score);
     }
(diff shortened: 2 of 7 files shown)


debnatkh:
This deterministic balancer assumes that strict filesystem enforcement is enabled – it ignores fileSize, desiredFreeSpaceReserve, and minFreeSpaceReserve

Let us make a critical event if we encounter SBP_WEIGHTED_DETERMINISTIC + no strict enforcement

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+    Y_UNUSED(precisionBytes);
+
+    // Update with empty stats.
+    TVector<TShardStats> shardStats(Metas.size());
+    for (ui64 i = 0; i < shardStats.size(); ++i) {
+        shardStats[i] = Metas[i].Stats;
+    }
+    Update(shardStats, {}, {});
+
+    ShardSelector = Metas.size() - 1;
```
debnatkh: underflow in case of an empty set of shards. Why dont you want to start with 0?

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+    TVector<TString> shardIds)
+    : TShardBalancerBase(
+          blockSize,
+          precisionBytes,
+          maxFileBlocks,
+          desiredFreeSpaceReserve,
+          minFreeSpaceReserve,
+          std::move(shardIds))
+{
+    Y_UNUSED(precisionBytes);
```
debnatkh: used

cloud/filestore/libs/storage/tablet/tablet_state.cpp
```diff
@@ -132,6 +132,13 @@ void TIndexTabletState::InitInMemoryIndexState(const TStorageConfig& config)
 
 void TIndexTabletState::InitShardBalancer(const TStorageConfig& config)
 {
+    // When a filesystem is just created, GetBlockSize() is zero.
+    // BlockSize later gets its correct value and InitShardBalancer is called
+    // again.
+    if (GetBlockSize() == 0) {
+        return;
```
debnatkh: doesnt this leave Impl->ShardBalancer null with no null-checks anywhere downstream?

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -64,19 +64,25 @@ TShardBalancerBase::TShardBalancerBase(
...
+
+    TShardMetaComp metaCmp(PrecisionBytes, BlockSize);
+    Metas.reserve(shardIds.size());
     for (ui32 i = 0; i < shardIds.size(); ++i) {
+        emptyShardStats.ShardId = shardIds[i];
         Metas.emplace_back(
             i,
-            // Before the first update shards are treated like empty with
-            // infinite capacity. maxFileBlocks is used as infinity here.
-            TShardStats{
```
debnatkh, line 72: I think that explicitly constructing TShardStats and calculating its score each time is more readable than patching a fake "empty" shard

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -167,7 +168,62 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
     NProto::TError SelectShard(ui64 fileSize, TString* shardId) final;
 };
 
-/////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+class TShardBalancerWeightedDeterministic: public TShardBalancerBase
```
debnatkh, line 192: Let us add a small description of how this balancer works

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -167,7 +168,62 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+{
+    static constexpr ui64 ScoreLevelsCount = 8;
+    static constexpr ui64 MaxScore()
```
debnatkh: static constexpr ui64 ScoreLevelsCount = 8;
static constexpr ui64 MaxScore = ScoreLevelsCount - 1;

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -167,7 +168,62 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+    ui64 CurrentScore;
+
+    // Actually, it's a two-dimensional array of size:
+    // ScoreLevelsCount * Metas.size().
+    // It's a mapping from (score, shardIdx) -> nextShardIdx.
+    TVector<ui64> NextShard;
+
+    const ui64& GetNextShard(ui64 shardIdx, ui64 score) const
+    {
+        return NextShard[shardIdx * ScoreLevelsCount + score];
```
debnatkh: Why do you flatten the array? Why is it not a vector of arrays, or vice versa?

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -167,7 +168,62 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+public:
+    TShardBalancerWeightedDeterministic(
+        ui32 blockSize,
+        ui64 precisionBytes,
+        ui32 maxFileBlocks,
+        ui64 desiredFreeSpaceReserve,
+        ui64 minFreeSpaceReserve,
+        TVector<TString> shardIds);
+
+public:
```
debnatkh: no need in second public

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
 
 ////////////////////////////////////////////////////////////////////////////////
 
+TShardBalancerWeightedDeterministic::TShardBalancerWeightedDeterministic(
+    ui32 blockSize,
+    ui64 precisionBytes,
+    ui32 maxFileBlocks,
```
debnatkh: tabs for consistency

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+    Metas.reserve(stats.size());
+    for (ui64 i = 0; i < stats.size(); ++i) {
+        const ui64 score = Min<ui64>(
+            MaxScore(),
+            (maxBlocksCount - stats[i].UsedBlocksCount) / step);
+        Metas.emplace_back(i, stats[i], score);
+    }
+}
+
+void TShardBalancerWeightedDeterministic::CalcNextShard()
```
debnatkh, line 308: also need some comment explaining non-trivial logic :)

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -167,7 +168,62 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+class TShardBalancerWeightedDeterministic: public TShardBalancerBase
+{
+    static constexpr ui64 ScoreLevelsCount = 8;
+    static constexpr ui64 MaxScore()
+    {
+        return ScoreLevelsCount - 1;
+    }
+
+    // Iterator state.
+    ui64 ShardSelector;
```
debnatkh: is default init needed here?

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+{
+    NextShard.assign(ScoreLevelsCount * Metas.size(), Metas.size());
+
+    auto setNextShard = [this](ui64 shard, ui64 nextShard, ui64 fromScore) {
```
debnatkh: there are nextShard and NextShard variables here
might be misleading

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+          blockSize,
+          precisionBytes,
+          maxFileBlocks,
+          desiredFreeSpaceReserve,
+          minFreeSpaceReserve,
+          std::move(shardIds))
+{
+    Y_UNUSED(precisionBytes);
+
+    // Update with empty stats.
```
debnatkh: it is not empty

might be "zero stats"

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +253,192 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+{
+    Y_UNUSED(desiredFreeSpaceReserve);
+    Y_UNUSED(minFreeSpaceReserve);
+
+    if (stats.size() != Metas.size()) {
+        return MakeError(E_ARGUMENT, TStringBuilder() << "stats.size() "
+            << stats.size() << " != Metas.size() " << Metas.size());
+    }
+
+    if (stats) {
```
debnatkh: let us explicitly cast to bool

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
ruslan-gar: It's not a problem as SelectShard returns error when it's called with Metas.size() == 0
ShardSelector represents not a shard to be selected but a shard that has already been seleted in the previous call
of SelectShard. In order to select the first shard during the first call the SelectShard is initialized this wey

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
ruslan-gar: This fields are initialized in the only ctor TShardBalancerWeightedDeterministic has.
It look like default init doesn't make sense.

cloud/filestore/libs/diagnostics/critical_events.h
```diff
@@ -66,6 +66,7 @@ namespace NCloud::NFileStore{
     xxx(BadValueInFeatureConfig)                                               \
     xxx(ResetSessionInterrupted)                                               \
     xxx(DestroySessionInterrupted)                                             \
+    xxx(WeightedDeterministicBalancerRequiresStrictEnforcement)                \
```
debnatkh: I would rather have a more generic name, something like "incompatible features"

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
debnatkh: Changing semantics to "next position to be selected" seems reasonable

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -246,6 +251,200 @@ NProto::TError TShardBalancerWeightedRandom::SelectShard(
...
+            {
+                setNextShard(shardIdx, meta.ShardIdx, meta.Score);
+            }
+            stack.pop_back();
+        }
+
+        const ui64 leftBorder = !stack.empty() ? stack.back()->ShardIdx : 0;
+        for (ui64 shardIdx = leftBorder; shardIdx < meta.ShardIdx; ++shardIdx) {
+            setNextShard(shardIdx, meta.ShardIdx, meta.Score);
+        }
```
debnatkh, line 323: Using a stack algorithm in this method is overkill; it might be worth simplifying

```
void TShardBalancerWeightedDeterministic::CalcNextShard()
{
    NextShard.SetSizes(Metas.size(), ScoreLevelsCount);
    NextShard.FillEvery(Metas.size());

    for (ui64 score = 0; score < ScoreLevelsCount; ++score) {
        ui64 firstQualifying = Metas.size();
        for (ui64 shardIdx = 0; shardIdx < Metas.size(); ++shardIdx) {
            if (Metas[shardIdx].Score >= score) {
                firstQualifying = shardIdx;
                break;
            }
        }
        if (firstQualifying == Metas.size()) {
            continue;
        }

        ui64 nextQualifying = firstQualifying;
        for (ui64 shardIdx = Metas.size(); shardIdx-- > 0;) {
            NextShard[score][shardIdx] = nextQualifying;
            if (Metas[shardIdx].Score >= score) {
                nextQualifying = shardIdx;
            }
        }
    }
}
```

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
ruslan-gar: We agreed to leave the semantica as is.
ShardSelector is renamed to LastSelectedShard

debnatkh:
I’d leave **one correctness comment before approving**, plus a couple of smaller test/design comments. The latest visible Filestore CI for commit `89b1838` is green. ([[GitHub](https://github.com/ydb-platform/nbs/pull/6784)][1])

* **[P1/P2] `CalcScore()` appears to assume that `stats[0].TotalBlocksCount` is the total capacity of the whole shard group.** The code computes `maxBlocksCount = max(stats[0].TotalBlocksCount / stats.size(), minBlocksCount)`. But `TShardStats::TotalBlocksCount` is populated independently from each shard’s own `GetStorageStats` response; the aggregation code copies each shard’s `src.GetTotalBlocksCount()` into that shard’s entry. ([[GitHub](https://github.com/ydb-platform/nbs/pull/6784/files)][2]) If `TotalBlocksCount` is genuinely per-shard capacity, dividing the first shard’s capacity by the number of shards makes the target smaller by a factor of N. It also makes the result depend on which shard happens to be at index 0 and silently assumes homogeneous capacities. I’d ask for the invariant to be demonstrated/documented, or change the calculation to use the actual global limit / capacities intentionally. The current unit test unintentionally bakes in the assumption by giving every shard `15_TB`/`25_TB` as `TotalBlocksCount` and then dividing it by five.  A useful review comment would be: **“Why is `stats[0].TotalBlocksCount / stats.size()` the per-shard target? `TotalBlocksCount` comes from the individual shard stats, so this seems to divide one shard’s capacity by N and makes scoring depend on shard 0. Can we either document the invariant that every shard reports the global filesystem size here, or derive this from the actual filesystem limit?”**

* **[Test] Add a heterogeneous-capacity test.** For example `{total=5TB, 10TB, 10TB}` with comparable utilization. The requirements only say “prefer shards with more free space”, so the new policy should not produce dramatically different behavior merely because `stats[0]` changes. The current deterministic test covers several `UsedBlocksCount` distributions but keeps `TotalBlocksCount` identical across all entries. ([[GitHub](https://github.com/ydb-platform/nbs/issues/6783)][3])

* **[Minor] The “incompatible features” check only reports a critical event but still enables the balancer.** That may be intentional, but the deterministic implementation explicitly ignores `fileSize`, `desiredFreeSpaceReserve`, and `minFreeSpaceReserve`, while the old policies use those values to avoid selecting shards that cannot fit the file. ([[GitHub](https://github.com/ydb-platform/nbs/pull/6784/files)][2]) With strict enforcement disabled, therefore, behavior is knowingly unsafe rather than merely suboptimal. The PR already reports `IncompatibleFeatures` for this combination, which is good; I’d just verify that “emit a critical event and continue” is really the desired rollout behavior rather than falling back to the previous policy. The service test currently verifies only that the counter increments. 

The rest looks reasonably structured. The simplified `CalcNextShard()` is easy to reason about and, because `ScoreLevelsCount` is fixed at 8, remains O(N) with respect to shard count, satisfying the issue’s linear-complexity requirement. ([[GitHub](https://github.com/ydb-platform/nbs/pull/6784/files)][2]) I also like that identical score updates no longer rebuild the transition matrix.

**Verdict:** I’d mark it **request changes / clarification on the `TotalBlocksCount` invariant**. If that invariant is valid and established elsewhere, I don’t see another obvious blocker from the diff I reviewed.

[1]: https://github.com/ydb-platform/nbs/pull/6784 "issue-6783: [Filestore] Deterministic weighted shard balancer by ruslan-gar · Pull Request #6784 · ydb-platform/nbs · GitHub"
[2]: https://github.com/ydb-platform/nbs/pull/6784/files "issue-6783: [Filestore] Deterministic weighted shard balancer by ruslan-gar · Pull Request #6784 · ydb-platform/nbs · GitHub"
[3]: https://github.com/ydb-platform/nbs/issues/6783 "[Filestore] Deterministic weighted shard balancer · Issue #6783 · ydb-platform/nbs · GitHub"

---

* For the first point it might make sense to add validation and critical event reporting in case of differing TotalBytesCount. Or even better, explicitly passing TotalBytesCount as a constructor param – same as done with precisionBytes,maxFileBlocks,desiredFreeSpaceReserve,minFreeSpaceReserve,

Other stuff is not significant

debnatkh (approved):

yegorskii (approved):

PR #6788, opened 2026-08-14
[Filestore] add logging for tablet draining process start and completion

debnatkh:
### Notes
We have noticed some drains taking longer than expected. We would like to add more logging to `service_actor_actions_drain_tablets.cpp`

### Issue
Small logging change – no issue

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_actions_drain_tablets.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_actions_drain_tablets.cpp
@@ -34,6 +34,13 @@ struct TDrainTabletsActionActor final
             ReplyWithError(ctx, MakeError(E_ARGUMENT, "Failed to parse input"));
             return;
         }
+
+        LOG_INFO(
+            ctx,
+            TFileStoreComponents::SERVICE,
+            "Draining tablets started, KeepDown=%d",
+            request.GetKeepDown());
+
         DrainTablets(ctx, request.GetKeepDown());
         Become(&TThis::StateWork);
     }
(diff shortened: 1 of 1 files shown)


WilyTiger (approved):

debnatkh:
I pulled up the diff (+17 −0 in `cloud/filestore/libs/storage/service/service_actor_actions_drain_tablets.cpp`). It's a small, low-risk logging-only change. Here's the review:

**What it does**
- Adds a `LOG_INFO` right after request parsing, before `DrainTablets(...)` is called, logging `"Draining tablets started, KeepDown=%d"`.
- Adds a `LOG_WARN` in the completion handler's error branch, logging `"Draining tablets failed: %s"` with the formatted/quoted error, before the existing `ReplyWithError`.
- Adds a `LOG_INFO` on the success path, logging `"Draining tablets finished"`, before `ReplyWithSuccess`.

**Assessment: looks good to merge.** A few minor, optional thoughts:

1. **`%d` for `KeepDown`** — if `GetKeepDown()` returns a `bool`/proto bool field, it'll promote fine to `%d`, but double check the proto field type is actually a bool/int and not something like a `TMaybe<bool>` that needs unwrapping first.
2. **Duration isn't explicit** — since the motivating issue is "some drains taking longer than expected," you get start/finish timestamps implicitly via the log timestamps, but it might be slightly more useful to log the elapsed duration directly in the "finished" message (e.g., using an actor-local start time) so it's greppable in one line rather than requiring correlating two log lines.
3. **No count/target info** — the "started" log only has `KeepDown`; if this actor knows how many tablets it's about to drain (or a batch/request ID), including that would make correlating "started" and "finished" log lines across concurrent drains easier, especially if this actor can run more than once in overlapping windows.
4. Everything else — using `TFileStoreComponents::SERVICE`, quoting the error with `FormatError(...).Quote()`, and placement before the reply calls — matches existing conventions in the file.

This is exactly the kind of low-risk, single-purpose PR that's fine to approve as-is; the above are "nice to have" suggestions rather than blockers, and CI already passed with one approval (WilyTiger) — just needs a second approver per the branch protection rule shown on the PR.

gy2411 (approved):

PR #6789, opened 2026-08-14
Filestore: reject create node if lock exists

neihar:
### Notes
Return "E_REJECTED" if node ref locked during node creation
Added unit test

### Issue
#6798

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
@@ -2665,16 +2665,14 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
             getNodeRefResponse->GetErrorReason());
     }
 
-    TABLET_TEST_4K_ONLY(ShouldRejectCreateNodeWhileDirectoryIsBeingCreatedInShard)
+    TABLET_TEST_4K_ONLY(
+        ShouldRejectCreateNodeWhileDirectoryIsBeingCreatedInShard)
     {
         // Sharded directory creation is two-phase: the directory tablet
         // commits the NodeRef (parent, name) -> (shardId, shardNodeName) and
         // only then creates the node in the shard. Until the shard node is
         // created the ref stays locked and a concurrent CreateNode with the
-        // same name must get E_REJECTED, not E_FS_EXIST - otherwise the
-        // caller's follow-up GetNodeAttr, which follows the ref into the
-        // shard, gets E_FS_NOENT (the mkdir=EEXIST + stat=ENOENT pair
-        // observed by concurrent mkdir -p clients). E_FS_EXIST may only be
+        // same name should get E_REJECTED. E_FS_EXIST should only be
         // returned once the ref is unlocked and the entry is fully visible.
 
         NProto::TStorageConfig storageConfig;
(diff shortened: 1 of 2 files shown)


cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -2664,6 +2664,187 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+            UNIT_ASSERT_VALUES_EQUAL(shardId, node.GetShardFileSystemId());
+            shardNodeName = node.GetShardNodeName();
+            UNIT_ASSERT(!shardNodeName.empty());
+            UNIT_ASSERT(response->Record.GetIsNodeRefLocked());
+        }
+
+        //
+        // ...and the shard hop (emulated here the same way the service actor
+        // does it) finds nothing - the ENOENT that reaches the client right
+        // after its mkdir got EEXIST.
```
qkrorlqr: let's replace the comments describing the bug with the comments describing how the code is supposed to work or with just the comments describing the stages of this test scenario

after this pr no one gets `EEXIST`
and at this layer it's not `EEXIST` but `E_FS_EXIST`

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -2664,6 +2664,187 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+        env.GetRuntime().DispatchEvents(TDispatchOptions{
+            .CustomFinalCondition = [&] ()
+            { return shardCreateEvent != nullptr; }});
+
+        //
+        // Client B: mkdir of the same name - the ref is locked, so the
+        // request should be rejected for a retry instead of EEXIST.
+        //
+
+        auto mkdirStatus = [&] (ui64 requestId) {
```
qkrorlqr: strange name - what does `Status` mean here? it's just mkdir or, better, `createNode` or `createDir` - at this layer there's no mkdir

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
```diff
@@ -2664,6 +2664,187 @@ Y_UNIT_TEST_SUITE(TIndexTabletTest_NodesInternal)
...
+        {
+            auto response = tablet.RecvCreateNodeResponse();
+            UNIT_ASSERT_VALUES_EQUAL_C(
+                S_OK,
+                response->GetStatus(),
+                FormatError(response->GetError()));
+        }
+
+        {
+            const auto error = mkdirStatus(102);
```
qkrorlqr: if we want to emulate the actual scenario then `requestId` should stay the same (101) because `E_REJECTED` is retried in `TDurableClient` and thus `requestId` doesn't change because it's generated by the guest

neihar: it should be indifferent for requestId, it can be new request from different client

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
neihar: ok

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
neihar: that we care only about returned Error, not the whole response

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
neihar: added check for 101 and 102

cloud/filestore/libs/storage/tablet/tablet_ut_nodes_internal.cpp
neihar: decided to make it more standard way, mkdir and return response and use .getError where needed.

qkrorlqr (approved):

yegorskii (approved):

SvartMetal (approved):

PR #6800, opened 2026-08-15
issue-5895: fastshard layout dump

qkrorlqr:
### Notes
Adding methods for fastshard layout dump. Will use them in the next PR to display fastshard layout on tablet monpage and will probably allow to fetch it via `ExecuteAction` as well.

This PR:
* moves `simple_template.h/cpp` to `cloud/storage/core/libs/common`
* supports for loops in `simple_template.h/cpp`
* adds unit tests for `simple_template.h/cpp` (now the code is complex enough to need them)
* adds `DumpLayoutHtml` and `DumpLayoutJson` methods to `IFileSystemShard`
* implements them for fastshard: memshard implementation is dummy, persistent shard implementation outputs per-component geometry

### Issue


qkrorlqr:
Claude review:

> I pulled the PR branch (head 00e66e7, 3 commits, base b6ad17e) and reviewed the full diff: the simple_template move + loop support, the new DumpLayoutHtml/Json interface methods, the fastshard implementation, resources, and tests. Overall it's clean work — the template engine is well-tested (nesting, shadowing, unterminated loops), all three IFileSystemShard implementers override the new pure virtuals, and the ya.make wiring is complete. I found two real issues, both in the layout numbers rather than the plumbing.
>
> 1. DataPages size/slot count is derived from BitCount, which overstates the actual data region by up to 16×
>
> TPageAllocator::Init does BitCount = RoundUp(pageClusterCount, bitsPerPage) — i.e. it rounds up to a whole 4 KB bitmap page, which is 32768 bits, so BitCount comes in multiples of 32768 clusters (1 GB of data at 32 KB/cluster). But the space Init actually reserves is pageClusterCount * PageClusterPageCount pages. With the test's ExpectedGroupCapacity = 64_MB, CalcPageClusterCount gives 2048 clusters (64 MB reserved), while the new GetDataSize() = BitCount * PageClusterSize reports 1 GB, and slotCount reports 32768 instead of 2048. The test hard-codes EXPECT_EQ(1_GB, ...), so it enshrines the inflated number rather than catching it. If the intent is "layout of what's reserved on the group," this should be pageClusterCount-based (which probably means storing pageClusterCount/bitmapPageCount in the allocator instead of deriving from BitCount). If the intent really is "bitmap-addressable space," that's worth a comment — and note it also means the dump silently reflects the pre-existing oddity that the bitmap can hand out clusters far past the reserved region.
>
> Related guard: GetBitmapSize() = PageSize * (BitCount / bitsPerPage) only works because BitCount is currently an exact multiple of bitsPerPage. If issue 1 is fixed by setting BitCount = pageClusterCount, this division truncates to 0 (2048/32768). Storing bitmapPageCount from Init avoids both problems at once.

This is a bug but it's not a bug in this PR - it's a bug in the bitmap initialization. I'll fix it separately. Bitmap visible bit count should be capped by the actual max page cluster count.

> 2. The test's contiguity claim doesn't match its own assertions
>
> shard_ut_layout.cpp says "The components must cover the group contiguously in the documented order, starting at offset 0," but the asserted values contain a gap: PageAllocatorBitmap ends at 76 KB + 4 KB = 80 KB, while DataPages starts at 96 KB. The 16 KB hole comes from cluster alignment plus the (pre-existing, not introduced here) double-count in FirstStoragePageClusterId = RoundUp(firstPageNo + bitmapPageCount, ...) — firstPageNo was already advanced by bitmapPageCount two lines earlier. Since this PR is exactly the thing that makes the layout visible, I'd either fix the misleading test comment, or — better — check whether that double-add is intentional; if it isn't, fixing it changes the expected offsets this test just locked in. Either way the dump currently presents a layout with an unexplained hole while the test asserts it's gap-free.
>
> Things I checked and deliberately did not flag: the per-iteration copy of vars in the loop renderer (fine at this scale), no HTML escaping in the template (all substituted values here are fixed names and integers; same behavior as the pre-move code), the "" vs "{}" asymmetry between empty HTML and JSON dumps (documented and tested), and the mixed gtest/unittest frameworks (already the repo's pattern).

This one is about a tiny mismatch between a unit test comment and the actual logic. Isn't really impactful.

WilyTiger (approved):

debnatkh (approved):

yegorskii (approved):

PR #6804, opened 2026-08-17
[Disk Manager] Remove snapshot legacy storage

BarkovBG:
Remove snapshot legacy storage


cloud/disk_manager/internal/pkg/dataplane/snapshot/config/config.proto
```diff
@@ -11,7 +11,7 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
 message SnapshotConfig {
     required persistence.PersistenceConfig PersistenceConfig = 1;
     optional string StorageFolder = 2 [default = "snapshot"];
-    optional string LegacyStorageFolder = 3 [default = "hardware/default/snapshot"];
+    reserved 3;
```
jkuradobery: Do you use any protobuf binary serialization for snapshot config? because otherwise we can just change numbers here and remove the field.

cloud/disk_manager/internal/pkg/services/images/protos/create_image_from_image_task.proto
```diff
@@ -9,13 +9,12 @@ option go_package = "github.com/ydb-platform/nbs/cloud/disk_manager/internal/pkg
...
 
 message CreateImageFromImageRequest {
-    reserved 4, 5, 7;
+    reserved 4, 5, 7, 8;
 
     string SrcImageId = 1;
     string DstImageId = 2;
     string FolderId = 3;
     repeated types.DiskPool DiskPools = 6;
-    bool UseS3 = 8;
```
jkuradobery, line 18: Please, avoid removing this flag. We did not fully migrate to S3,  migration to S3 and legacy storage are separate concerns.

jkuradobery (dismissed):

cloud/disk_manager/internal/pkg/services/pools/create_base_disk_task.go
```diff
@@ -105,51 +103,23 @@ func (t *createBaseDiskTask) Run(
 						DstDisk:                 t.request.BaseDisk,
 					},
 				)
-				if err != nil {
-					return err
-				}
```
gy2411, line 110: Why we remove this `if`? Maybe we should check this error right here. And similarly for error on the line 107.

Argumentatin: it is more reliable. If the code changes and we add change the code and we use := instaed of =, the error will not be handeled.

cloud/disk_manager/internal/pkg/services/images/create_image_from_snapshot_task.go
gy2411: We should also get rid of the `UseDataplaneTasks` flag (here and in other places), shouldn't we?
(Maybe in separate pr.)

gy2411 (dismissed):

jkuradobery (approved):

gy2411 (approved):

PR #6806, opened 2026-08-17
issue-6208: [Filestore] Introduce a new api action for diagnostic stats

szumal:
### Notes
The goal is to be able to send a request to each shard, instead of having `getStorageStats` on the main tablet be responsible for aggregating results from all shards.

### Issue


szumal:
This change was already included in https://github.com/ydb-platform/nbs/pull/6670

PR #6810, opened 2026-08-17
issue-5894: fixing an accept/shutdown race in fastshard server and sn server

qkrorlqr:
### Notes
Caught an unexpected `ECONNRESET` error instead of the expected `EIO` (`EIO` is generated by the `RecvAll()` code itself when it receives 0 bytes from the socket) in the `ServerTest.StopClosesIdleConnection` sn server test. This is the result of returning `RST` by the server instead of gracefully shutting down the connection (`FIN -> FIN+ACK -> ACK`). Happens because there can be a pending connection in the listen socket's backlog at the time when we shut the server down and thus there's a small time window between shutdown and accept4 during which we might end up realizing that we need to shut down.

Fixing this simply by `accept4`-ing everything that we can accept before closing the listen socket. There's actually still a small time window within which a connection can fall into socket's backlog but then get `RST`-ed but now there's a guarantee that this doesn't happen for the connections that were received before `Stop()` was called. And it seems to be impossible to guarantee `FIN`s anyway because there's no way to force `connect()` to fail via Linux socket API. Anyway, any sane client needs to be able to process `RST` so this is not an issue.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/fastshard/server/server.cpp
+++ b/cloud/filestore/libs/storage/fastshard/server/server.cpp
@@ -447,6 +447,32 @@ int AcceptFiberMain(TAcceptParams* params) noexcept
         if (which == 1) {
             acceptFuture.cancel();
             acceptFuture.wait();
+
+            //
+            // Drain connections whose handshake has already completed:
+            // their clients saw connect() succeed, so they must go
+            // through the regular handler shutdown (FIN via
+            // THandlerRegistry::Stop) instead of being reset when the
+            // listening socket is closed. Stop() waits for this fiber
+            // before stopping the registry, so every handler
+            // registered here is shut down and waited for.
+            //
+
+            for (;;) {
+                int cfd = ::accept4(
+                    lfd,
+                    nullptr /* addr */,
+                    nullptr /* addrlen */,
+                    SOCK_NONBLOCK | SOCK_CLOEXEC);
+                if (cfd >= 0) {
+                    RegisterHandler(cfd, *handlers, registry);
+                    continue;
+                }
+                if (errno == EINTR || errno == ECONNABORTED) {
+                    continue;
+                }
+                break;
+            }
             return 0;
         }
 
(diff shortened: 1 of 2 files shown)


WilyTiger (approved):

debnatkh (approved):

qkrorlqr:
Something's wrong with github checks - probably some effects of the yesterday's github inc. I tried rerunning them 10+ times - doesn't seem to work. I'll merge this and then will immediately pull the main branch and check locally that everything's okay. The scope of this change is pretty small - fastshard-only, affects only the connections that got into socket backlog right before shutdown was initiated.

qkrorlqr:
AI review:

> Two real issues, both small but worth addressing:
> 
> 1. The drain is unbounded under sustained inbound connections. Nothing stops the kernel from completing new handshakes into the backlog while the drain runs — ListenFd stays open and listening until the destructor. A client connecting in a tight loop (very plausible over loopback, which is exactly the test environment this fix targets) can keep the backlog non-empty, so the loop may not reach EAGAIN for a long time, and every accepted connection spawns a handler fiber that Handlers->Stop() must then wake and join. Before this PR Stop() was bounded by the number of live handlers; now its latency is coupled to the inbound connect rate. Since the PR description already accepts that post-Stop() connections may get RST, capping the drain at the listen backlog size (or a fixed iteration count) preserves the intended guarantee — "everything that completed its handshake before Stop() gets a FIN" — while restoring bounded shutdown.
> 
> 2. Unexpected accept4 errors are swallowed silently. The main accept loop returns errno, and the sn server's Stop() even logs that exit code specifically so unclean shutdowns are visible. The drain loop just breaks on anything outside EINTR/ECONNABORTED/EAGAIN — so under, say, EMFILE, the connections that were supposed to be drained get RST with no trace in the logs. That silently reproduces the exact flaky symptom this PR exists to eliminate, in a form that would be much harder to diagnose next time. A single SILK_WARN when errno != EAGAIN && errno != EWOULDBLOCK before breaking fixes it.
> 
> Not flagged: the 26-line copy-paste between the two files follows the pre-existing pattern (the whole AcceptFiberMain is already duplicated), the drain skipping Prune() is harmless since Handlers->Stop() runs immediately after, and the residual RST window the author describes is genuinely unavoidable with the Linux socket API — the analysis in the PR description is correct.

IMO not really important for this part of the code

PR #6811, opened 2026-08-17
Add OTLP trace primitives and validation

librarian:
### Notes

### Issue

### DIFF
--- a/.github/scripts/tracing/otlp/time.py
+++ b/.github/scripts/tracing/otlp/time.py
@@ -1,54 +1,91 @@
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass
-from typing import Any
+from decimal import Decimal, InvalidOperation, ROUND_HALF_UP
 
 
-class Ns(int):
-    """A non-negative nanosecond value."""
+def _rounded_ns(value: object, multiplier: int) -> int | None:
+    if isinstance(value, bool):
+        return None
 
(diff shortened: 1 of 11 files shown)


.github/scripts/tracing/otlp/time.py
```diff
@@ -0,0 +1,80 @@
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+from typing import Any
+
+
+class Ns(int):
```
WilyTiger: Ns does not provide meaningful unit safety despite being introduced as a dedicated nanosecond type. It accepts float and str values, silently truncates fractional values, loses its type after arithmetic, and automatically treats any integer passed to Interval as nanoseconds.
It can also lose precision when from_s() or from_ms() receives a floating-point value, especially one parsed from JSON. For example, 1.001 * 1_000_000_000 may evaluate to 1000999999.9999999, which int() truncates to 1_000_999_999 instead of 1_001_000_000. Absolute Unix timestamps are even more susceptible because a float cannot represent arbitrary nanoseconds at that magnitude.
Consider using plain integers with explicit _ns field names and separate, explicitly rounded conversion functions. If exact decimal input must be preserved, parse it as Decimal. If strict unit safety is required, Nanoseconds should be a proper value object rather than an int subclass

WilyTiger (dismissed):

robot-nbs (dismissed):

PR #6812, opened 2026-08-17
Normalize ya test trace inputs

librarian:
### Notes

### Issue


.github/scripts/tracing/yatrace/trace_loader_test.py
```diff
@@ -0,0 +1,329 @@
+import json
```
WilyTiger, line 1: looks like this test is not run in ci, workflow only collects .github/scripts/*_test.py and .github/scripts/tests, nested tracing/yatrace/trace_loader_test.py would be ignored

librarian: They will be enabled here https://github.com/ydb-platform/nbs/pull/6819/changes#diff-e266ed77ccb04479528a73cc9d6077c801064fef4bef6333ce38234356f7afe1R332-R333

If they will be enabled here they won't work because not everything is introduced yet.

.github/scripts/tracing/yatrace/trace_loader.py
```diff
@@ -0,0 +1,455 @@
...
+    tests = [event for event in events if event.name.startswith("subtest-")]
+    test_attempts = _attempt_groups(tests)
+    if not chunk_events:
+        attempts = _attempts(test_attempts)
+        timestamps = _attempt_timestamps(test_attempts)
+        return (Chunk(None, None, "", None, attempts, timestamps),) if tests else ()
+
+    chunks: dict[tuple[int, int, str] | None, list[_Event]] = {}
+    for event in chunk_events:
+        chunks.setdefault(event.chunk_identity, []).append(event)
```
WilyTiger: This can split two snapshots of the same chunk into two separate Chunk objects. For example, an early chunk-event may identify the chunk as (index=0, total=1, filename=""), while a later event identifies it as (index=0, total=1, filename="chunk-0"). These are two updates for chunk 0 of 1, but using the full chunk_identity as the dictionary key puts them into different groups. As a result, start and finish metrics are not merged, and lookup by the public (0, 1) key finds two candidates instead of one. Could we treat a missing filename as matching the single non-empty filename for the same (index, total)?

librarian: This will not happen in our codebase, but can happen if we will decide to run contrib/ydb or yt/ tests. Because this filename is property of chunks produced by FORK_TEST_FILES. Fixed nevertheless.

WilyTiger (approved):

PR #6813, opened 2026-08-17
Parse ya event logs and build statistics

librarian:
### Notes

### Issue

### DIFF
--- a/.github/scripts/tracing/yatrace/critical_path.py
+++ b/.github/scripts/tracing/yatrace/critical_path.py
@@ -1,0 +1,279 @@
+from __future__ import annotations
+
+import re
+from collections import defaultdict
+from dataclasses import dataclass
+from typing import Any, Mapping, Sequence
+
+from ..otlp import (
+    Interval,
+    Ns,
+    Span,
+    Trace,
+    update_span_attributes,
+)
+from .metrics import finite_number
+from .node import TEST_NODE_MARKERS, YaNode
+from .span_chunk import TestChunk
+
+CRITICAL_TASK_SUFFIX_RE = re.compile(r"(?:(?:-CACHED|-DYN_UID_CACHE))+$")
+
+
+@dataclass(frozen=True, slots=True)
+class YaCriticalPathEntry:
+    index: int
+    base_type: str
+    raw_type: str
+    text: str
+    uid: str
+    elapsed_ms: int | float | None
+    start_ms: int | float | None
+    end_ms: int | float | None
+    interval: Interval | None
+    is_test: bool
+
+    @classmethod
+    def from_raw(
+        cls,
+        index: int,
+        value: Mapping[str, Any],
+    ) -> YaCriticalPathEntry:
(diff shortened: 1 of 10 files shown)


.github/scripts/tracing/yatrace/critical_path.py
```diff
@@ -0,0 +1,279 @@
...
+        entry: YaCriticalPathEntry,
+    ) -> tuple[int, int, int, int] | None:
+        if entry.interval is None:
+            return None
+        overlap_ns = entry.interval.overlap(node.interval)
+        distance_ns = entry.interval.boundary_distance(node.interval)
+        return (
+            int(entry.base_type in {node.tool, node.tag}),
+            overlap_ns,
+            -distance_ns,
```
WilyTiger: -distance_ns raises TypeError because the Ns value object (otlp/time.py) has no __neg__

.github/scripts/tracing/yatrace/critical_path.py
```diff
@@ -0,0 +1,279 @@
...
+                candidate_indices = [
+                    index for index in records_by_uid.get(uid, ()) if index in available
+                ]
+            else:
+                candidate_indices = list(available)
+            scored = (
+                (score, index)
+                for index in candidate_indices
+                if (score := self._node_score(build_records[index], entry))
+                if score is not None and (uid or score[1] > 0)
```
WilyTiger: score[1] > 0 compares an Ns to a plain int, the dataclass-generated (order=True) comparison operators only work between two Ns instances

.github/scripts/tracing/yatrace/critical_path.py
```diff
@@ -0,0 +1,279 @@
...
+                        ),
+                    )
+                return max(
+                    matching_uid,
+                    key=lambda record: len(record.interval),
+                )
+        if interval is None:
+            return None
+        overlapping = [
+            record for record in test_nodes if interval.overlap(record.interval)
```
WilyTiger: the filters if interval.overlap(record.interval) / if chunk.overlap(interval) rely on Ns truthiness, but Ns defines no __bool__/__len__, so any instance (including Ns(0)) is truthy. The "overlap" filtering never actually filters anything, so an unrelated test/chunk can get matched

.github/scripts/tracing/yatrace/node.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def _chunk_index(outputs: tuple[str, ...]) -> int | None:
+    for output in outputs:
+        _, marker, relative = output.partition("/test-results/")
+        parts = relative.split("/")
+        if (
+            marker
+            and len(parts) > 1
+            and (match := re.fullmatch(r"chunk(\d+)", parts[1]))
```
WilyTiger: checks parts[1] instead of parts[0]. For a typical path like .../test-results/chunk3/run.json, the chunk number is never recognized (always None), even though the sibling _test_identity correctly uses parts[0]

WilyTiger (approved):

PR #6814, opened 2026-08-17
Project ya test and build spans

librarian:
### Notes

### Issue

### DIFF
--- a/.github/scripts/tracing/yatrace/build_operations.py
+++ b/.github/scripts/tracing/yatrace/build_operations.py
@@ -1,0 +1,251 @@
+from __future__ import annotations
+
+import math
+from collections import Counter
+from dataclasses import dataclass
+from typing import Any, Mapping, Sequence
+
+from ..otlp import Interval, Ns, Span
+from ..projector import SpanProjector
+from . import limits
+from .critical_path import YaCriticalPath, YaCriticalPathEntry
+from .node import YaNode
+from .statistics import YaBuildStatistics
+from .worker_spans import WorkerSpanProjector
+
+
+@dataclass(frozen=True, slots=True)
+class YaBuildOperations:
+    nodes: Sequence[YaNode]
+    failures: Mapping[str, int | None]
+    statistics: YaBuildStatistics
+    critical_path: YaCriticalPath
+    test_starts: Sequence[Ns]
+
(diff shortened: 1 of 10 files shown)


.github/scripts/tracing/yatrace/build_operations.py
```diff
@@ -0,0 +1,251 @@
...
+        attributes.update(
+            self.statistics.build_attributes(nodes, self.critical_path.build_entries)
+        )
+        return attributes
+
+    def _failures(self, nodes: Sequence[YaNode]) -> tuple[set[str], set[int]]:
+        uids = {node.uid for node in nodes if node.uid and node.uid in self.failures}
+        representatives = {
+            max(
+                (index for index, node in enumerate(nodes) if node.uid == uid),
```
WilyTiger, line 85: Are we worried about a huge evlog here? If ya make breaks badly (a shared header or a base library fails, and a large part of the build and tests go down with it), nodes can get very large. _failures then rescans the whole list once per failed uid, so the cost is O(failed_uids*nodes)

It could be done in one pass by grouping node indices by uid first:
```
indices_by_uid: dict[str, list[int]] = defaultdict(list)
for index, node in enumerate(nodes):
    if node.uid and node.uid in self.failures:
        indices_by_uid[node.uid].append(index)

representatives = {
    max(
        indices,
        key=lambda i: (nodes[i].kind == "execute", len(nodes[i].interval), nodes[i].end_ns),
    )
    for indices in indices_by_uid.values()
}
return set(indices_by_uid), representatives
```

It may be a non-issue in practice often. But since the cost is easy to avoid, it seemed worth asking

.github/scripts/tracing/yatrace/build_operations.py
```diff
@@ -0,0 +1,251 @@
...
+            sum(index in failed and index not in selected_set for index in candidates),
+        )
+
+    @staticmethod
+    def _select_commands(
+        nodes: Sequence[YaNode],
+        selected: Sequence[int],
+        critical: Mapping[int, YaCriticalPathEntry],
+        failed: set[int],
+    ) -> tuple[list[tuple[int, int, YaNode]], int]:
```
WilyTiger, line 147: such complex tuples are very hard to read, maybe worth to add some small dataclass for readability?

WilyTiger (approved):

Issue #6825, opened 2026-08-17
[Filestore] implement availability counters


SvartMetal:
A Shared Filesystem is considered **unavailable** during any `N-minute` interval in which, for at least one filesystem request type, at least one request was **outstanding** and every such request either failed with an **EIO** error, or **hung**.

**Note**: A request submitted during an interval and still outstanding when the interval ends is not classified for that interval. It does not contribute to determining whether the interval is available or unavailable

In all `N-minute` intervals in which the unavailability condition is not met, a Shared Filesystem is considered **available**.

Shared Filesystem availability for the measurement period shall be calculated as:
Number of `N-minute` intervals in which the Shared Filesystem was Available / Total number of `N-minute` intervals in the measurement period × 100%

### Definitions

A request is **outstanding** at a given point in time if it was submitted but had received no response, success or error, by that point.
A request is **hung** if it was **outstanding** for the entire duration of the `N-minute` interval.
Request type - a category of filesystem operation, Fuse request types:
```
readdir
readdirplus
opendir
releasedir
lookup
getattr
setattr
write
write_buf
read
open
create
release
mkdir
rmdir
unlink
rename
link
symlink
readlink
flush
fsync
```

EIO error - POSIX EIO error code (see https://man7.org/linux/man-pages/man3/errno.3.html)

SvartMetal:
It might be better to base the availability request types not on FUSE requests but on the request names already exposed in user stats, for example https://github.com/ydb-platform/nbs/blob/360c8a46179a00b2ad84471a6c43875af602c449/cloud/filestore/libs/diagnostics/user_counter.cpp#L48

This way there would be no need to refactor the existing user stats for consistency with FUSE requests

PR #6826, opened 2026-08-17
[Filestore] implement availability counters

SvartMetal:
### Notes

A Shared Filesystem is considered **unavailable** during any `N-minute` interval in which, for at least one filesystem request type, at least one request was **outstanding** and every such request either failed with an **EIO** error, or **hung**.

**Note**: A request submitted during an interval and still outstanding when the interval ends is not classified for that interval. It does not contribute to determining whether the interval is available or unavailable

In all `N-minute` intervals in which the unavailability condition is not met, a Shared Filesystem is considered **available**.

Shared Filesystem availability for the measurement period shall be calculated as:
Number of `N-minute` intervals in which the Shared Filesystem was Available / Total number of `N-minute` intervals in the measurement period × 100%

### Definitions

A request is **outstanding** at a given point in time if it was submitted but had received no response, success or error, by that point.
A request is **hung** if it was **outstanding** for the entire duration of the `N-minute` interval.
Request type - a category of filesystem operation, Fuse request types:
```
readdir
readdirplus
opendir
releasedir
lookup
getattr
setattr
write
write_buf
read
open
create
release
mkdir
rmdir
unlink
rename
link
symlink
readlink
flush
fsync
```

EIO error - POSIX EIO error code (see https://man7.org/linux/man-pages/man3/errno.3.html)

### Issue
#6825

### DIFF
--- a/cloud/filestore/libs/diagnostics/availability_counters.cpp
+++ b/cloud/filestore/libs/diagnostics/availability_counters.cpp
@@ -97,7 +97,7 @@ void TAvailabilityCounters::Register(NMonitoring::TDynamicCounters& counters)
         "Availability_MissingIntervals",
         true);   // derivative
 
-    // no intervals have been reported yet - start as available
+    // No intervals have been reported yet - start as available.
     *LastIntervalAvailableCounter = 1;
 
     // index 0 is EFileStoreAvailabilityRequestType::None and stays unused
@@ -264,20 +264,15 @@ void TAvailabilityCounters::FinishInterval()
 {
     bool intervalAvailable = true;
 
-    // index 0 is EFileStoreAvailabilityRequestType::None and stays unused
+    // Index 0 is EFileStoreAvailabilityRequestType::None and stays unused.
     for (size_t i = 1; i < RequestTypeStates.size(); ++i) {
         auto& state = RequestTypeStates[i];
 
         TGuard g{state.Lock};
 
-        // Inflight >= InflightStartedInInterval holds because every
-        // registered completion consumes exactly one registration stamp:
-        // completions of requests started in the current interval decrement
-        // both counters, completions of older requests decrement Inflight
-        // only, and there can be no more of the latter than there are older
-        // requests still counted in Inflight. The subtraction is clamped
-        // defensively so that a violation can never wrap it into a huge
-        // hung count.
+        // Inflight >= InflightStartedInInterval holds but the subtraction is
+        // clamped defensively so that a violation can never wrap it into a
+        // huge hung count.
         Y_DEBUG_ABORT_UNLESS(
             state.Inflight >= state.InflightStartedInInterval);
         // Requests that were outstanding at the interval start and are still
@@ -301,8 +296,8 @@ void TAvailabilityCounters::FinishInterval()
         const bool requestTypeAvailable =
             !hadFailedRequest || hadNonFailedRequest;
         if (!requestTypeAvailable) {
-            // the aggregated interval availability is the logical AND over
-            // the request types
+            // The aggregated interval availability is the logical AND over
+            // the request types.
             intervalAvailable = false;
         }
 
--- a/cloud/filestore/libs/diagnostics/availability_counters.cpp
+++ b/cloud/filestore/libs/diagnostics/availability_counters.cpp
@@ -15,11 +15,11 @@ namespace {
 
 // Limits the number of intervals finished by a single UpdateStats() call.
 // Protects against unbounded catch-up loops after large forward clock jumps.
-// With the default 5-minute interval this covers a stats-updater stall of up
+// With the default 2-minute interval this covers a stats-updater stall of up
 // to one hour, anything beyond that realigns without evaluation.
-constexpr size_t MaxIntervalsPerUpdate = 12;
+constexpr size_t MaxIntervalsPerUpdate = 30;
 
-constexpr TDuration DefaultIntervalDuration = TDuration::Minutes(5);
+constexpr TDuration DefaultIntervalDuration = TDuration::Minutes(2);
 
 }   // namespace
 
--- a/cloud/filestore/libs/diagnostics/availability_counters.cpp
+++ b/cloud/filestore/libs/diagnostics/availability_counters.cpp
@@ -97,7 +97,7 @@ void TAvailabilityCounters::EnableAndRegister(
     // No intervals have been reported yet - start as available.
     *LastIntervalAvailableCounter = 1;
 
-    // index 0 is EFileStoreAvailabilityRequestType::None and stays unused
+    // Index 0 is EFileStoreAvailabilityRequestType::None and stays unused.
     for (size_t i = 1; i < RequestTypeStates.size(); ++i) {
         auto& state = RequestTypeStates[i];
         auto requestCounters = counters.GetSubgroup(
@@ -121,7 +121,9 @@ void TAvailabilityCounters::EnableAndRegister(
     CountersRegistered.store(true, std::memory_order_release);
 }
 
-void TAvailabilityCounters::RequestStarted(TCallContext& callContext)
+void TAvailabilityCounters::RequestStarted(
+    TCallContext& callContext,
+    TInstant now)
 {
     if (!CountersRegistered.load(std::memory_order_acquire)) {
         return;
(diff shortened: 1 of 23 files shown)


qkrorlqr (dismissed):

yegorskii (dismissed):

SvartMetal:
AI review: https://chatgpt.com/share/6a8adb8c-dd6c-83eb-9a75-18e1201f8ebd

> 
```
1. Blocker: request events can be assigned to the wrong interval
RequestStarted() and RequestCompleted() obtain now, call RollIntervals(now), and only afterwards acquire the per-request-type lock. Meanwhile, another thread may roll that state into the next interval. The retry condition only checks whether now is at or beyond the current interval end; it does not protect against an event whose timestamp belongs to the already-published previous interval.
A concrete completion race:
Current interval ends at E.

completion:
    now = E - 1
    RollIntervals(now) returns because now < E
    pauses before acquiring the request-type lock

updater:
    advances past E
    rolls the request-type state
    classifies the outstanding request as hung
    publishes the interval as unavailable

completion:
    acquires the request-type lock
    now is still less than the current interval end
    records success in the new interval
The previous interval is reported unavailable even though the request completed before its boundary according to the supplied timestamp. The success is instead applied to the next interval.
The analogous start race is also wrong:
start:
    now = E - 1
    passes RollIntervals()
    pauses

updater:
    rolls the interval

start:
    increments InflightStartedInInterval in the new interval
That request is treated as having started in the new interval. If it remains outstanding for that entire interval, it can be incorrectly excluded from the hung-request count because it is considered “fresh.”
This contradicts the interface contract that says the supplied timestamp determines the wall-clock interval to which an event belongs. Production also samples Timer->Now() before calling the availability-counter methods, so this interleaving is possible in the real request path.
A second check of CurrentIntervalEndUs is insufficient: once the old interval has been classified and its counters published, detecting that the event is stale cannot repair it. Event processing and interval rollover need an actual linearization protocol, for example serialization through RollLock, or another generation-based design that prevents the interval from being published while a pre-boundary event is pending.
The tests need deterministic synchronization hooks around:
Return from the rollover fast path.
Acquisition of the request-type lock.
Publication of the rolled interval.
Required cases:
completion timestamp is boundary - 1µs, rollover wins the state lock, previous interval must still be available;
start timestamp is boundary - 1µs, rollover wins the state lock, request must be treated as originating in the previous interval.
The existing stress test is very unlikely to detect this reliably.
```
Implementation is not perfect but good enough

>
```
2. Test bug: request-name tests use the production mapping as their oracle
TEnv::RequestTypeByName() converts a string such as "lookup" to an enum by iterating over enum values and calling the production GetAvailabilityRequestTypeName(). The test then publishes counters using that same production mapping and asserts against the original string.
For example, suppose production accidentally swaps the names for Lookup and GetAttr:
Lookup  -> "getattr"
GetAttr -> "lookup"
Then:
env.Start("lookup");
selects GetAttr, because the production mapping says GetAttr is "lookup". The GetAttr counter is also published under "lookup", so the assertion for "lookup" passes.
Thus, the test suite can accept a completely wrong enum-to-metric-label mapping. The comment claiming that literal names make the assertions independent from production is only half true: assertion lookup is literal, but test setup is not independent.
Only Write and WriteBuf appear to have direct name checks that do not go through this helper; the rest of the mapping remains vulnerable to this self-confirming behavior.
Use an independent compile-time table in tests:
struct TExpectedRequestType
{
    ERequestType Type;
    TStringBuf Name;
};

constexpr TExpectedRequestType ExpectedTypes[] = {
    {ERequestType::Lookup, "lookup"},
    {ERequestType::GetAttr, "getattr"},
    // all remaining values
};
Then:
start requests directly by enum;
assert the exact expected label from the test table;
verify every supported enum value appears exactly once.
```
It is okay

>
```
3. Test bug: the successful-request integration test is vacuous
ShouldTrackAvailabilityOfSuccessfulRequests sends one successful create request and then checks that the aggregate interval and the create subgroup are available. Its comment claims this covers the request itself rather than merely an idle interval. It does not.
The implementation reports an interval as available for every request type that has no failure evidence. An entirely idle create type therefore receives the exact same result.
Deleting the actual FUSE request from this test would leave all asserted values unchanged. The test would also pass if ops.create stopped registering availability requests entirely.
The unit tests named ShouldReportSuccessfulRequestsAsAvailable and ShouldReportNonEioErrorsAsAvailable have the same weakness: removing their request start/completion calls makes them equivalent to the empty-interval test.
A meaningful successful-request integration test must create a result that differs from idleness. For example:
Interval N:
    start create request
    leave it outstanding

Interval N+1:
    request is initially carried as hung
    complete it successfully before interval end

Expected:
    create interval is available because a successful completion exists
Alternatively, issue one create request returning EIO and another successful create request in the same interval, then verify that the operation type is available. That proves successful completion actually affects classification.
```
Good enough

>
```
4. Test bug: the concurrency test checks a nearly tautological invariant
ShouldEnableConcurrentlyWithRequestProcessing finally asserts approximately:
total >= 2
total == available + unavailable
But RollInterval() increments TotalIntervals and then immediately increments exactly one of AvailableIntervals or UnavailableIntervals. The equality is structurally guaranteed by the implementation branch being tested; it says almost nothing about whether requests reached the correct interval or were classified correctly.
The test does not detect:
pre-boundary events applied after rollover;
requests leaked in the in-flight count;
unexpectedly unavailable intervals;
skipped intervals;
wrong interval assignment;
duplicate or partial registration;
incorrect per-request-type counters.
It also gives the same timestamp to each request’s start and completion while another thread advances fake time in a tight loop. The resulting number and placement of intervals are highly scheduler-dependent, but the test then deliberately avoids asserting any meaningful classification.
There are also unbounded busy-wait loops. A worker failure, scheduling issue, or future code change can leave the test spinning forever rather than failing with a useful diagnostic.
This test should be split:
A deterministic concurrent-enable test that checks exactly one set of counters is registered.
Deterministic rollover races using barriers and exact expected counters.
A bounded stress test with deadlines that asserts:
no unexpected unavailable intervals;
no missing intervals under its controlled clock progression;
the exact expected number of rolled intervals;
no remaining in-flight requests.
At present, this is the most obvious “bullshit” test in the suite: it creates lots of concurrency but validates almost none of the behavior affected by concurrency.
```
Good enough - the main purpose of this test is to exercise data races

>
```
5. MissingIntervals still does not raise the promised critical event
You previously said that missing intervals should not occur and that you would add a critical event. The current head still only increments Availability_MissingIntervals, realigns the interval boundary, and continues. There is no corresponding missing-interval critical event in critical_events.h; only the double-registration event was added.
The tests now normalize this behavior by deliberately skipping 31 or 60 intervals and checking ordinary counter continuation, without checking any critical event.
There is a related metric-semantic problem. The header describes Availability_LastIntervalAvailable as the state of the last finished interval. After a gap, however, the implementation does not know or publish the state of the actual last finished wall-clock interval; it retains the value from the last evaluated interval before the gap. The gauge is therefore stale while still appearing current.
At minimum:
emit the intended critical event when missingIntervals != 0;
test the event;
either document the gauge as “last evaluated interval” or explicitly invalidate/reset it after a gap.
```
I will add it in the next PR

>
```
6. Twenty-one of the twenty-two FUSE mappings have no end-to-end protection
The PR introduces a manual mapping from FUSE callbacks to 22 availability request types. The integration tests exercise only create; the availability-counter unit tests bypass the FUSE InitOps mapping and operate on availability enums directly.
I did not spot an obviously incorrect mapping in the current table, but the tests would not detect errors such as:
ops.lookup    -> GetAttr
ops.getattr   -> Lookup
ops.releasedir -> Release
The self-oracle problem from finding 2 makes this coverage gap worse.
The mapping should be extracted into a table that can be independently inspected and tested. A test should verify all 22 callback-to-enum-to-label triples. Full FUSE integration for all operations may be excessive, but the hardcoded table itself must have complete independent coverage.
```
I will fix it in the next PR

**UPDATE**: actually, I don't think I'll change this yet, because: 1. the mapping is easy to verify for correctness 2. we may change the SLA and base it not on FUSE requests but on the existing request names exposed in user stats

>
```
7. Minor: documentation still says five-minute intervals
The public header describes Availability_TotalIntervals as counting finished five-minute intervals, while the default interval configured by this implementation is two minutes.
This should describe configurable availability intervals, or explicitly say two minutes.
```
I will fix it in the next PR

neihar (approved):

yegorskii (approved):

cloud/filestore/libs/diagnostics/availability_counters.h
```diff
@@ -0,0 +1,234 @@
...
+// type, and requests outside the SLA (AvailabilityRequestType == None) are
+// ignored entirely.
+//
+// The EIO classification is based on TCallContext::GuestReplyErrno - the
+// errno actually sent to the guest - because the internal request error does
+// not always match the guest-visible outcome.
+//
+// Published sensors (registered on the per-filesystem per-client counters
+// subgroup of the request stats):
+//  * Availability_TotalIntervals       - derivative, finished 5-min intervals;
```
neihar, line 64: 2 mins?

cloud/filestore/libs/diagnostics/availability_counters.cpp
```diff
@@ -0,0 +1,423 @@
...
+    TCallContext& callContext,
+    TRequestTypeState& state)
+{
+    // A repeated registration without a completion in between would leak
+    // the previous registration's accounting: only one completion follows,
+    // so the extra inflight unit would be reported as hung forever.
+    if (callContext.AvailabilityIntervalSeqNo != 0) {
+        ReportAvailabilityCountersDoubleRegistration(
+            TStringBuilder() << "request type: "
+                << static_cast<ui32>(callContext.AvailabilityRequestType));
```
neihar, line 173: Print name?

cloud/filestore/libs/diagnostics/availability_counters.cpp
```diff
@@ -0,0 +1,423 @@
...
+    if (startSeqNo == state.IntervalSeqNo) {
+        Y_DEBUG_ABORT_UNLESS(state.InflightStartedInInterval);
+        // The request started in the current interval and thus can no longer
+        // be counted as a fresh outstanding request at the interval end.
+        if (state.InflightStartedInInterval > 0) {
+            --state.InflightStartedInInterval;
+        }
+    }
+
+    if (callContext.GuestReplyErrno == EIO) {
```
neihar, line 253: probably is fine, but will it count canceled request as "passed sla?"

cloud/filestore/libs/diagnostics/availability_counters.h
neihar: I will fix it in the next PR - I see this last string in mesage - ok

PR #6829, opened 2026-08-18
issue-5895: getaddrinfo shouldn't be called from fiber scheduler threads and from actor system threads

qkrorlqr:
### Notes
* calling `getaddrinfo()` only in thread-mode fibers (in silk's thread-mode workers, not in fiber scheduler threads)
* fetching own hostname upon filestore-server startup to make sure that `FQDNHostName()` uses a cached hostname instead of calling `getaddrinfo()` from a fiber or an actor system thread

### Issue

### DIFF
--- a/cloud/filestore/libs/daemon/server/bootstrap.cpp
+++ b/cloud/filestore/libs/daemon/server/bootstrap.cpp
@@ -22,6 +22,7 @@
 #include <cloud/filestore/libs/storage/core/probes.h>
 #include <cloud/filestore/libs/storage/fastshard/bootstrap/core.h>
 
+#include <cloud/storage/core/libs/common/hostname.h>
 #include <cloud/storage/core/libs/common/task_queue.h>
 #include <cloud/storage/core/libs/common/thread_pool.h>
 #include <cloud/storage/core/libs/diagnostics/stats_updater.h>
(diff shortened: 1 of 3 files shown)


qkrorlqr:
Claude:

> Small, focused PR (+36/−2): wraps two getaddrinfo calls in FiberScheduler::ThreadModeScope and warms the FQDNHostName() cache at filestore-server startup. The approach is sound. Serious issues I'd raise:
> 
> 1. Startup now hard-depends on DNS. The bootstrap comment says "failure fails the daemon." Previously, hostname resolution happened lazily; now a DNS outage during a restart prevents the daemon from coming up at all. For a storage daemon, restarting into a degraded-DNS environment is exactly the scenario you don't want to add a hard startup dependency to. At minimum, consider a fallback to the non-FQDN hostname (or an empty/placeholder value for BuildBackendInfo) after retries are exhausted, rather than failing boot. Also verify the retry policy in GetFqdnHostNameWithRetries is bounded — an unbounded retry loop here silently hangs startup.
> 

It's okay to expect dns to work on startup. If it doesn't and we absolutely need to launch this specific instance of filestore-server then we can add the proper line into /etc/hosts.

> 2. ThreadModeScope from non-fiber contexts. TClient::Connect and OpenTcp are wrapped unconditionally. If these are ever invoked from a plain thread or an actor thread (unit tests, other callers, future call sites), what does constructing ThreadModeScope outside a fiber do — no-op, assert, or UB? If it isn't a guaranteed safe no-op, this needs a guard, because the whole point of the PR is that these functions get called from mixed contexts.

It will work via the proxy fibers mechanism. And it's actually not supposed to be called from the outside of the fiber runtime so it doesn't even matter that much.
 
> 3. The fix is partial on these paths. In OpenTcp, the blocking connect() (and any subsequent blocking socket I/O) immediately follows the resolution and will stall a scheduler thread just as badly as getaddrinfo — a slow/unreachable peer blocks until the TCP timeout, which is typically much longer than a DNS lookup. If this path runs on fibers, the scope should cover the connect as well, or that should be an explicit follow-up in issue #5895.
> 

That's a nice thing to fix - in a separate PR.

> 4. Warm-up only covers filestore-server. The comment notes BuildBackendInfo calls FQDNHostName() on response paths. If any other binary (vhost, client, etc.) shares that code path, they still hit the original bug; the cache warm-up lives only in this one bootstrap.
> 

Yes, but the scope of this PR is filestore-server only.

> Not blocking, but I'd confirm: the retry callback logs via STORAGE_WARN capturing this — make sure logging is actually initialized at InitActorSystemPrerequisites time, since only InitConfigs() has run.

WilyTiger (approved):

yegorskii (approved):

PR #6830, opened 2026-08-18
issue-1751: [Filestore] Implement correct behavior for WriteBackCache when corruption is detected

e673:
### Notes

Implement correct behavior for WriteBackCache when corruption is detected.

Previous behavior:
- Incomplete object creation causing null pointer access in various scenarios, for example, metrics collection.
Both resulted in entering pod into crash loop;
- Triggering `Y_ABORT_UNLESS` when trying to add a write request.
Both resulted in a crash loop.

Expected behavior:
- WriteBackCache is fully constructed;
- WriteBackCache doesn't serve requests and hangs;
- No errors are sent to the guest;
- A critical event is reported.

Manual resolution should be conducted in order to resolve the corrupted state.

Note: this PR covers only the case when corruption is detected at initialization. If corruption happens during the operation, `Y_ABORT_UNLESS` will cause a one-time crash. This is acceptable because the logic to handle all corruption paths would be complicated. The goal is to avoid entering a crash loop.

### Issue

### DIFF
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -41,6 +41,7 @@ namespace NCloud::NFileStore{
     xxx(WriteBackCacheCorruptionError)                                         \
     xxx(WriteBackCacheDataLossError)                                           \
     xxx(WriteBackCacheImpossibleState)                                         \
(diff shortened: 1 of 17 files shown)


e673:
This PR was split into two PRs

PR #6831, opened 2026-08-18
issue-5432: retry operations on unconfirmed async handles during tablet restart grace period

budevg:
An async CreateHandle may return before its transaction is committed. If the
tablet restarts during this window, the returned handle is lost until the client
replays the queued ConfirmCreateHandle request.

During a configurable grace period after tablet startup, return E_REJECTED with
EF_INSTANT_RETRIABLE for operations referencing an unknown handle. This allows
reads and lock operations to wait for handle confirmation instead of failing
permanently with E_FS_BADHANDLE.


cloud/filestore/libs/storage/tablet/tablet_ut_handles.cpp
```diff
@@ -11,6 +12,62 @@ using namespace NKikimr;
 
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace {
+
```
qkrorlqr, line 14: nit: /////... should be here, not before `namespace {`

qkrorlqr (dismissed):

SvartMetal (dismissed):

SvartMetal (dismissed):

yegorskii (dismissed):

yegorskii (approved):

SvartMetal (approved):

PR #6835, opened 2026-08-18
issue-6608: implement quota usage tracking and management in TQuotaStore

debnatkh:
### Notes
In the same manner as the UsedNodes/UsedBytes count is modified, now per-quota usage is only modified

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/tablet/quota.cpp
+++ b/cloud/filestore/libs/storage/tablet/quota.cpp
@@ -4,6 +4,23 @@
 
 namespace NCloud::NFileStore::NStorage {
 
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+inline void ClampedAdd(ui64& value, i64 delta)
+{
+    if (delta >= 0) {
+        value += static_cast<ui64>(delta);
+        return;
+    }
+
+    const ui64 decrement = static_cast<ui64>(-delta);
+    value = Y_UNLIKELY(decrement > value) ? 0 : value - decrement;
+}
+
+}   // namespace
+
 ////////////////////////////////////////////////////////////////////////////////
 
 void TQuotaStore::UpdateQuota(const NProto::TQuota& quota)
@@ -42,18 +59,23 @@ void TQuotaStore::LoadUsage(const TQuotaUsage& usage)
     UsageByQuotaId[usage.QuotaId] = usage;
 }
 
-void TQuotaStore::UpdateUsage(ui32 quotaId, i64 bytesDelta, i64 nodesDelta)
+const TQuotaUsage* TQuotaStore::UpdateUsage(
+    ui32 quotaId,
+    i64 bytesDelta,
+    i64 nodesDelta)
 {
     if (quotaId == 0) {
-        return;
+        return nullptr;
     }
 
     auto [it, inserted] = UsageByQuotaId.try_emplace(quotaId);
     if (inserted) {
         it->second.QuotaId = quotaId;
     }
-    it->second.UsedBytes += bytesDelta;
-    it->second.UsedNodes += nodesDelta;
+    ClampedAdd(it->second.UsedBytes, bytesDelta);
+    ClampedAdd(it->second.UsedNodes, nodesDelta);
+
+    return &it->second;
 }
 
 const TQuotaUsage* TQuotaStore::FindUsage(ui32 quotaId) const
--- a/cloud/filestore/libs/storage/tablet/quota.cpp
+++ b/cloud/filestore/libs/storage/tablet/quota.cpp
@@ -14,6 +14,7 @@ void TQuotaStore::UpdateQuota(const NProto::TQuota& quota)
 void TQuotaStore::RemoveQuota(ui32 quotaId)
 {
     QuotaById.erase(quotaId);
+    UsageByQuotaId.erase(quotaId);
 }
 
 const NProto::TQuota* TQuotaStore::FindQuota(ui32 quotaId) const
@@ -34,6 +35,45 @@ TVector<NProto::TQuota> TQuotaStore::GetQuotas() const
     });
 
     return quotas;
+}
+
+void TQuotaStore::LoadUsage(const TQuotaUsage& usage)
+{
+    UsageByQuotaId[usage.QuotaId] = usage;
+}
+
+void TQuotaStore::UpdateUsage(ui32 quotaId, i64 bytesDelta, i64 nodesDelta)
+{
+    if (quotaId == 0) {
+        return;
+    }
+
+    auto [it, inserted] = UsageByQuotaId.try_emplace(quotaId);
+    if (inserted) {
+        it->second.QuotaId = quotaId;
+    }
+    it->second.UsedBytes += bytesDelta;
+    it->second.UsedNodes += nodesDelta;
+}
+
+const TQuotaUsage* TQuotaStore::FindUsage(ui32 quotaId) const
+{
+    return UsageByQuotaId.FindPtr(quotaId);
+}
+
+TVector<TQuotaUsage> TQuotaStore::GetUsages() const
+{
+    TVector<TQuotaUsage> usages;
+    usages.reserve(UsageByQuotaId.size());
+    for (const auto& [_, usage]: UsageByQuotaId) {
+        usages.push_back(usage);
+    }
+    // Make sure the usages are returned in a deterministic order
+    Sort(usages.begin(), usages.end(), [](const auto& lhs, const auto& rhs) {
+        return lhs.QuotaId < rhs.QuotaId;
+    });
+
+    return usages;
 }
 
 }   // namespace NCloud::NFileStore::NStorage
--- a/cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp
@@ -52,7 +52,42 @@ const NProto::TQuota& TIndexTabletState::SetQuota(
 void TIndexTabletState::DeleteQuota(IIndexTabletDatabase& db, ui32 quotaId)
 {
     db.DeleteQuota(quotaId);
+    db.DeleteQuotaUsage(quotaId);
     Impl->Quotas.RemoveQuota(quotaId);
+}
+
+void TIndexTabletState::LoadQuotaUsages(const TVector<TQuotaUsage>& usages)
+{
+    for (const auto& usage: usages) {
+        Impl->Quotas.LoadUsage(usage);
+    }
+}
+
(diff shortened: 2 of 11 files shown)


debnatkh:
Found three correctness issues in PR #6835:

* **[P1] Existing filesystems are not backfilled.** Startup only loads rows from the new `QuotaUsages` table; it never derives usage from existing nodes. After upgrading, quotas already containing nodes therefore report zero usage. A subsequent truncate/unlink applies a negative delta to zero, and because the counters are `ui64`, they wrap to enormous values. A migration or rebuild from existing nodes is required. ([[load path](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_actor_loadstate.cpp#L456-L464)](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_actor_loadstate.cpp#L456-L464), [[unsigned update](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/quota.cpp#L55-L56)](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/quota.cpp#L55-L56))

* **[P1] `ListQuotas` reports only the main tablet’s local usage.** Quota definitions are propagated to every shard, where each shard maintains its own counters, but listing simply returns `GetQuotaUsages()` locally. For sharded filesystems, usage from all shard tablets is omitted instead of being summed by quota ID. ([[listing implementation](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_actor_quota.cpp#L251-L257)](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_actor_quota.cpp#L251-L257), [[quota propagation](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_actor_quota.cpp#L171-L214)](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_actor_quota.cpp#L171-L214))

* **[P1] Deleting and recreating a quota loses usage for surviving nodes.** `DeleteQuota` removes the usage row without rejecting deletion of a non-empty quota or clearing quota IDs from its nodes. Updates while the quota is absent are ignored. Recreating the same ID consequently starts at zero despite those nodes still belonging to it; later deletion or shrinking of such a node can again underflow the unsigned counters. Either prevent [deletion](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp#L55-L60) while `UsedNodes != 0`, detach the nodes, or preserve/rebuild the accounting state. ([deletion](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp#L55-L60), [[ignored updates](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp#L79-L87)](https://github.com/ydb-platform/nbs/blob/b52f45ba0b99cb985a89388fabbde87ce97d5d1a/cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp#L79-L87))

---

1. Intentionally attaching is allowed for EMPTY dirs
2. QuotaUsageSharing is to be done in the next PR
3. Seamless detach is also intentional

cloud/filestore/libs/storage/tablet/quota.cpp
```diff
@@ -36,4 +37,43 @@ TVector<NProto::TQuota> TQuotaStore::GetQuotas() const
...
 }
 
+void TQuotaStore::LoadUsage(const TQuotaUsage& usage)
+{
+    UsageByQuotaId[usage.QuotaId] = usage;
+}
+
+void TQuotaStore::UpdateUsage(ui32 quotaId, i64 bytesDelta, i64 nodesDelta)
+{
+    if (quotaId == 0) {
```
SvartMetal, line 67: Can it really happen?

Y_DEBUG_ABORT_UNLESS?

cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp
```diff
@@ -52,7 +52,42 @@ const NProto::TQuota& TIndexTabletState::SetQuota(
...
+    return Impl->Quotas.GetUsages();
+}
+
+void TIndexTabletState::UpdateQuotaUsage(
+    IIndexTabletDatabase& db,
+    ui32 quotaId,
+    i64 bytesDelta,
+    i64 nodesDelta)
+{
+    if (quotaId == 0 || (bytesDelta == 0 && nodesDelta == 0)) {
```
SvartMetal, line 77: Same question about `quotaId == 0`

cloud/filestore/libs/storage/tablet/quota.cpp
debnatkh: It can, it is intentional – no quota => usage update is noop

debnatkh:
I pulled the diff and the surrounding code at the PR's head commit (`cae2414`). The PR itself is a clean, focused change — in-memory usage tracking in `TQuotaStore`, persistence via `WriteQuotaUsage`/`ReadQuotaUsages` (added by an earlier PR in the series), hooks in `CreateNodeWithId`/`UpdateNode`/`RemoveNode`, and reporting in `ListQuotas`. The node lifecycle hooks are placed well: creation via both `CreateNode` and `CreateHandle` inherits the parent's `QuotaId`, deferred deletion of open-but-unlinked files still flows through `RemoveNode` in the destroy-handle path, and `AddBlob` size extensions go through `UpdateNode`, so writes are counted. That said, here are the problems I found, roughly by severity:

1. **`ui64` underflow on negative deltas.** `TQuotaUsage::UsedBytes/UsedNodes` are `ui64`, but `UpdateUsage` adds signed `i64` deltas with no clamp. Any decrement below zero wraps to ~1.8×10¹⁹ and — worse — `UpdateQuotaUsage` then persists the wrapped value via `WriteQuotaUsage`. This isn't theoretical: quota *marking* of nodes already exists in main, so any filesystem that marked nodes before this PR deploys starts with zero counters; the first unlink or detach of a pre-existing marked node underflows immediately. There's no migration/recount of existing marked nodes, and no clamp-at-zero (or debug-assert) as a safety net.

> I made it clamped (same as normal UsedBytesCount)

2. **`DeleteQuota` never cleans up usage.** The DB layer already has `DeleteQuotaUsage(quotaId)`, but nothing calls it, and `TQuotaStore::RemoveQuota` doesn't erase the `UsageByQuotaId` entry either. After deleting a quota, its usage row lives forever in memory and on disk and keeps showing up in `ListQuotas`. And since nodes still carry the deleted `QuotaId`, any later size change on them resurrects/updates the ghost row.

> fixed

3. **Rename doesn't reconcile `QuotaId`.** `ExecuteTx_RenameNode` only touches parent timestamps and node refs — the moved child keeps its old `QuotaId`. So a file renamed out of a quota-marked subtree stays counted against that quota, and one renamed in is never counted. Given the series rejects conflicting markings elsewhere (`ShouldRejectMarkingWithDifferentExistingQuotaId`), rename across quota boundaries probably needs either reassignment or an explicit rejection; right now it silently corrupts the subtree invariant.

> rename is out of the picture for the scope of this PR

4. **Sharded filesystems aren't aggregated.** `SetQuota` propagates to shards and `CreateHandle` forwards `QuotaId` to shard requests, so shard-local nodes get marked and each shard tracks its own usage — but `HandleListQuotas` only reports the local tablet's usage. For multi-shard filesystems the leader's numbers undercount. If aggregation is planned for a later PR, worth a TODO.

> outside of this PR scope

5. **Zero rows and dead code in `UpdateQuotaUsage`.** Once usage drops to 0/0 the row is persisted and reported forever instead of being deleted. Also, the `if (usage)` check after `UpdateUsage` is dead — for `quotaId != 0`, `UpdateUsage` always creates the entry — and if it ever *were* null you'd silently get memory/DB divergence; having `UpdateUsage` return the updated struct (or asserting) would be tighter. The `quotaId == 0` guard is also duplicated between `TIndexTabletState::UpdateQuotaUsage` and `TQuotaStore::UpdateUsage`.

> also ok

6. **Semantics worth a comment: checkpoints and sparse files.** `RemoveNode` decrements usage even when a checkpoint keeps a history version that still occupies storage — defensible ("usage = live namespace"), but undocumented. Similarly, `UsedBytes` tracks logical `GetSize()`, so a sparse write at a large offset consumes quota without consuming storage. Both deserve an explicit note since this feeds future enforcement.

> same as other usedbytes convention in this PR

7. **Test gaps mirror the bugs above:** nothing covers delete-quota-with-usage, rename across quota boundaries, decrement-below-zero, or pre-existing marked nodes gaining counters; `ShouldPersistQuotaUsageAcrossReboot` checks `UsedNodes` but never `UsedBytes`.

> fixed

8. **Housekeeping:** the PR description is still the template placeholder ("Describe your PR…", "Put links to the related issues here") — it should link issue #6608 — and it's still marked draft, so some of the above (especially shard aggregation and enforcement) may be intentionally deferred.

If I had to pick the two must-fixes before this merges: the underflow-plus-persistence in (1), since it permanently corrupts stored counters, and the missing usage cleanup in (2), since the API for it already exists and the leak is user-visible via `ListQuotas`.

cloud/filestore/libs/storage/tablet/quota.cpp
SvartMetal: Comment is welcome

cloud/filestore/libs/storage/tablet/tablet_state_quotas.cpp
SvartMetal: Please add a comment at least

SvartMetal (approved):

agalibin (approved):

cloud/filestore/libs/storage/tablet/quota.cpp
debnatkh: Will add in the following PR

PR #6846, opened 2026-08-19
issue-1559: [Disk Manager] Add filesystem snapshot barrier mechanism

jkuradobery:
### Notes
Change Lock behavior to barrier with multiple acquirers behavior. LockFilesystemSnapshot() -> AcquireFilesystemSnapshotBarrier() UnlockFilesystemSnapshot() -> ReleaseFilesystemSnapshotBarrier().

When restoring a filesystem shard (which is exceptional situation caused by an incident), we need to be sure snapshot is not deleted. We want to be able to restore different shards simultaneously.

On the contrary to filesystem shard restoration, which is an incident mitigation method, filesystem from snapshot restoration should not require locking (we do not lock disk snapshots).
### Issue
#1559

### DIFF
--- a/cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/schema/schema.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/schema/schema.go
@@ -35,6 +35,22 @@ func Create(
 	err = db.CreateOrAlterTable(
 		ctx,
 		storageFolder,
+		"filesystem_snapshot_lock_holders",
+		persistence.NewCreateTableDescription(
+			persistence.WithColumn("snapshot_id", persistence.Optional(persistence.TypeUTF8)),
+			persistence.WithColumn("lock_task_id", persistence.Optional(persistence.TypeUTF8)),
+			persistence.WithPrimaryKeyColumn("snapshot_id", "lock_task_id"),
+		),
+		dropUnusedColumns,
+	)
+	if err != nil {
+		return err
+	}
+	logging.Info(ctx, "Created filesystem_snapshot_lock_holders table")
+
+	err = db.CreateOrAlterTable(
+		ctx,
+		storageFolder,
 		"deleting",
 		persistence.NewCreateTableDescription(
 			persistence.WithColumn("deleting_at", persistence.Optional(persistence.TypeTimestamp)),
@@ -176,7 +192,13 @@ func Drop(
 
 	logging.Info(ctx, "Dropping schema for dataplane filesystem snapshot storage in %v", db.AbsolutePath(storageFolder))
 
-	err := db.DropTable(ctx, storageFolder, "filesystem_snapshots")
+	err := db.DropTable(ctx, storageFolder, "filesystem_snapshot_lock_holders")
+	if err != nil {
+		return err
+	}
+	logging.Info(ctx, "Dropped filesystem_snapshot_lock_holders table")
+
+	err = db.DropTable(ctx, storageFolder, "filesystem_snapshots")
 	if err != nil {
 		return err
 	}
--- a/cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/schema/schema.go
+++ b/cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/schema/schema.go
@@ -35,7 +35,7 @@ func Create(
 	err = db.CreateOrAlterTable(
 		ctx,
 		storageFolder,
-		"filesystem_snapshot_lock_holders",
+		"filesystem_snapshot_locks",
 		persistence.NewCreateTableDescription(
 			persistence.WithColumn("snapshot_id", persistence.Optional(persistence.TypeUTF8)),
 			persistence.WithColumn("lock_task_id", persistence.Optional(persistence.TypeUTF8)),
@@ -46,7 +46,7 @@ func Create(
 	if err != nil {
 		return err
 	}
-	logging.Info(ctx, "Created filesystem_snapshot_lock_holders table")
+	logging.Info(ctx, "Created filesystem_snapshot_locks table")
 
 	err = db.CreateOrAlterTable(
 		ctx,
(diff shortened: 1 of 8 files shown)


debnatkh (dismissed):

SvartMetal:
Could you explain what these locks and barriers are and why you need them?

I also couldn't find any explanation in the comments or docs - could you add one?

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -288,46 +377,84 @@ func (s *storageYDB) deletingFilesystemSnapshot(
...
 }
 
-func (s *storageYDB) lockFilesystemSnapshot(
+func (s *storageYDB) acquireFilesystemSnapshotBarrier(
 	ctx context.Context,
 	session *persistence.Session,
 	snapshotID string,
 	lockTaskID string,
-) (locked bool, err error) {
+) error {
```
SvartMetal, line 384: \n 

and in other places

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -58,6 +58,96 @@ func (s *storageYDB) fetchFilesystemSnapshotByID(
...
+		if res.Err() != nil {
+			return "", task_errors.NewRetriableError(res.Err())
+		}
+
+		if ctx.Err() != nil {
+			return "", task_errors.NewRetriableError(ctx.Err())
+		}
+
+		return "", nil
+	}
```
SvartMetal, line 138: I'd like to see this logic encapsulated in the persistence module eventually

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -344,24 +471,29 @@ func (s *storageYDB) lockFilesystemSnapshot(
...
+	return nil
 }
 
-func (s *storageYDB) unlockFilesystemSnapshot(
+func (s *storageYDB) releaseFilesystemSnapshotBarrier(
 	ctx context.Context,
 	session *persistence.Session,
 	snapshotID string,
 	lockTaskID string,
 ) error {
```
SvartMetal, line 491: \n

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/schema/schema.go
```diff
@@ -176,7 +192,13 @@ func Drop(
 
 	logging.Info(ctx, "Dropping schema for dataplane filesystem snapshot storage in %v", db.AbsolutePath(storageFolder))
 
-	err := db.DropTable(ctx, storageFolder, "filesystem_snapshots")
+	err := db.DropTable(ctx, storageFolder, "filesystem_snapshot_lock_holders")
```
SvartMetal: Would be clearer to name it `filesystem_snapshot_barriers` or `filesystem_snapshot_locks`

It is OK to have multiple locks for one snapshot, I think

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -288,46 +377,84 @@ func (s *storageYDB) deletingFilesystemSnapshot(
 	return state.toFilesystemSnapshotMeta(), tx.Commit(ctx)
 }
 
-func (s *storageYDB) lockFilesystemSnapshot(
```
SvartMetal, line 291: lockFilesystemSnapshot was clearer

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -381,21 +514,47 @@ func (s *storageYDB) unlockFilesystemSnapshot(
...
+		snapshotID,
+	)
+	if err != nil {
+		return err
 	}
 
-	state.lockTaskID = ""
+	state.lockTaskID = nextLockTaskID
+	if nextLockTaskID != "" {
+		err = s.deleteFilesystemSnapshotLock(
```
SvartMetal, line 552: I don't understand this branch

If we get here, it means `state.lockTaskID == lockTaskID`. As far as I understand, we should delete the `lockTaskID` lock, not `nextLockTaskID`, which could be different from `lockTaskID`

jkuradobery: No, nextLockTaskID != "" when state.lockTask
== lockTaskID means "we release the current owner of snapshot so the next lock in queue becomes the snapshot owner". So we delete it from queue and update the snapshot record within the same transaction.

SvartMetal: Let's also add a comment that `state.lockTaskID` is not present in the `filesystem_snapshot_locks` table by construction

SvartMetal (dismissed):

SvartMetal (dismissed):

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -293,41 +381,80 @@ func (s *storageYDB) lockFilesystemSnapshot(
...
 	}
 
 	if len(states) == 0 {
-		return false, tx.Commit(ctx)
+		err = tx.Commit(ctx)
+		if err != nil {
+			return err
+		}
+
+		return task_errors.NewNonRetriableErrorf(
```
gy2411: Why this error is not silent? Seems that it can occur is some valid scenatio with concurrent deletion and lock acquire...

cloud/disk_manager/internal/pkg/dataplane/filesystem/snapshot/storage/storage_ydb.go
```diff
@@ -293,41 +381,80 @@ func (s *storageYDB) lockFilesystemSnapshot(
...
-	}
-
-	if len(state.lockTaskID) != 0 {
 		err = tx.Commit(ctx)
 		if err != nil {
-			return false, err
+			return err
 		}
 
+		return task_errors.NewNonRetriableErrorf(
```
gy2411: Why this error is not silent?

gy2411 (approved):

SvartMetal (approved):

PR #6847, opened 2026-08-19
[Disk Manager]  fix ydbd pg port collision

jkuradobery:
### Notes
When ydbd mon port is selected as 5432 and pg port is set as 5432 there is a port collision. fix it by using port-manager for pgwire port selection by setting YDB_ALLOCATE_PGWIRE_PORT environment variable.

### DIFF
--- a/cloud/disk_manager/test/recipe/ydb_launcher.py
+++ b/cloud/disk_manager/test/recipe/ydb_launcher.py
@@ -28,6 +28,8 @@ class YDBLauncher:
         self.__dynamic_storage_pools = dynamic_storage_pools
 
     def start(self):
+        # Make ydbd use the harness-allocated PGWire port; otherwise its default
+        # port 5432 may collide with a dynamically allocated monitoring port.
         os.environ["YDB_ALLOCATE_PGWIRE_PORT"] = "true"
         self.__cluster.start()
         for kimimr_node in list(self.__cluster.nodes.values()):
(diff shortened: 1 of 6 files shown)


cloud/disk_manager/test/recipe/ydb_launcher.py
```diff
@@ -26,6 +28,7 @@ def __init__(self, ydb_binary_path):
         self.__dynamic_storage_pools = dynamic_storage_pools
 
     def start(self):
+        os.environ["YDB_ALLOCATE_PGWIRE_PORT"] = "true"
```
SvartMetal, line 33: We need a comment that fully explains the problem and the workaround

gy2411 (dismissed):

contrib/ydb/public/tools/ydb_recipe/recipe.inc
```diff
@@ -1,4 +1,5 @@
 ENV(YDB_DRIVER_BINARY="contrib/ydb/apps/ydbd/ydbd")
+ENV(YDB_ALLOCATE_PGWIRE_PORT=true)
```
SvartMetal: This is YDB code

Are you planning to create a PR in the YDB repo?

jkuradobery: No, I'd rather change it in-place.

SvartMetal (approved):

gy2411 (approved):

PR #6861, opened 2026-08-21
[Blockstore] split local nonrepl loadtests by backend

kobzonega:
## Notes

Splits the shared local nonreplicated load test into explicit backend targets.

- add `NBS_LOCAL_NONREPL_BACKEND` to select the backend used by a target;
- keep the existing `dedicated` and `embedded` targets on AIO;
- add matching `dedicated-io-uring` and `embedded-io-uring` targets;
- register both new targets in `RECURSE_FOR_TESTS`;
- run RDMA cases only in AIO targets, while io_uring targets collect only io_uring cases;
- skip io_uring targets on Linux kernels older than 5.10.

## Issue

Arcadia Sandbox tasks run on an older Linux kernel by default. The io_uring targets are gated at Linux 5.10, the minimum version validated for these tests, so enabling them in Arcadia also requires infrastructure that provides a suitable VM and kernel environment. This PR contains only the common changes that can be upstreamed to GitHub.

### DIFF
--- a/cloud/blockstore/tests/loadtest/local-nonrepl/test.py
+++ b/cloud/blockstore/tests/loadtest/local-nonrepl/test.py
@@ -1,5 +1,7 @@
 import logging
 import os
+import re
+
 import pytest
 
 from cloud.blockstore.config.client_pb2 import TClientConfig
@@ -32,6 +34,17 @@ import yatest.common as yatest_common
 DEFAULT_BLOCK_SIZE = 4096
 DEFAULT_DEVICE_COUNT = 8
 DEFAULT_BLOCK_COUNT_PER_DEVICE = 262186  # 262144 + 42
+
+
+def kernel_version():
+    match = re.match(r"(\d+)\.(\d+)(?:\.(\d+))?", os.uname().release)
+    return tuple(int(value or 0) for value in match.groups())
+
+
+requires_disabled_ring = pytest.mark.skipif(
+    kernel_version() < (5, 10, 0),
+    reason="IORING_SETUP_R_DISABLED requires Linux >= 5.10",
+)
 
 
 def enable_lwtrace(kikimr, query_file):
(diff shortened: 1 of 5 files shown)


cloud/blockstore/tests/loadtest/local-nonrepl/ya.make
```diff
@@ -1,4 +1,5 @@
 RECURSE_FOR_TESTS(
     dedicated
     embedded
+    embedded-io-uring
```
budevg, line 5: what about `dedicated` tests ... they are using uring as well

cloud/blockstore/tests/loadtest/local-nonrepl/test.py
```diff
@@ -370,6 +370,10 @@ def __run_test(test_case, backend, use_rdma):
     'io_uring': DISK_AGENT_BACKEND_IO_URING,
 }
 
+SELECTED_BACKEND = os.getenv('NBS_LOCAL_NONREPL_BACKEND')
+if SELECTED_BACKEND:
+    BACKENDS = {SELECTED_BACKEND: BACKENDS[SELECTED_BACKEND]}
+
```
budevg, line 406: maybe just detect kernel version and skip io_uring test if kernel version is lower then min version required

budevg (changes requested):

cloud/blockstore/tests/loadtest/local-nonrepl/ya.make
kobzonega: Added a separate dedicated-io-uring test target

cloud/blockstore/tests/loadtest/local-nonrepl/test.py
kobzonega: We do not want to skip or mute these tests. We want to run them on the same kernel version used in Cloud. In Arcadia, the io_uring-specific ya.make targets will be patched to run inside a QEMU VM with Linux kernel 6.12

kobzonega: So I also added an explicit kernel version check in `test.py`

`io_uring` tests will be skipped if the tests infrastructure does not support the required environment and newer kernel

sharpeye (approved):

budevg (approved):

komarevtsev-d (approved):

PR #6866, opened 2026-08-21
issue-3781: TFileRingBuffer - operations should return verbose errors

e673:
### Notes
Current API of `TFileRingBuffer` makes it impossible to figure out if a method was completed with success or failed, and if failed — if it was caused by a corruption or an invalid argument.

This PR proposes the following changes in API:

- PushBack: `bool` -> `TPushBackResult` (bool, TError)
- Commit: `bool` -> `TError`
- Free: `bool` -> `TError`
- GetTag: `ui32` -> `TGetTagResult` (ui32, TError)
- SetTag: `void` -> `TError`
- Front: `TStringBuf` -> `TFrontResult` (TStringBuf, TError)
- PopFront: `void` -> `TPopFrontResult` (bool, TError)
- Visit: `void` -> `TError`
- GetMetadata: `TStringBufer` -> `TGetMetadataResult` (TStringBuf, TError)
- SetMetadata: `bool` -> `TSetMetadata` (bool, TError)
- IsCorrupted, SetCorrupted: these methods become thread-safe

General rule: an error is returned in the case of corruption or invalid argument.

This is needed for `WriteBackCache` and `HandleOpsQueue` so they can handle the situation when a corruption happens in the middle of processing.

Additional changes:
- Return an error when corruption is detected in Visit or Free
- Added missing UNIT_ASSERT check

### Issue
https://github.com/ydb-platform/nbs/issues/3781


e673:
AI review:

I pulled the PR (2 commits, 6 files: `file_ring_buffer.{h,cpp}`, two test files, and the two call sites `handle_ops_queue.cpp` and `persistent_storage.cpp`) and reviewed the full diff. Overall the direction is good, but there are several places where the implementation contradicts the new documented contracts, and — ironically — a few places where the new error information is dropped exactly where the PR says it's needed. The PR's stated goal is to make it possible to distinguish success from failure, and corruption from invalid argument, so WriteBackCache and HandleOpsQueue can handle mid-processing corruption — I'll judge it against that.

## Bugs / contract violations

**1. `SetTag` violates its own new documentation.** The header promises: "Returns an error if the pointer is invalid, **the tag value exceeds the value returned by GetMaxTag()** or if the buffer is corrupted." But the implementation never checks the tag, and it discards the `bool` returned by `WriteEntryHeader`:

```cpp
auto eh = Data()->ReadEntryHeader(it->second);
eh.Tag = tag;
Data()->WriteEntryHeader(it->second, eh);   // returns false if Tag > MaxTag — ignored
return {};
```

`WriteEntryHeader` in `file_ring_buffer_format.cpp` refuses to write when `header.Tag > MaxTag` (and MaxTag is 0 for pre-V5 formats), so `SetTag(ptr, 100)` is a **silent no-op reported as success**. Compare with `Commit`, which does check the same return value (`Y_ABORT_UNLESS(written)`).

> I've added an argument check and a test for SetTag

**2. Corruption detected *mid-operation* is still returned as success.** This is the exact scenario the PR exists for:

- `Visit()`: `VisitEntries` can hit an invalid entry, call `SetCorrupted("Invalid entry detected at VisitEntries")`, and stop early — but `Visit` then returns `{}`. The caller gets S_OK with a silently truncated iteration. It should return an error if `IsCorrupted()` after `VisitEntries` (this matters for WriteBackCache recovery, which replays entries via Visit).
- `Free()`: `EraseFreeEntriesFromFront` can set the corrupted flag ("Invalid front entry"), yet `Free` returns `{}` — and consequently `PopFront` returns `true` as well.

> Visit and Free should return an error in this case.

**3. A failed `Commit` leaks `CurrentAllocation` with no way to recover.** `Commit` can now fail via `ValidateAccess` (e.g., corruption flagged between `Alloc` and `Commit`, which is now possible since `SetCorrupted` is callable from another thread). `CurrentAllocation` isn't cleared on that path and there is no Discard/Abort API, so every subsequent `Alloc`/`PushBack` fails — and, because the `CurrentAllocation.HasValue()` check in `Alloc` comes **before** `ValidateAccess`, the caller gets a misleading `E_INVALID_STATE "Previous allocation is not committed"` instead of the corruption error, and the `ReportAccessToCorruptedFileRingBuffer` critical event never fires from that path.

> I've switched checks in `Alloc` — now it will fail with a corruption error.

**4. The error-code taxonomy undercuts the PR's goal.** Corruption maps to `E_INVALID_STATE`, but so does the "Previous allocation is not committed" usage error — a caller trying to distinguish corruption from misuse by code can't. Meanwhile `Commit` with no pending allocation returns `E_ARGUMENT "No allocation to commit"`, though `Commit` takes no arguments — that's a state error. I'd pick a consistent mapping (e.g., corruption → one dedicated code, API misuse → another) or callers will end up string-matching messages.

> This is not needed. In future, there will be a mandatory argument in `Commit`.

## Call sites still drop the new errors

**5. `THandleOpsQueue::PopFront` discards the result** (`RequestsToProcess.PopFront();`). Combined with `fs_impl.cpp::ProcessHandleOpsQueue`, a corrupted buffer produces an endless reschedule loop: `Empty()` is false, `Front()` returns `nullopt`, the recovery `PopFront()` silently does nothing on a corrupted buffer, and the function reschedules — firing `ReportHandleOpsQueueProcessError` on every tick. The loop shape is pre-existing, but the information needed to break out of it now exists and is thrown away.

> This is out of scope from this PR and will be resolved separately https://github.com/ydb-platform/nbs/pull/6666

**6. `THandleOpsQueue::Front` counts a corruption error as `IncrementParseErrorCount()`** — conflating precisely the two failure modes this PR separates. A corruption counter (or at least distinct handling) would be more honest.

> This is out of scope from this PR and will be resolved separately https://github.com/ydb-platform/nbs/pull/6666

**7. `persistent_storage.cpp`:** `Visit` and `SetTag` discard the returned errors, while `Commit`/`Free` convert corruption into `Y_ENSURE` exceptions. If the graceful handling is planned for a follow-up WriteBackCache PR, fine — but worth saying so explicitly, since until then a corruption detected between `Alloc` and `Commit` becomes a throw where it previously couldn't happen.

> I've replaced `Y_ENSURE` with a proper error return.

## Thread-safety commit

**8. The docs overpromise.** "All further operations on the buffer will fail once the flag is set" isn't guaranteed cross-thread: `IsCorrupted` uses a relaxed load with no synchronization, and an operation already past its `ValidateAccess` check will complete normally. `SetCorrupted` from another thread is not a safe "cancel" — I'd soften the wording so WriteBackCache authors don't rely on it. (The atomics themselves are fine; `exchange` defaults to seq_cst and the dedup of the critical event is correct.)

> The wording is:
> All **further** operations on the buffer will fail once the flag is set.
> It is ok if currently executing operations will complete successfully.

## Tests

- `while (rb.PushBack("ABCD").GetResult())` in `file_ring_buffer_accessor_ut.cpp` (twice) swallows errors — an error is indistinguishable from "buffer full", so the loop exits on a wrong premise. Assert `!HasError` inside the loop.

> Done

- `file_ring_buffer_accessor_ut.cpp` uses `UNIT_ASSERT_VALUES_EQUAL(true, <TResultOrError<bool>>)` but the required `Out<NCloud::TResultOrError<bool>>` specialization is only defined at the bottom of `file_ring_buffer_ut.cpp`. It links only because both TUs live in the same UNITTEST target — a fragile cross-file dependency (this smells like what broke the first CI build). Move the `operator==` helpers and `Out<>` specializations into a shared test header.

> Done

- The assertion sweep missed a few calls: `rb->PopFront()` in the Alloc-while-PopFront test and `rb->SetMetadata("abc")` in the migration test are still unchecked.

> Added asserts

## Nits

The variable `pushPackResult` in `handle_ops_queue.cpp` should be `pushBackResult` (appears twice). Header doc grammar is inconsistent ("On success, return true" → "returns"; `Free`'s doc is missing a final period), and the `Alloc` doc doesn't mention that a pending version migration makes it return `nullptr` even on an empty buffer — that surprised me reading the code and will surprise callers. Also, `PushBack` runs `ValidateAccess` and then `Alloc` repeats it, which gives corruption precedence over "uncommitted allocation" in `PushBack` but not in bare `Alloc` — worth aligning. In the PR description itself: "there methods become thread-safe" → "these", and "TStringBufer" → "TStringBuf".

> Unrelated to this PR

On CI: the latest run for f5419ac showed tasks+storage and disk_manager green with blockstore and filestore still in progress at the last bot update, and the previous revision had an actual build failure in tasks+storage plus one failed-build entry in the blockstore and filestore suites — so I'd wait for the remaining two suites before merging.

The most impactful fixes: check `WriteEntryHeader`'s result (and tag range) in `SetTag`, return an error from `Visit`/`Free` when corruption was detected mid-call, and stop discarding `PopFront`'s error in `THandleOpsQueue`.

cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp
```diff
@@ -99,15 +99,19 @@ class TFileRingBufferStorage: public IPersistentStorage
 
     void Commit() override
     {
-        bool success = Storage.Commit();
-        Y_ENSURE(success, "Failed to commit allocation");
+        auto res = Storage.Commit();
+        Y_ENSURE(
+            !HasError(res),
+            "Failed to commit allocation: " << FormatError(res));
```
SvartMetal, line 114: Can we get rid of exceptions usage?

e673: Yes, But this is unrelated to this PR and will be done in the next one
https://github.com/ydb-platform/nbs/pull/6867/changes#diff-4fd6698976acda0e55fdcf9397499b66e0c11ccd7e4517758b7d2296129c5ff6R102

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_accessor_ut.cpp
```diff
@@ -602,7 +615,8 @@ Y_UNIT_TEST_SUITE(TFileRingBufferAccessorTest)
     {
         TBootstrap b;
         b.Execute(
-            [](TFileRingBuffer& rb) { UNIT_ASSERT(rb.PushBack("ABC")); },
+            [](TFileRingBuffer& rb)
+            { UNIT_ASSERT_VALUES_EQUAL(true, rb.PushBack("ABC")); },
```
SvartMetal: Why not UNIT_ASSERT?

e673: `rb.PushBack` returns `TResultOrError<bool>`.
`operator == (bool value, TResultOrError<bool>)` checks that there is no error and the result equals to the value.

cloud/filestore/libs/vfs_fuse/handle_ops_queue.cpp
```diff
@@ -74,7 +74,7 @@ std::optional<NProto::TQueueEntry> THandleOpsQueue::Front()
     const auto req = RequestsToProcess.Front();
 
     NProto::TQueueEntry entry;
-    if (!entry.ParseFromArray(req.data(), req.size())) {
+    if (!entry.ParseFromArray(req.Data.data(), req.Data.size())) {
```
SvartMetal, line 84: Should we check `Front()` result for an error?

e673: Yes, but in a different PR:
https://github.com/ydb-platform/nbs/pull/6666

This PR doesn't change the bevahior of `THandleOpsQueue` and `TWriteBackCache`.

e673:
New AI review

I'll take a look at that PR.I reviewed the PR at its current head (`850ecb0`, 6 commits — note it's been force-pushed several times since the description was written). Here's what I found, roughly by severity.

## Actual bugs

**Wrong position in `Free()` corruption message** (`file_ring_buffer.cpp` ~L664)
```cpp
bool written = Data()->WriteEntryHeader(it->second, eh);
if (!written) {
    SetCorrupted(TStringBuilder() << "Cannot write entry header at "
                                  << CurrentAllocation.ActualPos);   // should be it->second
```
This was copy-pasted from `Commit()`. In `Free()`, `CurrentAllocation` is normally `CreateInvalid()`, so the critical event will report `INVALID_POS` instead of the entry that failed.

> Resolved

**`SetTag()` still ignores the `WriteEntryHeader` return value** while `Commit()` and `Free()` were changed to treat it as corruption. Same argument applies (position comes from `EntryMap`), so either all three should check it or none.

> Resolved

## Behavioral changes callers may not expect

- `Free()` now returns an error when `EraseFreeEntriesFromFront()` (or the migration it can trigger) detects corruption — even though the entry *was* successfully flagged free. In `persistent_storage.cpp` this hits `Y_ENSURE(...)` and throws `"Failed to free pointer ..."` from `TFileRingBufferStorage::Free`, which is called from `TWriteDataRequestManager::Evict`. Before this PR that path returned `true` and continued. The message is also misleading, since the free itself didn't fail.

> This is expected. Successful free on corruption should not be possible.

- `PopFront()` now calls `SetCorrupted()` on an invalid front entry where it previously returned silently. Intended per the commit log, but it's a new side effect.

> This is expected.

- `file_ring_buffer.h` swapped `#include <.../common/error.h>` for `error.pb.h`, so anything that got `HasError`/`MakeError`/`TResultOrError` transitively now breaks (the PR had to add the include to both `_ut.cpp` files for exactly this reason; the first CI run had build failures).

> No need to include broader files.

## API / design

- **Description doesn't match the code.** The description still says `TResultOrError<bool>`, `TResultOrError<TStringBuf>`, etc., but the last commit replaced all of that with seven macro-generated structs (`TPushBackResult`, `TFrontResult`, ...). `Alloc`'s change (`TResultOrError<char*>` → `TAllocResult`) isn't listed at all. A reviewer reading the description will review the wrong thing.

> Fixed

- The codebase already has `TResultOrError<T>` for exactly this "value or error" shape; introducing a macro that stamps out seven bespoke structs with `const` members diverges from convention. The `const` members also make the structs non-assignable (can't reassign an `auto res = rb.Front();`, can't sit in most containers).

> This is intended. The results are not expected to sit in containers.

- Both constructors in the macro are implicit. Because `TPopFrontResult`, `TPushBackResult` and `TSetMetadataResult` wrap `bool`, any pointer or integer silently converts: `return somePtr;` in a function returning `TPopFrontResult` compiles and yields `Removed = true`. Likewise `return -1;` into `TGetTagResult`.

> I've made them explicit but now more code is needed.

- None of the new error-returning methods are `[[nodiscard]]`. The stated purpose is to surface errors, yet in this very PR results are silently dropped by `Storage.Visit(...)` and `Storage.SetTag(...)` in `persistent_storage.cpp`, `RequestsToProcess.PopFront()` in `handle_ops_queue.cpp`, and `rb.PopFront()` in the test helper `PopAll`.

> I've made all methods [[nodiscard]]

- Callers weren't updated to use the new information. `THandleOpsQueue::AddCreateRequest` still maps every `!Pushed` to `QueueOverflow` (including corruption and `E_ARGUMENT`), and `THandleOpsQueue::Front()` never looks at `req.Error`. Fine if that's planned as a follow-up, but worth saying so.

> This is out of scope for this PR

- The macro invocations end with `);` after a body that already ends in `};`, producing `};;` inside the class — legal but trips `-Wextra-semi`.

> Fixed

## Documentation errors in the header

- `PushBack` doc says "`TAllocResult::Error` is set" — should be `TPushBackResult::Error`.

> Fixed

- `PopFront` doc says "`PopFrontResult::Removed`" (missing `T`) and "`TFrontResult::Error` is set" — should be `TPopFrontResult::Error`

> Fixed

- `GetMetadata` doc says "`TGetMetadataResult::Data`" but the field is named `Metadata`.

> Fixed

- `SetMetadata` doc says "if is cannot be set" and again references `TFrontResult::Error`.

> Fixed

- Field naming is inconsistent: `Front` → `Data`, `GetMetadata` → `Metadata`, `GetTag` → `Tag`.

> The proposed naming is correct

- `Validate` doc was rewritten into mixed imperative/third-person ("Set … and returns false. Return true … Fire … and doesn't visit").

> Fixed

## Test gaps

- No test for the new `Free()` → `EraseFreeEntriesFromFront()` corruption path (the "Return an error when corruption is detected in Visit or Free" item). `ShouldDetectCorruptionOnPopFront` corrupts the *front* entry, so it exercises `PopFront`'s own `IsInvalid()` check, not `Free`'s. Corrupting the second entry and freeing the first would cover it.

> I've added a test case for corrupting the second element

- `ForbidModificationOfCorruptedBuffer` dropped the old `Free` assertion and doesn't check `Free`, `GetTag`, `SetTag`, `GetMetadata`, or `SetMetadata` on the corrupted buffer.

> Added

- `GetTag`/`SetTag` with an invalid pointer aren't tested for `E_ARGUMENT`

> Added checks

- The randomized test compares only `.Pushed` between the reference and real implementations; the reference now returns `E_ARGUMENT`, so comparing `Error.GetCode()` would be cheap and catch drift.
- `PopAll` still loops on `!rb.Empty()` while ignoring `PopFront()`'s result — if a pop fails, the test hangs instead of failing.

## Minor

- The thread-safety change to `Corrupted` (relaxed load / seq_cst exchange) only makes the flag itself safe; the description's phrasing "become thread-safe" is fine, but the class-level `// Non-thread safe` comment plus per-method notes could be clearer that this is the *only* concurrent guarantee.

> The current comment is ok

- `std::move` is used in the header without an explicit `<utility>` include.

> Not needed

- Commit history contains several "AI review fix" commits that should probably be squashed before merge.

> Will be done

The core refactor is sound and the tests were updated thoroughly; the two things I'd block on are the `CurrentAllocation.ActualPos` copy-paste in `Free()` and the description/code mismatch.

SvartMetal (approved):

yegorskii (approved):

PR #6867, opened 2026-08-21
issue-1751: [Filestore] WriteBackCache should not crash on corruption

e673:
### Notes

Implement correct behavior for WriteBackCache when corruption is detected.

Previous behavior:
- Incomplete object creation caused null pointer access in various scenarios, for example, metrics collection.
- Multiple `Y_ABORT_UNLESS` in different places, for example, when trying to add a write request.
Both resulted in a crash loop.

Expected behavior when storage corruption happens:
- WriteBackCache is fully constructed;
- WriteBackCache doesn't serve requests and hangs;
- No errors are sent to the guest;
- Critical events are reported instead of `Y_ABORT_UNLESS`.

Manual resolution should be conducted in order to resolve the corrupted state.

### Issue

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
@@ -33,6 +33,9 @@ struct TCachedData
     // This is needed to avoid truncation when there are unflushed data parts
     // beyond the requested range.
     ui64 ReadDataByteCount = 0;
+
+    // WriteBackCache is in failed state
+    bool Failed = false;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
@@ -33,9 +33,6 @@ struct TCachedData
     // This is needed to avoid truncation when there are unflushed data parts
     // beyond the requested range.
     ui64 ReadDataByteCount = 0;
-
-    // WriteBackCache is in failed state
-    bool Failed = false;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
@@ -342,14 +342,8 @@ TReadResponseBuilder::TReadResponseBuilder(
 
 std::optional<NProto::TReadDataResponse>
 TReadResponseBuilder::TryFullyServeFromCache(
-    TWriteBackCacheState& state,
-    TNodeCachedDataPin pin) const
+    const TCachedData& cachedData) const
 {
-    auto cachedData = state.GetCachedData(NodeId, Offset, Length, pin);
-    if (cachedData.Failed) {
-        return std::nullopt;
-    }
-
     Validate(cachedData, Length);
 
     ui64 contiguousCachedDataByteCount = 0;
@@ -371,22 +365,16 @@ TReadResponseBuilder::TryFullyServeFromCache(
 
 bool TReadResponseBuilder::AugmentResponseWithCachedData(
     NProto::TReadDataResponse& response,
-    TWriteBackCacheState& state,
-    TNodeCachedDataPin pin) const
+    const TCachedData& cachedData) const
 {
-    auto cachedData = state.GetCachedData(NodeId, Offset, Length, pin);
-    if (cachedData.Failed) {
-        return false;
-    }
-
     Validate(cachedData, Length);
 
-    AugmentResponseWithCachedData(response, cachedData);
+    DoAugmentResponseWithCachedData(response, cachedData);
 
     return !cachedData.Parts.empty();
 }
 
-void TReadResponseBuilder::AugmentResponseWithCachedData(
+void TReadResponseBuilder::DoAugmentResponseWithCachedData(
     NProto::TReadDataResponse& response,
     const TCachedData& cachedData) const
 {
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
@@ -346,6 +346,10 @@ TReadResponseBuilder::TryFullyServeFromCache(
     TNodeCachedDataPin pin) const
 {
     auto cachedData = state.GetCachedData(NodeId, Offset, Length, pin);
+    if (cachedData.Failed) {
+        return std::nullopt;
+    }
+
     Validate(cachedData, Length);
 
     ui64 contiguousCachedDataByteCount = 0;
@@ -371,6 +375,10 @@ bool TReadResponseBuilder::AugmentResponseWithCachedData(
     TNodeCachedDataPin pin) const
 {
     auto cachedData = state.GetCachedData(NodeId, Offset, Length, pin);
+    if (cachedData.Failed) {
+        return false;
+    }
+
     Validate(cachedData, Length);
 
     AugmentResponseWithCachedData(response, cachedData);
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
@@ -80,8 +80,8 @@ private:
     const TString LogTag;
     const TString FileSystemId;
     const TString FilePath;
+    const IPersistentStoragePtr PersistentStorage;
 
-    IPersistentStoragePtr PersistentStorage;
     TWriteBackCacheState State;
 
     std::atomic<bool> DrainRequested = false;
@@ -262,7 +262,7 @@ public:
              promise = std::move(promise)](
                 const TFuture<NProto::TReadDataResponse>& future) mutable
         {
-            auto response = UnsafeExtractValue(future);
+            auto response = ExtractResponse(future);
             bool shouldHang = false;
 
             if (auto self = ptr.lock()) {
@@ -454,7 +454,7 @@ private:
         auto batchBuilder =
             RequestBuilder->CreateWriteDataRequestBatchBuilder(nodeId);
 
-        State.VisitUnflushedRequestsFromFrontFlushBatch(
+        bool success = State.VisitUnflushedRequestsFromFrontFlushBatch(
             nodeId,
             [&batchBuilder](const TCachedWriteDataRequest* request)
             {
@@ -463,11 +463,31 @@ private:
                     request->GetBuffer());
             });
 
+        if (!success) {
+            // This may happen if the storage is corrupted or an internal
+            // problem in the logic has happened. This is fatal and should
+            // result in suspending all operations. Flush will remain in-flight
+            // until manual resolution takes place.
+            return;
+        }
+
         auto writeDataBatch = batchBuilder->Build();
 
         if (writeDataBatch.Requests.empty()) {
-            // This may happen when the cache entered failed state after
-            // flush has been scheduled
+            // Flush can be scheduled only when there are requests to flush.
+            // The only reason why VisitUnflushedRequestsFromFrontFlushBatch()
+            // may return an empty batch for a non-empty unflushed queue is a
+            // presence of a barrier with BarrierId less than SequenceId for all
+            // unflushed requests. This cannot happen because:
+            // - newly added barriers cannot have BarrierId less than SequenceId
+            //   for any existing WriteData request;
+            // - flush cannot be scheduled if an existing barrier prevents it.
+            ReportWriteBackCacheImpossibleState(Sprintf(
+                "Flush is scheduled for node %lu but flush batch is empty",
+                nodeId));
+
+            // We do not fail flush because it may cause dropping WriteData
+            // requests
             State.FlushSucceeded(nodeId, 0);
             return;
         }
@@ -509,9 +529,7 @@ private:
 
             handle = State.GetLiveHandle(flushState->GetNodeId());
             if (handle == NProto::E_INVALID_HANDLE) {
-                // This may happen when WriteBackCache entered failed state
-                // concurrently after FlushFailed was called
-                // We stop any further processing including metrics reporting
+                ScheduleRetryFlush(flushState);
                 return;
             }
         }
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
@@ -109,61 +109,36 @@ public:
               args.ClientId.c_str()))
         , FileSystemId(args.FileSystemId)
         , FilePath(args.FilePath)
+        , PersistentStorage(CreateFileRingBufferPersistentStorage(
+              args.Stats->GetPersistentStorageStats(),
+              {.FilePath = args.FilePath, .DataCapacity = args.CapacityBytes},
+              Log,
+              LogTag))
         , State(
               *this,
+              PersistentStorage,
               Timer,
               args.Stats->GetWriteBackCacheStateStats(),
               args.Stats->GetWriteDataRequestManagerStats(),
               args.Stats->GetNodeStateHolderStats(),
               BuildFlushBatchLimits(args),
               LogTag)
-    {
-        auto createPersistentStorageResult =
-            CreateFileRingBufferPersistentStorage(
-                args.Stats->GetPersistentStorageStats(),
-                {.FilePath = args.FilePath, .DataCapacity = args.CapacityBytes},
-                Log,
-                LogTag);
-
-        if (HasError(createPersistentStorageResult)) {
-            ReportWriteBackCacheCorruptionError(
-                TStringBuilder()
-                << LogTag
-                << " WriteBackCache persistent storage initialization failed: "
-                << createPersistentStorageResult.GetError()
-                << ", FilePath: " << args.FilePath.Quote());
-            return;
-        }
-
-        PersistentStorage = createPersistentStorageResult.ExtractResult();
-
-        // File ring buffer should be able to store any valid TWriteDataRequest.
-        // Inability to store it will cause this and future requests to remain
-        // in the pending queue forever (including requests with smaller size).
-        // Should fit 1 MiB of data plus some headers (assume 1 KiB is enough).
-        Y_ABORT_UNLESS(
-            PersistentStorage->GetMaxSupportedAllocationByteCount() >=
-            1024 * 1024 + 1016);
-    }
+    {}
 
     // This method should be called outside TImpl constructor because
     // it captures weak_from_this()
     void Init()
     {
-        if (!PersistentStorage) {
-            return;
-        }
+        auto error = State.Init();
 
-        if (!State.Init(PersistentStorage)) {
-            ReportWriteBackCacheCorruptionError(
+        if (HasError(error)) {
+            ReportWriteBackCacheInitializationError(
                 TStringBuilder()
-                << LogTag
-                << " WriteBackCache failed to deserialize requests from the "
-                   "persistent storage due to corruption"
-                << ", FilePath: " << FilePath.Quote());
+                << LogTag << " WriteBackCache initialization failure: "
+                << FormatError(error) << ", FilePath: " << FilePath.Quote());
+        } else {
+            ScheduleAutomaticFlushIfNeeded();
         }
-
-        ScheduleAutomaticFlushIfNeeded();
     }
 
     void ScheduleAutomaticFlushIfNeeded()
@@ -453,20 +428,8 @@ private:
         auto writeDataBatch = batchBuilder->Build();
 
         if (writeDataBatch.Requests.empty()) {
-            // Flush can be scheduled only when there are requests to flush.
-            // The only reason why VisitUnflushedRequestsFromFrontFlushBatch()
-            // may return an empty batch for a non-empty unflushed queue is a
-            // presence of a barrier with BarrierId less than SequenceId for all
-            // unflushed requests. This cannot happen because:
-            // - newly added barriers cannot have BarrierId less than SequenceId
-            //   for any existing WriteData request;
-            // - flush cannot be scheduled if an existing barrier prevents it.
-            ReportWriteBackCacheImpossibleState(Sprintf(
-                "Flush is scheduled for node %lu but flush batch is empty",
-                nodeId));
-
-            // We do not fail flush because it may cause dropping WriteData
-            // requests
+            // This may happen when the cache entered failed state after
+            // flush has been scheduled
             State.FlushSucceeded(nodeId, 0);
             return;
         }
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
@@ -223,48 +223,86 @@ public:
             return MakeFuture(std::move(response));
         }
 
+        auto promise = NewPromise<NProto::TReadDataResponse>();
+
         // Prevent unflushed data parts from being evicted from storage until
         // the response is completed
         const auto pin = State.PinCachedData(request->GetNodeId());
 
         TReadResponseBuilder responseBuilder(*request);
-        if (auto response = responseBuilder.TryFullyServeFromCache(State, pin))
+
+        auto cachedData = State.GetCachedData(
+            request->GetNodeId(),
+            request->GetOffset(),
+            request->GetLength(),
+            pin);
+
+        if (!cachedData) {
+            // WriteBackCache is in failed state - hang the request
+            State.UnpinCachedData(request->GetNodeId(), pin);
+            State.AddHangingRequest(promise);
+            return promise.GetFuture();
+        }
+
+        if (auto response = responseBuilder.TryFullyServeFromCache(*cachedData))
         {
             State.UnpinCachedData(request->GetNodeId(), pin);
             InternalStats->AddReadDataStats(
                 EReadDataRequestCacheStatus::FullHit);
-            return MakeFuture(std::move(*response));
+            promise.SetValue(std::move(*response));
+            return promise.GetFuture();
         }
 
-        auto callback = [ptr = weak_from_this(),
-                         responseBuilder = std::move(responseBuilder),
-                         pin](TFuture<NProto::TReadDataResponse> future)
+        auto future = promise.GetFuture();
+
+        auto callback =
+            [ptr = weak_from_this(),
+             responseBuilder = std::move(responseBuilder),
+             pin,
+             promise = std::move(promise)](
+                const TFuture<NProto::TReadDataResponse>& future) mutable
         {
             auto response = UnsafeExtractValue(future);
+            bool shouldHang = false;
 
             if (auto self = ptr.lock()) {
                 if (!HasError(response)) {
-                    bool cachedDataApplied =
-                        responseBuilder.AugmentResponseWithCachedData(
-                            response,
-                            self->State,
-                            pin);
+                    auto cachedData = self->State.GetCachedData(
+                        responseBuilder.GetNodeId(),
+                        responseBuilder.GetOffset(),
+                        responseBuilder.GetLength(),
+                        pin);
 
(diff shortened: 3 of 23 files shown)


cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
```diff
@@ -33,6 +33,9 @@ struct TCachedData
     // This is needed to avoid truncation when there are unflushed data parts
     // beyond the requested range.
     ui64 ReadDataByteCount = 0;
+
+    // WriteBackCache is in failed state
```
SvartMetal: Let's add a comment about what failed state is (corruption, etc.)

cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
```diff
@@ -371,6 +375,10 @@ bool TReadResponseBuilder::AugmentResponseWithCachedData(
     TNodeCachedDataPin pin) const
 {
     auto cachedData = state.GetCachedData(NodeId, Offset, Length, pin);
+    if (cachedData.Failed) {
+        return false;
+    }
```
SvartMetal: Looks dangerous

cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
```diff
@@ -346,6 +346,10 @@ TReadResponseBuilder::TryFullyServeFromCache(
     TNodeCachedDataPin pin) const
 {
     auto cachedData = state.GetCachedData(NodeId, Offset, Length, pin);
+    if (cachedData.Failed) {
+        return std::nullopt;
+    }
```
SvartMetal: Also dangerous

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
```diff
@@ -453,20 +428,8 @@ class TWriteBackCache::TImpl final
...
         if (writeDataBatch.Requests.empty()) {
-            // Flush can be scheduled only when there are requests to flush.
-            // The only reason why VisitUnflushedRequestsFromFrontFlushBatch()
-            // may return an empty batch for a non-empty unflushed queue is a
-            // presence of a barrier with BarrierId less than SequenceId for all
-            // unflushed requests. This cannot happen because:
-            // - newly added barriers cannot have BarrierId less than SequenceId
-            //   for any existing WriteData request;
-            // - flush cannot be scheduled if an existing barrier prevents it.
-            ReportWriteBackCacheImpossibleState(Sprintf(
```
SvartMetal, line 464: Do we need to delete it from critical_events.h?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
```diff
@@ -87,17 +92,33 @@ TFuture<TWriteDataResponse> TWriteBackCacheState::AddWriteDataRequest(
                 "WriteBackCache doesn't accept new WriteData requests"));
     }
 
-    auto variant = RequestManager.AddRequest(std::move(request));
+    if (FailedFlag) {
```
SvartMetal: IsFailed?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.h
```diff
@@ -67,52 +75,69 @@ class TWriteDataRequestManager
...
+     * TAddRequestResult::Failed == false if the storage is full, backpressure
+     * is in effect or the pending queue is empty.
      *
-     * Returns nullptr if there are no pending requests or the persistent
-     * storage is full or backpressure is in effect.
+     * Returns result with empty TAddRequestResult::CachedRequest and
+     * TAddRequestResult::Failed == true if the storage is in failed state.
      */
-    std::unique_ptr<TCachedWriteDataRequest> TryProcessPendingRequest();
+    [[nodiscard]] TProcessPendingRequestRequest TryProcessPendingRequest();
```
SvartMetal: TProcessPendingRequestResult

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.h
```diff
@@ -33,9 +33,17 @@ class TWriteDataRequestManager
     THashSet<ui64> NodesWithBackpressure;
 
 public:
-    using TAddRequestResult = std::variant<
-        std::unique_ptr<TPendingWriteDataRequest>,
-        std::unique_ptr<TCachedWriteDataRequest>>;
+    struct TAddRequestResult
+    {
```
SvartMetal, line 37: Let's add `bool Failed` here?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -97,17 +98,47 @@ TWriteDataRequestManager::TWriteDataRequestManager(
...
+    // in the pending queue forever (including requests with smaller size).
+    // Should fit 1 MiB of data plus some headers (assume 1 KiB is enough).
+    const ui64 maxAllocationByteCount = 1024 * 1024 + 1016;
+
+    const ui64 maxSupportedAllocationByteCount =
+        PersistentStorage->GetMaxSupportedAllocationByteCount();
+
+    if (maxSupportedAllocationByteCount < maxAllocationByteCount) {
+        return MakeError(
+            E_FAIL,
```
SvartMetal: E_ARGUMENT

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -97,17 +98,47 @@ TWriteDataRequestManager::TWriteDataRequestManager(
     , Stats(std::move(stats))
 {}
 
-bool TWriteDataRequestManager::Init(const TCachedRequestVisitor& visitor)
+NProto::TError TWriteDataRequestManager::Init(
+    const TCachedRequestVisitor& visitor)
 {
-    bool success = true;
+    if (PersistentStorage->IsCorrupted()) {
+        return MakeError(E_FAIL, "Persistent storage is corrupted");
```
SvartMetal: E_INVALID_STATE?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -97,17 +98,47 @@ TWriteDataRequestManager::TWriteDataRequestManager(
...
 
-    PersistentStorage->Visit(
-        [this, &success, &loadedRequests](ui32 tag, const TStringBuf allocation)
+    auto visitResult = PersistentStorage->Visit(
+        [this, &error, &loadedRequests](ui32 tag, const TStringBuf allocation)
         {
             if (tag > static_cast<ui32>(ECachedWriteDataRequestTag::Max)) {
-                success = false;
+                error = MakeError(
+                    E_FAIL,
```
SvartMetal: E_INVALID_STATE?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -117,7 +148,7 @@ bool TWriteDataRequestManager::Init(const TCachedRequestVisitor& visitor)
                 allocation);
 
             if (!request) {
-                success = false;
+                error = MakeError(E_FAIL, "Request deserialization error");
```
SvartMetal: E_INVALID_STATE?

cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
```diff
@@ -33,6 +33,9 @@ struct TCachedData
...
+    bool Failed = false;
```
SvartMetal: const bool

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -266,30 +311,51 @@ void TWriteDataRequestManager::Remove(
...
     request->Time = Timer->Now();
     FlushedRequestsPushBack(request);
 
-    PersistentStorage->SetTag(
+    auto setTagResult = PersistentStorage->SetTag(
         request->GetAllocationPtr(),
         static_cast<ui32>(ECachedWriteDataRequestTag::Flushed));
+
+    if (HasError(setTagResult)) {
+        // log
```
SvartMetal: Critical event?

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -266,30 +311,51 @@ void TWriteDataRequestManager::Remove(
...
+bool TWriteDataRequestManager::SetHandleReleased(
     TCachedWriteDataRequest* request)
 {
-    PersistentStorage->SetTag(
+    auto setTagResult = PersistentStorage->SetTag(
         request->GetAllocationPtr(),
         static_cast<ui32>(ECachedWriteDataRequestTag::UnflushedHandleReleased));
+
+    if (HasError(setTagResult)) {
+        // log
```
SvartMetal: Here also

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -266,30 +311,51 @@ void TWriteDataRequestManager::Remove(
...
-void TWriteDataRequestManager::Evict(
+bool TWriteDataRequestManager::Evict(
     std::unique_ptr<TCachedWriteDataRequest> request)
 {
     FlushedRequestsRemove(request.get());
-    PersistentStorage->Free(request->GetAllocationPtr());
+    auto freeResult = PersistentStorage->Free(request->GetAllocationPtr());
+
+    if (HasError(freeResult)) {
+        // log
```
SvartMetal: And here

cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
e673: `TReadResponseBuilder` doesn't do this anymore

cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
e673: There is no more failed state in `TCachedData`

cloud/filestore/libs/vfs_fuse/write_back_cache/read_response_builder.cpp
e673: `TReadResponseBuilder` doesn't do this anymore

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache.cpp
e673: No, it is used in other places

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_state.cpp
e673: Renamed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.h
e673: Fixed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.h
e673: Added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Changed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Changed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Changed

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Changed

cloud/filestore/libs/vfs_fuse/write_back_cache/node_cache.h
e673: I've removed this field from `TCachedData` and implemented it differently

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Not needed - it is reported from `TPersistentStorage` now

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Not needed - it is reported from `TPersistentStorage` now

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Not needed - it is reported from `TPersistentStorage` now

e673:
AI review

I'll start by pulling the PR, its description, comments, and the diff.Now let me pull the actual diff and any review comments via the GitHub API.Got the diff. Let me look at the full change.Now let me get the full source files for context — the diff alone won't reveal races or ordering issues.`AddHandingReadDataResponse` is defined but never called. Let me check how `ReadData` behaves in the failed state and how the test session works.Let me check whether CI has finished on the latest commit, since I suspect the new read assertion fails.I reviewed PR #6867 (commit `142c30e`, the head after the Aug 26 15:59 force-push) against the full sources of `write_back_cache/`, `file_backed_containers/`, and the FUSE call sites. There are no human review comments on the PR yet — only CI bot posts; CI for the latest commit was still pending when I checked. Note that the last force-push un-commented one test assertion, and the two new "hang" unit tests have not been run by CI yet.

## Findings

### 1. `ReadData` does not hang in the failed state — it silently returns stale data (and the new test asserting a hang will fail)

`TWriteBackCacheState::AddHandingReadDataResponse()` (sic) is declared and implemented but never called. In `TImpl::ReadData`, when `GetCachedData` returns `Failed=true`, `TryFullyServeFromCache` returns `nullopt` and the code falls through to `Session->ReadData(...)`; the callback's `AugmentResponseWithCachedData` then returns `false` and the backend response is handed to the guest as-is.

> Yes, you reviewed incomple PR - now `ReadData` hangs too

Consequences:
- The guest reads data that lacks writes it already received ACKs for — silent inconsistency, which is worse than the intended hang and contradicts "doesn't serve requests and hangs".
- `TestHangWhenFileRingBufferIsCorrupted` now asserts `!readFuture.HasValue()`. With `TFileStoreTest`, `ReadDataHandler` returns `MakeFuture(response)` synchronously and `.Apply` runs immediately, so `readFuture` *will* have a value and both new tests (`...AtInitialization`, `...AtFlush`) should fail. In the previous commit (`36b5061`) this assertion was commented out, presumably for exactly this reason.

Fix: in `ReadData`, when `cachedData.Failed` (both in the full-hit path and in the augment callback), return `State.AddHandingReadDataResponse()` instead of forwarding to the session.

Related: `GetMaxWrittenOffset` returns `0` when failed, so `AdjustNodeSize` in `fs_impl.cpp` reports file sizes smaller than what the guest wrote. The in-memory node state is intact in the runtime-failure case, so returning the real value is strictly safer than 0.

### 2. New race → `Y_ABORT_UNLESS` crash in `ExecuteFlush`

```cpp
auto retryStatus = State.FlushFailed(nodeId, ...);   // lock released here
if (retryStatus == ShouldNotRetry) return;
handle = State.GetLiveHandle(nodeId);
Y_ABORT_UNLESS(handle != E_INVALID_HANDLE, ...);
```

`GetLiveHandle` now returns `E_INVALID_HANDLE` when `FailedFlag` is set. If another thread sets the flag between `FlushFailed` and the second `GetLiveHandle` (e.g. a `ReadData` → `UnpinCachedData` → `EvictUnpinnedFlushedEntries` → `Free` failure), this aborts — a crash-on-corruption path introduced by the PR. The window is narrow but real. Return instead of aborting when the handle is invalid after `FlushFailed`, or don't consult `FailedFlag` in `GetLiveHandle`.

> Fixed

### 3. Non-corruption errors are promoted to a permanent, global hang

`TryStoreRequestInPersistentStorage` treats *any* `Alloc` error as `Failed=true` → `SetFailedFlag()`. `TFileRingBuffer::Alloc` returns `E_ARGUMENT` for `size > MaxAllocationByteCount` without marking the buffer corrupted, and `ValidateWriteDataRequest` does not bound the request size. One oversized write freezes all I/O for the whole filesystem until manual intervention, and `TFileRingBufferStorage::Alloc` reports it as `WriteBackCacheCorruptionError` although nothing is corrupted. The same applies to the `Init` capacity check (a config error becomes "hang forever" instead of failing loudly at startup). Consider rejecting the individual request with an error and reserving the failed state for `IsCorrupted()`-backed errors.

> This is expected. Storage corruption and logic failures are treated the same way

### 4. `Drain()` / session shutdown hangs forever; `IsDrained()` is inconsistent with the failed state

`AddFlushAllRequest` checks `FailedFlag` before the "nothing to flush" early-return, so `Drain()` never resolves once failed. In `loop.cpp`, `StopAsyncOnCompletionQueueStopped` calls `Drain()` when `!IsDrained()` and waits on it — graceful stop never completes, so the "manual resolution" cannot be performed without SIGKILL. Meanwhile `IsDrained()` ignores `FailedFlag`: if the cache failed at `Init` before loading anything (corrupted file), `DrainingMode && !HasPendingOrUnflushedRequests()` is true, `GetMode()` reports `Drained`, and `GetWriteBackCacheRequestStrategy` returns `DoNotUse` — I/O proceeds directly to the backend while unflushed guest data may still sit in the corrupted file and would be replayed over newer writes after repair. Suggest: `IsDrained()` should be false when failed, and `Drain()` should complete with an error in that state so shutdown can proceed.

> `IsDrained` now checks for `IsFailed` flag.
> Yes, it is ok that manual resolution will require service restart.
> Dropping the requests on intentional session stop is to be made separately

### 5. Storage keeps being accessed after failure → repeated critical events and progressive in-memory data loss

`EvictUnpinnedFlushedEntries` and `ProcessPendingRequests` (reached via `UnpinCachedData` on every read) are not guarded by `FailedFlag`. Each call re-attempts `Free`/`Alloc` on the corrupted buffer, which fires `AccessToCorruptedFileRingBufferError` plus `WriteBackCacheCorruptionError` on every read, and each failed `Evict` drops another `TCachedWriteDataRequest` from memory. Add an early return on `FailedFlag` in `EvictUnpinnedFlushedEntries` / `UnpinCachedData` and `ProcessPendingRequests`.

> I've added a guard to `EvictUnpinnedFlushedEntries`

### 6. Lost observability in `ScheduleFlushNode`

The `ReportWriteBackCacheImpossibleState` for an empty flush batch was removed and replaced with a comment that it "may happen when the cache entered failed state". The genuinely impossible (non-failed) case is now silent. 

> The impossible state has become possible but it is now a sign of a failure.

Better: have `VisitUnflushedRequestsFromFrontFlushBatch` signal the failed case explicitly and keep the report for the other one. Also, in the failed state `Stats->FlushStarted()` is never balanced by `FlushCompleted()`, so the in-progress-flush metric drifts.

> You are right but I see no reason to overcomplicate the logic in this case (corruption + extremely unlikely race).
> Manual resolution will imply service restart that will reset the metrics.

### Minor

- Naming/typos: `AddHandingReadDataResponse` → `AddHangingReadDataResponse`; `TProcessPendingRequestRequest` → `...Result`; the header doc for `TryProcessPendingRequest` references `TAddRequestResult::CachedRequest/Failed`, which don't exist on that type.

> Fixed

- `// log` placeholder comments left in `SetFlushed`, `SetHandleReleased`, `Evict`.

> Fixed

- `#include <atomic>` added to `write_back_cache_state.h` but unused (`FailedFlag` is a plain `bool` — correctly always accessed under the state lock, so no data race there).

> Removed

- `TWriteDataRequestManager` still exposes a default ctor and move ops that are no longer needed now that it's constructed in the initializer list.

> Removed

- `SetFlushed` mutates the unflushed/flushed lists before `SetTag`; on failure the request manager's `Stats` counters are left skewed (no `RemovedUnflushedRequest`). Harmless in the hang design but worth a comment.

> Changed the order

- Critical events are emitted while holding the `TAdaptiveLock` spinlock (`Sprintf` + `FormatError` + logging); not a bug, but heavier than ideal under a spinlock.

> This is not a big issue because it doesn't affect other filesystems

What I checked and found OK: `FailedFlag`/`THangingRequests` are only touched under the lock; `TFileRingBuffer` construction cannot throw (`Map()` catches); stats getters are null-safe when unmapped, so the metrics-collection crash from the description is fixed; intrusive-list items unlink on destruction so dropped requests don't dangle; never-fulfilled promises are silently discarded on destruction (no broken-promise exception); member init/destruction order in `TImpl` and `TWriteBackCacheState` is correct.

e673:
Second AI review after the changes have been made

I've read the full PR (head 038cd60, three commits on top of 1fcef7a), the surrounding code on the branch, and the TFileRingBuffer / loop.cpp call sites it interacts with. No memcpy/memmove/memset or raw buffer indexing is introduced or changed; the mmap-touching paths are unchanged and every new branch only adds early returns, so I found no out-of-bounds risk. The problems are in behavior and lost diagnostics.

Significant issues

1. Graceful shutdown deadlocks in the failed state. loop.cpp:1415 (StopAsyncOnCompletionQueueStopped) calls WriteBackCache.Drain() and only proceeds to unmount when that future completes. With this PR, IsDrained() is false forever once IsFailed is set, and Drain() → AddFlushAllRequest() returns a promise that is never set. So DestroySession/Stop never finishes — the operator who is supposed to do "manual resolution" can't stop the process without SIGKILL. Drain() is internal, not guest-facing, so failing it (e.g. E_INVALID_STATE) doesn't violate "no errors to the guest"; StopAsyncOnWriteBackCacheDrained already handles the error path ("unflushed data will be lost"). A separate Drain-specific check in TImpl::Drain() avoids touching FlushAllData().

> This is not an issue but an expected behavior.
> I've added a comment to `Drain()`.
>
> The `DestroySession`/`Stop` should be handled separately but a different logic — this is out of scope for this PR.

2. Two invariant violations became silent instead of reported.

ScheduleFlushNode: the empty-batch branch used to fire WriteBackCacheImpossibleState. Now it's attributed solely to the failed state, but in a non-failed cache an empty batch is still the impossible case the old comment described, and it now goes unnoticed. Have VisitUnflushedRequestsFromFrontFlushBatch return bool (or expose State.IsFailed()) and keep the report when not failed.

> Done

ExecuteFlush: the second GetLiveHandle() returning E_INVALID_HANDLE used to Y_ABORT_UNLESS; now it silently returns and the node is stuck in FlushRequested with unflushed data forever — with no critical event, even in a non-failed cache (the race with FailAllPendingRequests erasing a handle that only had pending requests is possible). ScheduleRetryFlush(flushState) is a simpler and robust replacement: in the failed state the retry hits FlushFailed → ShouldNotRetry and terminates; in the race case it retries correctly.

> Yes, I know about the race and will address it is a separate PR https://github.com/ydb-platform/nbs/pull/6965
> Replaced with `ScheduleRetryFlush(flushState)`

3. ReadData: Apply → Subscribe changes exception semantics. With Apply, an exception from UnsafeExtractValue(future) (session future carrying an exception) was captured into the returned future. With Subscribe, it escapes into the thread that fulfilled the session promise and the read promise is never set. ExtractResponse(future) from error.h gives you the old safety.

> Replaced with `ExtractResponse`.

5. GetMaxWrittenOffset returns 0 when failed. This value adjusts GetAttr/ReadDir sizes; it's purely in-memory and doesn't touch storage. Returning 0 makes files visibly shrink below what the guest already wrote and had acknowledged — silently wrong data, which is worse than the intended "hang". Just drop the guard (same for ResetMaxWrittenOffset, which is harmless either way).

> Agree

Smaller correctness points
TWriteDataRequestManager::Init visitor doesn't short-circuit after the first error: it keeps deserializing, consumes sequence IDs, and a later entry can overwrite error. Add if (HasError(error)) return; at the top of the lambda.

> Done

Flush in-flight counter drifts in the failed state: UpdateFlushStatus/TriggerFlushAll call Stats->FlushStarted(), then FlushSucceeded(nodeId, 0) early-returns without FlushCompleted(). TriggerPeriodicFlushAll also isn't guarded, so this repeats every period.

> Added a guard to `TriggerPeriodicFlushAll`.

Doc inaccuracies: IPersistentStorage::Visit says "does not visit anything ... if corrupted", but TFileRingBuffer::VisitEntries can visit some entries and then detect corruption (Init handles this correctly, the doc doesn't).

> Fixed

GetMaxSupportedAllocationByteCount mentions PushBack, which isn't in this interface.

> Fixed

TryProcessPendingRequest's comment refers to TAddRequestResult fields but returns TProcessPendingRequestResult.

> Fixed

TTestStorage::Free/SetTag still Y_ENSURE-throw on a bad pointer while the interface now promises an error return.

> Fixed

write_back_cache_state.h: the AddHangingRequest declaration is 83 columns.

> Fixed

Complexity that can go
THangingRequests is dead weight. In library/cpp/threading/future, destroying a TPromise without setting it leaves the future unfulfilled anyway (~TFutureState only destroys a set value). Nothing ever resolves these promises, so the class, the four vectors, AddHangingRequest, and the pre-created promise in ReadData only add an unbounded-growing container. Replace with NewPromise<T>().GetFuture() at the call sites (the ReadData callback can just not set its promise). If the plan is to resolve them later (e.g. from Drain, which would also fix issue 1), say so in a comment and land that together.

> This is to be used later, added a comment

TAddRequestResult loosened a std::variant into three fields where "neither pointer set" already means failure, making Failed redundant there. TryStoreRequestInPersistentStorage returning TProcessPendingRequestResult is a misnomer. TResultOrError<std::unique_ptr<TCachedWriteDataRequest>> (nullptr = full/backpressure, error = failed) mirrors IPersistentStorage::Alloc, keeps the error text, and removes both structs.

> This change was requested by another reviewer.

TImpl::PersistentStorage is now used only to seed State; make it const or construct it inside State's initializer.

> Done

Tests
TestHangWhenFileRingBufferIsCorrupted does Sleep(TDuration::Seconds(1)), but TBootstrap uses TTestScheduler (tasks run only on RunAllScheduledTasks) and TFileStoreTest handlers return MakeFuture synchronously. Nothing can happen during that second; it just adds ~2 s across the two tests. Remove it.

> Done

ShouldHangWhenFileRingBufferBecomesCorruptedAtFlush asserts flushFuture.HasValue() while the comment says the flush succeeded — assert !HasError(flushFuture.GetValue()). Neither hang test checks a critical-event counter, which is half the stated goal.

> Done

Coverage gap: TTestStorage can't fail, so the SetTag failure branches (FlushSucceeded, FlushFailed→SetHandleReleased, DropCachedData), Alloc/Commit failures (AddWriteDataRequest, ProcessPendingRequests), and the Init tag/deserialization/min-size errors are only reachable through the two file-level tests, which exercise just IsCorrupted() at init and Free during eviction. A SetFailed() switch on TTestStorage would let you unit-test the rest cheaply.
node_cache_ut.cpp's Add helper now silently returns nullptr where std::get<1> used to throw on a pending result.

> Will do it later

One consequence worth a conscious decision rather than a fix: with ServerWriteBackCacheEnabled=false, loop.cpp:1087 calls Drain() at startup so the cache runs in Draining mode (UseDirect). A corrupted file now makes every direct read/write hang in AcquireBarrier, whereas the intent of draining mode is pass-through. That's defensible (the file may hold unflushed data), but it's a broader blast radius than the description implies.

SvartMetal (approved):

yegorskii (approved):

PR #6870, opened 2026-08-21
Fix use-after-dtor in cloud filestore client

makxenov:
`Stop()` must run on the throwing path too. Otherwise it will happen only on static destruction phase, and the threads then may access some global variables that were already destroyed at that point.

### DIFF
--- a/cloud/filestore/apps/client/lib/command.cpp
+++ b/cloud/filestore/apps/client/lib/command.cpp
@@ -15,6 +15,7 @@
 
 #include <util/datetime/base.h>
 #include <util/generic/guid.h>
+#include <util/generic/scope.h>
 #include <util/system/env.h>
 #include <util/system/fs.h>
 #include <util/system/sysstat.h>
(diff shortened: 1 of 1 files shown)


cloud/filestore/apps/client/lib/command.cpp
```diff
@@ -124,13 +124,21 @@ int TCommand::Run(int argc, char** argv)
...
 
-    if (!Execute()) {
-        // wait until operation completed
-        with_lock (WaitMutex) {
-            while (ProgramShouldContinue.PollState() == TProgramShouldContinue::Continue) {
-                WaitCondVar.WaitT(WaitMutex, WaitTimeout);
+    // Stop() must run on the throwing path too. Otherwise it will happen only
+    // on static destruction phase, and the threads then may access some
+    // global variables that were already destroyed at that point.
+    try {
```
efimov-ivan-yteam: Возможно, тут хорошо бы смотрелся `Y_DEFER`

efimov-ivan-yteam (dismissed):

PR #6873, opened 2026-08-21
Fix use-after-dtor in TTabletMetrics

efimov-ivan-yteam:
### Notes
MaxUsedQuota, ReadDataPostponed and WriteDataPostponed each keep a shared_ptr to FsRegistry, so merely releasing the FsRegistry member does not destroy it - its REGISTER_LOCAL registrations would stay live in the process-wide registry while the counters declared after those three members are already destroyed, and the metrics service's periodic Update() would read them.
Drop every registration in ~TTabletMetrics(), while all members are still alive.

Credits to @makxenov

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_counters.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_counters.cpp
@@ -91,6 +91,16 @@ TTabletMetrics::TTabletMetrics(IMetricsRegistryPtr metricsRegistry)
     , AggregatableFsRegistry(CreateMetricsRegistryStub())
 {}
 
+TTabletMetrics::~TTabletMetrics()
+{
+    MaxUsedQuota.Unregister(MaxUsedQuotaKey);
+    ReadDataPostponed.Unregister(ReadDataPostponedKey);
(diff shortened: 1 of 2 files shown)


nisqatsi (dismissed):

cloud/filestore/libs/storage/tablet/tablet_counters.cpp
```diff
@@ -91,6 +91,16 @@ TTabletMetrics::TTabletMetrics(IMetricsRegistryPtr metricsRegistry)
     , AggregatableFsRegistry(CreateMetricsRegistryStub())
 {}
 
+TTabletMetrics::~TTabletMetrics()
+{
+    MaxUsedQuota.Unregister(MaxUsedQuotaKey);
+    ReadDataPostponed.Unregister(ReadDataPostponedKey);
+    WriteDataPostponed.Unregister(WriteDataPostponedKey);
+
+    AggregatableFsRegistry.reset();
```
qkrorlqr: why do we need these explicit `reset()`s?

qkrorlqr (approved):

nisqatsi (approved):

PR #6875, opened 2026-08-21
Tests: Fix high memory usage on dumping profile log

proller:
### Avoid OOM when parsing large profile logs
Flappy test:  cloud/filestore/tests/fio/qemu-local-noserver-mq-test

Root cause: `yatest.common.execute` materialized the complete `dumpevents` output in memory, after which the parser created another full list of decoded events. In the failed QEMU run, a 29 MB profile log expanded to approximately 798 MB and 4.4 million lines, increasing Python RSS to about 3.7 GB and triggering the guest OOM killer.

The fix writes `dumpevents` output to a caller-owned temporary file and parses it incrementally. High-volume MQ tests now consume events through an iterator and retain only the required thread IDs.

Added focused tests for streaming, parsing, filtering, and backward compatibility.


cloud/filestore/tools/testing/profile_log/common.py
```diff
@@ -1,19 +1,51 @@
...
-         "--profile-log", profile_log_path,
-         "--fs-id", fs_name])
-
     type_dict = {}
-    for line in proc.stdout.decode("utf-8").splitlines():
+    for line in _iter_profile_log_lines(
+            profile_tool_bin_path,
+            profile_log_path,
+            fs_name):
         request_type = line.rstrip().split("\t")[2]
```
debnatkh: you can reuse _parse_profile_log_event

cloud/filestore/tools/testing/profile_log/ut/common_ut.py
```diff
@@ -0,0 +1,161 @@
...
+        "--profile-log",
+        "/test/profile.log",
+        "--fs-id",
+        "nfs_test",
+    ]
+    assert execute_mock.call_args.kwargs["stdout"] is streams[0]
+    assert streams[0].closed
+
+
+def test_iter_profile_log_events_parses_incrementally():
```
debnatkh: re-tests laziness that's already implied by _iter_profile_log_lines being a generator and iter_profile_log_events being a generator wrapping it. 

Generators are lazy by construction in Python; you don't need to prove that a yield-based function doesn't eagerly consume its input

cloud/filestore/tools/testing/profile_log/ut/common_ut.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+        assert iter(events) is events
+        assert consumed == []
+        assert next(events) == ("ReadData", {"loop_thread_id": "1"})
+        assert consumed == [lines[0]]
+        assert list(events) == [("WriteData", {"loop_thread_id": "2"})]
+
+
+def test_get_profile_log_events_remains_a_list_wrapper():
```
debnatkh: This test checks that list(x) returns a list— it asserts Python's own behavior, dressed up with a mock to prove delegation. Given that get_profile_log_events's body is literally return list(iter_profile_log_events(...)), this test has near-zero chance of ever catching a real bug; the only way it fails is if someone deletes the list() call, which would be caught by literally any other caller

cloud/filestore/tools/testing/profile_log/ut/common_ut.py
```diff
@@ -0,0 +1,161 @@
...
+    assert isinstance(events, list)
+    assert events == expected
+    iter_events.assert_called_once_with(
+        "/test/filestore-profile-tool",
+        "/test/profile.log",
+        "nfs_test",
+    )
+
+
+def test_analyze_profile_log_filters_by_node_name():
```
debnatkh: This is the only test really testing the logic here :) I would argue that the majority of other tests are useless

SvartMetal (dismissed):

cloud/filestore/tools/testing/profile_log/common.py
proller: _parse_profile_log_event splits all fields, runs a regex substitution for every body field, creates several intermediate strings and lists, and constructs a dictionary that analyze_profile_log immediately discards. The node filter is also checked only after all this work. The PR description gives a real input size of approximately 4.4 million lines, so this adds millions of unnecessary regex operations and dictionary allocations to an already expensive repeated test operation.

On a representative event line, a local synthetic microbenchmark showed this full parser to be about 15 times slower than extracting only field 2. The exact factor depends on event bodies, but the regression is clearly substantial.

cloud/filestore/tools/testing/profile_log/ut/common_ut.py
proller: 3. [P2] The remaining unit test does not protect the memory regression being fixed

The final test file contains only one three-line test, and it calls analyze_profile_log. It verifies that a caller-owned stdout object is supplied and that ExecutionResult.stdout is not accessed, but:

it never calls iter_profile_log_events, although all three changed MQ tests use that function;
it never verifies parsed event bodies such as loop_thread_id;
it never verifies that parsed events are consumed lazily;
it never verifies that get_profile_log_events remains a list-returning compatibility wrapper;
it would still pass if the implementation started doing an unbounded output.read() and materialized the entire temporary file in memory.

The first commit actually contained direct tests for the iterator, incremental parsing and the compatibility wrapper. The second commit deleted 109 test lines, but the PR description still says that focused tests for “streaming, parsing, filtering, and backward compatibility” were added.

At minimum, add:

A direct iter_profile_log_events test checking the parsed loop_thread_id and other body fields.
A laziness test proving that requesting one event consumes only one input line.
A stream implementation that raises on read() and readlines(), so only line iteration is permitted.
A compatibility test asserting that get_profile_log_events still returns a list.

SvartMetal (approved):

komarevtsev-d (approved):

PR #6878, opened 2026-08-22
issue-6608: Implement quota usage aggregation across shards

debnatkh:
### Notes
Second-to-last PR to get quotas _enforcement (very first version)_ working: now stats are aggregated

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
+++ b/cloud/filestore/libs/storage/service/service_ut_sharding.cpp
@@ -671,13 +671,47 @@ Y_UNIT_TEST_SUITE(TStorageServiceShardingTest)
                 TSetNodeAttrArgs(fileId).SetSize(fileSize));
         }
 
-        // each shard should have received at least one of the files -
-        // otherwise this test wouldn't actually be exercising cross-shard
-        // aggregation
+        // a second, distinct quota that only ever lands on one shard (a
+        // single file)
+        {
+            NProtoPrivate::TSetQuotaRequest request;
+            request.SetFileSystemId(fsConfig.FsId);
+            request.SetQuotaId(43);
+            request.SetMaxBytes(1_GB);
+            request.SetMaxNodes(100);
+
+            TString buf;
+            google::protobuf::util::MessageToJsonString(request, &buf);
+            service.ExecuteAction("setquota", buf);
+        }
+
+        constexpr ui64 otherQuotaFileSize = 50;
+        auto otherQuotaDirId =
+            service
+                .CreateNode(
+                    headers,
+                    TCreateNodeArgs::Directory(RootNodeId, "other_quota_dir"))
+                ->Record.GetNode()
+                .GetId();
+        service.SetNodeAttr(
+            headers,
+            fsConfig.FsId,
+            TSetNodeAttrArgs(otherQuotaDirId).SetQuotaId(43));
+        auto otherQuotaFileId =
+            service
+                .CreateNode(
+                    headers,
+                    TCreateNodeArgs::File(otherQuotaDirId, "file"))
+                ->Record.GetNode()
+                .GetId();
+        service.SetNodeAttr(
+            headers,
+            fsConfig.FsId,
+            TSetNodeAttrArgs(otherQuotaFileId).SetSize(otherQuotaFileSize));
+
         for (const auto& shardId: fsConfig.ShardIds()) {
             const auto stats = GetStorageStats(service, shardId);
             const auto& usages = stats.GetStats().GetQuotaUsages();
-            UNIT_ASSERT_VALUES_EQUAL_C(1, usages.size(), shardId);
             UNIT_ASSERT_C(usages.contains(42), shardId);
             const auto& usage = usages.at(42);
             UNIT_ASSERT_C(usage.GetUsedNodes() > 0, shardId);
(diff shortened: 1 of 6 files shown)


debnatkh:
Here are the concrete issues I can point to in this diff:

**1. Monitoring page's "Filesystem" columns may be silently stale/misleading**
```cpp
const auto& aggregateUsages = CachedAggregateStats.GetQuotaUsages();
```
This new `FilesystemUsedBytes`/`FilesystemUsedNodes` display reads from `CachedAggregateStats`, not from a fresh forced fetch. If that cache is only refreshed lazily/periodically (as "cached" implies), an operator looking at the monitoring page could see stale or all-zero filesystem-level numbers even though `TabletUsedBytes`/`TabletUsedNodes` next to them are live. Worth confirming: does whatever populates `CachedAggregateStats` actually fan out to shards and merge `QuotaUsages`, or does it only do that on a forced/explicit request? If it's the latter, these two new columns could be misleading by default rather than genuinely useful — the diff doesn't show that code path, so this needs verifying before merge.

**2. Test's cross-shard loop hard-assumes every shard receives ≥1 file**
```cpp
for (const auto& shardId: fsConfig.ShardIds()) {
    ...
    UNIT_ASSERT_VALUES_EQUAL_C(1, usages.size(), shardId);
    UNIT_ASSERT_C(usages.contains(42), shardId);
```
With `fileCount = 4` round-robined across shards, this only holds if `ShardIds().size() <= 4` and the round-robin actually touches every shard. It's not testing the merge logic in a way that's robust to that fixture changing (e.g., if `TShardedFileSystemConfig` defaults ever grow shard count, or round-robin starts skewed), so this test could start failing for reasons unrelated to the feature it's meant to verify.

**3. No test isolates `MergeQuotaUsages` itself**
The only coverage is the full end-to-end sharding test. There's no direct/unit-level test of `MergeQuotaUsages` for edge cases it will actually hit in production:
- a quota ID present in `src` but not yet in `dst` (first-shard-to-report case)
- multiple distinct quota IDs merged simultaneously
- a shard reporting an empty `QuotaUsages` map (e.g., no files under that quota landed there)
- what happens to the aggregate when one shard fails to respond at all — does the aggregate quietly under-report usage for that quota, with no signal that data is partial?

That last one especially seems like a real gap: nothing in the diff indicates the aggregate result is flagged as partial/incomplete if a shard's `GetStorageStats` fails or times out during fan-out.

**4. No cleanup path for quota deletion is visible or tested**
If a quota is removed while nodes still reference it (or after all nodes are removed), it's not clear from this diff whether `QuotaUsages` entries get cleared from either the per-tablet state or the aggregate, versus lingering as `usage=0` entries forever. Might be pre-existing behavior unrelated to this PR, but the PR doesn't add a test proving stale entries eventually disappear.

**5. Naming collision risk: `TQuotaUsage` (C++ struct) vs. `TQuotaUsageValue` (new proto message)**
```cpp
TVector<TQuotaUsage> GetQuotaUsages() const;   // existing C++ struct
...
message TQuotaUsageValue { uint64 UsedBytes = 1; uint64 UsedNodes = 2; }  // new proto type
```
Not a bug, but genuinely confusing for future readers — two very similarly named types with overlapping fields (`UsedBytes`/`UsedNodes`) but one also carries `QuotaId` as a struct field while the other relies on the map key. Easy to introduce a bug later by mixing them up.

**6. PR description is still the unfilled template**
"Notes: Describe your PR..." / "Issue: Put links to the related issues here" — the issue-6608 link should be filled in before this leaves Draft, especially since reviewers landing on the PR page get zero context.

None of these are correctness show-stoppers in the merge logic itself — the map-based `MergeQuotaUsages` and `FillSelfStorageStats` additions look right for the happy path the test covers. The real risk areas are **#1** (whether the new monitoring columns are trustworthy) and **#3** (no coverage of partial-failure/multi-quota scenarios), since those are exactly the cases that won't show up until this hits production with real shard failures or several quotas in play.


---

1. This is acceptable
2. This implicit expectation is present in a lot of tests – we will not get rid of round robin by default anytime soon
3. This is a small method, tested by tablet-level test
4. Deletion is deliberately out of the scope for the initial state
5. I find this naming acceptable
6. fixed

debnatkh:
Script to check it is working:

```

echo "hello world!" > /tmp/quota-test-data

echo "1. Setting quota 42"
./filestore-client executeaction --action setquota --input-json \
    '{"FileSystemId":"nfs","QuotaId":42,"MaxBytes":1073741824,"MaxNodes":100}'

echo "2. Creating directory /quota-test-dir"
./filestore-client mkdir --filesystem nfs --path /quota-test-dir

echo "3. Resolving NodeId of /quota-test-dir"
DIR_NODE_ID=$(./filestore-client stat --filesystem nfs --path /quota-test-dir --json \
    | python3 -c 'import json,sys; print(json.load(sys.stdin)["Id"])')
echo "   NodeId=$DIR_NODE_ID"

echo "4. Attaching quota 42 to node $DIR_NODE_ID"
./filestore-client setnodeattr --filesystem nfs --node-id "$DIR_NODE_ID" --quota-id 42

echo "5. Creating and writing /quota-test-dir/file.txt"
./filestore-client write --filesystem nfs --path /quota-test-dir/file.txt --data /tmp/quota-test-data

echo "6. Listing quotas to confirm usage"
./filestore-client executeaction --action listquotas --input-json '{"FileSystemId":"nfs"}'

echo "Done."
```

Result:
image

debnatkh:
I'll fetch the PR to review it.The PR page doesn't show the diff itself, so let me pull it directly.I have the diff (5 source files + 1 test). To review it properly I need the surrounding code, so let me pull the relevant files at the PR's commit.Now let me look at the quota model and the test helpers the new test depends on.I pulled the diff (5 source files + 1 test) and the surrounding code at commit `f28e3a5`. The core idea is sound, but there's one real bug plus several smaller things.

**1. Double-counting when a shard handles `FORCE_FETCH_SHARDS` (bug)**

In `TIndexTabletActor::HandleGetStorageStats`, when the tablet is a shard and needs to fan out, the flow is: `FillSelfStorageStats(stats)` (which now appends this shard's `QuotaUsages`), then the `!IsMainTablet()` branch clears the `FILESTORE_TABLET_STATS` counters because `TAggregateStatsActor` will re-fetch self stats as part of the fan-out (the shard's `ShardIds` includes itself). But the clear macro only covers `FILESTORE_TABLET_STATS`; `QuotaUsages` is left in place. Then `MergeQuotaUsages` adds the shard's own `GET_ONLY_SELF` response on top, so its usage is counted twice.

Fix is one line in that branch: `stats->ClearQuotaUsages();`. Note `SevenBytesHandlesCount` has the identical pre-existing bug (`dst = dst + src` with no clear), so it's worth fixing both while you're there.

The test doesn't catch this because it only exercises main. Adding `GetStorageStats(service, fsConfig.Shard1Id, 0, STATS_REQUEST_MODE_FORCE_FETCH_SHARDS)` and asserting `UsedNodes == 1 + fileCount` would fail on the current branch (you'd get `1 + fileCount + <files on shard1>`).

**2. Test coverage gaps**

- `UsedBytes` is only asserted to be `0`. Nothing writes data, so the bytes half of the merge is untested. Writing a few blocks to files on different shards and checking the sum would cover it.
- The cached/background path (`HandleUpdateCounters` → `HandleAggregateStatsCompleted` → `CachedAggregateStats`) isn't tested. A `GetStorageStats` with `cacheTTL > 0` after the background tick would verify quota usages survive the cache round-trip.
- Only one quota ID. With two quotas where one exists only on some shards, you'd exercise the "not found in dst → `AddQuotaUsages`" branch after dst is already non-empty.

**3. Non-deterministic order in the aggregate**

`TQuotaStore::GetUsages()` deliberately sorts by `QuotaId`, but `MergeQuotaUsages` produces first-seen order, which depends on which shard replies first. The test survives because there's one quota and it uses `usages[0]`. Consider sorting in `ReplyAndDie` (or building via a map) so consumers and future tests get stable output.

**4. Monitoring page shows local usage on the main tablet**

`RenderHttpInfo_QuotasTab` uses `FindQuotaUsage`, i.e. this tablet's local usage. On main that's just the attach-point directories, not the number an operator would expect. `CachedAggregateStats` is already accessible in this file (it's used at line ~1109), so on main you could read `CachedAggregateStats.GetQuotaUsages()` instead, or at least label the columns "Local UsedBytes/Nodes".

**5. Minor**

- `MergeQuotaUsages` is O(|dst| × |src|) per shard response. Fine today; if quotas become per-directory and number in the thousands, a `THashMap<ui32, int>` index into `dst` would keep it linear.
- `usage ? FormatByteSize(usage->UsedBytes) : FormatByteSize(0)` — simpler as `FormatByteSize(usage ? usage->UsedBytes : 0)`.
- `MergeQuotaUsages` will happily include a quota ID that a shard reports but main has already deleted. Whether that's desired depends on how quota deletion propagates to shards, but worth a conscious decision.
- The PR description is still the untouched template and there's no link to #6608. Understandable for a draft, but worth filling in before review.

If you want, I can sketch the extra test cases for #2 in the style of the existing `ShouldAggregateQuotaUsageAcrossShards` test.

---

1. not new to this PR
2. some tests added
3. not important
4. not true. CachedAggregateStats is used
5. a. intentional, b. fixed, c. intentional

SvartMetal (approved):

agalibin (approved):

PR #6881, opened 2026-08-23
[Filestore] expose availability counters to user stats

SvartMetal:
### Notes
Right now the availability counters are exposed only in aggregated form, because I'm still not sure about the request names (whether we should use FUSE callbacks for availability request names or some other naming)

### Issue
#6825

### DIFF
--- a/cloud/filestore/libs/diagnostics/user_counter.cpp
+++ b/cloud/filestore/libs/diagnostics/user_counter.cpp
@@ -31,6 +31,14 @@ constexpr TStringBuf FILESTORE_INDEX_LATENCY         = "filestore.index_latency"
 constexpr TStringBuf FILESTORE_INDEX_ERRORS          = "filestore.index_errors";
 constexpr TStringBuf FILESTORE_INDEX_CUMULATIVE_TIME = "filestore.index_cumulative_time";
 
+// Availability counters
+constexpr TStringBuf FILESTORE_AVAILABILITY_TOTAL_INTERVALS =
+    "filestore.availability_total_intervals";
+constexpr TStringBuf FILESTORE_AVAILABILITY_AVAILABLE_INTERVALS =
+    "filestore.availability_available_intervals";
+constexpr TStringBuf FILESTORE_AVAILABILITY_UNAVAILABLE_INTERVALS =
+    "filestore.availability_unavailable_intervals";
+
 TLabels MakeFilestoreLabels(
     const TString& cloudId,
     const TString& folderId,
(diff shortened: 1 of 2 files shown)


yegorskii (approved):

SvartMetal:
>
```
1. [P2] Missing intervals are invisible in user stats
The availability tracker publishes Availability_MissingIntervals specifically because it may skip intervals when more than 30 intervals need to be processed in one update. Those skipped intervals are deliberately excluded from Availability_TotalIntervals, with the documented expectation that the SLA calculation decides how to account for them. This PR exports only total, available, and unavailable intervals, so the user-stats consumer cannot distinguish complete measurements from partially missing data.
Concrete counterexample:
60 intervals pass without an update
30 intervals are evaluated and all are available
30 intervals are recorded as missing
exported user stats contain available=30, unavailable=0, total=30
the apparent availability is 100%, although half of the measurement period was not evaluated
This remains a problem even if a critical event is added: the event can alert operators, but the exported historical data is still not self-describing.
Suggested review comment:
Please also expose Availability_MissingIntervals, or explicitly include missing intervals in the denominator/unavailable count. The tracker excludes them from TotalIntervals, so without this metric user stats can report 100% availability for a partially observed measurement period.
```
We are not going to expose missing intervals
>
```
2. [P2] The unregistration part of the test performs zero assertions
After unregistering, the test passes {} to ValidateTestResult. ValidateJsons only loops over expectedJson["sensors"]; with {}, the loop is empty. Therefore, the test passes even if all three new RemoveUserMetric calls are deleted and the metrics remain published forever.
Suggested review comment:
This check is vacuous: ValidateJsons iterates only expected sensors, and {} contains none. The test remains green if none of the availability metrics are removed. Please inspect the actual sensor list and assert that these three names are absent, or make the validator compare the complete sensor sets.
The existing ShouldMultipleRegister test has the same problem, but the new test should not repeat it.
```
Yes, also encountered this. I will fix it in a separate PR as this is pre-existing problem
>
```
3. [P2] The value test bypasses the fragile tracker-to-exporter contract
The test manually obtains the source subgroup and writes values into counters using the same string literals as RegisterFilestore. It never enables TAvailabilityCounters, rolls an interval, or verifies that the tracker and user-stats exporter are actually connected. A future intentional rename in the tracker, accompanied by updated tracker tests but not updated user-stats code, would leave this test green while production user metrics stayed permanently at zero.
The assertion also looks up metrics only by labels.name and compares only the value. It does not verify:
kind == "RATE"
service, project, cluster, filestore, or instance labels
absence of unexpected metrics
the complete emitted metric set
The normal expected JSON fixtures contain this information, but the validator ignores it. Consequently, an incorrect derivative flag or incorrect tenant labels would not be detected by this test.
A meaningful integration test should retain the stats returned by GetFileSystemStats, enable availability tracking through it, advance a controllable timer across an interval, call UpdateStats, and then verify complete emitted records. A smaller direct-wrapper test can remain, but it should assert full labels, RATE kind, and exact removal after unregistering.
```
Right now it is good enough

neihar (approved):

SvartMetal:
image

Not relevant to this change

PR #6882, opened 2026-08-23
[Filestore] add critical events for availability counters 

SvartMetal:
### Notes
Adding `AvailabilityCountersUnavailableInterval, AvailabilityCountersMissingIntervals` critical events.

The missing intervals critical event fires when the availability counters stall for some time and can't publish some counters on time

### Issue
#6825

### DIFF
--- a/cloud/filestore/libs/diagnostics/availability_counters.cpp
+++ b/cloud/filestore/libs/diagnostics/availability_counters.cpp
@@ -64,6 +64,7 @@ const char* GetAvailabilityRequestTypeName(
 ////////////////////////////////////////////////////////////////////////////////
 
 void TAvailabilityCounters::EnableAndRegister(
+    TString fileSystemId,
     TDuration intervalDuration,
     NMonitoring::TDynamicCounters& counters)
 {
@@ -73,6 +74,8 @@ void TAvailabilityCounters::EnableAndRegister(
     if (CountersRegistered.load(std::memory_order_acquire)) {
         return;
     }
+
+    FileSystemId = std::move(fileSystemId);
 
     // A zero interval duration selects the default one.
     IntervalDuration = intervalDuration == TDuration::Zero()
@@ -327,6 +330,9 @@ void TAvailabilityCounters::RollIntervals(TInstant now)
             (alignedNow - CurrentIntervalStart).MicroSeconds() /
             IntervalDuration.MicroSeconds();
         MissingIntervalsCounter->Add(missingIntervals);
+        ReportAvailabilityCountersMissingIntervals(
+            TStringBuilder() << "filesystem " << FileSystemId << ": "
+                << missingIntervals << " intervals were missed");
 
         CurrentIntervalStart = alignedNow;
     }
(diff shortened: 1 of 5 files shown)


yegorskii (approved):

neihar (approved):

SvartMetal:
>
```
1. High: catch-up can generate a large burst of critical events while registry locks are held
RollIntervals() processes up to 30 overdue intervals in one call. RollInterval(true) now reports a critical event for every unavailable interval. Therefore, when a hung request survives a sufficiently long updater stall, one call can report up to 30 identical AvailabilityCountersUnavailableInterval events, followed by a AvailabilityCountersMissingIntervals event when the gap exceeds the limit.
This is performed independently for every enabled (filesystem, client) statistics object. TRequestStatsRegistry::UpdateStats() holds the registry read lock while updating the entire StatsMap, and every critical-event report performs a logging call; the fallback path explicitly flushes Cerr. Consequently, a process-wide updater stall can produce up to 30 × number_of_enabled_clients log records in one update and extend the registry lock hold time, blocking registration and unregistration while recovery is in progress. MaxIntervalsPerUpdate bounds work per tracker, but not across the registry.
Suggested review comment:
RollIntervals() can synchronously emit up to 30 unavailable-interval critical events per (filesystem, client) tracker during catch-up. Since the registry updates every tracker under its read lock, a process stall can result in O(30 * clients) reports in one update and delay registration/unregistration. Please aggregate catch-up results and emit at most one event per tracker/update, including the number of unavailable intervals and affected request types, or add explicit rate limiting
```
It is okay
>
```
2. Medium: the event loses the client dimension
TAvailabilityCounters implements a per-client (instance, filesystem) metric, and the registry stores separate statistics objects using (fileSystemId, clientId) as the key. However, the PR passes only FileSystemId into TAvailabilityCounters, and both new messages identify only the filesystem.
As a result:
events from multiple clients of the same filesystem are indistinguishable
repeated identical events cannot be attributed to a specific client
"filesystem X, unavailable..." can misleadingly sound like filesystem-wide unavailability, although the failed metric belongs to one client instance
The constructor should receive both FileSystemId and ClientId. Alternatively, reporting should be aggregated at filesystem level before emitting a filesystem-level event.
Suggested review comment:
Availability accounting is per (filesystem, client), but the event contains only FileSystemId. With several clients of the same filesystem, the reports are indistinguishable and a client-specific problem looks filesystem-wide. Please include ClientId, or aggregate the reports per filesystem before raising the event
```
It is okay for now
>
```
3. Medium: historical unavailable intervals are reported without their interval timestamps
During catch-up, CurrentIntervalStart points to a historical interval and advances after each RollInterval() call. The critical-event message does not include that start time or the interval end. Thus, several old intervals are logged together when the updater resumes, with only the current emission time and otherwise identical messages. An operator cannot tell whether the filesystem is unavailable now or whether the event describes a historical interval rolled after a stall.
The message should include something such as:
interval [2026-08-23T14:10:00Z, 2026-08-23T14:15:00Z)
For an aggregated catch-up report, it should include the first and last affected interval and the total unavailable-interval count.
Suggested review comment:
Please include the evaluated interval boundaries in the unavailable event. Catch-up evaluates historical intervals, but all reports are emitted at recovery time, so without CurrentIntervalStart they look like current unavailability and cannot be correlated with the actual incident
```
It is okay for now
```
4. Tests do not cover the risky behavior or the new message contract
The unavailable-event test checks only one normal interval and asserts that the global event counter becomes 1. The missing-interval test uses one tracker with no outstanding request and checks only whether the sensor increments once. Neither test verifies:
the actual message, including the newly added filesystem ID
the missing client ID
the interval timestamp
event cardinality when a request remains hung across 30 catch-up intervals
multiplication across several client trackers
Therefore, the second commit, whose purpose is to add the filesystem ID to the message, is effectively untested.
```
It is okay

PR #6885, opened 2026-08-24
issue-5895: fastshard - enforcing proper bit count limit in TPersistentBitmap

qkrorlqr:
### Notes
Fastshard's allocator bitmap bit count used to be inflated - it thought that it had BitmapPageCount * BitsPerPage bits available whereas the actual limit could've been lower (BitmapPageCount * BitsPerPage is greater than the actual limit because it rounds the number of bits up to a multiple of BitsPerPage). Fixed this - now the tail of the last bitmap page is filled with 1s upon initialization.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/fastshard/impl/naive_mirrored/persistent_bitmap.cpp
+++ b/cloud/filestore/libs/storage/fastshard/impl/naive_mirrored/persistent_bitmap.cpp
@@ -1,5 +1,7 @@
 #include "persistent_bitmap.h"
 
+#include <silk/util/logger.h>
+
 #include <util/string/builder.h>
 
 namespace NCloud::NFileStore::NStorage::NFastShard {
@@ -251,8 +253,7 @@ NProto::TError TPersistentBitmap::InitIfNeeded() const
             const ui64 endBit = MaxBits % bitsPerPage;
             if (endBit) {
                 if (endBit % 8 != 0) {
-                    return MakeError(E_INVALID_STATE, TStringBuilder()
-                        << "unaligned max bit count: " << MaxBits);
+                    SILK_WARN("unaligned max bit count: %lu", MaxBits);
                 }
 
                 const ui64 endByte = endBit / 8;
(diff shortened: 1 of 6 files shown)


qkrorlqr:
Claude:

> First, MaxBits % 8 != 0 produces a lazy E_INVALID_STATE: construction succeeds and every bitmap operation then fails at runtime. All current configs happen to be byte-aligned, but nothing enforces it — either mask the partial byte (~2 lines) or validate eagerly at construction. This is my one remaining pre-merge ask. Second, the PR changes the on-disk layout relative to main for configs where the old rounded-up bit count differed (test totals shifted 17680 → 16384); fine if fastshard's format isn't frozen, but it should be an explicit acknowledgment. A related design note worth a comment: the 0xFF tail is persisted once the last page is written, so a future increase of MaxBits on an existing shard would see stale tail bits as permanently allocated.

Both problems are not a big deal: fastshard is used only in tests so masking the last bits of the bitmap is not a problem from the backwards compat standpoint; `E_INVALID_STATE` upon lazy-init is also not a big deal - it will be noticed immediately if for whatever reason someone decides to make the number of page clusters not divisible by 8

qkrorlqr:
> Claude:
> 
> > First, MaxBits % 8 != 0 produces a lazy E_INVALID_STATE: construction succeeds and every bitmap operation then fails at runtime. All current configs happen to be byte-aligned, but nothing enforces it — either mask the partial byte (~2 lines) or validate eagerly at construction. This is my one remaining pre-merge ask. Second, the PR changes the on-disk layout relative to main for configs where the old rounded-up bit count differed (test totals shifted 17680 → 16384); fine if fastshard's format isn't frozen, but it should be an explicit acknowledgment. A related design note worth a comment: the 0xFF tail is persisted once the last page is written, so a future increase of MaxBits on an existing shard would see stale tail bits as permanently allocated.
> 
> Both problems are not a big deal: fastshard is used only in tests so masking the last bits of the bitmap is not a problem from the backwards compat standpoint; `E_INVALID_STATE` upon lazy-init is also not a big deal - it will be noticed immediately if for whatever reason someone decides to make the number of page clusters not divisible by 8

Actually I decided to drop the `MaxBits % 8 == 0` requirement

WilyTiger (approved):

arigachnyy (approved):

Issue #6485, opened 2026-07-12
[Filestore] non-blocking open: delay physical cleanup for zero-link nodes


budevg:
This is part of the  #5432.

With non-blocking `CreateHandle`, filestore-vhost may acknowledge an open before the tablet DB transaction has committed. If the tablet restarts in that window, it can temporarily forget the handle until filestore-vhost replays or confirms the pending open.

The dangerous race is:

1. filestore-vhost acknowledges `CreateHandle`.
2. The tablet restarts before the handle is persisted.
3. Another client unlinks the same node.
4. The tablet sees no open handle and physically deletes the zero-link node.

Blocking unlink only in the same filestore-vhost is insufficient because the unlink may come from another client on another host.

Unlink should remove the directory entry immediately, but physical zero-link node/data cleanup should be delayed long enough for async-open recovery to complete.

The first version may use a bounded timing window. 

## Proposed behavior

When unlink makes a node zero-link:

1. Remove the directory entry as today.
2. Do not immediately physically remove the node/data if fast-open recovery may be in progress.
3. Enqueue delayed zero-link cleanup.
4. After the recovery window expires, recheck the node before deleting it.

Cleanup may delete the node only if:

* the node still has zero links;
* `HasOpenHandles(nodeId) == false`;
* no async-open recovery barrier relevant to the filesystem/session is active, if such a barrier is implemented.

## Scope

* Add a delayed cleanup queue or reuse an existing orphan cleanup mechanism if it fits.
* Add a configurable recovery delay/window.
* Preserve current user-visible unlink semantics: the name disappears immediately.
* Prefer leaking or delaying cleanup over deleting too early.

## Metrics

Add counters/gauges for:

* delayed zero-link cleanup queue depth;
* oldest delayed cleanup age;
* number of nodes deleted by delayed cleanup;
* number of cleanup attempts skipped because handles or recovery barriers were still present.

PR #6901, opened 2026-08-25
[Filestore] issue-5432: defer node destruction after a tablet restart

budevg:
#5432
#6485
#6486

Async CreateHandle replies before its tx commits, so a restart loses the handle
until the client's queued ConfirmCreateHandle recreates it. Within that window
an unlink - or a rename over the node, or another session closing its last
handle - destroys the node together with its data, and the confirm that follows
fails with E_FS_NOENT.

Such a node is now unlinked but kept until UnconfirmedCreateHandleGraceTimeout
after the tablet start expires. It is recorded in the new
DeferredNodeDestruction table and destroyed later by a background task, unless a
confirm claimed it by then.

### DIFF
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -270,6 +270,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
                                        TDuration,  TDuration::Zero()          )\
     xxx(UnconfirmedCreateHandleGraceTimeout,                                   \
                                        TDuration,  TDuration::Minutes(2)      )\
+    xxx(MaxDeferredNodeDestructionCount,                ui64,       100'000   )\
+    xxx(MaxDeferredNodeDestructionsPerTx,               ui32,       1'000     )\
                                                                                \
     xxx(NodeRegistrationMaxAttempts,         ui32,      10                    )\
     xxx(NodeRegistrationTimeout,             TDuration, TDuration::Seconds(5) )\
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -674,7 +674,7 @@ bool TIndexTabletActor::NeedsNodeDestructionDeferral(
         && IsInUnconfirmedCreateHandleGracePeriod(ctx)
         && !GetFileSystem().GetIsFastShard()
         && node.Attrs.GetType() == NProto::E_REGULAR_NODE
-        && node.NodeId <= MaxNodeIdAtStart;
+        && node.NodeId <= MaxNodeIdAtTabletStart;
 }
 
 bool TIndexTabletActor::ShouldDeferNodeDestruction(
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -314,18 +314,15 @@ void TIndexTabletActor::DestroySessionHandlesAndRemoveNodes(
                 nodeId,
                 it->Attrs.GetSize());
 
-            auto e = RemoveNode(
+            const bool removed = DeferNodeDestructionOrRemoveNode(
                 db,
+                ctx,
                 *it,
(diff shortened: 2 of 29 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_destroydeferrednodes.cpp
```diff
@@ -0,0 +1,126 @@
...
+void TIndexTabletActor::ExecuteTx_DestroyDeferredNodes(
+    const TActorContext& ctx,
+    TTransactionContext& tx,
+    TTxIndexTablet::TDestroyDeferredNodes& args)
+{
+    auto db = CreateIndexTabletDatabaseProxy(tx.DB, args.NodeUpdates);
+
+    const ui64 commitId = GenerateCommitId();
+    if (commitId == InvalidCommitId) {
+        return ScheduleRebootTabletOnCommitIdOverflow(
```
qkrorlqr: in other txns we use `args.OnCommitIdOverflow()` - it also properly updates args.Error

cloud/filestore/libs/storage/tablet/tablet_state_nodes.cpp
```diff
@@ -120,6 +120,11 @@ NProto::TError TIndexTabletState::RemoveNode(
     ui64 minCommitId,
     ui64 maxCommitId)
 {
+    // the node is gone - nothing to destroy later, no matter how this call ends
+    if (HasDeferredNodeDestruction(node.NodeId)) {
+        RemoveDeferredNodeDestruction(db, node.NodeId);
```
qkrorlqr: doesn't look required here - you already manage this in the `DestroyDeferredNodes` tx

qkrorlqr (commented):
looks good overall

probably makes sense to add private api methods for deferred node destruction entry management similarly to response log entry management / op log entry management - in a separate PR

cloud/filestore/libs/storage/tablet/tablet_actor_loadstate.cpp
```diff
@@ -409,11 +417,14 @@ void TIndexTabletActor::CompleteTx_LoadState(
         args.TabletStorageInfo,
         args.LargeDeletionMarkers,
         args.OrphanNodeIds,
+        args.DeferredNodeDestructionIds,
         args.OpLog,
         args.ResponseLog,
         config);
     UpdateLogTag();
 
+    MaxNodeIdAtStart = ShardedId(GetLastNodeId(), GetFileSystem().GetShardNo());
```
SvartMetal: MaxNodeIdAtTabletStart

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1930,6 +2000,7 @@ bool TIndexTabletActor::BehaveAsShard(const NProto::THeaders& headers) const
 void TIndexTabletActor::RunRegularTasks(const TActorContext& ctx)
 {
     DeleteOldResponseLogEntries(ctx);
+    DestroyDeferredNodes(ctx);
```
SvartMetal, line 2003: Let's add a unit tests with 2+ transactions in-flight

cloud/filestore/libs/storage/core/config.cpp
```diff
@@ -270,6 +270,8 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
                                        TDuration,  TDuration::Zero()          )\
     xxx(UnconfirmedCreateHandleGraceTimeout,                                   \
                                        TDuration,  TDuration::Minutes(2)      )\
+    xxx(MaxDeferredNodeDestructionCount,                ui64,       100'000   )\
```
SvartMetal, line 273: Are we sure we should enable it by default?

@budevg @qkrorlqr

budevg: Actually the deferred node mechanism is active only when AsyncCreateHandleEnabled is set to true. Since it defaults to false, deferred node destruction is not enabled by default.

SvartMetal (approved):

yegorskii (approved):

Issue #6685, opened 2026-08-04
[NBS] Implement dynamic configuration


nisqatsi:
**RFC:** https://github.com/ydb-platform/ydb-rfc/blob/main/nbs/003_dynamic_configuration.md
**Related internal ticket:** NBS-7044

---

## Required YDB Backports for current main branch YDB version 25.1.4

**Console and ConfigsDispatcher fixes for YAML configs after 25.1.4:**

1. [x] https://github.com/ydb-platform/ydb/pull/16242
2. [x] https://github.com/ydb-platform/ydb/pull/16869
3. [x] https://github.com/ydb-platform/ydb/pull/41424

**Implement NBS dynamic configuration support and fix related YAML configs bugs:**

1. [x] https://github.com/ydb-platform/ydb/pull/41432
2. [x] https://github.com/ydb-platform/ydb/pull/41434
3. [x] https://github.com/ydb-platform/ydb/pull/41435
4. [x] https://github.com/ydb-platform/ydb/pull/43238
5. [x] https://github.com/ydb-platform/ydb/pull/43323
6. [x] https://github.com/ydb-platform/ydb/pull/44420
7. [ ] https://github.com/ydb-platform/ydb/pull/45952
8. [ ] https://github.com/ydb-platform/ydb/pull/46142
9. [ ] https://github.com/ydb-platform/ydb/pull/46337

## Implementation steps

TODO

PR #6902, opened 2026-08-25
[NBS] Add reusable storage config ICB controls

nisqatsi:
### Notes

- Add reusable `TStorageConfigControls` for read-write storage parameters, which can be shared between several `TStorageConfig` objects to deliver ICB parameter change to each object.
- Preserve config-bound controls for existing behavior and add config-independent controls for `TStorageConfig` objects that share one control set. Without an explicit ICB override, each such `TStorageConfig` uses its own protobuf value.
- Return a control to its mode-specific initial state in `RestoreDefault()`. The control then has no explicit override in either mode, and each `TStorageConfig` reads its own raw protobuf value.
- Existing runtime configuration flows remain unchanged. Re-registering the same controls set on a different `TControlBoard` now aborts instead of silently rebinding it, but no current in-repository code path performs such re-registration.

#### `TStorageConfig` copy behavior

- Copy the raw `TStorageServiceConfig` by value and share the current Features object and the same live `TStorageConfigControls` object in both control modes.
- Keep copies connected to current and future ICB changes without another `Register()` call, including after the source `TStorageConfig` is destroyed. Keep later raw protobuf changes local to each object.
- Share config-bound controls only through `TStorageConfig` copy construction, where the source raw proto is copied at the same time. `GetStorageConfigControls()` still returns null for config-bound controls, and the public three-argument constructor still accepts only config-independent controls or null. Callers therefore cannot attach config-bound controls to an unrelated proto.

#### `TControlBoard` registration behavior

- Bind every `TStorageConfigControls` instance, in either mode, to the first `TControlBoard` passed to `Register()`. Re-registering the same instance on that board is a no-op; registering it on another board aborts in both modes.
- Preserve the existing same-board behavior for config-bound controls. During their first registration, wrappers may adopt controls already registered under the same names on that board.
- Before this change, the same config-bound wrappers could be registered on several boards. A later registration could silently repoint the wrappers to controls already stored on another board, while the first board retained controls that the `TStorageConfig` no longer read. This was possible through the API but was not an explicit supported contract.
- No current server or DiskAgent path re-registers one controls instance on multiple boards. Each existing config-bound object registers once with its node-local board, so the known in-repository runtime behavior remains unchanged.

#### Config-independent control collisions

- Initialize config-independent controls with the reserved `NoOverride` default. The value means that no explicit ICB override exists and is outside the range set for ICB Control.
- `RegisterSharedControl()` resolves a name collision by replacing the wrapper's underlying `TControl` with the object already registered on the board; it does not verify the control mode or compatibility.
- If a config-independent wrapper adopted a config-bound control, it would also adopt that control's protobuf-derived default, bounds, and current value. The protobuf-derived default differs from `NoOverride`, so it would be treated as an explicit override. `RestoreDefault()` would restore that value and could no longer remove the override.
- Require the first registration of a config-independent set to insert every control name. Any collision aborts, ensuring that the registered controls retain the `NoOverride` semantics.
- Adopting another compatible config-independent set could be valid in principle, but `TControlBoard` does not identify the mode or provide a complete compatibility contract. The implementation therefore rejects all collisions instead of assuming that existing controls are compatible. Register one config-independent set per board and share it between `TStorageConfig` instances.

#### Target controls model

- The config-independent mode is the target model: register one control set per board, share it between configurations, and keep protobuf defaults in each `TStorageConfig` while ICB stores only explicit overrides.
- Keep config-bound controls in this PR for compatibility with existing construction and registration behavior. After callers migrate to config-independent controls, remove the config-bound mode and simplify the constructors, default handling, and mode-specific control logic.

### Issue

#6685

### DIFF
--- a/cloud/blockstore/libs/storage/core/config.cpp
+++ b/cloud/blockstore/libs/storage/core/config.cpp
@@ -9,6 +9,7 @@
 #include <library/cpp/monlib/service/pages/templates.h>
 
 #include <util/generic/size_literals.h>
+#include <util/system/mutex.h>
 
 #include <google/protobuf/text_format.h>
 #include <google/protobuf/util/message_differencer.h>
@@ -1003,32 +1004,191 @@ constexpr TAtomicBase ConvertToAtomicBase(const TDuration& value)
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct TStorageConfig::TImpl
+// The private control collection owned by one TStorageConfigControls. It keeps
+// one TControlWrapper for every read-write storage setting and applies one of
+// these default-value policies:
+//  - a config-bound set:
+//      - created by the two-argument TStorageConfig constructor
+//      - uses values from its initial proto as ICB defaults
+//      - may be shared only by TStorageConfig copy construction, which also
+//        copies the raw proto
+//  - a config-independent set:
+//      - uses a reserved default to represent "no override" (represents only
+//        explicit ICB overrides)
+//      - may be used by (shared between) multiple TStorageConfig instances
+struct TStorageConfigControls::TImpl
 {
-    NProto::TStorageServiceConfig StorageServiceConfig;
-    NFeatures::TFeaturesConfigPtr FeaturesConfig;
+    // The reserved default used by config-independent controls to mean
+    // "no override". Their lower bound is NoOverride + 1, so ICB cannot set
+    // this value itself.
+    static constexpr TAtomicBase NoOverride = Min<TAtomicBase>();
 
-#define BLOCKSTORE_CONFIG_CONTROL(name, type, value)                           \
-    TControlWrapper Control##name{                                             \
-        ConvertToAtomicBase<type>(BLOCKSTORE_CONFIG_GET_CONFIG_VALUE(          \
-            StorageServiceConfig,                                              \
-            name,                                                              \
-            type,                                                              \
-            value)),                                                           \
-        Min<i64>(),                                                            \
-        Max<i64>()};                                                           \
-    // BLOCKSTORE_CONFIG_CONTROL
+    // The control-set mode: true for config-independent defaults that represent
+    // absent overrides; false for defaults bound to one configuration proto.
+    const bool ConfigIndependent;
+
+    // Registration-state lock. It serializes the first registration and
+    // protects RegisteredBoard.
+    TMutex RegistrationLock;
+
+    // Non-owning address of the registered board. It is null before Register(),
+    // then remains equal to that board for the lifetime of this control set.
+    // Access requires RegistrationLock.
+    TControlBoard* RegisteredBoard = nullptr;
+
+    // One wrapper per BLOCKSTORE_STORAGE_CONFIG_RW field. Each wrapper stores
+    // the field's ICB default, allowed range, and current ICB value.
+#define BLOCKSTORE_CONFIG_CONTROL(name, ...) TControlWrapper Control##name;
 
     BLOCKSTORE_STORAGE_CONFIG_RW(BLOCKSTORE_CONFIG_CONTROL)
 
 #undef BLOCKSTORE_CONFIG_CONTROL
 
+    explicit TImpl(
+        const NProto::TStorageServiceConfig& storageServiceConfig,
+        bool configIndependent)
+        : ConfigIndependent(configIndependent)
+    {
+#define BLOCKSTORE_CONFIG_RESET(name, type, value)                             \
+    Control##name.Reset(                                                       \
+        ConfigIndependent                                                      \
+            ? NoOverride                                                       \
+            : ConvertToAtomicBase<type>(BLOCKSTORE_CONFIG_GET_CONFIG_VALUE(    \
+                  storageServiceConfig,                                        \
+                  name,                                                        \
+                  type,                                                        \
+                  value)),                                                     \
+        ConfigIndependent ? NoOverride + 1 : Min<TAtomicBase>(),               \
+        Max<TAtomicBase>());                                                   \
+    // BLOCKSTORE_CONFIG_RESET
+
+        BLOCKSTORE_STORAGE_CONFIG_RW(BLOCKSTORE_CONFIG_RESET)
+
+#undef BLOCKSTORE_CONFIG_RESET
+    }
+
+    std::optional<i64> ReadOverride(const TControlWrapper& control) const
+    {
+        const TAtomicBase value = static_cast<TAtomicBase>(control);
+        if (value == control.GetDefault()) {
+            return std::nullopt;
+        }
+
+        return static_cast<i64>(value);
+    }
+};
+
+TStorageConfigControls::TStorageConfigControls()
+    : Impl(new TImpl(NProto::TStorageServiceConfig(), true))
+{}
+
+TStorageConfigControls::TStorageConfigControls(
+    const NProto::TStorageServiceConfig& storageServiceConfig)
+    : Impl(new TImpl(storageServiceConfig, false))
+{}
+
+TStorageConfigControls::~TStorageConfigControls() = default;
+
+void TStorageConfigControls::Register(TControlBoard& controlBoard)
+{
+    TGuard<TMutex> guard(Impl->RegistrationLock);
+
+    // Another board could replace wrappers with controls registered under the
+    // same names there, so this control set remains bound to its first board.
+    Y_ABORT_UNLESS(
+        !Impl->RegisteredBoard || Impl->RegisteredBoard == &controlBoard);
+
+    if (Impl->RegisteredBoard) {
+        return;
+    }
+
+    bool allControlsWereInserted = true;
+
+#define BLOCKSTORE_CONFIG_REGISTER(name, ...)                                  \
+    allControlsWereInserted &= controlBoard.RegisterSharedControl(             \
+        Impl->Control##name,                                                   \
+        "BlockStore_" #name);                                                  \
+    // BLOCKSTORE_CONFIG_REGISTER
+
+    BLOCKSTORE_STORAGE_CONFIG_RW(BLOCKSTORE_CONFIG_REGISTER)
+
+#undef BLOCKSTORE_CONFIG_REGISTER
+
+    // On a name collision, RegisterSharedControl replaces the wrapper's control
+    // with the TControl object already registered by another control set.
+    //  - a config-independent wrapper would then lose its NoOverride default
+    //    and adopt that object's default, bounds, and current value
+    //  - config-bound wrappers intentionally preserve the existing
+    //    shared-registration behavior.
+    // Register one config-independent set per board and share it between
+    // TStorageConfig instances; abort if this contract is violated.
+    Y_ABORT_UNLESS(!Impl->ConfigIndependent || allControlsWereInserted);
+    Impl->RegisteredBoard = &controlBoard;
+}
+
+std::optional<i64> TStorageConfigControls::GetOverride(TStringBuf name) const
+{
+#define BLOCKSTORE_CONFIG_GET_OVERRIDE(field, ...)                             \
+    if (name == #field) {                                                      \
+        return Impl->ReadOverride(Impl->Control##field);                       \
+    }                                                                          \
+    // BLOCKSTORE_CONFIG_GET_OVERRIDE
+
+    BLOCKSTORE_STORAGE_CONFIG_RW(BLOCKSTORE_CONFIG_GET_OVERRIDE)
+
+#undef BLOCKSTORE_CONFIG_GET_OVERRIDE
+
+    return std::nullopt;
+}
+
+bool TStorageConfigControls::RestoreDefault(TStringBuf name)
+{
+    {
+        TGuard<TMutex> guard(Impl->RegistrationLock);
+        if (!Impl->RegisteredBoard) {
+            return false;
+        }
+    }
+
+#define BLOCKSTORE_CONFIG_RESTORE_DEFAULT(field, ...)                          \
+    if (name == #field) {                                                      \
+        /* Assign only GetDefault(), because operator= rewrites the control's  \
+           Default as well as control's Value */                               \
+        Impl->Control##field = Impl->Control##field.GetDefault();              \
+        return true;                                                           \
+    }                                                                          \
+    // BLOCKSTORE_CONFIG_RESTORE_DEFAULT
+
+    BLOCKSTORE_STORAGE_CONFIG_RW(BLOCKSTORE_CONFIG_RESTORE_DEFAULT)
+
+#undef BLOCKSTORE_CONFIG_RESTORE_DEFAULT
+
+    return false;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TStorageConfig::TImpl
+{
+    NProto::TStorageServiceConfig StorageServiceConfig;
+    NFeatures::TFeaturesConfigPtr FeaturesConfig;
+
+    // Non-null ICB controls for all read-write storage fields. The
+    // three-argument constructor can reuse a config-independent collection;
+    // the two-argument constructor creates a config-bound collection; the copy
+    // constructor shares the collection in either mode.
+    TStorageConfigControlsPtr Controls;
+
     TImpl(
-            NProto::TStorageServiceConfig storageServiceConfig,
-            NFeatures::TFeaturesConfigPtr featuresConfig)
+        NProto::TStorageServiceConfig storageServiceConfig,
+        NFeatures::TFeaturesConfigPtr featuresConfig,
+        TStorageConfigControlsPtr controls)
         : StorageServiceConfig(std::move(storageServiceConfig))
         , FeaturesConfig(std::move(featuresConfig))
-    {}
+        , Controls(std::move(controls))
+    {
+        Y_ABORT_UNLESS(Controls);
+    }
 
     void SetFeaturesConfig(NFeatures::TFeaturesConfigPtr featuresConfig)
     {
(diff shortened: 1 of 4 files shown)


nisqatsi:
## Summary

Проверил полный diff на head `3e5decc1a9a5cdab5ab7274b6ef77af9088779ef`, семантику `TControl`/`TControlBoard`, копирование и runtime call sites. Новая модель direct reuse/copy в целом последовательна, но нашёл два важных пробела корректности: **0 🔴 / 2 🟡 / 0 🟢 / 0 💡**.

## Issues

🟡 **[важно]** [`config.cpp:1064`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config.cpp#L1064-L1068) — проверка override и получение его значения выполняются двумя независимыми атомарными чтениями

`HasOverride()` сначала читает control, после чего [`GetOverride()`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config.cpp#L1120-L1125), [`GetStorageConfigProto()`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config.cpp#L1195-L1202) и RW-геттеры читают его повторно. Между чтениями ICB может выполнить `RestoreDefault(s)`:

1. первое чтение видит обычный override;
2. restore записывает `NoOverride`;
3. второе чтение получает sentinel и обрабатывает его как реальное значение.

**Риск:** `GetOverride()` может вернуть `optional(NoOverride)` вместо `nullopt`; геттер может вернуть `true` для `bool`, `0` для `ui32`, `2^63` для `ui64` или некорректную/обёрнутую duration; `GetStorageConfigProto()` может записать такое значение в proto. ICB специально использует атомарные чтения/записи, поэтому параллельный restore с чтением конфигурации является штатным сценарием.

**Решение:** заменить `HasOverride() -> bool` на helper, который один раз читает current value и возвращает `std::optional<i64>`. Все три consumer-а должны проверять и использовать один и тот же snapshot. Для config-independent режима snapshot сравнивается с `NoOverride`, для config-bound — с default.

🟡 **[важно]** [`config.cpp:1224`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config.cpp#L1224-L1255) — новый config-independent режим теряется при `TStorageConfig::Merge`

[`Merge()`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config.cpp#L1404-L1420) сначала материализует активные ICB overrides через `GetStorageConfigProto()`, затем накладывает patch и создаёт результат двухаргументным конструктором с новым config-bound control set. Поэтому любой реально изменивший proto patch:

- превращает получившийся effective snapshot в raw proto (поля из patch при этом перезаписывают текущие ICB-значения тех же полей);
- теряет исходный shared `TStorageConfigControls`;
- перестаёт видеть последующие ICB updates и restore.

**Риск:** после перехода global config на reusable controls volume configs со storage patch будут незаметно отключены от live ICB. Именно эти объекты создаются существующими call sites в [`volume_actor_loadstate.cpp:112`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/volume/volume_actor_loadstate.cpp#L110-L113) и [`volume_actor_change_storage_config.cpp:60`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/volume/volume_actor_change_storage_config.cpp#L59-L61). В текущем commit production ещё использует legacy-конструктор, но новый публичный режим уже не компонуется с основным factory для patched configs.

**Решение:** сохранить live controls в результате и явно зафиксировать precedence для поля, одновременно заданного patch и ICB. Если для нового режима действует заявленный приоритет ICB над proto, нужно применять patch к raw `Impl->StorageServiceConfig` и передавать тот же controls pointer в трёхаргументный конструктор. Если должна сохраниться текущая семантика `patch > ICB` из [`ShouldOverrideConfigsViaImmediateControlBoard`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config_ut.cpp#L249-L307), потребуется маска patched fields (shared controls применяются только к остальным полям). В обоих вариантах нужен тест `shared controls + Merge + последующие SetValue/RestoreDefault`, включая patch того же поля; legacy config-bound ветку можно оставить без изменения.

cloud/blockstore/libs/storage/core/config.cpp
```diff
@@ -1003,32 +1003,179 @@ constexpr TAtomicBase ConvertToAtomicBase(const TDuration& value)
...
+
+        BLOCKSTORE_STORAGE_CONFIG_RW(BLOCKSTORE_CONFIG_RESET)
+
+#undef BLOCKSTORE_CONFIG_RESET
+    }
+
+    bool HasOverride(const TControlWrapper& control) const
+    {
+        return ConfigIndependent
+                   ? static_cast<TAtomicBase>(control) != NoOverride
```
nisqatsi: 🟡 **Раздельные чтения control позволяют применить `NoOverride` как значение**

**Severity:** Major | **Confidence:** High

`HasOverride()` выполняет первое атомарное чтение текущего значения `TControlWrapper`. После результата `true` методы `GetOverride()`, `GetStorageConfigProto()` и RW-геттеры независимо читают тот же `TControlWrapper` ещё раз, чтобы получить применяемое значение. Если между чтениями `RestoreDefault` запишет `NoOverride`, место повторного чтения обработает sentinel как настоящий override.

**Риск:**
- `TStorageConfigControls::GetOverride()` вернёт `optional(NoOverride)` вместо `nullopt`.
- RW-геттер преобразует sentinel в `true` для `bool`, `0` для `ui32`, `2^63` для `ui64` либо в некорректную duration.
- `TStorageConfig::GetStorageConfigProto()` запишет преобразованный sentinel в protobuf.

**Предлагаемое решение:** Считать текущее значение `TControlWrapper` один раз, вернуть этот snapshot как `std::optional<i64>` и использовать его во всех трёх местах чтения.

```cpp
std::optional<i64> ReadOverride(const TControlWrapper& control) const
{
    const TAtomicBase value = static_cast<TAtomicBase>(control);
    const bool hasOverride = ConfigIndependent
        ? value != NoOverride
        : value != control.GetDefault();

    if (!hasOverride) {
        return std::nullopt;
    }
    return static_cast<i64>(value);
}
```

nisqatsi (commented):
## AI Review Summary

**Verdict:** ✅ No critical issues found

### Direct issues
- 🟡 **Major | High:** Раздельные чтения control позволяют применить `NoOverride` как значение — `cloud/blockstore/libs/storage/core/config.cpp:1067`

### Affected-context issues
- 🟡 **Major | High:** `TStorageConfig::Merge()` теряет общие ICB controls — `cloud/blockstore/libs/storage/core/config.cpp:1228`

### Positive observations
- Новые тесты отдельно проверяют config-bound и config-independent режимы, совместное использование controls при копировании и сохранение live ICB-связи у копии после уничтожения исходного `TStorageConfig`.

cloud/blockstore/libs/storage/core/config.cpp
```diff
@@ -1003,32 +1003,179 @@ constexpr TAtomicBase ConvertToAtomicBase(const TDuration& value)
...
-struct TStorageConfig::TImpl
+// The private control collection owned by one TStorageConfigControls. It keeps
+// one TControlWrapper for every read-write storage setting and applies one of
+// these default-value policies:
+//  - a config-bound set:
+//      - created by the two-argument TStorageConfig constructor
+//      - uses values from its initial proto as ICB defaults
+//      - may be shared only by TStorageConfig copy construction, which also
+//        copies the raw proto
+//  - a config-independent set:
```
nisqatsi: (test)

cloud/blockstore/libs/storage/core/config.cpp
```diff
@@ -1066,13 +1213,47 @@ struct TStorageConfig::TImpl
...
+{}
+
+TStorageConfig::TStorageConfig(
+    NProto::TStorageServiceConfig storageServiceConfig,
+    NFeatures::TFeaturesConfigPtr featuresConfig,
+    TStorageConfigControlsPtr controls)
+{
+    Y_ABORT_UNLESS(!controls || controls->Impl->ConfigIndependent);
+
+    if (!controls) {
```
nisqatsi, line 1244: **[OUTER CONTEXT]**

🟡 **`TStorageConfig::Merge()` теряет общие ICB controls**

**Severity:** Major | **Confidence:** High

**Уже в целевой ветви (`main`):** [`TStorageConfig::Merge()`](https://github.com/ydb-platform/nbs/blob/b647e50987197120c1d6254da8efe27b48e41b34/cloud/blockstore/libs/storage/core/config.cpp#L1228-L1244) формирует эффективный protobuf через `GetStorageConfigProto()`, накладывает patch через `MergeFrom()` и создаёт результат двухаргументным конструктором. Для прежней config-bound модели этот путь фиксирует текущие эффективные ICB-значения в новом config. Кроме того, для volume с `StorageConfigPatch` созданные `Merge()` config-bound controls не зарегистрированы в host `TControlBoard`, поэтому результат перестаёт получать последующие изменения ICB.

Связанная проблема: [#6927 — Volume StorageConfigPatch disables live ICB updates](https://github.com/ydb-platform/nbs/issues/6927).

**Добавлено или изменено в PR:** [новые конструкторы `TStorageConfig`](https://github.com/ydb-platform/nbs/blob/3e5decc1a9a5cdab5ab7274b6ef77af9088779ef/cloud/blockstore/libs/storage/core/config.cpp#L1215-L1239) позволяют хранить общий config-independent `TStorageConfigControls`, но двухаргументный путь передаёт `nullptr`, а ветка на строках 1231–1233 создаёт новый config-bound control set. Поскольку существующий `Merge()` не передаёт controls исходного config, результат теряет общий `TStorageConfigControls`. Кроме того, `Merge()` сначала материализует ICB override в protobuf, а затем patch того же поля перезаписывает его, поэтому patch получает приоритет над явным override. Так PR делает прежнюю реализацию `Merge()` несовместимой с новым reusable-controls режимом.

**Риск:** Результат `Merge()` перестаёт видеть последующие `SetValue` и `RestoreDefault`; patch того же поля может стать видимым даже при активном явном override. После перевода global config на reusable controls это затронет volume configs со storage patch.

**Предлагаемое решение:** Для config-independent режима накладывать patch на raw proto и передавать те же controls в результат; для config-bound режима сохранить текущее поведение.

```cpp
auto controls = config->GetStorageConfigControls();
const auto configProto = controls
    ? config->Impl->StorageServiceConfig
    : config->GetStorageConfigProto();

auto patchedConfigProto = configProto;
patchedConfigProto.MergeFrom(patch);
if (google::protobuf::util::MessageDifferencer::Equals(
        patchedConfigProto,
        configProto))
{
    return config;
}

return std::make_shared<TStorageConfig>(
    std::move(patchedConfigProto),
    config->Impl->FeaturesConfig,
    std::move(controls));
```

Добавьте тест для `shared controls + Merge + последующие SetValue/RestoreDefault`, включая patch того же поля.

nisqatsi:
>Summary

>No crash paths, OOB, or memcpy/memset anywhere in the diff; the only aborts are the intentional Y_ABORT_UNLESS checks. The real problems are one newly-introduced race that can leak the sentinel into getters (already flagged in the PR thread — I agree with it), plus a fair amount of mode-dependent branching that isn't actually needed.

>Correctness

>1. Sentinel leak via check-then-read (config.cpp:1067, 1197, 1291) — already flagged in the thread, and I want to stress it's new, not pre-existing. Before this PR the !IsDefault() → read race was benign: a re-read after RestoreDefault returns the proto-equivalent default. Now in config-independent mode the re-read returns INT64_MIN, which ConvertValue turns into true for all 99 bool RW fields, 0 for ui32, 2^63 for ui64, and TDuration::Zero() for durations ((ui64)INT64_MIN * 1000 wraps to 0). The fix is a single-read std::optional<i64> ReadOverride(const TControlWrapper&) used by the getter macro, GetStorageConfigProto(), and GetOverride() — which also removes three copies of the same pattern.

fixed

>2. HasOverride doesn't need the mode branch. NoOverride is installed as the wrapper's Default, and Default never changes after construction (RestoreDefault(name) writes GetDefault() back; collision adoption aborts). So value != NoOverride ≡ !IsDefault() in both modes. I verified this against a faithful copy of TControl on every mutation path (SetFromHtmlRequest, RestoreDefault, Set, sentinel write). Collapsing it leaves ConfigIndependent affecting only the initial default and Register's collision policy.

fixed

>3. The NoOverride + 1 lower bound doesn't do what the comment says. TControl::SetFromHtmlRequest bypasses bounds entirely when newValue == Default, so ICB can write INT64_MIN. It's harmless (equivalent to restore-default), but the comment at config.cpp:1024 and the PR description ("outside the range accepted from ICB") are wrong. Simplest is to drop the mode-dependent lower bound; otherwise fix the comment.

`NoOverride` действительно можно явно передать через ICB, потому что значение, равное `Default`, не ограничивается bounds. Функционально это эквивалентно Restore и не создаёт отдельного override.
Комментаирй поправлен.

>4. RestoreDefault(name) requires RegisteredBoard but never uses it. Looks vestigial. It conflates "unregistered" with "unknown field" in the return value, and the test locks that in. Drop the check. Separately: it restores via operator=, which is TControl::Set and rewrites Default as well as Value — correct only because the written value equals Default. Worth a one-line comment so nobody later "optimizes" it into a value-only write with a different argument.

Текущий заложенный Restore-контракт и тест оставлены до уточнения Minor 2. Комментарий добавлен.

>5. Merge vs. copy constructor have opposite sharing semantics (also flagged in-thread). The copy ctor exists so a derived config keeps live controls, but the one in-repo derivation path — Merge, used by both volume actors for StorageConfigPatch — still materializes overrides and creates fresh unregistered config-bound controls. Nothing hits it today since no production code uses config-independent mode, but either wire Merge through the three-arg ctor now (with a test, and an explicit decision on patch-vs-override precedence — today patch wins, with shared controls the override would win) or defer the copy ctor to #6903 where it has a consumer. Right now the copy ctor's only user is the test.

fixed

>6. Register has no guard against concurrent first registration of a shared set. Two threads pass the null check, both call RegisterSharedControl, the loser gets false → config-independent abort. Init-time only in current callers, but sharing is the point of the class, so a comment stating the single-threaded-init requirement (or a guard) is worthwhile.

fixed

>Tests

>1. The property the sentinel exists for is untested: in config-independent mode, set ICB to a value that equals one config's proto value (e.g. 100) and assert the 300-config also reads 100. That's the observable difference from config-bound mode; ShouldSetAndRestoreConfigIndependentControlsViaIcb sets 200, which distinguishes nothing.

fixed

>2. ShouldCopyConfigAndShareControlsViaIcb: static_cast<bool>(copy->GetStorageConfigControls()) == static_cast<bool>(controls) should be pointer equality for the non-null case.

fixed

>3. UNIT_ASSERT(!controls->RestoreDefault(...)) before Register pins the vestigial contract from item 4.

>4. GetStorageConfigProto() in config-independent mode has no coverage; it drives Merge and the config dump in config_initializer.cpp.

>Minor

>1. Y_ABORT_UNLESS(!controls || controls->Impl->ConfigIndependent) in the public ctor is unreachable — config-bound controls are never handed out (GetStorageConfigControls returns null, ctor is private). Fine as a cheap assert, but the header comment implies it's a meaningful runtime guard.

Недостижимо сейчас, но может быть нарушено последующими правками. Оставлено. Убран только избыточный комментарий из .h.

>2. Design consequence of the target model, not a bug here: with the sentinel, the ICB HTML page will show -9223372036854775808 as Default/Current for all 426 BlockStore_* controls, so operators lose the effective-value view they have today.

В текущей реализации ICB в YDB оптимальное решение невозможно. Данный вопрос будет дополнительно проработан на следующих этапах разработки.

cloud/blockstore/libs/storage/core/config.cpp
nisqatsi: fixed

cloud/blockstore/libs/storage/core/config.cpp
nisqatsi: fixed

cloud/blockstore/libs/storage/core/config.cpp
```diff
@@ -1066,13 +1226,47 @@ struct TStorageConfig::TImpl
...
+          other.Impl->FeaturesConfig,
+          other.Impl->Controls))
 {}
 
 TStorageConfig::~TStorageConfig() = default;
 
+TStorageConfigControlsPtr
+TStorageConfig::GetStorageConfigControls() const
+{
+    return Impl->Controls->Impl->ConfigIndependent ? Impl->Controls : nullptr;
```
SvartMetal: Impl->Controls->Impl->

Looks cryptic

nisqatsi: Ну как есть. Это не в public, только в cpp. Добавлять специальные геттеры - будет больше кода ради одной строчки.

SvartMetal: В остальных местах тоже есть всякие `Impl->Controls->Impl->Control##name`

SvartMetal (approved):

agalibin (approved):

PR #6903, opened 2026-08-25
[NBS]  Add aggregate blockstore configuration with atomic access

nisqatsi:
### Notes

Introduce the aggregate BlockStore configuration boundary required for dynamic configuration. This preparatory PR defines immutable merged snapshots, atomic snapshot publication, and process-wide read access.

The provider is not yet initialized from bootstrap or used by runtime components, so the existing configuration delivery path remains active. This PR builds on the reusable Storage ICB controls from #6902, which is already merged into `main`.

### Changes

- Add a 19-section aggregate protobuf and `MakeBlockstoreConfig()` factories for regular and bootstrap construction.
- Apply standard protobuf merge rules to ordinary fields. Merge Features by name: the first record within each source wins, a complete dynamic record replaces the matching static record at its existing position, and dynamic-only records are appended in source order.
- Add `IBlockstoreConfig`, which owns const pointers to runtime wrappers and protobuf sections. Readers can retain a complete snapshot or an individual section after a newer snapshot is published.
- Add `TBlockstoreConfigHolder` as the single-writer publication point that atomically replaces the current snapshot.
- Add `GetCurrentBlockstoreConfig()` and bootstrap-only provider initialization. Bootstrap must bind the process-wide holder once, before concurrent reads start; reset and rebind attempts abort.
- Preserve live Storage ICB controls across snapshots in both config-bound and config-independent modes. Rename `GetStorageConfigProto()` to `GetEffectiveStorageConfigProto()` to make its existing ICB-aware serialization behavior explicit.

<details>
<summary>File summaries</summary>

| File or group | Summary |
| --- | --- |
| `cloud/blockstore/config/blockstore.proto` | New aggregate schema for the 19 supported top-level BlockStore configuration sections. |
| `cloud/blockstore/libs/config/blockstore_config.h` | New immutable aggregate interface, factory contracts, merge API, and extra runtime parameters. |
| `cloud/blockstore/libs/config/blockstore_config.cpp` | New snapshot ownership, aggregate merge, wrapper construction, and Feature overlay implementation. |
| `cloud/blockstore/libs/config/blockstore_config_holder.h` | New single-writer, multi-reader holder interface for atomic snapshot publication. |
| `cloud/blockstore/libs/config/blockstore_config_holder.cpp` | New `THotSwap`-based snapshot load and publication implementation. |
| `cloud/blockstore/libs/config/blockstore_config_provider.h` | New process-wide read API for the current BlockStore configuration. |
| `cloud/blockstore/libs/config/blockstore_config_provider_private.h` | New bootstrap-only API for one-time provider initialization. |
| `cloud/blockstore/libs/config/blockstore_config_provider.cpp` | New process-lifetime holder binding and current-snapshot access implementation. |
| `cloud/blockstore/libs/config/ut/blockstore_config_ut.cpp` | New aggregate merge, ownership, ICB, publication, and provider coverage. |
| Runtime wrapper pointer aliases | Add const shared-pointer aliases for wrapper types stored by immutable aggregate snapshots. |
| Proto-backed runtime wrappers | Expose the source protobuf already owned by each wrapper for aggregate configuration assembly. |
| `cloud/blockstore/libs/storage/core/config.{h,cpp}` | Accept const Features ownership and rename the Storage protobuf accessor to clarify its existing ICB-aware behavior. |
| `cloud/blockstore/libs/storage/core/config_ut.cpp` | Update renamed calls and verify that effective protobuf serialization includes live ICB values. |
| Storage configuration accessor call sites | Update YDB and DiskAgent initializers, Storage service and Volume consumers, monitoring, and existing tests for the accessor rename without changing their behavior. |
| `cloud/storage/core/libs/features/features_config.{h,cpp,features_config_ut.cpp}` | Own and expose the source Features protobuf and verify that it remains independent of the caller. |
| Build files | Register the aggregate protobuf, configuration library, and unit-test target. |

</details>

### Issue

#6685

### DIFF
--- a/cloud/blockstore/libs/config/blockstore_config.cpp
+++ b/cloud/blockstore/libs/config/blockstore_config.cpp
@@ -1,0 +1,376 @@
+/*******************************************************************************
+
+Configuration construction owns immutable runtime wrappers for all merged
+sections. Storage configuration copies retain the shared ICB controls and use
+the features wrapper created from the merged protobuf configuration.
+
+*******************************************************************************/
+
+#include "blockstore_config.h"
+
+#include <util/generic/hash.h>
+#include <util/generic/hash_set.h>
+#include <util/generic/vector.h>
+#include <util/system/yassert.h>
+
+namespace NCloud::NBlockStore {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+// The private IBlockstoreConfig implementation and owner of its runtime
+// wrappers and protobuf sections. Only MakeBlockstoreConfig() creates
+// instances.
+class TBlockstoreConfigImpl final: public IBlockstoreConfig
+{
+public:
+    TBlockstoreConfigImpl(
+        const NProto::TBlockstoreConfig& config,
+        NFeatures::TFeaturesConfigConstPtr featuresConfig,
+        NStorage::TStorageConfigConstPtr storageConfig,
+        NStorage::TDiskAgentConfigConstPtr diskAgentConfig);
+
+    const NServer::TServerAppConfigConstPtr& GetServerConfig() const override;
+
+    const NFeatures::TFeaturesConfigConstPtr&
+    GetFeaturesConfig() const override;
+
+    const NStorage::TStorageConfigConstPtr& GetStorageConfig() const override;
+
+    const TDiagnosticsConfigConstPtr& GetDiagnosticsConfig() const override;
+
+    const NDiscovery::TDiscoveryConfigConstPtr&
+    GetDiscoveryServiceConfig() const override;
+
+    const NClient::TClientAppConfigConstPtr& GetEndpointConfig() const override;
+
+    const NStorage::TDiskAgentConfigConstPtr&
+    GetDiskAgentConfig() const override;
+
+    const NStorage::TDiskRegistryProxyConfigConstPtr&
+    GetDiskRegistryProxyConfig() const override;
+
+    const NSpdk::TSpdkEnvConfigConstPtr& GetSpdkEnvConfig() const override;
+
+    const NRdma::TRdmaConfigConstPtr& GetRdmaConfig() const override;
+
+    const NYdbStats::TYdbStatsConfigConstPtr&
+    GetYdbStatsConfig() const override;
+
+    const NLogbroker::TLogbrokerConfigConstPtr&
+    GetLogbrokerConfig() const override;
+
+    const NNotify::TNotifyConfigConstPtr& GetNotifyConfig() const override;
+
+    const NIamClient::TIamClientConfigConstPtr&
+    GetIamClientConfig() const override;
+
+    const TGrpcClientConfigConstPtr& GetKmsClientConfig() const override;
+
+    const TGrpcClientConfigConstPtr& GetComputeClientConfig() const override;
+
+    const TRootKmsConfigConstPtr& GetRootKmsConfig() const override;
+
+    const NCells::TCellsConfigConstPtr& GetCellsConfig() const override;
+
+    const TLocalNVMeConfigConstPtr& GetLocalNVMeConfig() const override;
+
+private:
+    const NServer::TServerAppConfigConstPtr ServerConfig;
+    const NFeatures::TFeaturesConfigConstPtr FeaturesConfig;
+    const NStorage::TStorageConfigConstPtr StorageConfig;
+    const TDiagnosticsConfigConstPtr DiagnosticsConfig;
+    const NDiscovery::TDiscoveryConfigConstPtr DiscoveryServiceConfig;
+    const NClient::TClientAppConfigConstPtr EndpointConfig;
+    const NStorage::TDiskAgentConfigConstPtr DiskAgentConfig;
+    const NStorage::TDiskRegistryProxyConfigConstPtr DiskRegistryProxyConfig;
+    const NSpdk::TSpdkEnvConfigConstPtr SpdkEnvConfig;
+    const NRdma::TRdmaConfigConstPtr RdmaConfig;
+    const NYdbStats::TYdbStatsConfigConstPtr YdbStatsConfig;
+    const NLogbroker::TLogbrokerConfigConstPtr LogbrokerConfig;
+    const NNotify::TNotifyConfigConstPtr NotifyConfig;
+    const NIamClient::TIamClientConfigConstPtr IamClientConfig;
+    const TGrpcClientConfigConstPtr KmsClientConfig;
+    const TGrpcClientConfigConstPtr ComputeClientConfig;
+    const TRootKmsConfigConstPtr RootKmsConfig;
+    const NCells::TCellsConfigConstPtr CellsConfig;
+    const TLocalNVMeConfigConstPtr LocalNVMeConfig;
+};
+
+TBlockstoreConfigImpl::TBlockstoreConfigImpl(
+    const NProto::TBlockstoreConfig& config,
+    NFeatures::TFeaturesConfigConstPtr featuresConfig,
+    NStorage::TStorageConfigConstPtr storageConfig,
+    NStorage::TDiskAgentConfigConstPtr diskAgentConfig)
+    : ServerConfig(
+          std::make_shared<NServer::TServerAppConfig>(config.GetServer()))
+    , FeaturesConfig(std::move(featuresConfig))
+    , StorageConfig(std::move(storageConfig))
+    , DiagnosticsConfig(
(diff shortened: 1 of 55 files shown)


cloud/blockstore/libs/config/blockstore_config.cpp
```diff
@@ -0,0 +1,355 @@
...
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+NProto::TBlockstoreConfig MergeBlockstoreConfig(
+    const NProto::TBlockstoreConfig& staticConfig,
+    const NProto::TBlockstoreConfig& dynamicConfig)
+{
+    auto result = staticConfig;
+    result.MergeFrom(dynamicConfig);
```
nisqatsi, line 321: **[OUTER CONTEXT]**

🟡 **Dynamic-слой не переопределяет существующую feature с тем же именем**

**Severity:** Major | **Confidence:** High

**Уже в целевой ветви (`users/usenkoav/0100-26_2_1-nbs_dynconfig-01-storage_config`, `f423c0c56c55e557c9d6c48682f46e5ea77dd993`):** `TFeaturesConfig` обходит `Features` по порядку и использует `emplace`, поэтому первая запись с данным `Name` остаётся в lookup map, а последующие дубликаты игнорируются (`cloud/storage/core/libs/features/features_config.cpp:35-40`).

**Добавлено или изменено в PR:** `MergeBlockstoreConfig()` вызывает `MergeFrom()` (`cloud/blockstore/libs/config/blockstore_config.cpp:297-303`), поэтому protobuf добавляет repeated `dynamicConfig.Features` после static-элементов. Затем `MakeBlockstoreConfig()` строит runtime `TFeaturesConfig` из этого списка (`cloud/blockstore/libs/config/blockstore_config.cpp:314-316`). Если оба слоя содержат feature с одинаковым `Name`, lookup сохраняет static-версию и полностью игнорирует dynamic-версию.

**Риск:** Публикация новой конфигурации завершится успешно, но изменение value, whitelist, blacklist или probability существующей feature фактически не вступит в силу. Сервис продолжит работать со static-настройкой без ошибки или диагностического сигнала.

**Предлагаемое решение:** Определить merge-политику для `Features` как для списка, ключом которого является `Name`, и обеспечить приоритет dynamic-записи над static-записью. Добавить тест, где оба слоя задают одно имя с разными `Value` или фильтрами, и проверить результат через runtime-accessor `TFeaturesConfig`, а не только порядок элементов protobuf.

nisqatsi (commented):
## AI Review Summary

**Verdict:** ✅ No critical issues found

### Affected-context issues
- 🟡 **Major | High:** Dynamic-слой не переопределяет существующую feature с тем же именем — `cloud/blockstore/libs/config/blockstore_config.cpp:302`

### Positive observations
- Новый holder публикует целую конфигурацию через `THotSwap`, а тесты отдельно проверяют согласованность секций при конкурентном чтении.
- Bootstrap-фабрика сохраняет host-specific параметры DiskAgent и общие ICB controls; эти контракты покрыты отдельными тестами.

cloud/blockstore/libs/config/blockstore_config.cpp
nisqatsi: fixed

nisqatsi:
>Problems and recommendations

>1. Build-hygiene regression: heavy include moved into a hot header. #include <contrib/ydb/core/control/immediate_control_board_impl.h> was moved from storage/core/config.cpp into storage/core/config.h. Nothing in the header needs the complete type — it only takes NKikimr::TControlBoard& by reference, and the forward declaration already comes via cloud/blockstore/libs/kikimr/public.h (that's how the base compiled). config.h is included by essentially the entire blockstore storage layer, so this drags ydb-core headers into hundreds of TUs. Move it back to the .cpp (which also left a stray double blank line where the include was removed).

fixed

>2. Repeated-field merge is a footgun outside Features. result.MergeFrom(dynamicConfig) concatenates every repeated field per standard protobuf rules: a dynamic layer that re-specifies DiskAgent.FileDevices, Cells, certificate lists, etc. will duplicate entries rather than replace them. Features was special-cased precisely because append semantics are wrong for dynamic overrides, but every other repeated field has the same problem and gets none of the treatment. This is stated in the description ("standard protobuf merge rules to ordinary fields"), so it's deliberate for this preparatory PR — but before any real dynamic delivery lands, the replace-vs-append policy for other repeated fields needs to be decided, and it's worth a comment on MergeBlockstoreConfig now.

Будет решаться постепенно при переводе компонентов на динамическое использование соответствующих частей конфигурации.

>3. Getters returning const shared_ptr& undermine the retention contract. The API's whole point is "retain a section after a newer snapshot is published", yet const auto& server = config->GetServerConfig(); dangles the moment the caller drops config — a reference into a destroyed aggregate. The tests copy correctly, but the signature invites the broken pattern. For a snapshot API, return the shared_ptr by value; one refcount bump per access is noise compared to a use-after-free class of bugs.

Возврат shared_ptr по значению добавляет атомарные inc/dec счётчика ссылок на каждое чтение. На hot path это лишняя стоимость, если вызывающий использует доступ через корневой TBlockstoreConfigConstPtr.
При необходимости указатель будет сохраняться через копирование явно, но не подобной конструкцией `const auto& server = config->GetServerConfig()` 

>4. blockstore_config.h couples the world together unnecessarily. It includes 17 full wrapper-config headers, but the interface only needs the ...ConstPtr aliases and incomplete types — which is exactly what the PR's new public.h aliases provide. Include the public.h files (plus blockstore.pb.h) in the header and keep the full includes in the .cpp. As written, every consumer of the aggregate config transitively recompiles against every subsystem's config header.

Увеличенная пересборка при изменениях редко меняемых proto и их оберток - не проблема. Определения разных типов указателей на сабконфиги уже были в их собственных хедерах, к ним рядом добавлены ...ConstPtr. Разносить такие определения типов по разным файлам нецелесообразно.

>5. Provider global: shutdown and misuse behavior. The function-local static TBlockstoreConfigHolderPtr is destroyed at exit, so any thread still calling GetCurrentBlockstoreConfig() during teardown hits a destroyed holder (THotSwap dtor stores nullptr → abort at best, UB at worst). For process-lifetime infrastructure the usual fix is to intentionally leak the holder. Also note Y_ABORT_UNLESS(!holder) / the null check are plain unsynchronized reads — they only work under the documented "bind before readers start" contract; a violation is a silent data race, not a guaranteed abort, so the check gives a slightly false sense of runtime detection.

Контракт уже явно описан.

>6. Provider is one-shot with no test escape hatch. InitializeBlockstoreConfigProvider aborts on a second call and there's no reset, so ShouldExposeCurrentBlockstoreConfig permanently binds the global for the whole test binary — it works today only because it's the sole provider test and the ut target doesn't fork subtests. The next person who adds a second provider test gets an opaque process abort. Add a private/test-only reset, or fork that test.

Добавлен test-only reset для использования в текущем и в последующих тестах для переводимых на новую дин. конфигурацию компонентов. В дальнейшем конкретный вариант может быть уточнен и изменен (например, использование SIMPLE_UNIT_FORKED_TEST).

>7. Minor dead code in MergeFeatures. (a) The early return duplicates the caller's guard in MergeBlockstoreConfig and is unreachable-true there — keep the caller's check (it correctly avoids MutableFeatures() setting the has-bit) and drop the inner one. (b) featureIndices.emplace(...) in the dynamic-append branch is never read: static processing is finished and dynamicFeatureNames already dedupes dynamic names. (c) Each feature is copied twice (into the vector, then AddFeatures()->CopyFrom); building the result's repeated field directly, or swapping in a locally built RepeatedPtrField, would halve that. All harmless, all easy simplifications.

(a) - fixed
(b) - fixed
(c) - править не нужно, не на hot path, и текущая реализация более понятная

>8. Nits. TBlockstoreConfigHolder's ctor initializes Current(initialConfig) before the Y_ABORT_UNLESS(initialConfig) in the body — works (the by-value parameter is still checkable), but check first for clarity. The two MakeBlockstoreConfig overloads have meaningfully different semantics (the bootstrap one ignores the merged proto's StorageService/DiskAgent sections in favor of the passed adapters — tested, but easy to pick the wrong overload); a distinct name like MakeBootstrapBlockstoreConfig would prevent that. TFeaturesConfig now stores every feature twice (proto + parsed map) for all users including filestore — negligible, just noting it's intentional.

1 - fixed
2 - no fix, незачем
3 - no fix, вроде очевидно, что proto нужен для экспорта, внутренняя map - для быстрого поиска

SvartMetal (approved):

agalibin (approved):

Issue #6550, opened 2026-07-17
[Filestore] Shard balancer should limit the set of selected shards to based on the parentNodeId


debnatkh:
The file system currently selects a shard from all shards with enough space when a tablet creates a file. This can spread one node’s children across many shards. As a result, listing requests may need `TListNodesActor` to send `GetNodeAttrBatch` requests to thousands of shards in case of a filesystem with thousands of shards

We want to limit the set of selected shards to a configurable value. The set should be selected using `parentNodeId`. A limit of 100 or 200 shards should be considered reasonable for now.

qkrorlqr:
To make the system more predictable and the debugging - easier I suggest to not use anything like hash(parentNodeId) but just use the next 100-200 shards that go after the shardNo of the shard that owns the parent node

e.g. if the filesystem has 1000 shards and the fs is configured to use no more than 200 shards for direct children and parentNodeId is managed by shard 750, then children can be created in shards 751 - 950

if parentNodeId is managed by shard 950, then children can be created in shards 951-1000 U 1-150, etc

ruslan-gar:
Then, we may find ourselves in a situation when shards added as a result of resizing are not used at all.
Let's say we added 400 new shards and no new directories are created.
In this case the last 200 shards will not be used.

PR #6907, opened 2026-08-25
issue-6550: [Filestore] Deterministic shard balancer limits the set of selected shards

ruslan-gar:
### Notes
Parent node shard is used select shards for child nodes. ParentNodeId is used as hint or, in other words,
as an in initial index for allowed shards window.
Shards are selected in [Hint % ShardCount +1, Hint % ShardCount + ShardsPerDitrectory] range.

### Issue
#6550


cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -1572,7 +1573,8 @@ NProto::TError TIndexTabletState::SelectShard(
         balancer = Impl->FileShardBalancer.get();
     }
 
-    auto e = balancer->SelectShard(fileSize, shardId);
+    auto e =
+        balancer->SelectShard(ExtractShardNo(parentNodeId), fileSize, shardId);
```
debnatkh: /* argname */ comment would be nice here

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -1627,10 +1629,16 @@ NProto::TError TIndexTabletState::UpdateShardBalancer(
...
-TVector<TShardStats>
-TIndexTabletState::MakeOrderedShardList() const
+TString TIndexTabletState::DescribeShardBalancers() const
 {
-    return Impl->ShardBalancer->MakeOrderedShardList();
+    TString ret = Impl->ShardBalancer->Describe();
+
+    if (Impl->FileShardBalancer) {
+        ret += " File shard balancer: ";
+        ret += Impl->FileShardBalancer->Describe();
```
debnatkh: TStringBuilder

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -128,23 +131,35 @@ size_t TShardBalancerBase::FindUpperBoundAmongAllShardsToFitFile(
...
+            << " suffer=" << s.Suffer;
     }
-    return shardStats;
+    return sb;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
 
 NProto::TError TShardBalancerRoundRobin::SelectShard(
+    ui64 hint,
```
debnatkh: i believe hint is better as a last arg

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -191,12 +199,49 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+        bool IsInside(const ui32 idx) const
+        {
+            if (Left <= Right) {
+                return idx >= Left && idx <= Right;
+            } else {
+                return (idx >= Left && idx < ShardCount) || idx <= Right;
+            }
+        }
+
+        ui64 Unwrap(ui64 coord)
```
debnatkh: const

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -191,12 +199,49 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+        {
+            if (coord >= Left) {
+                return coord;
+            } else {
+                return coord + ShardCount;
+            }
+        }
+    };
+
+    const ui64 ShardsPerDirectoryCount;
```
debnatkh: isn't it supposed to be ui32?

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -191,12 +199,49 @@ class TShardBalancerWeightedRandom: public TShardBalancerBase
...
+
+    TVector<TIterator> Iterators;
```
debnatkh, line 257: Internal state should be private

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -223,9 +272,10 @@ class TShardBalancerWeightedDeterministic: public TShardBalancerBase
         std::optional<ui64> desiredFreeSpaceReserve,
         std::optional<ui64> minFreeSpaceReserve) final;
 
-    NProto::TError SelectShard(ui64 fileSize, TString* shardId) final;
-
-    TVector<TShardStats> MakeOrderedShardList() const final;
+    NProto::TError SelectShard(
+        ui64 hint,
```
debnatkh: Generally, I believe it is reasonable to have logic like "If there are fewer shards than shardsPerDirectoryCount, then just do not use this logic at all

debnatkh:
Without DirectoryCreationInShardsEnabled, every directory is in the main tablet, so every file-create uses hint=0 → iterator 0 → window [1 .. ShardsPerDirectoryCount]. The entire filesystem is silently confined to SPD shards; the rest stay empty. Probably we need to respect ShardsPerDirectoryCount only with DirectoryCreationInSHardsEnabled

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -348,13 +400,36 @@ NProto::TError TShardBalancerWeightedDeterministic::Update(
...
 
+void TShardBalancerWeightedDeterministic::SelectShard(TIterator& it)
+{
+    ui64 nextShardIdx = NextShard[it.CurrentScore][it.LastSelectedShard];
+
+    if (!it.IsInside(nextShardIdx) ||
+        (it.IsInside(it.LastSelectedShard) &&
+         it.Unwrap(nextShardIdx) <= it.Unwrap(it.LastSelectedShard)))
+    {
+        it.CurrentScore = (++it.CurrentScore) % ScoreLevelsCount;
```
debnatkh: `(it.CurrentScore + 1) % ScoreLevelsCount`

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -332,6 +355,35 @@ void TShardBalancerWeightedDeterministic::CalcNextShard()
     }
 }
 
+void TShardBalancerWeightedDeterministic::UpdateIterators()
+{
+    if (ShardsPerDirectoryCount && ShardsPerDirectoryCount < Metas.size()) {
+        if (Iterators.size() == Metas.size()) {
+            return;
+        }
+        Iterators.resize(Metas.size());
```
debnatkh, line 370: On any shard-count change, UpdateIterators rebuilds all iterators and resets LastSelectedShard, but keeps the stale CurrentScore of surviving iterators — inconsistent, though low impact since shard count rarely changes.

cloud/filestore/libs/storage/tablet/model/shard_balancer.h
```diff
@@ -194,9 +202,45 @@ class TShardBalancerWeightedDeterministic: public TShardBalancerBase
...
+        bool IsInside(const ui32 idx) const
+        {
+            if (Left <= Right) {
+                return idx >= Left && idx <= Right;
+            } else {
+                return (idx >= Left && idx < ShardCount) || idx <= Right;
+            }
+        }
+
+        ui64 Unwrap(ui64 coord) const
```
debnatkh: IsInside(idx) / Unwrap(coord) — two names (idx, coord) for the same concept across two methods; pick one.

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -332,6 +355,35 @@ void TShardBalancerWeightedDeterministic::CalcNextShard()
...
+        for (ui64 i = 0; i < Iterators.size(); ++i) {
+            auto& it = Iterators[i];
+            it.ShardCount = Metas.size();
+            it.Left = (i + 1) % Metas.size();
+            it.Right = (it.Left + ShardsPerDirectoryCount - 1) % Metas.size();
```
debnatkh: Poking fields directly. UpdateIterators sets it.ShardCount, it.Left, it.Right, (it.LastSelectedShard) field-by-field in two places. Give TIterator an Init(left, right, shardCount) / constructor and the invariant lives in one spot.

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -348,13 +400,36 @@ NProto::TError TShardBalancerWeightedDeterministic::Update(
...
         CalcNextShard();
     }
+    UpdateIterators();
 
     return {};
 }
 
+void TShardBalancerWeightedDeterministic::SelectShard(TIterator& it)
+{
+    ui64 nextShardIdx = NextShard[it.CurrentScore][it.LastSelectedShard];
```
debnatkh: Left, Right, ShardCount, LastSelectedShard, CurrentScore are ui32; IsInside takes ui32; but Unwrap takes/returns ui64, SelectShard declares ui64 nextShardIdx and ui64 ind even though NextShard is TArray2D<ui32> and Iterators is a vector. Every call site does an implicit narrowing/widening. Pick ui32

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -332,6 +355,35 @@ void TShardBalancerWeightedDeterministic::CalcNextShard()
...
+{
+    if (ShardsPerDirectoryCount && ShardsPerDirectoryCount < Metas.size()) {
+        if (Iterators.size() == Metas.size()) {
+            return;
+        }
+        Iterators.resize(Metas.size());
+        for (ui64 i = 0; i < Iterators.size(); ++i) {
+            auto& it = Iterators[i];
+            it.ShardCount = Metas.size();
+            it.Left = (i + 1) % Metas.size();
```
debnatkh: why do we need +1 here?

debnatkh: ind = 0 % N = 0 → iterator 0, whose window is Left = (0+1) % N = 1 (Metas index) to Right = SPD % N, i.e. shard numbers 2 … SPD+1, sounds weird

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -348,13 +400,36 @@ NProto::TError TShardBalancerWeightedDeterministic::Update(
     if (!stats.empty() && CalcScore(stats)) {
         CalcNextShard();
     }
+    UpdateIterators();
 
     return {};
 }
 
+void TShardBalancerWeightedDeterministic::SelectShard(TIterator& it)
```
debnatkh: SelectShard(TIterator&) — private void helper overloading the public virtual SelectShard(ui64, TString*, ui64). Legal but two very different operations under one name; AdvanceIterator(it) / Step(it) is clearer.

cloud/filestore/libs/storage/tablet/model/shard_balancer.cpp
```diff
@@ -364,44 +439,14 @@ NProto::TError TShardBalancerWeightedDeterministic::SelectShard(
...
-    // We are in an undefined part of the NextShard matrix.
-    // That means that we should start from the initial state.
-    if (NextShard[CurrentScore][LastSelectedShard] == Metas.size()) {
-        CurrentScore = 0;
-        LastSelectedShard = Metas.size() - 1;
-    }
-
-    LastSelectedShard = NextShard[CurrentScore][LastSelectedShard];
-    *shardId = Metas[LastSelectedShard].Stats.ShardId;
+    ui64 ind = hint % Iterators.size();
```
debnatkh: ind — abbreviation; the codebase spells Idx/index

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
debnatkh: not related to this PR, but

if (HasError(e)) return e ; return e;

kinda redundant

debnatkh (dismissed):

ruslan-gar:
I reviewed the current HEAD, 557d4922a5e8874a0651da9a418dbda963629cd1, including the PR description, visible review discussion, latest fixup commit, affected call sites, and the deterministic balancer implementation. I don’t have a confirmed correctness bug to report on HEAD.

The main issue raised in existing review — enabling ShardsPerDirectoryCount while DirectoryCreationInShardsEnabled=false would confine all files to the first N shards — is fixed on HEAD. InitShardBalancer() now passes 0 as the restriction unless directory creation in shards is enabled, and the new service test exercises this configuration. The PR discussion shows that this was an existing reviewer finding, so it should not be repeated.

I also specifically checked the dangerous indexing/arithmetic paths. In restricted mode, UpdateIterators() only constructs per-hint iterators when 0 < ShardsPerDirectoryCount < Metas.size(). Left is consequently a valid shard index and the circular Right = (Left + count - 1) % ShardCount describes a non-empty range; unrestricted/oversized counts instead use one full-range iterator. SelectShard() rejects an empty Metas before indexing, and a non-empty balancer always gets at least one iterator. I found no OOB access, bad pointer operation, memcpy/memmove/memset issue, or newly introduced lifetime/concurrency hazard in this patch.

There is one question I would leave on the PR, not a confirmed bug: at cloud/filestore/libs/storage/tablet/model/tablet_state.cpp:1452-1475, FileShardBalancer is constructed over FileShardFileSystemIds only, while the deterministic balancer interprets the hint modulo the size/order of its own Metas array. The creation path supplies the parent node ID as the hint (tablet_actor_createnode.cpp:3124-3132; the create-handle path similarly uses the parent ID). If FileShardFileSystemIds can be a non-prefix/non-contiguous subset of physical shard numbers, “parent shard + 1 … + N” and hint % FileShardBalancer.size() are different coordinate systems. For example, if file-only shards were physical shards {5,6,7,8}, a parent on physical shard 5 would normalize to index 5 % 4 == 1, i.e. start at physical shard 6 rather than the intended successor mapping implied by the PR description. The surrounding construction code does not document a contiguity/order invariant. If the filesystem guarantees that FileShardFileSystemIds are laid out so this normalization is intentional, there is no issue; otherwise this deserves a mapping by actual shard number. No severity assigned because I could not prove that the triggering configuration is valid.

debnatkh (dismissed):

yegorskii (dismissed):

debnatkh (approved):

yegorskii (approved):

PR #6915, opened 2026-08-26
issue-5895: fastshard - added logging context for all requests: logging NodeId/Handle/Name/Pages when possible

qkrorlqr:
### Notes
Logging NodeId, Name, Handle, PageClusterIds, StoragePageClusterIds everywhere where possible. If some of those items are not applicable, they are not logged. Introduced a `TLoggingContext` class for that and using it in all requests. Adds a little bit of overhead but doesn't look like a big deal for now.

### Issue


neihar (approved):
Looks questionable in some places, but ok for current stage of fastshard

arigachnyy (approved):

PR #6916, opened 2026-08-26
[Filestore] do not release barrier if unconfirmed data already deleted

neihar:
### Notes
Do not release barrier if unconfirmed data already deleted
Added ai-assisted UT

Case that happened in test:
```
• tablet restarted
• connection established with client
• client made WriteData
• AddDataUnconfirmed page faulted
• Client disconnected -> Unconfirmed data started deletion
• AddDataUnconfirmed restarted
```

### Issue
#2293

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_ut_unconfirmed_data.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_ut_unconfirmed_data.cpp
@@ -11,7 +11,9 @@
 #include <contrib/ydb/core/base/blobstorage.h>
 #include <contrib/ydb/core/base/logoblob.h>
 #include <contrib/ydb/core/mind/local.h>
+#include <contrib/ydb/core/tablet_flat/shared_cache_events.h>
 #include <contrib/ydb/core/testlib/basics/storage.h>
+#include <contrib/ydb/core/testlib/tx_helpers.h>
 
 #include <library/cpp/testing/unittest/registar.h>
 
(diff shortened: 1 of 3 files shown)


cloud/filestore/libs/storage/tablet/tablet_ut_unconfirmed_data.cpp
```diff
@@ -180,6 +182,76 @@ ui64 RebootTabletAndCreateHandle(TIndexTabletClient& tablet, ui64 nodeId)
...
+    TTestEnv& env,
+    TIndexTabletClient& tablet,
+    ui64 tabletId,
+    ui64 nodeId)
+{
+    auto& runtime = env.GetRuntime();
+
+    NTabletFlatScheme::TSchemeChanges scheme;
+    TString err;
+    const TString policyChange = Sprintf(R"___(
```
qkrorlqr, line 200: isn't there a better way to inject a page fault? e.g. via the recently introduced mechanism that randomly injects page faults in some tests

neihar: Unfortunately, in this specific case, because of FIFO nature of reschedule, I wasn't able to utilize it to guarantee that DeleteUnconfirmed will be right before AddUnconfirmed.
We need not just reschedule TX, but execute another specific TX exactly before.

qkrorlqr (approved):

yegorskii (approved):

Issue #4115, opened 2025-08-11
[Filestore] execute full compaction at the end of each load-test and validate the number of blobs owned by the tablet


SvartMetal:
Yet another check to ensure that the garbage is really being collected by the tablet

PR #6918, opened 2026-08-26
[filestore] support  forced flush, flush_bytes,  collect_garbage operations

tdiff:
### Notes
This change implements initial changes for supporting new types of forced operations on a tablet.

Before this change only "range" forced operations (cleanup, compaction, )  were supported. The linked issue requires support of other operations such as flush, flush_bytes and collect_garbage which naturally operate on whole tablet state, not on a specific range. New operation types do not map well on existing range-based logic.

Some notes:
- "non-range" operations are called "tablet" operations
-  only one forced operation can be running at the same time. This also prevents Flush from running in parallel with other ops (which is normally allowed).
-  there are still some issues (e.g. no way to query queued op state or reliably determine if op has failed), some of which will be fixed in future commits.   
- new mode will be tested with implementation of operations .

### Issue
https://github.com/ydb-platform/nbs/issues/4115

### DIFF
--- a/cloud/filestore/libs/storage/tablet/events/tablet_private.h
+++ b/cloud/filestore/libs/storage/tablet/events/tablet_private.h
@@ -37,6 +37,7 @@ namespace NCloud::NFileStore::NStorage {
     xxx(Flush,                                  __VA_ARGS__)                   \
     xxx(FlushBytes,                             __VA_ARGS__)                   \
     xxx(ForcedRangeOperation,                   __VA_ARGS__)                   \
+    xxx(ForcedTabletOperation,                  __VA_ARGS__)                   \
     xxx(Truncate,                               __VA_ARGS__)                   \
     xxx(ReadBlob,                               __VA_ARGS__)                   \
     xxx(WriteBlob,                              __VA_ARGS__)                   \
@@ -636,6 +637,33 @@ struct TEvIndexTabletPrivate
         {
         }
     };
+
+    //
+    // ForcedTabletOperation
+    //
+
+    enum EForcedTabletOperationMode
+    {
+        Flush = 0,
+        FlushBytes = 1,
+        CollectGarbage = 2,
+    };
+
+    struct TForcedTabletOperationRequest
+    {
+        EForcedTabletOperationMode Mode;
+        TString OperationId;
+
+        TForcedTabletOperationRequest(
+                EForcedTabletOperationMode mode,
+                TString operationId)
+            : Mode(mode)
+            , OperationId(std::move(operationId))
+        {}
+    };
+
+    using TForcedTabletOperationResponse = TEmpty;
+    using TForcedTabletOperationCompleted = TEmpty;
 
     //
     // NodeCreatedInShard
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1184,7 +1184,8 @@ TIndexTabletActor::ProcessForcedRangeOperationRequest(
     }
 
     if (e.GetCode() == S_OK && IsForcedOperationRunning()) {
-        const auto* state = std::get_if<TForcedRangeOperationState>(GetForcedOperationState());
+        const auto* state =
+            std::get_if<TForcedRangeOperationState>(GetForcedOperationState());
         if (state) {
             const auto currentMode = state->Mode;
             if (currentMode == mode) {
@@ -1249,7 +1250,8 @@ TIndexTabletActor::ProcessForcedTabletOperationRequest(
     }
 
     if (e.GetCode() == S_OK && IsForcedOperationRunning()) {
-        const auto* state = std::get_if<TForcedTabletOperationState>(GetForcedOperationState());
+        const auto* state =
+            std::get_if<TForcedTabletOperationState>(GetForcedOperationState());
         if (state) {
             const auto currentMode = state->Mode;
             if (currentMode == mode) {
@@ -1285,23 +1287,26 @@ void TIndexTabletActor::EnqueueForcedOperationIfNeeded(
         return;
     }
 
-    std::visit(TOverloaded{
-        [&](TPendingForcedRangeOperation& state) {
-            auto request =
-                    std::make_unique<TEvIndexTabletPrivate::TEvForcedRangeOperationRequest>(
-                        std::move(state.Ranges),
-                        state.Mode,
-                        std::move(state.OperationId));
-            ctx.Send(ctx.SelfID, request.release());
-        },
-        [&](TPendingForcedTabletOperation& state) {
-            auto request =
-                    std::make_unique<TEvIndexTabletPrivate::TEvForcedTabletOperationRequest>(
-                        state.Mode,
-                        std::move(state.OperationId));
-            ctx.Send(ctx.SelfID, request.release());
-        }
-    }, *pendingRequest.Get());
+    std::visit(
+        TOverloaded{
+            [&](TPendingForcedRangeOperation& state)
+            {
+                auto request = std::make_unique<
+                    TEvIndexTabletPrivate::TEvForcedRangeOperationRequest>(
+                    std::move(state.Ranges),
+                    state.Mode,
+                    std::move(state.OperationId));
+                ctx.Send(ctx.SelfID, request.release());
+            },
+            [&](TPendingForcedTabletOperation& state)
+            {
+                auto request = std::make_unique<
+                    TEvIndexTabletPrivate::TEvForcedTabletOperationRequest>(
+                    state.Mode,
+                    std::move(state.OperationId));
+                ctx.Send(ctx.SelfID, request.release());
+            }},
+        *pendingRequest.Get());
 }
 
 void TIndexTabletActor::HandleForcedOperationStatus(
@@ -1320,7 +1325,8 @@ void TIndexTabletActor::HandleForcedOperationStatus(
         if (rangeState) {
             response->Record.SetRangeCount(rangeState->RangesToCompact.size());
             response->Record.SetProcessedRangeCount(rangeState->Current);
-            response->Record.SetLastProcessedRangeId(rangeState->GetCurrentRange());
+            response->Record.SetLastProcessedRangeId(
+                rangeState->GetCurrentRange());
             response->Record.SetStatus(
                 rangeState->Current < rangeState->RangesToCompact.size()
                     ? TStatus::E_RUNNING
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1137,6 +1137,31 @@ void TIndexTabletActor::HandleForcedOperation(
     const TActorContext& ctx)
 {
     const auto& request = ev->Get()->Record;
+    using TResponse = TEvIndexTablet::TEvForcedOperationResponse;
+    std::unique_ptr<TResponse> response;
+    switch (request.GetOpType()) {
+    case NProtoPrivate::TForcedOperationRequest::E_COMPACTION:
+    case NProtoPrivate::TForcedOperationRequest::E_CLEANUP:
+    case NProtoPrivate::TForcedOperationRequest::E_DELETE_EMPTY_RANGES:
+        response = ProcessForcedRangeOperationRequest(request, ctx);
+        break;
+    case NProtoPrivate::TForcedOperationRequest::E_FLUSH:
+    case NProtoPrivate::TForcedOperationRequest::E_FLUSH_BYTES:
+    case NProtoPrivate::TForcedOperationRequest::E_COLLECT_GARBAGE:
+        response = ProcessForcedTabletOperationRequest(request, ctx);
+        break;
+    default:
+        response = std::make_unique<TResponse>(
+            MakeError(E_ARGUMENT, "unsupported mode"));
+    }
+    NCloud::Reply(ctx, *ev, std::move(response));
+}
+
+std::unique_ptr<TEvIndexTablet::TEvForcedOperationResponse>
+TIndexTabletActor::ProcessForcedRangeOperationRequest(
+    const NProtoPrivate::TForcedOperationRequest& request,
+    const TActorContext& ctx)
+{
     using EMode = TEvIndexTabletPrivate::EForcedRangeOperationMode;
     EMode mode{};
     NProto::TError e;
@@ -1145,33 +1170,30 @@ void TIndexTabletActor::HandleForcedOperation(
             mode = EMode::Compaction;
             break;
         }
-
         case NProtoPrivate::TForcedOperationRequest::E_CLEANUP: {
             mode = EMode::Cleanup;
             break;
         }
-
         case NProtoPrivate::TForcedOperationRequest::E_DELETE_EMPTY_RANGES: {
             mode = EMode::DeleteZeroCompactionRanges;
             break;
         }
-
         default: {
             e = MakeError(E_ARGUMENT, "unsupported mode");
         }
     }
 
     if (e.GetCode() == S_OK && IsForcedOperationRunning()) {
-        const auto* rangeState = std::get_if<TForcedRangeOperationState>(GetForcedOperationState());
-        if (rangeState) {
-            const auto currentMode = rangeState->Mode;
+        const auto* state = std::get_if<TForcedRangeOperationState>(GetForcedOperationState());
+        if (state) {
+            const auto currentMode = state->Mode;
             if (currentMode == mode) {
                 e = MakeError(S_ALREADY, "already launched");
-            } else {
-                e = MakeError(E_TRY_AGAIN, TStringBuilder() << "mode mismatch: "
-                    << static_cast<int>(mode)
-                    << " != " << static_cast<int>(currentMode));
             }
+        } else {
+            e = MakeError(E_TRY_AGAIN, TStringBuilder() << "mode mismatch: "
+                << static_cast<int>(mode));
+                // << " != " << static_cast<int>(currentMode));
         }
     }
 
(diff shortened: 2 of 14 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,81 @@
...
+            break;
+        case TEvIndexTabletPrivate::EForcedTabletOperationMode::FlushBytes:
+            break;
+        case TEvIndexTabletPrivate::EForcedTabletOperationMode::CollectGarbage:
+            break;
+        default:
+            TABLET_VERIFY_C(false, "unexpected forced tablet operation mode");
+    }
+
+    auto actorId = ctx.Register(actor.release());
```
neihar, line 270: in currect code it will crash, because std::unique_ptr<IActor> actor; is always null, we should not ship such kind of code even as intermediate

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,81 @@
...
+    WorkerActors.insert(actorId);
+}
+
+void TIndexTabletActor::HandleForcedTabletOperationCompleted(
+    const TEvIndexTabletPrivate::TEvForcedTabletOperationCompleted::TPtr& ev,
+    const TActorContext& ctx)
```
neihar, line 276: should we somehow convert error to answer/state?

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,81 @@
+#include "tablet_actor.h"
+
+#include <cloud/filestore/libs/diagnostics/profile_log.h>
```
neihar: do we need all includes?

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -1479,35 +1489,66 @@ void TIndexTabletState::StartForcedRangeOperation(
...
-    if (ForcedRangeOperationState
-            && ForcedRangeOperationState->OperationId == operationId)
+    auto checkId = [operationId](const TForcedOperationState& state) {
+        const auto& opId = std::visit(
+            [] (const auto& state)
+            {
+                return state.OperationId;
+            },
+            state);
+        return opId == operationId;
```
neihar: [&](const auto& s) { return s.OperationId == operationId; }

cloud/filestore/libs/storage/tablet/tablet_actor_forced_range_operation.cpp
```diff
@@ -329,6 +309,10 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         std::move(msg->Ranges),
         std::move(msg->OperationId));
 
+
```
neihar: extra line

cloud/filestore/libs/storage/tablet/tablet_actor_forced_range_operation.cpp
```diff
@@ -329,6 +309,10 @@ void TIndexTabletActor::HandleForcedRangeOperation(
...
+    const auto* state = std::get_if<TForcedRangeOperationState>(GetForcedOperationState());
```
neihar: looks like more than 80 symbols, I recommend to setup clang format

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
neihar: we can just complete early with E_NOT_IMPLEMENTED

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1188,14 +1161,17 @@ void TIndexTabletActor::HandleForcedOperation(
...
 
-    if (e.GetCode() == S_OK && IsForcedRangeOperationRunning()) {
-        const auto currentMode = GetForcedRangeOperationState()->Mode;
-        if (currentMode == mode) {
-            e = MakeError(S_ALREADY, "already launched");
-        } else {
-            e = MakeError(E_TRY_AGAIN, TStringBuilder() << "mode mismatch: "
-                << static_cast<int>(mode)
-                << " != " << static_cast<int>(currentMode));
+    if (e.GetCode() == S_OK && IsForcedOperationRunning()) {
```
neihar, line 1186: how it will work for new tablet operations? looks like we will always return E_NOT_FOUND

cloud/filestore/libs/storage/tablet/events/tablet_private.h
```diff
@@ -643,6 +638,33 @@ struct TEvIndexTabletPrivate
         }
     };
 
+    //
+    // ForcedTabletOperation
+    //
+
+    enum EForcedTabletOperationMode
```
neihar, line 645: I would in general either made only refactor without adding stub for new ops and then as separate PR only new ops. Or just one PR with all stuff. Current splitting looks strange for me.

neihar (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
tdiff: this code is unreachable in either case.

cloud/filestore/libs/storage/tablet/events/tablet_private.h
tdiff: It is hard to split the needed refactoring naturally.

Would you like to see ForcedOperationState to be converted to a std::variant with one alternative type? I don't think it would make a lot of sense.

I can spill everything including implementation of new ops into a single PR, but perhaps it would be harder to review.

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
tdiff: sorry, not sure I follow

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
tdiff: this is one of issues with this code that existed before these changed which I've noted on the commit summary page - errors are not propagated to clients.

I have a fix for that, but I deliberately put it out of scope of this pr.

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
tdiff: Ok, pushed additionally changes implementing new ops to resolve that kind of questions.

cloud/filestore/libs/storage/tablet/events/tablet_private.h
tdiff: Ok, pushed additionally changes implementing new ops to resolve that kind of questions.

Please note that tests for new functions are missing at the point.

neihar: It is okay to use std::vartiant only for 1 option, if we say that in next PR it will be more. PR will look complete and consistent in that way. If we add new function I would like to see test as well in the same PR

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
neihar: We can simplify this code

```
auto checkId = [&operationId](const TForcedOperationState& state) {
    return std::visit(
        [&](const auto& s) { return s.OperationId == operationId; },
        state);
};
```

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: I suppose this comment is outdated?

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1164,6 +1137,31 @@ void TIndexTabletActor::HandleForcedOperation(
...
+        break;
+    case NProtoPrivate::TForcedOperationRequest::E_FLUSH:
+    case NProtoPrivate::TForcedOperationRequest::E_FLUSH_BYTES:
+    case NProtoPrivate::TForcedOperationRequest::E_COLLECT_GARBAGE:
+        response = ProcessForcedTabletOperationRequest(request, ctx);
+        break;
+    default:
+        response = std::make_unique<TResponse>(
+            MakeError(E_ARGUMENT, "unsupported mode"));
+    }
```
SvartMetal, line 1156: clang-format

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1172,30 +1170,31 @@ void TIndexTabletActor::HandleForcedOperation(
...
+            const auto currentMode = state->Mode;
+            if (currentMode == mode) {
+                e = MakeError(S_ALREADY, "already launched");
+            }
         } else {
             e = MakeError(E_TRY_AGAIN, TStringBuilder() << "mode mismatch: "
-                << static_cast<int>(mode)
-                << " != " << static_cast<int>(currentMode));
+                << static_cast<int>(mode));
+                // << " != " << static_cast<int>(currentMode));
```
SvartMetal: ?

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1219,10 +1218,95 @@ void TIndexTabletActor::HandleForcedOperation(
...
+            std::get_if<TForcedTabletOperationState>(GetForcedOperationState());
+        if (state) {
+            const auto currentMode = state->Mode;
+            if (currentMode == mode) {
+                e = MakeError(S_ALREADY, "already launched");
+            }
+        } else {
+            e = MakeError(E_TRY_AGAIN, TStringBuilder() << "mode mismatch: "
+                << static_cast<int>(mode));
+                // << " != " << static_cast<int>(currentMode));
```
SvartMetal: Same question

Why do we need commented code?

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1219,10 +1218,95 @@ void TIndexTabletActor::HandleForcedOperation(
...
+            e = MakeError(E_ARGUMENT, "unsupported mode");
+        }
+    }
+
+    if (e.GetCode() == S_OK && IsForcedOperationRunning()) {
+        const auto* state =
+            std::get_if<TForcedTabletOperationState>(GetForcedOperationState());
+        if (state) {
+            const auto currentMode = state->Mode;
+            if (currentMode == mode) {
```
SvartMetal:
```
if (state->Mode == mode) {
```

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1219,10 +1218,95 @@ void TIndexTabletActor::HandleForcedOperation(
...
 
-    NCloud::Reply(ctx, *ev, std::move(response));
+std::unique_ptr<TEvIndexTablet::TEvForcedOperationResponse>
+TIndexTabletActor::ProcessForcedTabletOperationRequest(
+    const NProtoPrivate::TForcedOperationRequest& request,
+    const TActorContext& ctx)
+{
+    using EMode = TEvIndexTabletPrivate::EForcedTabletOperationMode;
+    EMode mode{};
+    NProto::TError e;
```
SvartMetal, line 1228: e -> error

We usually avoid one-letter abbreviations (there can be exceptions)

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1219,10 +1218,95 @@ void TIndexTabletActor::HandleForcedOperation(
...
+            e = MakeError(E_TRY_AGAIN, TStringBuilder() << "mode mismatch: "
+                << static_cast<int>(mode));
+                // << " != " << static_cast<int>(currentMode));
+        }
+    }
+
+    using TResponse = TEvIndexTablet::TEvForcedOperationResponse;
+    auto code = e.GetCode();
+    auto response = std::make_unique<TResponse>(std::move(e));
+    if (code == S_OK) {
```
SvartMetal, line 1260: Using `S_OK` in this context looks cryptic. You can use a lambda to construct the response from an error, and use an early return everywhere you set an error

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,297 @@
...
+                    std::move(requestInfo));
+            break;
+        case TEvIndexTabletPrivate::EForcedTabletOperationMode::CollectGarbage:
+            actor = std::make_unique<TForcedCollectGarbageActor>(
+                    ctx.SelfID,
+                    LogTag,
+                    Config->GetCompactionRetryTimeout(),
+                    std::move(requestInfo));
+            break;
+        default:
```
SvartMetal: It's better to avoid using default in a switch where it isn't needed. If we get rid of default, there would be no need to verify anything, since the compiler will check at compile time that all enum values are handled

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,297 @@
...
+void TIndexTabletActor::HandleForcedTabletOperationCompleted(
+    const TEvIndexTabletPrivate::TEvForcedTabletOperationCompleted::TPtr& ev,
+    const TActorContext& ctx)
+{
+    auto* msg = ev->Get();
+
+    TABLET_VERIFY(IsForcedOperationRunning());
+    const auto* state =
+        std::get_if<TForcedTabletOperationState>(GetForcedOperationState());
+    TABLET_VERIFY(state);
```
SvartMetal: I'd suggest avoiding verifies in non-critical situations, even when they check an invariant that's impossible by construction, to minimize the damage of one tablet affecting all the other tablets running on that node

Let's use an impossible critical event in situations like this (and guard the code below with an if to avoid crashes). This way, the impossible event will crash in debug builds and call our duty if it happens in testing or production

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -1527,48 +1499,85 @@ FILESTORE_DUPCACHE_REQUESTS(FILESTORE_DECLARE_DUPCACHE)
...
         TVector<ui32> Ranges;
         TString OperationId;
     };
 
-    TVector<TPendingForcedRangeOperation> PendingForcedRangeOperations;
-    TMaybe<TForcedRangeOperationState> ForcedRangeOperationState;
-    TVector<TForcedRangeOperationState> CompletedForcedRangeOperations;
+    struct TPendingForcedTabletOperation
+    {
+        TEvIndexTabletPrivate::EForcedTabletOperationMode Mode;
```
SvartMetal: Let's use default initialization to completely get rid of undefined behavior when we forget to initialize some POD member

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -1527,48 +1499,85 @@ FILESTORE_DUPCACHE_REQUESTS(FILESTORE_DECLARE_DUPCACHE)
...
 
-    const TForcedRangeOperationState* GetForcedRangeOperationState() const
+    const TForcedOperationState* GetForcedOperationState() const
     {
-        return ForcedRangeOperationState.Get();
+        return ForcedOperationState.Get();
     }
 
-    const TForcedRangeOperationState* FindForcedRangeOperation(
+    // TODO: FindForcedOperation does not perform lookup in pending
```
SvartMetal, line 1569: If it can lead to a bug or surprising behavior, let's at least link an issue in the TODO

SvartMetal (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: This PR was moved back to draft after previous comments (I've extended its scope).
There are still some unfinished pieces currently

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: just to be fair - this is a copy of pre-existing approach in the same file: https://github.com/ydb-platform/nbs/pull/6918/changes/BASE..75d06fee0a07ad230ffac192442ce33de3f9cfbf#diff-9f310c7c66b7a0efe972fbbefa6d2a2274707056edf48b1197a7d797e2e284d6L1176-L1178

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: done

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: same - this keeps existing naming style from 
https://github.com/ydb-platform/nbs/pull/6918/changes/BASE..75d06fee0a07ad230ffac192442ce33de3f9cfbf#diff-9f310c7c66b7a0efe972fbbefa6d2a2274707056edf48b1197a7d797e2e284d6L1142

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
tdiff: agree given this is a private event, so no forward compatibility is needed, unless we just want to stay defensive and safeguard against accidentally invalid input

cloud/filestore/libs/storage/tablet/tablet_state.h
tdiff: done

cloud/filestore/libs/storage/tablet/tablet_state.h
tdiff: added

cloud/filestore/libs/storage/tablet/events/tablet_private.h
tdiff: added new ops impl as well as some tests

tdiff:
re-opened as a new pr: https://github.com/ydb-platform/nbs/pull/7033

PR #6938, opened 2026-08-27
fix request resources deallocation race

yegorskii:
### Notes

Fix a use-after-free in the RDMA client that can occur when an endpoint
reconnects while a request is still holding buffers from the connection that's
being torn down.

**The problem.** On every reconnect `TClientEndpoint::CreateQP()` reinitializes
`SendBuffers`/`RecvBuffers`. `TBufferPool::Init()` replaces the whole internal
`Impl`, which destroys every `TChunk` of the previous generation — and
`~TChunk()` both deregisters the memory region and `free()`s the chunk memory.
A `TPooledBuffer` handed out earlier keeps a raw `TChunk*` and an address inside
that chunk, so after a reconnect both are dangling. Two places didn't know about
this:

1. `FreeRequest()` (called from `~TRequest()`, whenever the caller drops its
   `TClientRequestPtr`) unconditionally called
   `SendBuffers.ReleaseBuffer()`/`RecvBuffers.ReleaseBuffer()`. Request lifetime
   is controlled by the external caller, not by the endpoint, so a request that
   outlived a reconnect would dereference a freed `TChunk`.
2. Nothing on the send path checked which pool generation a request's buffers
   came from. If the endpoint reconnected between `AllocateRequest()` and the
   moment the request was actually processed — a window invisible to the
   endpoint's own tracking, since the request isn't enqueued anywhere yet — a
   stale request would still be accepted, and its buffers used for
   `IBV_WR_BIND_MW`/`PostSend` (`TBindWr` dereferences the dangling `TChunk*`
   via `GetMemoryRegion()`) or written to by `AbortRequest()`.

**The fix.** Added `BufferPoolGeneration`, an atomic counter bumped in
`CreateQP()`. The bump, both `Init()` calls and the `SendBuffer`/`RecvBuffer`
acquisition now happen under `AllocationLock`, so "generation + pool contents"
is a single atomic unit for every thread; `AllocateRequest()` stamps the request
with the current generation inside the same lock, together with its buffer
acquisition. Three call sites check the stamp:

- `HandleQueuedRequests()` checks it before `StartRequest()`, right after the
  request is dequeued. This is where a stale request is caught on the normal
  send path: the caller's thread can be preempted between the state check in
  `SendRequest()` and `InputRequests.Enqueue()`, so a request can land in the
  queue after `AbortRequests()` has already drained it and before the pools are
  replaced. The work request is returned to `SendQueue` and the request is
  aborted.
- `AbortRequest()` checks it before serializing the error. For a stale request
  `OutBuffer` points into a torn down pool, so instead of writing there it
  points `ResponseBuffer` at a preallocated, pre-serialized
  `E_RDMA_UNAVAILABLE` error with static storage duration. Putting the check
  here covers every abort path at once — the not-connected branch in
  `SendRequest()` and `HandleQueuedRequests()`, the reconnect drain in
  `AbortRequests()`, cancellation, and timeouts — rather than repeating it at
  each call site.
- `FreeRequest()` skips releasing buffers for a stale generation and just drops
  them: the chunk they came from no longer exists, so there is nothing to return
  and nothing to leak in the current pool.

Memory windows are only acquired in `BindBuffers()`, for requests that are
already in `ActiveRequests`, and those are drained through `AbortRequest()`
(which resets both windows) before the endpoint may reach `Disconnected` and
reconnect. A stale-generation request therefore never carries one, so the
memory-window handling in `FreeRequest()` is unreachable on that path and
needed no changes.

Also in this change: `AbortRequest()` now eagerly destroys both memory windows
(destroying a window invalidates it, which guarantees no remote write can
succeed afterwards, and lets the subsequent MR/PD deallocation succeed), and
`SerializeError()` gained a `TString`-returning overload for the preallocated
error above.

### Issue
#5456

### DIFF
--- a/cloud/storage/core/libs/rdma/iface/protobuf.cpp
+++ b/cloud/storage/core/libs/rdma/iface/protobuf.cpp
@@ -191,6 +191,19 @@ size_t SerializeError(ui32 code, TStringBuf message, TStringBuf buffer)
     return 0;   // will be interpreted as E_FAIL by ParseError
 }
 
+TString SerializeError(ui32 code, TStringBuf message)
+{
+    auto error = MakeError(code, TString(message));
+
+    TString result;
+    result.resize(error.ByteSizeLong());
+
+    bool succeeded = error.SerializeToArray(result.Detach(), result.size());
+    Y_ENSURE(succeeded, "could not serialize protobuf message");
+
+    return result;
+}
+
 NProto::TError ParseError(TStringBuf buffer)
 {
     NProto::TError error;
--- a/cloud/storage/core/libs/rdma/iface/protobuf.cpp
+++ b/cloud/storage/core/libs/rdma/iface/protobuf.cpp
@@ -198,8 +198,9 @@ TString SerializeError(ui32 code, TStringBuf message)
     TString result;
     result.resize(error.ByteSizeLong());
 
-    bool succeeded = error.SerializeToArray(result.Detach(), result.size());
-    Y_ENSURE(succeeded, "could not serialize protobuf message");
+    Y_ABORT_UNLESS(
+        error.SerializeToArray(result.Detach(), result.size()),
+        "could not serialize protobuf message");
 
     return result;
 }
--- a/cloud/storage/core/libs/rdma/impl/client.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client.cpp
@@ -164,6 +164,8 @@ struct TRequest
     NVerbs::TMemoryWindowPtr InMemoryWindow = NVerbs::NullPtr;
     NVerbs::TMemoryWindowPtr OutMemoryWindow = NVerbs::NullPtr;
 
+    ui64 BufferPoolGeneration = 0;
+
     ERequestState State = ERequestState::Init;
 
     TRequest(
@@ -597,6 +599,7 @@ private:
 
     TBufferPool SendBuffers;
     TBufferPool RecvBuffers;
+    std::atomic<ui64> BufferPoolGeneration = 0;
     TMutex AllocationLock;
 
     TPooledBuffer SendBuffer {};
@@ -800,6 +803,8 @@ void TClientEndpoint::ChangeState(
 
 void TClientEndpoint::CreateQP()
 {
+    ++BufferPoolGeneration;
+
     CompletionChannel = Verbs->CreateCompletionChannel(Connection->verbs);
     SetNonBlock(CompletionChannel->fd, true);
 
@@ -987,6 +992,7 @@ TResultOrError<TClientRequestPtr> TClientEndpoint::AllocateRequest(
         shared_from_this(),
         std::move(handler),
         std::move(context));
+    req->BufferPoolGeneration = BufferPoolGeneration.load();
 
     try {
         with_lock (AllocationLock) {
@@ -1033,6 +1039,15 @@ ui64 TClientEndpoint::SendRequest(
 
     auto clientReqId = GetNewReqId();
     req->ClientReqId = clientReqId;
+
+    if (req->BufferPoolGeneration != BufferPoolGeneration.load()) {
+        // Endpoint reconnected between AllocateRequest() and SendRequest():
+        // req->OutBuffer/OutMemoryWindow may belong to an already torn down
+        // pool generation/PD.
+        auto* handler = req->Handler.get();
+        handler->HandleResponse(std::move(req), RDMA_PROTO_FAIL, 0);
+        return clientReqId;
+    }
 
     if (!CheckState(EEndpointState::Connected)) {
         AbortRequest(std::move(req), E_RDMA_UNAVAILABLE, "endpoint is unavailable");
@@ -1957,6 +1972,19 @@ bool TClientEndpoint::FlushHanging() const
 void TClientEndpoint::FreeRequest(TRequest* req) noexcept
 {
     with_lock (AllocationLock) {
+        const bool sameBufferPoolGeneration =
+            req->BufferPoolGeneration == BufferPoolGeneration.load();
+
+        if (!sameBufferPoolGeneration) {
+            // Late request destruction after reconnect: buffers/windows
+            // belong to an older pool generation and cannot be safely
+            // released via the current pool/PD, so just drop them without
+            // touching SendBuffers/RecvBuffers or calling ibv_dealloc_mw.
+            req->InMemoryWindow.release();
+            req->OutMemoryWindow.release();
+            return;
+        }
+
         // destroy memory windows that haven't been properly invalidated
         if (req->InMemoryWindow) {
             req->InMemoryWindow.reset();
--- a/cloud/storage/core/libs/rdma/impl/client.cpp
+++ b/cloud/storage/core/libs/rdma/impl/client.cpp
@@ -867,13 +867,13 @@ void TClientEndpoint::CreateQP()
         ++BufferPoolGeneration;
         SendBuffers.Init(Verbs, Connection->pd, sendFlags);
         RecvBuffers.Init(Verbs, Connection->pd, recvFlags);
+
+        SendBuffer = SendBuffers.AcquireBuffer(
+            Config.SendQueueSize * sizeof(TRequestMessage), true);
+
+        RecvBuffer = RecvBuffers.AcquireBuffer(
+            Config.RecvQueueSize * sizeof(TResponseMessage), true);
     }
-
-    SendBuffer = SendBuffers.AcquireBuffer(
-        Config.SendQueueSize * sizeof(TRequestMessage), true);
-
-    RecvBuffer = RecvBuffers.AcquireBuffer(
-        Config.RecvQueueSize * sizeof(TResponseMessage), true);
 
     SendWrs.resize(Config.SendQueueSize);
     RecvWrs.resize(Config.RecvQueueSize);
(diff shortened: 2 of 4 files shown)


cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1030,6 +1040,15 @@ ui64 TClientEndpoint::SendRequest(
     auto clientReqId = GetNewReqId();
     req->ClientReqId = clientReqId;
 
+    if (req->BufferPoolGeneration != BufferPoolGeneration.load()) {
+        // Endpoint reconnected between AllocateRequest() and SendRequest():
+        // req->OutBuffer/OutMemoryWindow may belong to an already torn down
+        // pool generation/PD.
+        auto* handler = req->Handler.get();
+        handler->HandleResponse(std::move(req), RDMA_PROTO_FAIL, 0);
```
tpashkin: Did we check that all existing clients correctly handle `responseBytes=0` case? E.g. not trying to parse the response?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1953,6 +1972,19 @@ bool TClientEndpoint::FlushHanging() const
...
     with_lock (AllocationLock) {
+        const bool sameBufferPoolGeneration =
+            req->BufferPoolGeneration == BufferPoolGeneration.load();
+
+        if (!sameBufferPoolGeneration) {
+            // Late request destruction after reconnect: buffers/windows
+            // belong to an older pool generation and cannot be safely
+            // released via the current pool/PD, so just drop them without
+            // touching SendBuffers/RecvBuffers or calling ibv_dealloc_mw.
+            req->InMemoryWindow.release();
```
tpashkin: Wait, are we just leaking them? This feels wrong.

I'm not 100% sure that PD is ref-counted, but did we try to actually deallocate them? Even if calling `ibv_dealloc_mw` is unsafe, we need to at least free them

tpashkin: From the [man](https://man7.org/linux/man-pages/man3/ibv_alloc_mw.3.html)
```
ibv_dereg_mr() fails if any memory window is still bound to this MR
```

tpashkin: And also

```
ibv_dealloc_pd() may fail if any other resource is still associated with the PD being freed
```

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1953,6 +2004,17 @@ bool TClientEndpoint::FlushHanging() const
 void TClientEndpoint::FreeRequest(TRequest* req) noexcept
 {
     with_lock (AllocationLock) {
+        const bool sameBufferPoolGeneration =
+            req->BufferPoolGeneration == BufferPoolGeneration.load();
+
+        if (!sameBufferPoolGeneration) {
```
tpashkin: why bother saving this to a variable that's gonna be used just 1 time?

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1261,8 +1310,10 @@ void TClientEndpoint::AbortRequest(
...
 {
-    // destroying memory window automatically invalidates it. this ensures no
-    // remote write can succeed after this point
+    if (req->InMemoryWindow) {
+        req->InMemoryWindow.reset();
+        Counters->ReleaseMemoryWindow();
+    }
     if (req->OutMemoryWindow) {
         req->OutMemoryWindow.reset();
         Counters->ReleaseMemoryWindow();
```
tpashkin, line 1268: I think this comment is important and we shouldn't remove it

In fact I would add another one clarifying why we are deallocating `InMemoryWindow` specifically – "bound memory windows will prevent MR/PD deallocation, so make sure to free everything up before ReleaseResources"

tpashkin: As a side note – I just noticed, that MRs get deallocated only after we've already reconnected, which is not ideal, since if PDs before/after reconnect were genuinely different, it would in fact fail – I will fix that

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1953,6 +2004,17 @@ bool TClientEndpoint::FlushHanging() const
 void TClientEndpoint::FreeRequest(TRequest* req) noexcept
 {
     with_lock (AllocationLock) {
+        const bool sameBufferPoolGeneration =
+            req->BufferPoolGeneration == BufferPoolGeneration.load();
```
tpashkin: Out of curiosity – why call `load` explicitly?

yegorskii: "explicitly", just to indicate that this is not an generic variable, but with "additional" semantics

cloud/storage/core/libs/rdma/impl/client.cpp
yegorskii: I was afraid the 'if' will not fit 80 symbols limitation, plus I think compiler will be able to optimize this const

tpashkin: > I was afraid the 'if' will not fit 80 symbols limitation

Well, declaration doesn't fit either :shrug:

> I think compiler will be able to optimize this const

Agree, but I'm speaking entirely from semantics standpoint

tpashkin (dismissed):

tpashkin (dismissed):

cloud/storage/core/libs/rdma/iface/protobuf.cpp
```diff
@@ -191,6 +191,19 @@ size_t SerializeError(ui32 code, TStringBuf message, TStringBuf buffer)
...
 
+TString SerializeError(ui32 code, TStringBuf message)
+{
+    auto error = MakeError(code, TString(message));
+
+    TString result;
+    result.resize(error.ByteSizeLong());
+
+    bool succeeded = error.SerializeToArray(result.Detach(), result.size());
+    Y_ENSURE(succeeded, "could not serialize protobuf message");
```
komarevtsev-d: `Y_ENSURE` кидает экспешн. Лучше же тогда `Y_ABORT_UNLESS`?

cloud/storage/core/libs/rdma/iface/protobuf.cpp
```diff
@@ -191,6 +191,19 @@ size_t SerializeError(ui32 code, TStringBuf message, TStringBuf buffer)
...
 }
 
+TString SerializeError(ui32 code, TStringBuf message)
+{
+    auto error = MakeError(code, TString(message));
+
+    TString result;
+    result.resize(error.ByteSizeLong());
+
+    bool succeeded = error.SerializeToArray(result.Detach(), result.size());
```
komarevtsev-d: nit: const

cloud/storage/core/libs/rdma/iface/protobuf.cpp
yegorskii: Если здесь сломалось, то сломалось на старте и вообще выглядит как поломка protobuf.

cloud/storage/core/libs/rdma/impl/client.cpp
```diff
@@ -1030,6 +1049,12 @@ ui64 TClientEndpoint::SendRequest(
     auto clientReqId = GetNewReqId();
     req->ClientReqId = clientReqId;
 
+    if (req->BufferPoolGeneration != BufferPoolGeneration.load()) {
+        // Endpoint reconnected between AllocateRequest() and SendRequest()
+        RejectStaleGenerationRequest(std::move(req));
```
komarevtsev-d: В `TClientEndpoint::AbortRequest` что ниже, также
1) счетчики щелкаются 
2) OutMemoryWindow очищается

Это не надо сделать?

cloud/storage/core/libs/rdma/iface/protobuf.cpp
komarevtsev-d: Ну тем более причина в сторону `Y_ABORT_UNLESS` же?

cloud/storage/core/libs/rdma/impl/client.cpp
yegorskii: Здесь еще окна не выделен. Они выделяются в StartRequest

cloud/storage/core/libs/rdma/iface/protobuf.cpp
yegorskii: Я не очень понимаю в чем здесь разница между Y_ABORT_UNLESS и Y_ENSURE

cloud/storage/core/libs/rdma/impl/client.cpp
komarevtsev-d: Здесь нет, а в `TClientEndpoint::HandleQueuedRequests` уже могут быть выделены

cloud/storage/core/libs/rdma/iface/protobuf.cpp
komarevtsev-d: 1) `Y_ABORT_UNLESS` даст тебе понятное сообщение об ошибке
2) зачем кидать исключение, если ты точно знаешь что его никто не ловит?

cloud/storage/core/libs/rdma/impl/client.cpp
yegorskii: нет не могут, именно  там вызывается StartRequest

yegorskii: SendRequest мы просто кладем запрос в очередь, в HandleQueuedRequests мы разгребаем очередь и зовем StartRequest где происходит выделение окон. Но проверка стоит раньше

tpashkin (approved):

komarevtsev-d (approved):

PR #6952, opened 2026-08-28
issue-5895: fastshard - storage group retry policy

qkrorlqr:
### Notes
Storage group should retry storage device requests by itself. Even for the naive storage group implementation it's required because a temporary storage node outage will lead to data inconsistency among the devices in the group. In the real implementation it will be needed as well because even though it won't lead to data inconsistency, still, error propagation to the layer above will lead to the cancellation of the whole operation and then error propagation up to the client which will introduce too much overhead and extra latency.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/fastshard/impl/naive_mirrored/shard.cpp
+++ b/cloud/filestore/libs/storage/fastshard/impl/naive_mirrored/shard.cpp
@@ -2575,7 +2575,20 @@ struct TNaiveMirroredStorageGroupFactory: IStorageGroupFactory
             });
         }
 
-        return CreateNaiveMirroredStorageGroup(std::move(devices));
+        TStorageGroupRetryPolicy retryPolicy;
+        if (config.GetRetryTotalTimeoutMs()) {
+            retryPolicy.TotalTimeout =
+                TDuration::MilliSeconds(config.GetRetryTotalTimeoutMs());
+        }
+        if (config.GetRetryBackoffIncrementMs()) {
+            retryPolicy.BackoffIncrement =
+                TDuration::MilliSeconds(config.GetRetryBackoffIncrementMs());
+        }
+
+        return CreateNaiveMirroredStorageGroup(
+            std::move(devices),
+            retryPolicy,
+            CreateFiberTimer());
     }
 };
 
--- a/cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group.cpp
+++ b/cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group.cpp
@@ -14,18 +14,84 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+class TFiberTimer final: public ITimer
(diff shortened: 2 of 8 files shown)


qkrorlqr:
cloud/filestore/tests/fmdtest/qemu-kikimr-naive-mirrored-test/test.py.test_create_unlink_steal 	1m 13.250s 	FAIL 	[DIR](https://github-actions-s3.storage.eu-north2.nebius.cloud/ydb-platform/nbs/PR-check/33206441919/1/nebius-x86-64/test_data/filestore/1/actions-runner/_work/nbs/nbs/cloud/filestore/tests/fmdtest/qemu-kikimr-naive-mirrored-test/test-results/py3test/testing_out_stuff/index.html) | [LOG](https://github-actions-s3.storage.eu-north2.nebius.cloud/ydb-platform/nbs/PR-check/33206441919/1/nebius-x86-64/logs/filestore/1/cloud/filestore/tests/fmdtest/qemu-kikimr-naive-mirrored-test/test-results/py3test/testing_out_stuff/test.py.test_create_unlink_steal.log)

fixed in https://github.com/ydb-platform/nbs/pull/6929

qkrorlqr:
claude:

> Reviewed the diff (8 files: retry loop in storage_group.cpp, TFiberTimer, config plumbing, fake node response queues, 5 new tests). The overall shape is fine and the tests are solid. Three things I'd consider significant:
> 
> 1. Read retries back off before failing over to a healthy replica. ReadPages rotates the round-robin selector on each attempt, but CallWithRetries sleeps BackoffIncrement (500 ms) before the next attempt. So a single dead replica adds 500 ms to every read that lands on it, even though two healthy replicas are sitting right there. Given the PR's stated goal is avoiding extra latency, reads should try the other replicas immediately and only start backing off after a full pass over all devices fails.

Will be improved separately - probably within a real storage group implementation. This PR is only about a very basic retry and backoff policy.

> 
> 2. No cancellation / shutdown path. A permanently failing device turns every WriteLogRecord/AcquireDevices/ReleaseDevices into a 5‑minute stall (the whole write waits for all fibers), and there is no stop flag or cancellation token the group owner can set. Anything that tears down the shard while requests are in flight — shutdown, group reconfiguration, device replacement — can now block for up to TotalTimeout. Worth at least a stop signal checked in the loop, or making the deadline derive from a per-request deadline rather than a fixed policy.
> 

Yes, this will be solved by a real storage group implementation.

> 3. Write retry after E_TIMEOUT assumes the storage node write is idempotent. If the first WriteLogRecord actually landed but the response was lost, the retry resends the same LSN. That's correct only if the node treats a duplicate LSN as success (or a retriable error). If it returns something non-retriable, a successful write is reported as failed. I couldn't verify the node's behaviour from this diff — please confirm it, and ideally add a test for the duplicate-LSN case.

Each storage node write is supposed to be idempotent.

yegorskii (approved):

arigachnyy (approved):

PR #6953, opened 2026-08-28
issue-6608: Implement quota usage enforcement

debnatkh:
### Notes
Final step necessary for the basic quota scenarios to work – the enforcement itself

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1896,14 +1896,13 @@ void TIndexTabletActor::UpdateLogTag()
     }
 }
 
-bool TIndexTabletActor::HasBlocksLeft(
+bool TIndexTabletActor::HasSpaceLeft(
     ui64 prevSize,
     ui64 newSize,
     ui32 quotaId) const
 {
-    const ui64 blocksRequired = static_cast<ui64>(Max<i64>(
-        0,
-        GetBlocksDifference(prevSize, newSize, GetBlockSize())));
+    const ui64 blocksRequired = static_cast<ui64>(
+        Max<i64>(0, GetBlocksDifference(prevSize, newSize, GetBlockSize())));
 
     if (!GetFileSystem().GetStrictFileSystemSizeEnforcementEnabled()) {
         if (GetUsedBlocksCount() + blocksRequired > GetBlocksCount()) {
@@ -1951,14 +1950,6 @@ bool TIndexTabletActor::HasBlocksLeft(
     }
 
     return true;
-}
-
-bool TIndexTabletActor::HasSpaceLeft(
-    ui64 prevSize,
-    ui64 newSize,
-    ui32 quotaId) const
-{
-    return HasBlocksLeft(prevSize, newSize, quotaId);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1896,8 +1896,15 @@ void TIndexTabletActor::UpdateLogTag()
     }
 }
 
-bool TIndexTabletActor::HasBlocksLeft(ui64 blocksRequired) const
+bool TIndexTabletActor::HasBlocksLeft(
+    ui64 prevSize,
+    ui64 newSize,
+    ui32 quotaId) const
 {
+    const ui64 blocksRequired = static_cast<ui64>(Max<i64>(
+        0,
+        GetBlocksDifference(prevSize, newSize, GetBlockSize())));
+
     if (!GetFileSystem().GetStrictFileSystemSizeEnforcementEnabled()) {
         if (GetUsedBlocksCount() + blocksRequired > GetBlocksCount()) {
             return false;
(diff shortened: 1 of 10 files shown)


debnatkh:
Good — the `.diff` link gives the full patch, including `tablet_actor_createnode.cpp`, which GitHub's HTML view had cut off. That file has the most interesting change. Updated review:

## Summary
Adds per-directory quota enforcement (bytes + node count), keyed by `quotaId` inherited from a node's parent directory. Enforcement point is centralized in three tablet-level helpers (`HasBlocksLeft`, `HasSpaceLeft`, `HasNodesLeft`), now threaded with a `quotaId` param and called from every mutation path that can grow file size or create nodes: `AddData`, `AllocateData`, `CreateHandle`, `CreateNode`, `SetNodeAttr`, `WriteData`.

## The important part: `tablet_actor_createnode.cpp`

```cpp
-    if (!HasNodesLeft() && !behaveAsShard) {
-        args.Error = ErrorNoSpaceLeft();
-        return true;
-    }
-
     // validate parent node exists
     ...
     if (!behaveAsShard) {
         const ui32 quotaId = args.ParentNode->Attrs.GetQuotaId();
         args.Attrs.SetQuotaId(quotaId);
         args.Request.SetQuotaId(quotaId);
+
+        if (!HasNodesLeft(quotaId)) { ... }
+        if (!HasSpaceLeft(0, args.Attrs.GetSize(), quotaId)) { ... }
     }
```

This is the real fix, not just a signature change: previously `HasNodesLeft()` was called with no quota context (the function took no args at all before this PR), so node-count quota could never have been enforced on `CreateNode` — only on `CreateHandle` after the fact. Moving the check after the parent is read lets it use the *inherited* `quotaId` before the child is created, and the new `HasSpaceLeft(0, args.Attrs.GetSize(), quotaId)` call catches quota-byte violations from a node's *initial* size (relevant for symlinks/special files created with nonzero size, since regular files start at 0).

Things worth double-checking here:
- **Order-of-validation change**: the inode-limit check used to run before the parent-existence read; now it runs after. If `ReadNode` on the parent returns `false` ("not ready"), that's a retry path, so should be fine, but it's worth confirming there's no test relying on `ErrorNoSpaceLeft` being returned before other validation errors for a missing/invalid parent.
- **`HasSpaceLeft(0, size, quotaId)` semantics**: passing `prevSize=0` means this always charges the *full* initial size against the quota, which is correct for a brand new node, but confirm no call site could invoke `PrepareTx_CreateNode` for something that isn't actually brand-new (e.g. some sharded resend path) where charging the full size again would double count.

## Remaining points from before, still applicable

1. **Off-by-one style inconsistency** in `HasNodesLeft`: the global check uses `usedNodes >= GetNodesCount()`, the quota check uses `usedNodes + 1 > quota->GetMaxNodes()`. Same relation, different phrasing in the same function — harmless but worth aligning for readability.
2. **`Max(local usage, cached aggregate usage)` fallback**: this avoids undercounting from stale caches, but means quota checks can be more conservative than reality until `CachedAggregateStats` refreshes (the test needs `ForceRefreshAggregateQuotaUsage` to get deterministic assertions). Worth a short comment in the production code, not just the test, explaining this staleness window is expected.
3. **Fail-open on dangling `quotaId`**: `if (quota && quota->GetMaxBytes())` — if `FindQuota(quotaId)` returns null (quota deleted but still referenced by a node), enforcement silently no-ops rather than erroring or logging. Confirm that's intentional.
4. **`0` as "unlimited" convention**: correct per `ShouldNotEnforceUnlimitedQuota`, but confirm this is documented on the proto field itself, not just implied by test behavior.
5. **`HasNodesLeft`/`HasSpaceLeft` refactor in the strict-enforcement branch**: the `if/else` restructuring in `HasNodesLeft` is a genuine control-flow rewrite of an existing safety check (running out of inodes), worth a manual side-by-side trace against the pre-PR version rather than trusting the diff coloring, since it's easy to invert a branch here silently.

## Test coverage
The three added tests (`ShouldRejectCreateNodeWhenQuotaNodesExceeded`, `ShouldRejectGrowingFileBeyondQuotaBytes`, `ShouldNotEnforceUnlimitedQuota`) cover the `CreateNode`/`SetNodeAttr` paths well, including the boundary case and shrink-then-regrow. Given the `CreateNode` fix above is the most behaviorally significant part of this PR, I'd want a dedicated test asserting that **creating a node directly exceeds `MaxBytes` via its initial size** (i.e., exercising the new `HasSpaceLeft(0, ...)` call specifically, not just growth via `SetNodeAttr`/`WriteData`).

Still a draft with an empty description/no linked issue text — worth filling in before requesting review, especially to state the fail-open-on-dangling-quota and stale-aggregate-cache tradeoffs as intentional design decisions rather than leaving reviewers to infer them from the diff.

---

Both things noted as double-checking are invalid: first change is okay, since I moved validation within the same PrepareTx part. The second is invalid since all this change does is adding quotaId to all caller places

1. in this particular place it is consistent: for bytes we chack that current + delta <= limit, for nodes we chach that current + 1 <= limit
2. This is same as for regular limit enforcement. Some staleness is a necessary evil of the sharded approach
3. Intentional by design, detachment of quotas should be seamless
4. added comment
5. This change is valid, souble-checked it

---

Good, that all makes sense — thanks for the corrections. Updated take:

- **Order-of-validation / `HasSpaceLeft(0,...)` double-count concern**: retracted, since it's still within the same `PrepareTx_CreateNode` and not a repeatable/resendable call site — not a real risk.
- **1**: agreed, `current + delta <= limit` vs `current + 1 <= limit` is exactly consistent once you read it as "delta for nodes is always 1." Not actually inconsistent, just a phrasing difference I misread as a style mismatch.
- **2**: makes sense — staleness under the sharded aggregate model is an accepted tradeoff already present in the existing (non-quota) enforcement, so quotas inheriting the same property isn't a new risk to flag.
- **3**: good to know it's deliberate (fail-open on a dangling/deleted quota so removing a quota doesn't strand existing nodes in a broken state) — that's a reasonable design choice given the goal of "seamless detachment."
- **4**: 👍
- **5**: good, that's the one part of this diff I'd have wanted eyes-on regardless, glad it's already been traced.

So the only substantive item left standing from my read is really just the test-coverage suggestion: a test that exercises quota-bytes enforcement via `CreateNode`'s new `HasSpaceLeft(0, args.Attrs.GetSize(), quotaId)` path specifically (e.g. a symlink or other node type created with nonzero initial size under a tight `MaxBytes` quota), since the current three tests hit the bytes limit only through `SetNodeAttr`/growth, not through creation itself. Everything else here looks solid — nice fix in `createnode.cpp` moving the inode check to use the inherited quotaId instead of the old quota-blind global check.

cloud/filestore/libs/storage/tablet/tablet_actor_createnode.cpp
```diff
@@ -652,8 +645,21 @@ bool TIndexTabletActor::PrepareTx_CreateNode(
...
+        const ui32 quotaId = args.ParentNode->Attrs.GetQuotaId();
+        args.Attrs.SetQuotaId(quotaId);
+        args.Request.SetQuotaId(quotaId);
+
+        if (!HasNodesLeft(quotaId)) {
+            args.Error = ErrorNoSpaceLeft();
+            return true;
+        }
+
+        if (!HasSpaceLeft(0, args.Attrs.GetSize(), quotaId)) {
```
ruslan-gar: It looks like we don't need to check the available space when a node is created in the shard that manages the directory. This check will be performed in a shard where the actual file is created.

ruslan-gar (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1922,43 +1929,91 @@ bool TIndexTabletActor::HasBlocksLeft(ui64 blocksRequired) const
...
+bool TIndexTabletActor::HasSpaceLeft(
+    ui64 prevSize,
+    ui64 newSize,
+    ui32 quotaId) const
 {
-    return HasBlocksLeft(
-        static_cast<ui64>(Max<i64>(
-            0,
-            GetBlocksDifference(prevSize, newSize, GetBlockSize()))));
+    return HasBlocksLeft(prevSize, newSize, quotaId);
```
ruslan-gar: I would totally eliminate HasBlocksLeft

ruslan-gar (approved):

agalibin (approved):

PR #6961, opened 2026-08-29
issue-6608: add limits for the number of quotas

debnatkh:
### Notes
Small TODO completed – limiting the total number of quotas in the filesystem

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/service/filestore.h
+++ b/cloud/filestore/libs/service/filestore.h
@@ -38,6 +38,7 @@ constexpr ui32 MaxSymlink = NProto::E_FS_LIMITS_SYMLINK;
 constexpr ui64 MaxNodes = static_cast<ui32>(NProto::E_FS_LIMITS_INODES);
 constexpr ui64 MaxXAttrName = NProto::E_FS_LIMITS_XATTR_NAME;
 constexpr ui64 MaxXAttrValue = NProto::E_FS_LIMITS_XATTR_VALUE;
+constexpr ui32 MaxQuotas = NProto::E_FS_LIMITS_QUOTAS;
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/cloud/filestore/libs/storage/tablet/quota.cpp
+++ b/cloud/filestore/libs/storage/tablet/quota.cpp
@@ -54,6 +54,11 @@ TVector<NProto::TQuota> TQuotaStore::GetQuotas() const
     return quotas;
 }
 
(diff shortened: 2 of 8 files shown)


debnatkh:
I pulled the diff (4 files, +50/−1). Here's the change:

**What it does**
- Adds `E_FS_LIMITS_QUOTAS = 1000` to the `EFilestoreLimits` proto enum, and exposes it as `constexpr ui32 MaxQuotas` in `filestore.h`.
- In `TIndexTabletActor::HandleSetQuota`, rejects a `SetQuota` call with `E_ARGUMENT` if the quota ID is new (`!FindQuota(...)`) and the filesystem already has `MaxQuotas` quotas.
- Adds a unit test `ShouldRejectSetQuotaPastMaxQuotasLimit` covering: filling up to the limit, rejecting a new ID past the limit, allowing an *update* to an existing ID while at the limit, and freeing a slot via delete so a new ID can be added again.

**Overall**: small, focused, well-tested fix for the TODO in issue #6608. Logic is correct — the `!FindQuota(...) &&` guard properly distinguishes "new quota" from "update existing quota," so updates aren't blocked at the limit, and the test explicitly asserts that distinction plus the delete-then-add-again path. That's the trickiest part of this kind of limit check and it's covered.

**Things worth double-checking before merge:**

1. **Why 1000, and is it configurable?** The limit is a hardcoded proto constant with no comment justifying the number and no way to tune it per-cluster (unlike some other limits that go through storage config). If quota count needs to scale differently across deployments, this might be worth a config knob instead of/alongside the constant.

2. **Backward compatibility**: if any existing filesystem already has more than 1000 quotas (unlikely but worth confirming), this only blocks *new* quota creation, so it's non-breaking — updates and deletes still work. Good, but worth a sentence in the PR description confirming this was considered.

3. **`GetQuotas().size()` cost**: worth a quick check of `GetQuotas()`'s implementation — if it returns a copy/materializes a container rather than a reference, this adds overhead to every `SetQuota` call, not just ones near the limit.

4. **PR description is thin** ("Small TODO completed") — for a limits change like this, it'd help reviewers/future readers to state the reasoning behind the value 1000 and confirm there's no other code path that creates quotas bypassing `HandleSetQuota`.

Nothing blocking — the core logic and test are sound. The main gap is documentation/justification for the chosen limit rather than any functional issue.

---

3: fixed

neihar (approved):

agalibin (approved):

Issue #6956, opened 2026-08-29
[Filestore] fastshard - journalling layer


qkrorlqr:
Main issue: https://github.com/ydb-platform/nbs/issues/5895

Implement the journalling layer described here: https://github.com/ydb-platform/nbs/blob/main/doc/filestore/design/storage/fastshard/storage-node.md

Currently missing in the docs:
* prev-lsn should be passed to the storage device alongside lsn
* there should be a method for reading the tail of the journal (needed for storage group recovery)
* there should be a method for advancing lsn-low-watermark

PR #6970, opened 2026-08-31
Add PrevLogSequenceNumber to TWriteLogRecordRequest

arigachnyy:
### Notes
Add `PrevLogSequenceNumber` to the request and makes the disk agent enforce the chain.
The stored lsn is advanced only after the writes have actually succeeded, so a request rejected for an invalid order.

### Issue

### DIFF
--- a/cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_journalled_device_ut.cpp
+++ b/cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_journalled_device_ut.cpp
@@ -261,12 +261,14 @@ Y_UNIT_TEST_SUITE(TDiskAgentJournalledDeviceTest)
             {[&](auto& proto)
              {
                  proto.SetDeviceUUID(uuid);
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/storage/disk_agent/disk_agent_actor_journalled_device_ut.cpp
```diff
@@ -704,6 +749,129 @@ Y_UNIT_TEST_SUITE(TDiskAgentJournalledDeviceTest)
         }
     }
 
+    Y_UNIT_TEST_F(ShouldValidateLogSequenceNumber, TFixture)
```
qkrorlqr, line 752: it's better to extract the core part of the request processing logic into some component which is decoupled from the transport - then we'll be able to write unit tests that don't require bootstrapping of the server and the client

qkrorlqr (approved):

WilyTiger (approved):

PR #6975, opened 2026-08-31
[NBS] Fix stalled requests during tablet restart

### LINKED ISSUE #6936: [NBS] Stalled requests during tablet restart
Duplicate write requests may hang after a volume tablet restart.

With overlapping request handling enabled, duplicate writes wait for the original request in `DuplicateWriteAndZeroRequests`.
During tablet shutdown, `TVolumeActor::CancelRequests()` cancels the original requests in `VolumeRequests` but leaves their queued duplicates without a response. The durable client therefore does not receive `E_REJECTED` and cannot retry these writes, leaving them stuck even after the tablet recovers.
Expected behavior: When an original request is cancelled during tablet shutdown, all its queued duplicates should also receive `E_REJECTED`.

Tekha03:
## Summary

- Add regression coverage for duplicate `WriteBlocksLocal` requests during user and system volume actor shutdown.
- Reply with `E_REJECTED` to queued duplicate requests before cancelling the original request.
- Keep the regression tests and implementation fix in separate commits.

Fixes #6936

## Tests

`./ya make -t -j8 cloud/blockstore/libs/storage/volume/ut --fail-fast`

### DIFF
--- a/cloud/blockstore/libs/storage/volume/volume_ut.cpp
+++ b/cloud/blockstore/libs/storage/volume/volume_ut.cpp
@@ -10419,7 +10419,8 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
         DoShouldRejectRequestsWhenVolumeIsKilled(true, true);
     }
 
-    void DoShouldRejectDuplicateWriteBlocksLocalWhenVolumeIsKilled(bool rebootSysTablet) {
+    void DoShouldRejectDuplicateWriteBlocksLocalWhenVolumeIsKilled(bool rebootSysTablet)
+    {
         NProto::TStorageServiceConfig config;
         config.SetOverlappingRequestsPolicy(NProto::EOverlappingRequestsPolicy::ORP_ENABLE);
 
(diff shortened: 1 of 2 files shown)


komarevtsev-d (dismissed):

cloud/blockstore/libs/storage/volume/volume_ut.cpp
```diff
@@ -10419,6 +10419,93 @@ Y_UNIT_TEST_SUITE(TVolumeTest)
         DoShouldRejectRequestsWhenVolumeIsKilled(true, true);
     }
 
+    void DoShouldRejectDuplicateWriteBlocksLocalWhenVolumeIsKilled(bool rebootSysTablet) {
```
qkrorlqr: clang-format this section please

qkrorlqr (changes requested):

cloud/blockstore/libs/storage/volume/volume_ut.cpp
Tekha03: moved the brace to a new line

drbasic (approved):

qkrorlqr (approved):

PR #6976, opened 2026-08-31
[Filestore] read/write commands by node ID (not full path) in the filestore-client binary

debnatkh:
### Notes
Right now filestore-client needs path in order to be able to read/write

### Issue
Small change, thus no issue


debnatkh:
## PR #6976 Review: `[Filestore] read/write commands by node ID`

**Summary:** Adds a `--node <ID>` option to `filestore-client read` and `write`, as an alternative to `--path`, so callers can operate directly on a node ID instead of resolving a full path. `--path` and `--node` are made mutually exclusive.

### What looks good
- Clean, minimal diff (+155/−41) focused on a small, well-scoped feature.
- `Opts.MutuallyExclusive(PathOptionName, NodeOptionName)` correctly prevents both flags being passed together.
- Good use of `Y_ENSURE(Mode != EModeType::Unknown, ...)` to give a clear error message if neither flag is given, rather than silently defaulting.
- Test coverage: `test_read_write_by_node_id` creates a file by path, writes/reads it by node ID, and cross-checks against reading the same file by path — a solid round-trip test, plus a canonical-data update.
- `client.py` test helper changes (`write`/`read` now take `path=None, node=None`) are backward compatible with existing call sites that only pass `path`.

### Things worth double-checking

1. **`Y_ABORT("unreachable")` for `EModeType::Unknown` in `Execute()`**
   Since `Y_ENSURE(Mode != EModeType::Unknown, ...)` already runs earlier in the same function, this `case EModeType::Unknown: Y_ABORT(...)` is dead code — fine as a defensive switch-exhaustiveness guard, but worth a one-line comment noting that, so a future reader doesn't wonder why it's unreachable.

2. **No test for the mutually-exclusive/neither-flag error paths**
   The new test only covers the happy path (valid node ID). It'd strengthen coverage to also assert that passing neither `--path` nor `--node`, or both together, fails with the expected error — cheap to add given the harness already exists.

3. **`write.cpp`: node-mode doesn't validate the node exists before `CreateHandle`**
   In `Path` mode, the code explicitly checks `parent.Node.GetType() != NProto::E_INVALID_NODE` before proceeding. In `Node` mode, there's no equivalent check — it just calls `createRequest->SetNodeId(NodeId)` and lets `CreateHandle` fail server-side if the ID is bogus. That's probably fine (server will return an error), but it means the error message for "invalid node ID" will be less friendly/specific than the path-mode error. Minor, likely acceptable.

4. **`read.cpp`: directory-node check now uses `node.GetType()` instead of `resolved.back().Node.GetType()`** — this is correct after the refactor (since `node` unifies both resolution paths), just worth a careful look to confirm `GetNodeAttr`'s returned `NProto::TNodeAttr` populates `Size`/`Type` identically to what `ResolvePath` returns, since `Length` calculation and the directory check both depend on it.

5. **Test data file naming** — `results.txt` contains `0123456789` twice (once from node-ID read, once from path read), which is a nice cross-check, but the test itself doesn't assert equality in code — it relies on canonical-data diffing to catch any mismatch. That's consistent with the existing test style in this file, so not a blocker.

### Verdict
Small, well-tested, low-risk change. No blocking issues — the two suggestions (dead-code comment, negative-path test) are polish rather than correctness concerns.

---

1. Same pattern as in other client places
2. This test, IMO, will be excessive
3. there is O_CREAT
4. I got rid of resolve, yep
5. this is ya make format

---

Fair enough on all counts — none of those are real issues, they're just me not having the full context of the codebase's conventions. Sounds like this PR is in good shape to merge.

antonmyagkov (approved):

agalibin (approved):

PR #6981, opened 2026-08-31
issue-1751: [Filestore] Support multiple incomplete allocations by TFileRingBuffer

e673:
### Notes
The PR adds support for multiple incomplete allocations by `TFileRingBuffer`.
This enables its clients not to block on each other when adding data from multiple threads.

Implementation details:
- `Alloc` adds an entry with `FreeFlag` set in the entry header;
- `Commit` makes the entry visible by resetting `FreeFlag`.

### Issue

### DIFF
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
@@ -50,9 +50,6 @@ struct IPersistentStorage
      *
      * On failure, returns nullptr if the buffer is full or an error if
      * allocation is not possible due to corruption or invalid argument.
-     *
-     * Note: only one allocation is possible at a time. Repeated Alloc will
-     * return an error.
      */
     [[nodiscard]] virtual TResultOrError<char*> Alloc(size_t size) = 0;
 
@@ -72,22 +69,8 @@ struct IPersistentStorage
     [[nodiscard]] virtual NProto::TError Commit(const void* ptr) = 0;
 
     /**
-     * Completes the previously made allocation using checksum provided by
-     * caller and making the allocation visible.
-     *
-     * Since IPersistentStorage is non-thread safe, it requires external
-     * synchronization and executing its methods under a lock. Calculating
-     * checksums inside Commit() consumes some time and it will prevent other
-     * threads from using IPersistentStorage. The purpose of this method is to
-     * make multi-threaded work with IPersistentStorage more efficient by
-     * reducing the time spent inside the lock.
-     *
-     * Once committed, it is not allowed to modify the contents of the allocated
-     * entry. If there is a need to augment the allocation with additional data,
-     * GetTag/SetTag can be used.
-     *
-     * An error is returned if there is no incomplete allocation correponding to
-     * the provided pointer or the buffer is corrupted.
+     * Commits previously allocated memory buffer but takes a checksum provided
+     * by the caller instead of calculating it.
      *
      * Note: the checksum is not validated, the calling code has responsibility
      * to provide the correct Crc32c checksum. Passing an incorrect checksum may
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
@@ -64,8 +64,8 @@ struct IPersistentStorage
      * Memory that was allocated but not committed will be lost at buffer
      * recreation.
      *
-     * An error is returned if there is no incomplete allocation correponding to
-     * the provided pointer or the buffer is corrupted.
+     * An error is returned if there is no incomplete allocation corresponding
+     * to the provided pointer or the buffer is corrupted.
      */
     [[nodiscard]] virtual NProto::TError Commit(const void* ptr) = 0;
 
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
@@ -69,7 +69,33 @@ struct IPersistentStorage
      * An error is returned if there is no incomplete allocation or the buffer
      * is corrupted.
      */
-    [[nodiscard]] virtual NProto::TError Commit() = 0;
+    [[nodiscard]] virtual NProto::TError Commit(const void* ptr) = 0;
+
+    /**
+     * Completes the previously made allocation using checksum provided by
+     * caller and making the allocation visible.
+     *
+     * Since IPersistentStorage is non-thread safe, it requires external
+     * synchronization and executing its methods under a lock. Calculating
+     * checksums inside Commit() consumes some time and it will prevent other
+     * threads from using IPersistentStorage. The purpose of this method is to
+     * make multi-threaded work with IPersistentStorage more efficient by
+     * reducing the time spent inside the lock.
+     *
+     * Once committed, it is not allowed to modify the contents of the allocated
+     * entry. If there is a need to augment the allocation with additional data,
+     * GetTag/SetTag can be used.
+     *
+     * An error is returned if there is no incomplete allocation correponding to
+     * the provided pointer or the buffer is corrupted.
+     *
+     * Note: the checksum is not validated, the calling code has responsibility
+     * to provide the correct Crc32c checksum. Passing an incorrect checksum may
+     * lead to a corruption error.
+     */
+    [[nodiscard]] virtual NProto::TError Commit(
+        const void* ptr,
+        ui32 crc32c) = 0;
 
     /**
      * Frees a previously allocated and committed buffer.
--- a/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
+++ b/cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
@@ -54,7 +54,8 @@ struct IPersistentStorage
     [[nodiscard]] virtual TResultOrError<char*> Alloc(size_t size) = 0;
 
     /**
-     * Commits previously allocated memory buffer.
+     * Completes the previously made allocation by calculating checksum and
+     * making the allocation visible.
      *
      * Once committed, it is not allowed to modify the contents of the allocated
      * entry. If there is a need to augment the allocation with additional data,
@@ -63,14 +64,14 @@ struct IPersistentStorage
      * Memory that was allocated but not committed will be lost at buffer
      * recreation.
      *
-     * An error is returned if there is no incomplete allocation or the buffer
-     * is corrupted.
+     * An error is returned if there is no incomplete allocation correponding to
+     * the provided pointer or the buffer is corrupted.
      */
     [[nodiscard]] virtual NProto::TError Commit(const void* ptr) = 0;
 
     /**
-     * Commits previously allocated memory buffer but takes a checksum provided
-     * by the caller instead of calculating it.
+     * Commits the previously allocated memory buffer but takes a checksum
+     * provided by the caller instead of calculating it.
      *
      * Note: the checksum is not validated, the calling code has responsibility
      * to provide the correct Crc32c checksum. Passing an incorrect checksum may
--- a/cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
+++ b/cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
@@ -678,25 +678,11 @@ public:
             crc32c = Crc32c(ptr, eh.DataSize);
         }
 
-        if (Capabilities().EntryHeaderIsProcessedAtomically) {
-            eh.DataChecksum = *crc32c;
-            eh.FreeFlag = false;
+        eh.DataChecksum = *crc32c;
+        eh.FreeFlag = false;
 
-            if (!WriteEntryHeader(pos, eh)) {
-                return MakeBufferIsCorruptError();
-            }
-        } else {
-            eh.DataChecksum = *crc32c;
-
-            if (!WriteEntryHeader(pos, eh)) {
-                return MakeBufferIsCorruptError();
-            }
-
-            eh.FreeFlag = false;
-
-            if (!WriteEntryHeader(pos, eh)) {
-                return MakeBufferIsCorruptError();
-            }
+        if (!WriteEntryHeader(pos, eh)) {
+            return MakeBufferIsCorruptError();
         }
 
         return {};
@@ -728,25 +714,11 @@ public:
             return MakeBufferIsCorruptError();
         }
 
-        if (Capabilities().EntryHeaderIsProcessedAtomically) {
-            eh.DataChecksum = 0;
-            eh.FreeFlag = true;
+        eh.DataChecksum = 0;
+        eh.FreeFlag = true;
 
-            if (!WriteEntryHeader(pos, eh)) {
-                return MakeBufferIsCorruptError();
-            }
-        } else {
-            eh.FreeFlag = true;
-
-            if (!WriteEntryHeader(pos, eh)) {
-                return MakeBufferIsCorruptError();
-            }
-
-            eh.DataChecksum = 0;
-
-            if (!WriteEntryHeader(pos, eh)) {
-                return MakeBufferIsCorruptError();
-            }
+        if (!WriteEntryHeader(pos, eh)) {
+            return MakeBufferIsCorruptError();
         }
 
         EntryMap.erase(it);
--- a/cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
+++ b/cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
@@ -18,6 +18,7 @@
 #include <util/system/filemap.h>
 
 #include <atomic>
+#include <optional>
 
 namespace NCloud {
 
@@ -147,9 +148,6 @@ private:
 
     // Map of allocated entries: data ptr -> pos
     THashMap<const void*, ui64> EntryMap;
-
-    // Allocated entries that have not been committed yet
-    THashSet<ui64> IncompleteEntries;
 
 private:
     THeader* Header()
@@ -392,7 +390,7 @@ private:
     {
         auto front = GetFrontEntry();
         while (front.HasValue() && front.GetFreeFlag() &&
-               !IncompleteEntries.contains(front.ActualPos))
+               !EntryMap.contains(front.Data))
         {
             front = GetNextEntry(front);
         }
@@ -527,7 +525,8 @@ public:
 
         data.copy(allocationResult.AllocationPtr, data.size());
 
-        auto commitResult = Commit(allocationResult.AllocationPtr, 0, false);
+        auto commitResult =
+            Commit(allocationResult.AllocationPtr, std::nullopt);
 
         if (HasError(commitResult)) {
             return TPushBackResult(commitResult);
@@ -617,16 +616,8 @@ public:
             return TAllocResult(MakeBufferIsCorruptError());
         }
 
-        auto [it1, inserted1] = IncompleteEntries.insert(writePos);
-        if (!inserted1) {
-            SetCorrupted(
-                TStringBuilder()
-                << "Duplicate incomplete allocation at " << writePos);
-            return TAllocResult(MakeBufferIsCorruptError());
-        }
-
-        auto [it2, inserted2] = EntryMap.insert({ptr, writePos});
-        if (!inserted2) {
+        auto [_, inserted] = EntryMap.insert({ptr, writePos});
+        if (!inserted) {
             SetCorrupted(
                 TStringBuilder() << "Duplicate allocation at " << writePos);
             return TAllocResult(MakeBufferIsCorruptError());
@@ -653,7 +644,7 @@ public:
         return TAllocResult(ptr);
     }
 
-    NProto::TError Commit(const void* ptr, ui32 crc32c, bool hasCrc)
+    NProto::TError Commit(const void* ptr, std::optional<ui32> crc32c)
     {
         if (!ValidateAccess("Commit")) {
             return MakeBufferIsCorruptError();
@@ -663,12 +654,8 @@ public:
         if (it == EntryMap.end()) {
             return MakeInvalidPointerError();
         }
-        ui64 pos = it->second;
 
-        auto removed = IncompleteEntries.erase(pos);
-        if (!removed) {
-            return MakeInvalidPointerError();
-        }
+        ui64 pos = it->second;
 
         auto eh = Data()->ReadEntryHeader(pos);
         if (!eh.FreeFlag || eh.DataSize == 0 || eh.DataChecksum != 0 ||
@@ -680,7 +667,7 @@ public:
             return MakeBufferIsCorruptError();
         }
 
-        if (!hasCrc) {
+        if (!crc32c) {
             if (ptr != Data()->GetEntryDataPtr(pos, eh.DataSize)) {
                 SetCorrupted(
                     TStringBuilder()
@@ -692,14 +679,14 @@ public:
         }
 
         if (Capabilities().EntryHeaderIsProcessedAtomically) {
-            eh.DataChecksum = crc32c;
+            eh.DataChecksum = *crc32c;
             eh.FreeFlag = false;
 
             if (!WriteEntryHeader(pos, eh)) {
                 return MakeBufferIsCorruptError();
             }
         } else {
-            eh.DataChecksum = crc32c;
+            eh.DataChecksum = *crc32c;
 
             if (!WriteEntryHeader(pos, eh)) {
                 return MakeBufferIsCorruptError();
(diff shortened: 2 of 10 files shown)


cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h
```diff
@@ -69,7 +69,33 @@ struct IPersistentStorage
      * An error is returned if there is no incomplete allocation or the buffer
      * is corrupted.
      */
-    [[nodiscard]] virtual NProto::TError Commit() = 0;
+    [[nodiscard]] virtual NProto::TError Commit(const void* ptr) = 0;
+
+    /**
+     * Completes the previously made allocation using checksum provided by
+     * caller and making the allocation visible.
```
neihar: Just my opinion: very similar comments here and in the corresponding method in ringbuffer. And they are very long. I would like to see less duplicated info. Probably reference to one place or describe only pat that  layer adds, not all underlying stuff

e673: The API doesn't expose the explicit dependance on `TFileRingBuffer` (and the underlying storage may change in the future), so I'm in favor of comment duplication here instead of referring to a different location.

But I've reduced the comment for the second `Commit` method.

e673:
AI review 1

## Findings

1. [P2] `persistent_storage.cpp:109` does not refresh counters after `Alloc()`. `Alloc()` now reserves capacity and updates the maximum allocation size at `file_ring_buffer.cpp:609`, so usage metrics remain stale while allocations are incomplete—indefinitely if one is abandoned. A physically full buffer can still report zero usage. Call `SetCounters()` after allocation, including error paths that may mark storage corrupted.

> Done

2. [P2] `test_persistent_storage.cpp:34` no longer models the production commit contract. `Alloc()` immediately exposes entries through `Visit`, while both `Commit()` overloads at `line 52` ignore the pointer and always succeed. Consequently, wrong-pointer/double commits and `Free`/`SetTag` on incomplete allocations remain green in manager/state tests but fail in production. Track incomplete versus committed entries and validate the supplied pointer.

> This was pre-existing behavior and is not needed to be resolved

No confirmed data race, crash, OOB access, or wrap/recovery invariant violation was found under the documented external-lock model.

The PR’s sole human comment—removing duplicated `Commit` documentation—was addressed by the second commit. `git diff --check` passed, and all 76 focused `TFileRingBuffer` tests passed for V5/V6. Current PR workloads were still pending.

e673:
AI review 2

# Review result: **Request changes**

I reviewed the force-pushed head `d367d3f`. The PR description is still the default template, and I found no substantive human review comments to incorporate; GitHub showed the checks being rerun during this review. ([[GitHub](https://github.com/ydb-platform/nbs/pull/6981)][1])

The force-push **does fix** several issues from the previous pass:

* the pointer-bearing `Commit` API is propagated through Filestore, so the earlier compilation break is gone;
* `Front()` and `PopFront()` no longer expose an incomplete physical-head allocation;
* `Size()` excludes incomplete allocations;
* tag operations reject incomplete allocations;
* the checksum documentation now correctly says CRC32C. 

The following problems remain.

## 1. [P1] `Commit()` trusts a mutable mapped `DataSize`, enabling an out-of-bounds read and a possible data race

**Location:** `cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp`, `TFileRingBuffer::TImpl::Commit`

`Alloc(size)` records only `pointer -> position` in `EntryMap`. The original allocation size is not retained in trusted in-memory metadata; it exists only in the entry header inside the writable memory mapping. `Commit(ptr)` rereads that mapped header and then computes:

```cpp
Crc32c(ptr, entryHeader.DataSize)
```

It verifies that `DataSize` is nonzero and that the free/checksum/tag fields have the expected values, but it does not:

* compare `DataSize` with the size originally passed to `Alloc`;
* validate the resulting payload range with `GetEntryDataPtr`;
* verify that `ptr + DataSize` remains inside the allocation or even inside the mapping. 

This matters because the existing file-ring-buffer threat model explicitly includes the backing storage being modified while the service is using it. The format’s encoded header integrity is not a secret MAC: an external writer can construct a header with a different size while preserving the expected decoded zero checksum/tag state. ([[GitHub](https://github.com/ydb-platform/nbs/pull/3817)][2])

A failing sequence is:

```cpp
auto* a = ring.Alloc(16).AllocationPtr;

// The mapped header preceding a is externally modified:
// DataSize = a much larger value
// FreeFlag = true
// Tag = 0
// decoded DataChecksum = 0

ring.Commit(a);
```

Two memory-safety outcomes are possible:

1. If the forged length reaches beyond the mapping, `Crc32c` performs an out-of-bounds read and may crash with `SIGSEGV` or `SIGBUS`.
2. If it reaches another incomplete allocation that a different worker is filling outside the ring lock, `Commit(a)` reads that allocation while the other worker writes it. That is an unsynchronized read/write of the same memory and therefore a C++ data race.

The precomputed-checksum overload avoids the immediate CRC read, but it still commits the forged `DataSize`. Subsequent traversal or reopening will then interpret the wrong physical layout and report corruption.

### Required fix

Store immutable allocation metadata in memory:

```cpp
struct TIncompleteAllocation
{
    ui64 Position;
    ui32 Size;
    ui64 Generation; // useful for stale-handle/ABA protection
};
```

On commit:

1. find the allocation record;
2. validate that the mapped header still exactly represents an incomplete entry of the recorded size;
3. call the bounds-checked payload accessor with the recorded size and verify that it returns the same pointer;
4. calculate CRC32C using the recorded size, never the mapped size;
5. write the committed header;
6. remove the allocation from `IncompleteEntries` only after the transition succeeds.

The same size and header validation is necessary for `Commit(ptr, crc32c)`.

A regression test should corrupt an incomplete entry’s mapped `DataSize` before `Commit()` and verify under AddressSanitizer that commit returns a corruption error without reading outside the original allocation. A ThreadSanitizer variant should make the forged range overlap a second allocation being filled.

> I've added a pointer validation so no `SIGSEGV` will be possible.

> Regarding more strict validation: this is out of scope for this PR.
> This will be also needed for complete entries from `EntryMap` and will require adding more checks to each method.

## 2. [P2] An incomplete allocation cannot be canceled and can permanently pin the ring

**Locations:** `TFileRingBuffer::Free`, `EraseFreeEntriesFromFront`, and the public allocation API

The public API provides `Alloc`, `Commit`, and `Free`, but no way to abandon an allocation that will never be committed. `Free(ptr)` explicitly rejects incomplete allocations, and front reclamation deliberately stops when it encounters a position present in `IncompleteEntries`. 

Consider the intended asynchronous usage:

```text
A = Alloc(...)
worker A is canceled, throws, or loses its completion request

B = Alloc(...)
B is filled and committed
B may even be visited and freed
```

Because A remains incomplete:

* `Front()` and `PopFront()` remain blocked at A;
* freeing B does not advance `ReadPos` through A;
* the physical bytes occupied by A and all later entries remain in the used ring interval;
* continued allocations eventually exhaust the ring permanently.

Reopening the ring discards incomplete entries, but there is no live recovery path. One failed producer can therefore stall every other producer using the new concurrent-allocation facility.

The current write-back-cache caller still serializes and commits synchronously, so this failure path may not yet be reachable there. However, the linked issue explicitly calls for copying request buffers outside the lock with parallelism—the exact workflow in which cancellation and failed asynchronous work must be handled. 

### Required fix

Add an explicit `Abort`/`Cancel` operation, preferably through an opaque RAII allocation handle rather than a raw pointer. After the producer is known to be quiescent, abort should:

1. validate that the handle still refers to the same incomplete allocation;
2. remove it from `IncompleteEntries` and `EntryMap`;
3. leave or restore a valid free entry header;
4. invoke `EraseFreeEntriesFromFront()`.

A generation in the handle would also prevent a delayed completion for an old allocation from committing a new allocation that happens to reuse the same mapped address.

> Canceling incomplete allocations is not needed.

## 3. [P2] Allocation reserves physical capacity without updating storage counters

**Location:** `cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.cpp`, `TFileRingBufferStorage::Alloc`

With the new implementation, successful `Alloc()` immediately:

* advances the persistent `WritePos`;
* reduces available physical capacity;
* updates maximum observed occupancy.

However, `TFileRingBufferStorage::Alloc()` returns without calling `SetCounters()`. Counters are refreshed later by operations such as `Commit`, `Free`, or `Visit`. The exported values include raw used bytes and maximum observed usage. 

During the intended copy-outside-lock interval, monitoring therefore reports the pre-allocation capacity even though space has already been reserved. If an allocation is abandoned, the under-reporting can remain indefinitely while the ring approaches or reaches physical exhaustion.

Call `SetCounters()` after every successful allocation. Add a test that inspects raw-used and high-water counters after `Alloc()` but before `Commit()`.

> Done

## 4. [P3] `IPersistentStorage::Alloc` still documents the old single-allocation behavior

**Location:** `cloud/filestore/libs/vfs_fuse/write_back_cache/persistent_storage.h`

The interface documentation still states that only one allocation is possible at a time and that repeated `Alloc()` calls return an error. That directly contradicts the new implementation and the purpose of this PR. 

The comment should document:

* that multiple allocations may be outstanding;
* that each returned allocation must be individually committed;
* the lifetime of the returned pointer or handle;
* the required external synchronization;
* the cancellation behavior once an abort API is added.

> Done

## Concurrency coverage still needed

The added tests exercise out-of-order operations sequentially, but I did not find a true multithreaded test of the new feature. In particular, coverage should include:

1. two workers filling different allocations outside the ring lock and committing in reverse order;
2. `Visit`, `Front`, and `PopFront` executing under the external lock while another allocation is still being filled;
3. ThreadSanitizer validation of the above;
4. corrupted live-header size during commit under AddressSanitizer;
5. cancellation of an incomplete physical-head allocation followed by successful capacity reclamation and wraparound;
6. allocation-time counter updates. 

> I'll think about adding this test

Under correct API use and external serialization of ring-buffer methods, I did not identify another ordinary allocator-overlap or bounds-calculation bug. The direct `Front()` payload race from the first revision is fixed. The remaining P1 issue is specifically that the commit path bypasses the existing payload-range validation and uses mutable mapped metadata as a memory-read length.

I could not run the repository build or sanitizers locally because this environment failed to resolve `github.com`; this pass is based on the force-pushed source and GitHub diff. The remote checks were still rerunning when reviewed.

[1]: https://github.com/ydb-platform/nbs/pull/6981 "issue-1751: [Filestore] Support concurrent allocations by TFileRingBuffer by e673 · Pull Request #6981 · ydb-platform/nbs · GitHub"
[2]: https://github.com/ydb-platform/nbs/pull/3817 "issue-1751: TFileRingBuffer: resolve inconsistency in logic between PushBack, PopFront, Visit + deduplicate code by e673 · Pull Request #3817 · ydb-platform/nbs · GitHub"

e673:
AI review 3

I'll start by fetching the PR page to see what it's about.Now let me pull the actual diff and the full files using the container.Now let me read the full file_ring_buffer.cpp to understand invariants around the changed code.Let me look at the pre-PR version of the key functions and the format/accessor code to understand `FreeFlag`/checksum semantics.Now the accessor — its validation logic decides what happens to incomplete (FreeFlag=1, checksum=0) entries after a crash.The core logic is now clear. Let me look at the filestore-side changes (persistent_storage, the test double, and the caller).Now the test changes in `file_ring_buffer_ut.cpp`:Let me check how `Front`/`PopFront`/`Size`/`Empty` are used by real callers, since their semantics changed.Let me check the existing corruption/crash test harness in the ut file and whether it covers the new intermediate states.## Review of PR #6981 — multiple incomplete allocations in `TFileRingBuffer`

I read the full diff (3 commits, base `037c71175`), the untouched parts of `file_ring_buffer.cpp`, the V5/V6 data processors, and the accessor's validation code. I also built a Python model of the new state machine (positions, header states, `EntryMap`/`IncompleteEntries`, crash-snapshot at every intermediate write, reopen-and-continue) and fuzzed it across capacities 48–256. It found no violated invariants. Summary first, details below.

**Verdict: no crash, OOB access, or broken invariant found. Several design/robustness issues and a good amount of removable complexity.**

### Correctness (what I verified)

- **Memory safety.** The only raw byte access added is `Crc32c(ptr, eh.DataSize)` in `Commit`; it's guarded by `ptr == GetEntryDataPtr(pos, eh.DataSize)`, which bounds-checks `[pos, pos+hdr+DataSize)` against the mapping. `MemCopy`/`CopyMappedData` are untouched. Fine.
- **Crash consistency.** `Alloc` now writes `{size, crc=0, tag=0, Free=1}` and *then* bumps `WritePos` (fenced). Every intermediate state (slack marker written, header written, `WritePos` old/new) validates in `ValidateData`, and stale incomplete entries are skipped by `EraseFreeEntriesFromFront` on reopen (they're not in the fresh `IncompleteEntries`). The V5 two-step in `Commit`/`Free` correctly ensures each step changes only one 32-bit word and every intermediate state is a tolerated "free" header. Committed data is never lost in the model.
- **Bookkeeping.** `IncompleteEntries ⊆ values(EntryMap)` holds on all non-corrupting paths, so `Size()` doesn't underflow and the `Empty()` debug assert (`ReadPos==WritePos ⇔ EntryMap empty`) holds — including "free everything but a pinned incomplete entry at the front", wrap-around with a pinned `ReadPos`, and migration gating (never `Empty()` while an allocation is outstanding, so `TryMigrate` can't reset positions under a live allocation).

### Real issues

1. **No way to release an abandoned allocation (liveness).** `Free` on an incomplete entry returns `E_ARGUMENT` (and is tested to). Pre-PR, an abandoned `Alloc` cost nothing; now it permanently consumes capacity and, if it's at `ReadPos`, pins the read position so *no* later `Free` reclaims space — until process restart. Today's only caller (`TryStoreRequestInPersistentStorage`) can't fail between `Alloc` and `Commit` except via `Y_ABORT_UNLESS`, so it's not hit yet, but the whole point of the PR is to let other threads hold allocations open, and the first one that hits an error after `Alloc` will leak. Since the on-disk header is already `{Free=1, crc=0}`, cancelling is pure bookkeeping: erase from both maps and call `EraseFreeEntriesFromFront()`. I'd let `Free` accept incomplete pointers (it also deletes an error branch).

> This is an expected behavior. No `TFileRingBuffer` clients need freeing incomplete allocations.
> Also, accidental freeing the incomplete entry while another thread is being writing to it may cause real problems in `TWriteBackCache`.

2. **`Empty()`, `Size()`, `Front()`, `PopFront()` now disagree.** The new test literally asserts `Size()==0 && !Empty()`. `Front()` returns an empty `TStringBuf` with no error and `PopFront()` returns `Removed=false` with no error when the front is incomplete — indistinguishable from "buffer empty" while `Empty()` says false. A consumer written as `while (!Empty()) { Front(); PopFront(); }` (exactly the shape of `PopAll` in the ut file) busy-loops forever. `THandleOpsQueue` is safe only because it never uses `Alloc`. Suggest keeping the raw `ReadPos==WritePos` check internal (it's needed for position reset/migration) and making the public `Empty()` return `Size()==0`; and document on `Front`/`PopFront` that they can return "nothing" while `Size()>0`.

> I've changed `Size()` meaning — now it counts all allocations (including incomplete).
> The behavior of `Front` an `PopFront` is documented.

4. **Wrong caller-supplied CRC is a delayed, total loss.** `Commit(ptr, crc32c)` doesn't verify, and nothing in-process ever does; the mismatch surfaces only on the next open, where `ValidateData` fails and the **entire buffer** is declared corrupt (all cached writes discarded). The doc's "may lead to a corruption error" undersells that. Worth stating plainly, and worth a test (commit with a bad crc, reopen, assert `IsCorrupted()`).

> Added a test

6. **`Size()` can underflow in the corrupted path.** In `Alloc`, `IncompleteEntries.insert` happens before `EntryMap.insert`; if the second fails, `EntryMap.size() - IncompleteEntries.size()` wraps to ~2⁶⁴ and `TFileRingBufferStorage::Alloc` immediately publishes it via `SetCounters()`. Insert into `EntryMap` first (or collapse the two defensive checks into one — they detect the same impossible condition).

> Now it cannot because I've removed `IncompleteEntries`.

8. **`Commit` erases from `IncompleteEntries` before validating the header.** Every later failure happens to set `Corrupted`, so it's fine today, but the moment someone adds a non-corrupting early return, the entry silently becomes "committed" in `EntryMap` with `FreeFlag` still set on disk. Erase after the successful header write.

> `IncompleteEntries` has been removed.

### Complexity that can go

- **Move the V5 two-step into the V5 processor.** `Commit` and `Free` each carry a near-identical `if (EntryHeaderIsProcessedAtomically) {…} else {two writes}` block. The requirement is purely "when a V5 header transitions, the intermediate state must still read as free". `TFileRingBufferDataProcessorV5::WriteEntryHeader` can guarantee that itself: if `header.FreeFlag`, store the `DataSize` word first, signal-fence, then `Checksum`; otherwise store `Checksum` first, fence, then `DataSize`. I checked all four transitions (alloc, commit, free, set-tag) plus the slack marker: the intermediate states are identical to what the PR produces now. That deletes ~30 lines and the capability branching from `TImpl`, and the per-call fence in the new `WriteEntryHeader` wrapper becomes unnecessary.

> Done

- **`IncompleteEntries` is redundant with the on-disk `FreeFlag`.** "Incomplete" ⇔ "in `EntryMap` and header has `FreeFlag`". `Commit`/`Free`/`SetTag`/`GetTag` already read the header; `EraseFreeEntriesFromFront` has `front.Data` and can test `EntryMap.contains(front.Data)`. Keep a single `ui64 CommittedCount` for `Size()`. That removes the second container, the duplicate-insert check, the `Size()` subtraction (and issue 4).

> Done

- `Commit(ptr, crc32c, bool hasCrc)` → `std::optional<ui32>`.

> Done

### Tests

The new tests are sound and cover the headline behaviours (out-of-order commit, pinned front, precomputed crc survives reopen, stale entries dropped). Gaps worth filling given how much of this is about layout: (a) an incomplete entry pinning `ReadPos` through a wrap into the two-region state, then commit/free reclaiming; (b) crash with an incomplete entry *between* committed entries (only the tail case is covered); (c) the bad-crc reopen from issue 3; (d) V5 mid-commit intermediate state `{Free=1, crc≠0}` reopening cleanly — you have the `Corrupt(ofs)` harness already. Minor: the comment `// Already allocated` in `AllocShouldReturnExtendedStatus` means "already committed". `TTestStorage` still commits/frees anything (already discussed in the thread; I'd just note the divergence from production grew, since `Free` on an incomplete pointer is now `E_ARGUMENT` for real).

### Nits

`correponding` twice in `file_ring_buffer.h`; a double blank line after the `Commit(ptr, crc32c)` definition in the .cpp; the `IPersistentStorage::Commit(ptr)` doc still says "no incomplete allocation" without mentioning the pointer. The reviewer's comment-duplication request and the counters-after-`Alloc` point are both addressed in the follow-up commits.

> Fixed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -595,52 +609,82 @@ class TFileRingBuffer::TImpl
             Max(MaxObservedEntryByteCount, size);
 
         char* ptr = Data()->GetEntryDataPtr(writePos, size);
-        Y_ABORT_UNLESS(ptr != nullptr);
+        if (ptr == nullptr) {
+            SetCorrupted(
+                TStringBuilder() << "Cannot access data buffer at " << writePos
+                                 << ", size = " << size);
+            return TAllocResult(MakeBufferIsCorruptError());
```
SvartMetal: Do we want to pass a string with the concrete error message (one line above) so it gets logged by the calling code?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -506,7 +525,8 @@ class TFileRingBuffer::TImpl
 
         data.copy(allocationResult.AllocationPtr, data.size());
 
-        auto commitResult = Commit();
+        auto commitResult =
+            Commit(allocationResult.AllocationPtr, std::nullopt);
```
SvartMetal, line 518: Maybe add a Commit overload without crc32 (which calls `Commit(..., std::nullopt)`) to make the calling code less cryptic?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -255,6 +256,22 @@ class TFileRingBuffer::TImpl
             : TEntryInfo::CreateInvalid();
     }
 
+    // Sets corrupted state if Data()->WriteEntryHeader is failed
+    bool WriteEntryHeader(ui64 pos, const TFileRingBufferEntryHeader& header)
```
SvartMetal: It would be clearer to make it return an error, to preserve the invariant "if MakeBufferIsCorruptError is called, SetCorrupted is called nearby"

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
SvartMetal: Maybe store the last corruption error and return it in `MakeBufferIsCorruptError`?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
SvartMetal: Oh, looks like you already have this overload

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.h
```diff
@@ -117,10 +114,34 @@ class TFileRingBuffer
...
+     * checksums inside Commit() consumes some time and it will prevent other
+     * threads from using TFileRingBuffer. The purpose of this method is to make
+     * multi-threaded work with TFileRingBuffer more efficient by reducing the
+     * time spent inside the lock.
+     *
+     * Once committed, it is not allowed to modify the contents of the allocated
+     * entry. If there is a need to augment the allocation with additional data,
+     * GetTag/SetTag can be used.
+     *
+     * An error is returned if there is no incomplete allocation correponding to
```
SvartMetal: corresponding

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
SvartMetal, line 1: Are these incomplete allocations covered in the randomized test?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -255,6 +256,22 @@ class TFileRingBuffer::TImpl
...
+    {
+        // A compiler-only fence is sufficient here because there is no
+        // concurrent access to the memory and we just need to ensure
+        // that a compiler does not reorder writes.
```
neihar: IMHO: writing this near every fence looks strange

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -595,52 +609,82 @@ class TFileRingBuffer::TImpl
...
-            Crc32c(CurrentAllocation.Data, CurrentAllocation.Header.DataSize);
-
-        bool written = Data()->WriteEntryHeader(
-            CurrentAllocation.ActualPos,
-            CurrentAllocation.Header);
+        ui64 pos = it->second;
 
-        if (!written) {
+        auto eh = Data()->ReadEntryHeader(pos);
+        if (!eh.FreeFlag || eh.DataSize == 0 || eh.DataChecksum != 0 ||
```
neihar: if (!eh.FreeFlag) {
    return MakeInvalidPointerError();
}

as for Free with incomplete for symmetry, without corruption (because it is not)

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
neihar: especially interesting recovery from incomplete allocations

neihar:
It is easier to read all code twice than 3 rounds of posted AI review, looks strange for me

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: The error message is already logged inside `SetCorrupted` as a critical event.
Do you think we really need to propagate it to the calling code?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Yes, public methods have two overloads.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: I've added a method `SetCorruptedAndReturnError`

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: I've reverted adding this method - it is no more actual

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.h
e673: fixed

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
e673: Added a randomized test for this functionality

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: It is duplicated only once - let's keep it there.
I've only moved this from `Commit` to `Alloc`.

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Agree - fixed
(previously, there was a different implementation)

neihar (approved):

SvartMetal (approved):

yegorskii (approved):

PR #6985, opened 2026-09-01
[Filestore] support filestore-client alter; teach filestore-client executeaction to fail in case of an error

SvartMetal:
### Notes
Added alter command, by analogy with resize:
```
filestore-client alter --filesystem <fs-id> --cloud <cloud-id> --folder <folder-id> [--config-version N]
```

`TExecuteActionCommand::Execute` printed the error and returned exit code 0, so a failed action was indistinguishable from a successful one for scripts and tests. 
It now sets exit code 1 in case of an error by analogy with `filestore-client create`: https://github.com/ydb-platform/nbs/blob/bbbe4dcbc3c664e619f4f3037475dccdd576ddac/cloud/filestore/apps/client/lib/create.cpp#L110

### Issue

### DIFF
--- a/cloud/filestore/tests/scheme_cache/test.py
+++ b/cloud/filestore/tests/scheme_cache/test.py
@@ -77,16 +77,7 @@ def test_should_resize_and_alter():
     client.resize("fs0", shard_count * int(SHARD_SIZE / BLOCK_SIZE))
     verify_filesystem_topology(client, "fs0", shard_count)
 
-    client.execute_action(
-        "alterfilesystem",
-        {
-            "FileSystemId": "fs0",
-            "Config": {
-                "CloudId": "new_test_cloud",
-                "FolderId": "new_test_folder",
-            },
-        },
-    )
+    client.alter("fs0", "new_test_cloud", "new_test_folder")
 
     verify_io(client, "fs0")
 
--- a/cloud/filestore/apps/client/lib/alter.cpp
+++ b/cloud/filestore/apps/client/lib/alter.cpp
@@ -1,0 +1,69 @@
+#include "command.h"
+
+namespace NCloud::NFileStore::NClient {
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TAlterCommand final
+    : public TFileStoreCommand
+{
+private:
+    TString CloudId;
+    TString FolderId;
+    ui32 ConfigVersion = 0;
+
+public:
+    TAlterCommand()
+    {
+        Opts.AddLongOption("cloud")
+            .Required()
+            .RequiredArgument("STR")
+            .Help("new cloud id")
+            .StoreResult(&CloudId);
+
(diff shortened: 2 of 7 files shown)


cloud/filestore/tests/scheme_cache/test.py
```diff
@@ -77,16 +77,7 @@ def test_should_resize_and_alter():
...
-            "FileSystemId": "fs0",
-            "Config": {
-                "CloudId": "new_test_cloud",
-                "FolderId": "new_test_folder",
-            },
-        },
-    )
+    client.alter("fs0", "new_test_cloud", "new_test_folder")
 
     verify_io(client, "fs0")
```
WilyTiger, line 82: maybe we should also check here via describe what cloud and folder were changed?

SvartMetal: I'll add this check in the next PR, where alter will be supported for fs shards

WilyTiger (approved):

nisqatsi (approved):

SvartMetal:
AI review https://chatgpt.com/share/6a974155-7fd0-83eb-9c47-3e03493042b4:
```
I found no crash, OOB access, or broken production invariant in this PR. There are no memcpy/memmove/memset or equivalent raw-memory operations in the changed code.
The one thing I would change is the test:
The new alter test does not verify that anything was actually altered. test_should_resize_and_alter() calls client.alter("fs0", "new_test_cloud", "new_test_folder"), then only performs ordinary file I/O. If the server accepted the request but silently ignored CloudId/FolderId, or the CLI accidentally failed to populate those fields while still getting a successful response, the test would pass.
I would describe("fs0") after the alter and assert that CloudId == "new_test_cloud" and FolderId == "new_test_folder". Ideally also exercise --config-version, since it is part of the newly exposed CLI API but currently untested. The new command explicitly sets all three fields, including ConfigVersion.
The executeaction fix looks correct. Previously the error path printed an error and returned success; now it calls ProgramShouldContinue.ShouldStop(1) and returns false. TCommand::Run() waits on a false return only while the state is still Continue, then returns ProgramShouldContinue.GetReturnCode(), so this does not introduce a hang and correctly produces exit code 1. The modified sharding tests are also sensible: they explicitly expect common.ExecutionError only for actions against shards that are documented as not existing yet.

There is a small simplification available in TAlterCommand::Execute(): since it derives from TFileStoreCommand, the explicit
if (HasError(response)) {
    STORAGE_THROW_SERVICE_ERROR(response.GetError());
}
can just be CheckResponse(response), which already implements exactly that logic. I wouldn't block the PR on this. 
So my verdict is: no serious implementation problem found; one weak/nonsensical aspect in the test is that it tests “alter didn't break I/O” rather than “alter actually altered the requested attributes.”.
The large filestore test suite also passed on the PR commit.
```
I think it is good enough, test will be improved in a separate PR

PR #6990, opened 2026-09-01
[Filestore] add confirmation to filestore-client destroy

SvartMetal:
`filestore-client destroy` currently destroys the filesystem immediately, with no safeguard against a mistyped or wrong `--filesystem` id. This adds the same confirmation step that `blockstore-client destroyvolume` already has: the command prints a prompt and reads the filesystem id from stdin; if it doesn't match `--filesystem`, the command fails with exit code 1 and no request is sent.

### DIFF
--- a/cloud/filestore/tests/python/lib/client.py
+++ b/cloud/filestore/tests/python/lib/client.py
@@ -91,8 +91,18 @@ class FilestoreCliClient:
             "--filesystem", fs,
         ] + self.__cmd_opts()
 
-        logger.info("destroying filestore: " + " ".join(cmd))
-        return common.execute(cmd, env=self.__env, check_exit_code=self.__check_exit_code).stdout
+        # destroy asks to confirm by typing the filesystem id to stdin
+        with tempfile.NamedTemporaryFile(mode="w+") as confirmation:
+            confirmation.write(fs)
+            confirmation.flush()
+            confirmation.seek(0)
+
+            logger.info("destroying filestore: " + " ".join(cmd))
+            return common.execute(
+                cmd,
+                env=self.__env,
+                stdin=confirmation,
+                check_exit_code=self.__check_exit_code).stdout
 
     def mount(self, fs, path, mount_seqno=0, readonly=False):
         cmd = [
--- a/cloud/filestore/apps/client/lib/destroy.cpp
+++ b/cloud/filestore/apps/client/lib/destroy.cpp
@@ -1,5 +1,8 @@
 #include "command.h"
 
+#include <util/stream/input.h>
+#include <util/stream/output.h>
+
 namespace NCloud::NFileStore::NClient {
 
 namespace {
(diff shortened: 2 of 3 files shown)


SvartMetal:
image

Not relevant to this change

debnatkh (dismissed):

yegorskii (dismissed):

cloud/filestore/tests/python/lib/client.py
```diff
@@ -91,8 +91,18 @@ def destroy(self, fs):
             "--filesystem", fs,
         ] + self.__cmd_opts()
 
-        logger.info("destroying filestore: " + " ".join(cmd))
-        return common.execute(cmd, env=self.__env, check_exit_code=self.__check_exit_code).stdout
+        # destroy asks to confirm by typing the filesystem id to stdin
```
nisqatsi, line 94: negative-тест для несоответствия введенной и ожидаемой FileSystemId возможно написать?

SvartMetal: Good idea! Added

SvartMetal:
AI review https://chatgpt.com/share/6a974143-e700-83eb-a592-85e760a9e62f:
```
I reviewed the full PR, its description/comment thread, all three changed files, and the analogous blockstore-client destroyvolume implementation. I don’t see a correctness blocker in the current version.
A few observations:
No memory-safety concerns. The patch contains no memcpy/memmove/memset, pointer arithmetic, indexing, or other plausible OOB paths. The confirmation is read into a TString, so arbitrarily long stdin input is handled by the string implementation.
The important invariant is preserved: the destroy RPC is constructed/sent only after confirmation == FileSystemId. A mismatch therefore cannot accidentally reach DestroyFileStore.
The negative test added after the review comment is meaningful, not bullshit. It supplies not-fs0, requires a non-zero exit status, then verifies both describe("fs0") and list_filestores() still see the filesystem. It subsequently destroys the same filesystem through the regular helper, covering the positive path too.
Updating FilestoreCliClient.destroy() to feed the filesystem ID is necessary; otherwise a large number of existing tests using this wrapper would start failing/hanging on confirmation. The implementation does exactly that.
The one thing I investigated as potentially suspicious was Y_ENSURE on confirmation failure. Unlike blockstore, which logs and returns false, filestore throws; consequently TCommand::Run() doesn't execute its explicit Stop() after Execute(). However, this is not specific enough to reject this PR: filestore commands already use exceptions for ordinary command failures, TApp::Run() catches them and returns exit code 1, so this follows the existing client error model. I wouldn't request a local redesign here.
There are two minor simplifications I might consider, but neither deserves blocking the PR:
The test's client.describe("fs0") followed by assert "fs0" in client.list_filestores() is redundant. Either one demonstrates that failed confirmation did not destroy the filesystem; describe() is the stronger check because it itself must succeed.
The Python helper needs a temporary file solely to provide a few bytes of stdin. If yatest.common.execute has a convenient supported mechanism for passing an in-memory stdin payload, that would be simpler, but I wouldn't introduce a new mechanism merely for this patch.
So my review outcome would be LGTM / no actionable correctness issues found. In particular, I would not ask for more confirmation tests: mismatch/no-request and successful destruction are both already covered, and EOF/empty input would exercise essentially the same comparison branch without protecting an additional invariant.
```

nisqatsi (approved):

debnatkh (approved):

PR #7001, opened 2026-09-02
[NBS] Preserve skipped compaction counters

vladstepanyuk:
### Notes
Changed `=` -> `+=` here
https://github.com/ydb-platform/nbs/blob/e0051499b61fdf66557329ff5942a24cfda7b84b/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp#L119-L122
and  here
https://github.com/ydb-platform/nbs/blob/e0051499b61fdf66557329ff5942a24cfda7b84b/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp#L146-L149
because when a range compaction yields both a DataBlobId and a ZeroBlobId, both go into the same list (same ChannelDataKind), data first with {N, M}, then zero with {0, 0} (because of the !rc.DataBlobId guard in part_actor_compaction.cpp). With =, the zero blob overwrote the skipped counts with 0, so the compaction map under-counted BlobCount/BlockCount for such ranges. += fixes that.

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_ut.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_ut.cpp
@@ -6195,7 +6195,7 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
         bool firstCompactionRequestCaptured = false;
         bool compactionResultObserved = false;
 
-        runtime->SetObserverFunc(
+        auto observer = runtime->AddObserver(
             [&](TAutoPtr<IEventHandle>& event)
             {
                 switch (event->GetTypeRewrite()) {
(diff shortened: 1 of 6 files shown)


cloud/blockstore/libs/storage/partition/part_ut.cpp
```diff
@@ -6173,6 +6173,108 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
...
+        auto runtime = PrepareTestActorRuntime(config);
+
+        TPartitionClient partition(*runtime);
+        partition.WaitReady();
+
+        std::unique_ptr<IEventHandle> firstCompactionRequest;
+        bool firstCompactionRequestCaptured = false;
+        bool compactionResultObserved = false;
+
+        runtime->SetObserverFunc(
```
SvartMetal: It is deprecated

https://github.com/ydb-platform/nbs/blob/f0094052013483f49f76edcd685d6581f97081ca/contrib/ydb/library/actors/testlib/test_runtime.h#L291

SvartMetal (approved):

komarevtsev-d (approved):

PR #7009, opened 2026-09-02
update default rdma timeouts; resolves issue-2258

### LINKED ISSUE #2258: Set reasonable timeout in ibv_qp_attr
Current RDMA request timeout is 1 minute – https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/rdma/iface/client.h#L35. We don't set QP timeout and it seems that `timeout * retry_count` amounts to about 2.5 minutes

See https://www.rdmamojo.com/2013/01/12/ibv_modify_qp for the reference on timeout values

Log highlighting the problem: http://paste.org.ru/?8m0chl

tpashkin:
`min_rnr_timer` values are taken from this old version of a spec – https://www.afs.enea.it/asantoro/V1r1_2_1.Release_12062007.pdf (Encoding for RNR NAK Timer Field)
And cross-referenced with linux rxe implementation – https://elixir.bootlin.com/linux/v6.6.37/source/drivers/infiniband/sw/rxe/rxe_comp.c#L50


yegorskii (dismissed):

cloud/storage/core/libs/rdma/iface/client.h
```diff
@@ -44,10 +44,10 @@ struct TClientConfig
     ui32 RecvQueueSize = 0;
     TDuration ResolveTimeout = TDuration::Seconds(10);
     TDuration FlushTimeout = TDuration::Seconds(10);
-    ui8 QpRetryCount = 7;
-    ui8 QpRnrRetryCount = 7;
-    ui8 QpTimeout = 0;
-    ui8 QpMinRnrTimer = 0;
+    ui8 QpTimeout = 14;         // ~67ms
+    ui8 QpRetryCount = 7;       // QpTimeout * QpRetryCount = ~470ms
+    ui8 QpMinRnrTimer = 12;     // ~640us
```
budevg, line 49: it's now impossible to set `QpMinRnrTimer` = 0 and `QpTimeout` = 0 from config since 
```
#define SET(param, ...)                            \
    if (const auto& value = config.Get##param()) { \
        result.param = __VA_ARGS__(value);         \
    }
```
used
are we ok with this ?

budevg (changes requested):

cloud/storage/core/libs/rdma/iface/client.h
tpashkin: In practice – yes, but maybe not in theory. I'll check whether they've supported field presence for proto3

tpashkin: Okay, I've marked these 4 fields optional

tpashkin: I will probably do this for other fields as well, but in a separate PR

cloud/storage/core/libs/rdma/iface/config.h
```diff
@@ -151,4 +146,9 @@ inline TServerConfigPtr CreateServerConfigPtr(const NProto::TRdmaServer& config)
     return std::make_shared<TServerConfig>(CreateServerConfig(config));
 }
 
+////////////////////////////////////////////////////////////////////////////////
+
+#undef SET_NESTED
```
budevg, line 151: undef SET_OPTIONAL

cloud/storage/core/protos/rdma.proto
```diff
@@ -56,18 +56,23 @@ message TRdmaServer
...
-    uint32 QpRnrRetryCount = 19;
+    // 7 means infinite retries.
+    optional uint32 QpRnrRetryCount = 19;
+
     // Sets the "timeout" parameter in the "ibv_qp_attr" for
     // "ibv_modify_qp()" function. Valid values are in range [0, 31].
     // The resulting timeout is calculated as 4096*2^{QpTimeout} microseconds.
     // 0 means infinite timeout.
-    uint32 QpTimeout = 20;
+    optional uint32 QpTimeout = 20;
```
budevg, line 70: let's add test to config_ut.cpp to validate that optional value can be set to 0

budevg (changes requested):

cloud/storage/core/protos/rdma.proto
tpashkin: fine, but the entire premise of optional is to able to distinguish between default and explicitly set zeroes

yegorskii (approved):

budevg (approved):

PR #7012, opened 2026-09-03
issue-6608: do not allow to attach directory to non-existent quota

debnatkh:
### Notes
We should not allow attaching dirs to non-existent quotas

### Issue
#6608

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
@@ -147,6 +147,11 @@ bool TIndexTabletActor::PrepareTx_SetNodeAttr(
         } else if (args.Node->Attrs.GetType() != NProto::E_DIRECTORY_NODE) {
             args.Error = ErrorIsNotDirectory(args.NodeId);
             return true;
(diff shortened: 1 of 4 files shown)


neihar (approved):

cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
```diff
@@ -147,6 +147,11 @@ bool TIndexTabletActor::PrepareTx_SetNodeAttr(
         } else if (args.Node->Attrs.GetType() != NProto::E_DIRECTORY_NODE) {
             args.Error = ErrorIsNotDirectory(args.NodeId);
             return true;
+        } else if (!FindQuota(newQuotaId)) {
```
agalibin, line 150: "else" is extra if you have "return" from "if"

agalibin (approved):

cloud/filestore/libs/storage/tablet/tablet_actor_setnodeattr.cpp
debnatkh: True, but I would rather keep it consistent with other identical places for now

PR #7018, opened 2026-09-03
[Filestore] fix responsing to GetForcedOperationStatus for a forced operation with invalid number of ranges

tdiff:
### Notes
Before this change a forced operation with empty evaluated ranges would allocate an operationId and return a success response. However, during queue handling requests with empty responses are filtered out and silently dropped, so the client has no chance of getting status of this operation. 

### Issue
#4115

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
@@ -272,15 +272,15 @@ void TIndexTabletActor::EnqueueForcedRangeOperationIfNeeded(
     }
 
     auto pendingRequest = DequeueForcedRangeOperation();
-    if (pendingRequest.Ranges.empty()) {
+    if (!pendingRequest) {
         return;
     }
 
     auto request =
         std::make_unique<TEvIndexTabletPrivate::TEvForcedRangeOperationRequest>(
-            std::move(pendingRequest.Ranges),
-            pendingRequest.Mode,
-            std::move(pendingRequest.OperationId));
+            std::move(pendingRequest->Ranges),
+            pendingRequest->Mode,
+            std::move(pendingRequest->OperationId));
     ctx.Send(ctx.SelfID, request.release());
 }
 
@@ -295,9 +295,13 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         LogTag.c_str(),
         msg->Ranges.size());
 
-    auto replyError = [&] (
-        const NProto::TError& error)
+    auto replyError = [&](const NProto::TError& error)
     {
+        AbortForcedRangeOperation(
+            msg->Mode,
+            std::move(msg->Ranges),
+            std::move(msg->OperationId));
+
         if (ev->Sender == ctx.SelfID) {
             return;
         }
--- a/cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
@@ -1461,7 +1461,7 @@ TString TIndexTabletState::EnqueueForcedRangeOperation(
     return operationId;
 }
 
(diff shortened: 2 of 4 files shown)


cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1198,6 +1198,15 @@ void TIndexTabletActor::HandleForcedOperation(
         ranges.erase(ranges.begin(), b);
     }
 
+    if (ranges.empty()) {
+        replyError(MakeError(S_FALSE, "no ranges found"));
+        return;
+    }
+
+    if (ranges.size() > Max<ui32>()) {
+        replyError(MakeError(E_FAIL, "too many ranges to process"));
```
SvartMetal: E_FAIL is used only in high-severity cases, for example when something really unexpected happens, like `catch (...)` or corruption

E_ARGUMENT is a better choice here

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1198,6 +1198,15 @@ void TIndexTabletActor::HandleForcedOperation(
         ranges.erase(ranges.begin(), b);
     }
 
+    if (ranges.empty()) {
+        replyError(MakeError(S_FALSE, "no ranges found"));
```
SvartMetal: I wouldn't return S_FALSE here, to reduce the complexity of the client code (it would have to handle such an error differently, since it can't just use HasError to distinguish between success and failure)

I think it's fine to schedule the operation and finish it early in this case

tdiff: done

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -1486,6 +1486,18 @@ void TIndexTabletState::StartForcedRangeOperation(
         std::move(operationId));
 }
 
+void TIndexTabletState::AbortForcedRangeOperation(
+    TEvIndexTabletPrivate::EForcedRangeOperationMode mode,
+    TVector<ui32> ranges,
+    TString operationId)
+{
+    TABLET_VERIFY(!ForcedRangeOperationState.Defined());
```
SvartMetal, line 1494: Y_DEBUG_ABORT_UNLESS

neihar (dismissed):

cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
```diff
@@ -295,9 +295,13 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         LogTag.c_str(),
         msg->Ranges.size());
 
-    auto replyError = [&] (
-        const NProto::TError& error)
+    auto replyError = [&](const NProto::TError& error)
     {
+        AbortForcedRangeOperation(
```
tdiff, line 300: Codex review:

• Found one correctness issue:

  - [P2] Oversized requests appear unfinished forever — tablet_state_data.cpp:1494. AbortForcedRangeOperation() records rejected requests with Current == 0 and discards the rejection
    error. For Ranges.size() > Max<ui32>(), status therefore reports success with a positive total and zero progress permanently. The CLI polls until processed ≥ total, so it never
    finishes. Preserve and return a terminal error for this case. The empty-range test misses this because zero processed already equals its total.

  The new test reproduces the original empty-range bug; I found no apparent flakiness. No out-of-bounds access, unsafe move, or RAII defect found in the changes; no custom move
  assignment or raw memory operations were added.

tdiff: I'd prefer to ignore it as a corner case that should be fixed by adding an explicit status for the request instead of inferring it from the counts of handled ranges.

neihar (approved):

komarevtsev-d (approved):

PR #7020, opened 2026-09-03
[NBS] Local NVMe service: reconcile NVMe device states

sharpeye:
Add runtime state tracking for local NVMe devices.

* Continuously poll the device provider and reconcile device states.
* Mark missing devices as `OFFLINE`.
* Track `ONLINE`, `MAINTENANCE`, and `OFFLINE` device counts.
* Display device state and firmware revision on the Disk Agent monitoring page.
* Extend the test provider and unit tests to cover state transitions.

### DIFF
--- a/cloud/blockstore/libs/local_nvme/service_linux.cpp
+++ b/cloud/blockstore/libs/local_nvme/service_linux.cpp
@@ -255,6 +255,7 @@ private:
     auto RefreshDevice(const TSerialNumber& serialNumber)
         -> TResultOrError<NProto::TNVMeDevice>;
     auto UpdateDevices() -> NProto::TError;
+    void ReconcileDevices(TVector<NProto::TNVMeDevice> fetchedDevices);
     bool TryRestoreStateFromCache();
     void UpdateStateCache();
     auto CreateStateSnapshot() const -> NProto::TLocalNVMeServiceState;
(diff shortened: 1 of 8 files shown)


cloud/blockstore/libs/local_nvme/service_linux.cpp
```diff
@@ -594,17 +622,67 @@ auto TLocalNVMeService::EnsureIsReady(
...
+
+            if (oldState != newState) {
+                STORAGE_INFO(
+                    "NVMe device "
+                    << device.GetSerialNumber().Quote() << " changed state: "
+                    << NProto::ENVMeDeviceState_Name(oldState) << " -> "
+                    << NProto::ENVMeDeviceState_Name(newState));
+            }
+
+            device.SetDeviceState(newState);
```
dvrazumov, line 654: Why only the state field is updated? Don't we want to update the whole struct (like in case when PCI address suddenly changes)?

yegorskii (approved):

cloud/blockstore/libs/local_nvme/service_linux.cpp
sharpeye: I'd rather not complicate the code for now. Additional validation will be added in follow-up PRs.

dvrazumov (approved):

Issue #6977, opened 2026-08-31
[filestore] TForcedOperationStatusRequest does not return status for queued op


tdiff:
If a forced operation gets queued, following TForcedOperationStatusRequest requests won't return a valid status for it, because 
```
const TForcedOperationState* FindForcedOperation(
        const TString& operationId) const;
```
does not check the pending events queue.

Actually, the forced operation api looks half-broken: it mostly disallows queuing requests unless they come to fast one to another. Additionally, the queue is processed in LIFO order which is probably unexpected.

PR #7026, opened 2026-09-03
[Filestore] add status field to TForcedOperationStatusResponse

tdiff:
### Notes
This change adds a summary request state to TForcedOperationStatusResponse. It allows client to clearly understand the actual state of the operation.

Additionally, it fixes an issue with tablet changing operation id when re-queuing, and passes downstream operation error to the upstream (previously it was only logged).

### Issue
#4115, #6977

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1311,6 +1311,7 @@ void TIndexTabletActor::HandleForcedOperationStatus(
     const auto& request = ev->Get()->Record;
 
     using TResponse = TEvIndexTablet::TEvForcedOperationStatusResponse;
+    using TStatus = NProtoPrivate::TForcedOperationStatusResponse;
     auto response = std::make_unique<TResponse>();
 
     const auto* state = FindForcedRangeOperation(request.GetOperationId());
@@ -1318,11 +1319,17 @@ void TIndexTabletActor::HandleForcedOperationStatus(
         response->Record.SetRangeCount(state->RangesToCompact.size());
         response->Record.SetProcessedRangeCount(state->Current);
         response->Record.SetLastProcessedRangeId(state->GetCurrentRange());
-    } else {
+        response->Record.SetStatus(state->Status);
+        response->Record.MutableError()->CopyFrom(state->Error);
+    } else if (IsForcedRangeOperationPending(request.GetOperationId())) {
+        response->Record.SetStatus(TStatus::E_PENDING);
+    }
+    else {
+        response->Record.SetStatus(TStatus::E_UNKNOWN);
         response->Record.MutableError()->CopyFrom(MakeError(
             E_NOT_FOUND,
             TStringBuilder() << "forced operation with id "
-                << request.GetOperationId() << "not found"));
+                             << request.GetOperationId() << "not found"));
     }
 
     NCloud::Reply(ctx, *ev, std::move(response));
--- a/cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
@@ -300,7 +300,8 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         AbortForcedRangeOperation(
             msg->Mode,
             std::move(msg->Ranges),
-            std::move(msg->OperationId));
+            std::move(msg->OperationId),
+            error);
 
         if (ev->Sender == ctx.SelfID) {
             return;
@@ -311,7 +312,12 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         NCloud::Reply(ctx, *ev, std::move(response));
     };
 
-    if (msg->Ranges.empty() || msg->Ranges.size() > Max<ui32>()) {
+    if (msg->Ranges.empty()) {
+        replyError(MakeError(S_OK));
+        return;
+    }
+
+    if (msg->Ranges.size() > Max<ui32>()) {
         replyError(ErrorInvalidArgument());
         return;
     }
@@ -322,9 +328,11 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         msg->CallContext);
     requestInfo->StartedTs = ctx.Now();
 
-    // will lose original request info in case of enqueueing external request
     if (IsForcedRangeOperationRunning()) {
-        EnqueueForcedRangeOperation(msg->Mode, std::move(msg->Ranges));
+        EnqueueForcedRangeOperation(
+            msg->Mode,
+            std::move(msg->Ranges),
+            std::move(msg->OperationId));
         return;
     }
 
@@ -383,7 +391,7 @@ void TIndexTabletActor::HandleForcedRangeOperationCompleted(
     TABLET_VERIFY(IsForcedRangeOperationRunning());
     WorkerActors.erase(ev->Sender);
 
-    CompleteForcedRangeOperation();
+    CompleteForcedRangeOperation(msg->GetError());
     EnqueueForcedRangeOperationIfNeeded(ctx);
 }
 
--- a/cloud/filestore/libs/storage/tablet/tablet_state.h
+++ b/cloud/filestore/libs/storage/tablet/tablet_state.h
@@ -1525,6 +1525,9 @@ public:
 
         TInstant StartTime = TInstant::Now();
         ui32 Current = 0;
+        NProtoPrivate::TForcedOperationStatusResponse::EStatus Status =
+            NProtoPrivate::TForcedOperationStatusResponse::E_UNKNOWN;
+        NProto::TError Error;
 
         TForcedRangeOperationState(
                 TEvIndexTabletPrivate::EForcedRangeOperationMode mode,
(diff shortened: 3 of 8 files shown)


cloud/filestore/private/api/protos/tablet.proto
```diff
@@ -745,6 +745,17 @@ message TForcedOperationStatusResponse
...
+    // Operation status summary
+    enum EStatus
+    {
+        E_UNKNOWN = 0;
+        E_PENDING = 1;
+        E_RUNNING = 2;
+        E_COMPLETED = 3;
+        E_FAILED = 4;
+    }
+    EStatus Status = 5;
```
tdiff, line 758: Codex review:

• Found one compatibility issue in b452d6f791:

  - [P1] New status responses break older CLI clients — tablet_actor.cpp:1322. Setting a nonzero Status makes the service include "status" in its ExecuteAction JSON output. The previous
    CLI parses that JSON using JsonStringToMessage with unknown fields rejected. Consequently, an older CLI against an upgraded service aborts its first status poll with failed to parse
    response json, while compaction continues. The new E_UNKNOWN fallback only handles the opposite upgrade direction. Negotiate status-aware responses or version the action, and add a
    compatibility test parsing the response with the previous schema.

  The added tests meaningfully exercise empty completion, partial-progress failure, and pending-operation ID preservation. All nine targeted test variants passed using the existing test
  binary; I did not rebuild it. I found no definite new flakiness.

  No introduced out-of-bounds access, lifetime bug, or RAII misuse found. The diff contains no raw memory operations or custom move assignment.

  A small simplification: the CLI’s new E_FAILED throw is unreachable for responses produced here—operation failures already propagate through ExecuteAction’s error and throw at
  CheckResponse. Retain it only if deliberately supporting status-only failures.

tdiff: I'd say giving that client is a mostly support / dev tool the burden of introducing api versioning here is larger that simply upgrading a client if faced this issue. @SvartMetal ?

tdiff:
Codex review:

• Found one compatibility issue in https://github.com/ydb-platform/nbs/commit/b452d6f791dc33ad2eadcae6583881f341490543:

    [P1] New status responses break older CLI clients — tablet_actor.cpp:1322. Setting a nonzero Status makes the service include "status" in its ExecuteAction JSON output. The previous
    CLI parses that JSON using JsonStringToMessage with unknown fields rejected. Consequently, an older CLI against an upgraded service aborts its first status poll with failed to parse
    response json, while compaction continues. The new E_UNKNOWN fallback only handles the opposite upgrade direction. Negotiate status-aware responses or version the action, and add a
    compatibility test parsing the response with the previous schema.

The added tests meaningfully exercise empty completion, partial-progress failure, and pending-operation ID preservation. All nine targeted test variants passed using the existing test
binary; I did not rebuild it. I found no definite new flakiness.

No introduced out-of-bounds access, lifetime bug, or RAII misuse found. The diff contains no raw memory operations or custom move assignment.

A small simplification: the CLI’s new E_FAILED throw is unreachable for responses produced here—operation failures already propagate through ExecuteAction’s error and throw at
CheckResponse. Retain it only if deliberately supporting status-only failures.

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1311,18 +1311,25 @@ void TIndexTabletActor::HandleForcedOperationStatus(
...
     auto response = std::make_unique<TResponse>();
 
     const auto* state = FindForcedRangeOperation(request.GetOperationId());
     if (state) {
         response->Record.SetRangeCount(state->RangesToCompact.size());
         response->Record.SetProcessedRangeCount(state->Current);
         response->Record.SetLastProcessedRangeId(state->GetCurrentRange());
-    } else {
+        response->Record.SetStatus(state->Status);
+        response->Record.MutableError()->CopyFrom(state->Error);
```
neihar, line 1323: Status response's Error now means two different things, "status request failed" and "the operation failed". If a compaction ever fails with E_NOT_FOUND, the CLI will treat it as a reboot and re-launch.

neihar: we can add NProto.TError OperationError as additional field to distingluish it in CLI

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -1564,7 +1567,8 @@ FILESTORE_DUPCACHE_REQUESTS(FILESTORE_DECLARE_DUPCACHE)
 public:
     TString EnqueueForcedRangeOperation(
         TEvIndexTabletPrivate::EForcedRangeOperationMode mode,
-        TVector<ui32> ranges);
+        TVector<ui32> ranges,
+        TString operationId = {});
```
neihar, line 1571: let's not use default values and hidden logic

cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
```diff
@@ -322,9 +328,11 @@ void TIndexTabletActor::HandleForcedRangeOperation(
         msg->CallContext);
     requestInfo->StartedTs = ctx.Now();
 
-    // will lose original request info in case of enqueueing external request
```
neihar, line 325: Is it not true right now?

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1311,18 +1311,25 @@ void TIndexTabletActor::HandleForcedOperationStatus(
...
+    } else if (IsForcedRangeOperationPending(request.GetOperationId())) {
+        response->Record.SetStatus(TStatus::E_PENDING);
+    }
+    else {
```
neihar, line 1327: else style

cloud/filestore/libs/storage/tablet/tablet_actor_compactionforced.cpp
tdiff: not any more if I understand it correctly. `EnqueueForcedRangeOperation` now does not create a new operationid on each call, preserving the same id that was already reported to client.

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: Is it really needed?

The client is supposed to check status first. E_UNKNOWN means that the Error field cannot describe state of the operation (as it is unknown).

Otherwise, the error is only populated is status is E_FAILED, which is also kind of logical. We know status of the operation and we explain why it has failed.

cloud/filestore/libs/storage/tablet/tablet_state.h
```diff
@@ -1564,7 +1567,8 @@ FILESTORE_DUPCACHE_REQUESTS(FILESTORE_DECLARE_DUPCACHE)
...
     TMaybe<TPendingForcedRangeOperation> DequeueForcedRangeOperation();
```
neihar, line 1574: with pedning state, shouldn't we make FIFO instead or FILO in deque?

tdiff: this is a pre-existing bug in this code.

neihar: prexisting, but with pedning, behaviour can be different

neihar: agree that can be fixed separately

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
neihar: should we then check this in client? right now I don't see that we check it

cloud/filestore/libs/storage/tablet/tablet_state.h
tdiff: the behaviour of pending reuqests did not change, I only added explicit status reporting

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: fixed

cloud/filestore/libs/storage/tablet/tablet_state.h
tdiff: fixed

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: https://github.com/ydb-platform/nbs/pull/7026/changes#diff-fe817e9d263593fc74a2ebf969b63d364f28b956e4dd4a957ea0b4f7bfabdf65

neihar: at least we need to check status before CheckResponse

tdiff: After examining the CLI code a bit closer:
- Any error in ExecuteAction, regardless where it comes from, is written to statusResponse.Error. This also includes json parsing errors.
- we need to check these errors anyway, so CheckResponse() call is reasonable.

Additional handling of E_FAILED status here does not add anything to CheckResponse() and is just a dead code (assuming E_FAILED is always accompanied by an Error).

neihar (approved):

agalibin (approved):

PR #7033, opened 2026-09-04
[Filestore] implement forced ops: flush, flush_bytes, collect_garbage

tdiff:
### Notes
This change implements support of new types of forced operations on a tablet.

Before this change only "range" forced operations (cleanup, compaction, delete_empty_ranges)  were supported. The linked issue requires support of other operations such as flush, flush_bytes and collect_garbage which naturally operate on whole tablet state, not on a specific range. New operation types do not map well on existing range-based logic.

Some notes:
- "non-range" operations are called "tablet" operations
-  only one forced operation can be running at the same time. This also prevents Flush from running in parallel with other ops (which is normally allowed).

Previous version of the same PR: https://github.com/ydb-platform/nbs/pull/6918

### Issue
https://github.com/ydb-platform/nbs/issues/4115

### DIFF
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1409,23 +1409,24 @@ void TIndexTabletActor::HandleForcedOperationStatus(
 
     const auto* state = FindForcedOperation(request.GetOperationId());
     if (state) {
-        if (const auto* rangeState =
-                std::get_if<TForcedRangeOperationState>(state);
-            rangeState)
-        {
-            response->Record.SetRangeCount(rangeState->RangesToCompact.size());
-            response->Record.SetProcessedRangeCount(rangeState->Current);
-            response->Record.SetLastProcessedRangeId(
-                rangeState->GetCurrentRange());
-            response->Record.SetStatus(rangeState->Status);
-            response->Record.MutableError()->CopyFrom(rangeState->Error);
-        } else if (const auto* tabletState =
-                       std::get_if<TForcedTabletOperationState>(state);
-                   tabletState)
-        {
-            response->Record.SetStatus(tabletState->Status);
-            response->Record.MutableError()->CopyFrom(tabletState->Error);
-        }
+        std::visit(
+            TOverloaded{
+                [&](const TForcedRangeOperationState& state)
+                {
+                    response->Record.SetRangeCount(
+                        state.RangesToCompact.size());
+                    response->Record.SetProcessedRangeCount(state.Current);
+                    response->Record.SetLastProcessedRangeId(
+                        state.GetCurrentRange());
+                    response->Record.SetStatus(state.Status);
+                    response->Record.MutableError()->CopyFrom(state.Error);
+                },
+                [&](const TForcedTabletOperationState& state)
+                {
+                    response->Record.SetStatus(state.Status);
+                    response->Record.MutableError()->CopyFrom(state.Error);
+                }},
+            *state);
     } else if (IsForcedRangeOperationPending(request.GetOperationId())) {
         response->Record.SetStatus(TStatus::E_PENDING);
     } else {
--- a/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
+++ b/cloud/filestore/libs/storage/tablet/tablet_actor.cpp
@@ -1236,13 +1236,31 @@ void TIndexTabletActor::HandleForcedOperation(
     const auto& request = ev->Get()->Record;
 
     using TResponse = TEvIndexTablet::TEvForcedOperationResponse;
+    std::unique_ptr<TResponse> response;
+    switch (request.GetOpType()) {
+        case NProtoPrivate::TForcedOperationRequest::E_COMPACTION:
+        case NProtoPrivate::TForcedOperationRequest::E_CLEANUP:
+        case NProtoPrivate::TForcedOperationRequest::E_DELETE_EMPTY_RANGES:
+            response = ProcessForcedRangeOperationRequest(request, ctx);
+            break;
+        case NProtoPrivate::TForcedOperationRequest::E_FLUSH:
+        case NProtoPrivate::TForcedOperationRequest::E_FLUSH_BYTES:
+        case NProtoPrivate::TForcedOperationRequest::E_COLLECT_GARBAGE:
+            response = ProcessForcedTabletOperationRequest(request, ctx);
+            break;
+        default:
+            response = std::make_unique<TResponse>(
+                MakeError(E_ARGUMENT, "unsupported mode"));
+    }
+    NCloud::Reply(ctx, *ev, std::move(response));
+}
 
-    auto replyError = [&](NProto::TError error)
-    {
-        auto response = std::make_unique<TResponse>(std::move(error));
-        NCloud::Reply(ctx, *ev, std::move(response));
-    };
-
+std::unique_ptr<TEvIndexTablet::TEvForcedOperationResponse>
+TIndexTabletActor::ProcessForcedRangeOperationRequest(
+    const NProtoPrivate::TForcedOperationRequest& request,
+    const TActorContext& ctx)
+{
+    using TResponse = TEvIndexTablet::TEvForcedOperationResponse;
     using EMode = TEvIndexTabletPrivate::EForcedRangeOperationMode;
     EMode mode{};
     switch (request.GetOpType()) {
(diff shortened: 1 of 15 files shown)


tdiff:
Codex review:

• Found two correctness issues in 88520add3d:

  1. [P2] Forced flush-bytes reports completion before trimming finishes.
     tablet_actor_forced_tablet_operation.cpp:138 treats TEvFlushBytesResponse as completion, but the underlying handler replies before its asynchronous TTrimBytes transactions finish.
     Status can therefore become E_COMPLETED while fresh bytes remain and the flush is still running. Wait for trimming to finish before completing the forced operation. Test with
     delayed/paged trimming; the current ten-byte test doesn’t exercise this ordering reliably.

  2. [P2] Queuing a private request loses its reply destination.
     tablet_actor_forced_tablet_operation.cpp:232 saves only the mode and operation ID, discarding RequestInfo. When dequeued, the request is sent from the tablet itself, so ReplyAndDie
     suppresses the response. A private caller waiting behind another operation never receives its reply. Preserve the caller information or reject concurrent private requests with
     E_TRY_AGAIN. The queue test uses public requests and misses this.

  Additional test gap: the garbage-collection test creates no garbage, so it verifies successful no-op dispatch rather than actual collection.

tdiff:
Codex review:

• No newly introduced correctness or test-flakiness issues found in 3eb2ecb732.

  Reviewed operation dispatch, queue/state transitions, completion handling, retries, and the added tests. The flush-bytes completion policy correctly waits for an empty flush, covering
  multiple chunks and asynchronous trimming.

  Validation: 10 focused tablet tests passed, including flush, flush-bytes draining, GC, error propagation, and queued operation IDs.

  Remaining coverage gap: the added tests don’t explicitly exercise retries against a deliberately blocked background operation. A deterministic contention test would strengthen
  confidence in cleanup behavior under different event orderings.

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
```diff
@@ -1222,58 +1222,70 @@ void TIndexTabletActor::HandleForcedOperation(
...
     using TResponse = TEvIndexTablet::TEvForcedOperationResponse;
+    std::unique_ptr<TResponse> response;
+    switch (request.GetOpType()) {
+        case NProtoPrivate::TForcedOperationRequest::E_COMPACTION:
+        case NProtoPrivate::TForcedOperationRequest::E_CLEANUP:
+        case NProtoPrivate::TForcedOperationRequest::E_DELETE_EMPTY_RANGES:
+            response = ProcessForcedRangeOperationRequest(request, ctx);
+            break;
+        case NProtoPrivate::TForcedOperationRequest::E_FLUSH:
+        case NProtoPrivate::TForcedOperationRequest::E_FLUSH_BYTES:
```
neihar, line 1233: for forced flush bytes, shouldn't we respond from CompleteTx_TrimBytes? Because current reply paths reply earlier and can cause some strange effects for forced operation

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,367 @@
...
+                Config->GetCompactionRetryTimeout(),
+                std::move(requestInfo));
+            break;
+    }
+
+    auto actorId = ctx.Register(actor.release());
+    WorkerActors.insert(actorId);
+}
+
+void TIndexTabletActor::HandleForcedTabletOperationCompleted(
```
neihar, line 334: looks like it is identical to HandleForcedRangeOperationCompleted, is there a way to reduce copypaste?

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,367 @@
...
+            "got ForcedTabletOperationCompleted but no current op");
+        return;
+    }
+
+    auto* msg = ev->Get();
+    const auto* state =
+        std::get_if<TForcedTabletOperationState>(GetForcedOperationState());
+    if (!state) {
+        ReportForcedOperationUnexpectedState(
+            "got ForcedTabletOperationCompleted but current op is a range op");
```
neihar, line 349: should we return before cleaning state?

```
 WorkerActors.erase(ev->Sender);

    CompleteForcedOperation(msg->GetError());
```

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,367 @@
...
+        CreateRequestInfo(ev->Sender, ev->Cookie, msg->CallContext);
+    requestInfo->StartedTs = ctx.Now();
+
+    if (IsForcedOperationRunning()) {
+        EnqueueForcedTabletOperation(msg->Mode, std::move(msg->OperationId));
+        return;
+    }
+
+    const auto* state =
+        StartForcedTabletOperation(msg->Mode, std::move(msg->OperationId));
```
neihar, line 298: double move

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,367 @@
...
+        TFileStoreComponents::TABLET,
+        "%s ForcedTabletOperation mode=%u request",
+        LogTag.c_str(),
+        msg->Mode);
+
+    auto replyError = [&](const NProto::TError& error)
+    {
+        AbortForcedTabletOperation(
+            msg->Mode,
+            std::move(msg->OperationId),
```
neihar, line 276: double move

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,367 @@
...
+    const auto* state =
+        StartForcedTabletOperation(msg->Mode, std::move(msg->OperationId));
+    if (!state) {
+        replyError(MakeError(E_INVALID_STATE, "could not start the operation"));
+        return;
+    }
+
+    std::unique_ptr<IActor> actor;
+
+    switch (msg->Mode) {
```
neihar, line 306: default?

cloud/filestore/libs/storage/tablet/tablet_state_data.cpp
```diff
@@ -1584,38 +1585,84 @@ TString TIndexTabletState::EnqueueForcedRangeOperation(
...
+    TEvIndexTabletPrivate::EForcedTabletOperationMode mode,
+    TString operationId)
+{
+    if (ForcedOperationState.Defined()) {
+        ReportForcedOperationUnexpectedState("operation already running");
+        return nullptr;
+    }
+    TForcedTabletOperationState state(mode, std::move(operationId));
+    state.Status = NProtoPrivate::TForcedOperationStatusResponse::E_RUNNING;
+    ForcedOperationState.ConstructInPlace(std::move(state));
```
neihar, line 1648: move is not working for this class, because of default ctor, so, we need emplace or some other way

cloud/filestore/libs/storage/tablet/tablet_actor_monitoring.cpp
```diff
@@ -340,9 +340,16 @@ void DumpOperationState(
 
 void DumpCompactionInfo(
     IOutputStream& out,
-    const TIndexTabletState::TForcedRangeOperationState& state)
+    const TIndexTabletState::TForcedOperationState& state)
 {
-    DumpProgress(out, state.Current, state.RangesToCompact.size());
+    const auto* rangeOpState =
+        std::get_if<TIndexTabletState::TForcedRangeOperationState>(&state);
+    if (rangeOpState) {
```
neihar, line 347: we want to have information about new opses as well, now it will be misleading info in mon page

cloud/filestore/libs/storage/tablet/tablet_actor_forced_tablet_operation.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    FILESTORE_TRACK(
+        ResponseSent_TabletWorker,
+        RequestInfo->CallContext,
+        "ForcedTabletOperation");
+
+    if (RequestInfo->Sender != Tablet) {
+        // reply to caller
+        auto response = std::make_unique<
+            TEvIndexTabletPrivate::TEvForcedTabletOperationResponse>(error);
```
neihar, line 233: we need to add it to IngoreFunc in ZombyState as for 
```IgnoreFunc(TEvIndexTabletPrivate::TEvForcedRangeOperationProgress);```

File: cloud/filestore/libs/storage/tablet/tablet_state_data.cpp [hunk 10]
```diff
@@ -6,6 +6,7 @@
 #include <cloud/filestore/libs/storage/tablet/model/split_range.h>
 
 #include <util/generic/guid.h>
+#include "util/generic/overloaded.h"
```
neihar, line 9: icnlude style

neihar (changes requested):

cloud/filestore/libs/storage/tablet/tablet_actor.cpp
tdiff: I've spent some time trying to fix that, but turns out that even if we send response from CompleteTx_TrimBytes, there is no guarantee that a single flush_bytes opearation flushes all bytes (it only flushes one chunk).

Instead, a simplier and more reliable approach is retrying flush_bytes unless it returns S_FALSE, indicating that there are no more fresh bytes for processing.

PR #7046, opened 2026-09-04
issue-1559: [Disk Manager, Filestore] Add UsedNodesCount  to TGetFileSystemTopologyResponse

jkuradobery:
### Notes
We need UsedNodesCount to track the progress of filesystem snapshot.

### Issue
#1559


cloud/disk_manager/internal/pkg/clients/nfs/tests/client_metrics_test.go
```diff
@@ -303,6 +303,7 @@ func TestClientGetFileSystemTopologySuccess(t *testing.T) {
 				FileShardFileSystemIds:                 []string{"fs-1_file_s1"},
 				CompressNodeRef:                        true,
 				MainFileSystemId:                       "fs-1",
+				UsedNodesCount:                         42,
```
debnatkh, line 306: This is not a part of the topology, there is already a getstoragestats method for it

debnatkh (changes requested):

Issue #7022, opened 2026-09-03
[Filestore] Reduce HandleOpsQueue file size from 1GB to 100MB


budevg:
HandleOpsQueue is preallocated at 1GB per filesystem (HandleOpsQueueSize default). Target is 100MB. Two things are needed to get there without increasing the overflow rate:
- Smaller entries. The queued ConfirmCreateHandle entry stores redundant information
- Batched draining. The queue is drained one entry per scheduled tick, so under load it fills faster than it empties and capacity is what absorbs the burst. Sending confirmations in batches makes the queue short-lived, so the file no longer has to be sized for the burst.

PR #7047, opened 2026-09-05
issue-7022: drain HandleOpsQueue in batches

WilyTiger:
### Notes
Process HandleOpsQueue entries in configurable batches instead of one at a time. Requests within a batch are sent concurrently, and the entries are removed after all requests in the batch complete

### Issue
#7022


cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -503,25 +495,52 @@ void TFileSystem::ProcessHandleOpsQueue()
...
+    }
+
+    if (entries.empty()) {
         ScheduleProcessHandleOpsQueue(
             Config->GetAsyncHandleOperationIdlePeriod());
         return;
     }
 
+    TVector<TFuture<void>> futures;
+    futures.reserve(entries.size());
```
budevg, line 572: You need to preserve ordering for operations on the same handle. A queued confirm followed by destroy can now execute in reverse order: destroy removes or misses the handle, then confirm recreates it, leaving a live handle after close

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -406,16 +406,17 @@ void TFileSystem::CompleteAsyncCreateHandle(
...
-void TFileSystem::CompleteHandleOpsQueueEntry()
+void TFileSystem::CompleteHandleOpsQueueBatch(ui32 batchSize)
 {
     with_lock (HandleOpsQueueLock) {
-        HandleOpsQueue->PopFront();
+        HandleOpsQueue->PopFront(batchSize);
+    }
+
+    for (ui32 i = 0; i < batchSize; ++i) {
+        ProcessDelayedRelease();
```
budevg: Move the `batchSize` loop into `ProcessDelayedRelease()` and stop after the first false from  `ProcessAsyncRelease()`. false means `HandleOpsQueue` is full, so the remaining iterations only retry the same delayed request without any capacity becoming available.

Also can DelayedReleaseQueue remain non-empty after HandleOpsQueue becomes empty? If so, what schedules another attempt to process those delayed releases?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -888,6 +888,41 @@ class TFileRingBuffer::TImpl
...
+                if (visited == count) {
+                    return;
+                }
+            }
+
+            entry = GetNextEntry(entry);
+        }
+
+        if (entry.IsInvalid()) {
+            SetCorrupted("Invalid entry detected at VisitFirst");
```
budevg: Isn't it better to return error in case of corruption similar to `Visit()` or even make visit more generic and get maxVisitors param. If there is corruption we will constantly spam critical events without any handling of the problem since VisitFirst doesn't return error

budevg (changes requested):

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
WilyTiger: Fixed. When a queued confirm is followed later in the same batch by a destroy for the same handle, we now skip sending the confirm entirely and only send the destroy. This is safe in both cases: if the create's transaction was already durably completed on the tablet (the normal case), destroy removes it correctly, if it wasn't (i.e. tablet restart), destroy just gets S_ALREADY

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
WilyTiger: Fixed

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
WilyTiger: 1. Fixed - ProcessDelayedRelease() now returns bool, and the loop in CompleteHandleOpsQueueBatch stops on the first false, since it always retries the same head entry and further iterations after a failure were guaranteed to fail too.

2. You're right that DelayedReleaseQueue can remain non-empty after HandleOpsQueue becomes empty - there's a race between HandleOpsQueueLock and DelayedReleaseQueueLock: ProcessAsyncRelease releases HandleOpsQueueLock before returning false on overflow, and only then does the caller acquire the separate DelayedReleaseQueueLock to push. A concurrent CompleteHandleOpsQueueBatch can drain in that exact gap and  once HandleOpsQueue stays empty afterwards, nothing drains DelayedReleaseQueue. I'll fix it in a separate PR, since this problem was already present before this PR

cloud/filestore/libs/vfs_fuse/fs_ut.cpp
```diff
@@ -3307,6 +3307,147 @@ Y_UNIT_TEST_SUITE(TFileSystemTest)
...
+        features.SetAsyncDestroyHandleEnabled(true);
+        features.SetAsyncHandleOperationBatchSize(batchSize);
+        features.SetAsyncHandleOperationIdlePeriod(50);
+        features.SetAsyncHandleOperationDrainPeriod(0);
+
+        auto timer = std::make_shared<TTestTimer>();
+        auto scheduler = std::make_shared<TTestScheduler>(timer->Now());
+        TBootstrap bootstrap(timer, scheduler, features);
+
+        TVector<TPromise<NProto::TConfirmCreateHandleResponse>>
```
budevg, line 3329: let's add a case or a seperate test where open and release for same descriptor land in the same batch and skipConfirm fires. Assert no ConfirmCreateHandle is sent, the DestroyHandle is sent, and both entries are popped.

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -503,25 +502,93 @@ void TFileSystem::ProcessHandleOpsQueue()
...
+    TVector<bool> skipConfirm(entries.size(), false);
+    for (size_t i = 0; i < entries.size(); ++i) {
+        const auto& entry = entries[i];
+        if (!entry) {
+            continue;
+        }
+        if (entry->HasQueuedCreateHandleRequest()) {
+            pendingConfirmIndexByHandle[
+                entry->GetQueuedCreateHandleRequest().GetHandle()] = i;
+        } else if (entry->HasDestroyHandleRequest()) {
```
budevg: In theory due to bug we can have repeated ConfirmCreateHandle requests for the same handle and if there is Destroy Request in the same batch only the last one will be canceled. Maybe we can restructure the code loop to cover this case and it also simplifies the code

 ```cpp                                                                                                                       
   THashSet<ui64> destroyedHandles;                                                                                           
   for (const auto& entry: entries) {                                                                                         
       if (entry && entry->HasDestroyHandleRequest()) {                                                                       
           destroyedHandles.insert(entry->GetDestroyHandleRequest().GetHandle());                                             
       }                                                                                                                      
   }                                                                                                                          
 ```                                                                                                                          
                                                                                                                              
 and in the dispatch loop:                                                                                                    
                                                                                                                              
 ```cpp                                                                                                                       
   if (entry && entry->HasQueuedCreateHandleRequest()                                                                         
           && destroyedHandles.contains(                                                                                      
               entry->GetQueuedCreateHandleRequest().GetHandle()))                                                            
   {                                                                                                                          
       STORAGE_DEBUG(...);                                                                                                    
       futures.push_back(MakeFuture());                                                                                       
       continue;                                                                                                              
   }                                                                                                                          
 ```

cloud/filestore/libs/storage/tablet/tablet_actor_createsession.cpp
```diff
@@ -56,6 +56,8 @@ void FillFeatures(
         config.GetAsyncHandleOperationIdlePeriod().MilliSeconds());
     features->SetAsyncHandleOperationDrainPeriod(
         config.GetAsyncHandleOperationDrainPeriod().MilliSeconds());
+    features->SetAsyncHandleOperationBatchSize(
+        config.GetAsyncHandleOperationBatchSize());
```
budevg, line 60: if the BarchSize is not set in storage.prot it will set it to 0 here and the code in `ProcessHandleOpsQueue` will set it to 1 `Max<ui32>(1, ...)` . So we probably want something like
```cpp
   if (features.GetAsyncHandleOperationIdlePeriod()) {                                                                        
       config.SetAsyncHandleOperationIdlePeriod(                                                                              
           features.GetAsyncHandleOperationIdlePeriod());                                                                     
   }                                                                                                                          
```
to set it to default in config.cpp (32)

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -878,6 +883,38 @@ class TFileRingBuffer::TImpl
...
+
+        if (count == 0) {
+            return {};
+        }
+
+        size_t visited = 0;
+        VisitEntries(
+            [&](const TEntryInfo& e)
+            {
+                if (!e.GetFreeFlag()) {
```
budevg, line 911: We are skipping the free element  here but in the `THandleOpsQueue::PopFront` we don't try to skip already freed elements. Could it lead to some kind of bug ?

budevg (changes requested):

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
WilyTiger: THandleOpsQueue::PopFront(count) just calls the file ring buffer's own PopFront() sequentially, count times. The ring buffer's PopFront() in turn takes care of skipping already-free entries at the head itself (via Free() -> EraseFreeEntriesFromFront()), so there's nothing extra THandleOpsQueue::PopFront needs to do

budevg (approved):

yegorskii (approved):

Issue #7048, opened 2026-09-05
Create file-backed storage manager for WriteBackCache and DirectoryHandles


SvartMetal:
WriteBackCache and DirectoryHandles modules use state files in tmpfs in order to “persistently” store its state.

We need to create a manager that will handle creation/locking/deletion of module states, control its memory consumption, provide metrics and show stats on the monitoring page.

It should address the following problems:

* OOM prevention. No new state files should be created if total capacity of existing state file exceeds threshold. For example, WriteBackCache should be disabled if a state file cannot be allocated. State files should not be resized by modules themselves — they should use the manager instead.

* Reporting (and cleaning?) orphaned state files. It may happen that a session is destroyed but the state file is not removed. In this case, it will occupy memory indefinitely.

* Creation and usage of state files should be managed by modules themselves rather than external code. Example scenario: if WriteBackCache detects data corruption, in should stop and unlock the state file. But currently, the file will remain locked until the session is destroyed.

SvartMetal:
Need to add a doc which describes directory layout showing multiple session dirs, per-component dirs inside them, etc

PR #7049, opened 2026-09-05
[Filestore] vfs_fuse: introduce TPersistentStateManager which encapsulates per-session state files for write-back cache, handle ops queue and persistent directory handles storage

SvartMetal:
### Notes

`TFileSystemLoop` manages local per-session state files for three components - `HandleOpsQueue`, `WriteBackCache` and `DirectoryHandleStorage`. Each of them repeated the same routine by hand: build the `<basePath>/<fileSystemId>/<sessionId>/` path, call `CreateAndLockFile`, keep a `THolder<TFileLock>` plus an `*Initialized` flag as members, and call `UnlockAndDeleteFile` with a re-derived path on session destruction. 

This PR moves all of that into a new `TPeristentStateManager` class (`fs_persistent_state.h/.cpp`), owned by `TFileSystemLoop` and constructed from the VFS config.

Small behavior differences:
- Deleting the first component no longer wipes the others' state. The old cleanup did RemoveRecursive(<base>/<fs>/<session>) per component. With a shared directory, the first component's cleanup removed the session directory together with the state files of the other two components, from under them, while they were still held. Now only the requested file is ever removed, and the directory is removed with a non-recursive rmdir once it is empty, i.e. after the last component's file is gone. Unheld sibling files (e.g. of a component that has since been unconfigured) are never touched.
- Acquiring the same component for the same session twice now returns an `E_INVALID_STATE` error instead of silently dropping the previous lock.
- Acquiring an unconfigured component returns an error instead of hitting a debug assert.

### Issue

#7048

### DIFF
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -45,7 +45,7 @@ namespace NCloud::NFileStore{
     xxx(WriteBackCacheInitializationError)                                     \
     xxx(WriteBackCacheWritingNotAllowedInDrainingMode)                         \
     xxx(ErrorWasSentToTheGuest)                                                \
-    xxx(DirectoryHandlesStorageError)                                          \
+    xxx(DirectoryHandleStorageError)                                           \
     xxx(CalculateChecksumsBufferOverflow)                                      \
     xxx(UnexpectedFakeDescribeDataResponse)                                    \
     xxx(CreateLinkRequestWithShardNodeNameAndNoShardId)                        \
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -71,6 +71,7 @@ namespace NCloud::NFileStore{
     xxx(IncompatibleFeatures)                                                  \
     xxx(AvailabilityCountersUnavailableInterval)                               \
     xxx(AvailabilityCountersMissingIntervals)                                  \
+    xxx(PersistentStateSessionDirNotEmpty)                                     \
 // FILESTORE_CRITICAL_EVENTS
 
 #define FILESTORE_CRITICAL_EVENTS_WITHOUT_LOGGING(xxx)                         \
--- a/cloud/filestore/libs/vfs_fuse/loop.cpp
+++ b/cloud/filestore/libs/vfs_fuse/loop.cpp
@@ -6,6 +6,7 @@
 #include "fuse.h"
 #include "handle_ops_queue.h"
 #include "log.h"
+#include "persistent_state_manager.h"
 
 #include <cloud/filestore/libs/client/session.h>
 #include <cloud/filestore/libs/diagnostics/critical_events.h>
@@ -29,13 +30,10 @@
 #include <cloud/storage/core/libs/diagnostics/logging.h>
 
 #include <util/datetime/base.h>
-#include <util/folder/path.h>
 #include <util/generic/bitops.h>
 #include <util/generic/string.h>
 #include <util/generic/yexception.h>
 #include <util/system/event.h>
-#include "util/system/file_lock.h"
-#include <util/system/fs.h>
 #include <util/system/info.h>
 #include <util/system/rwlock.h>
 #include <util/system/spinlock.h>
@@ -59,49 +57,6 @@ using namespace NCloud::NFileStore::NVFS;
 namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
-
-static constexpr TStringBuf HandleOpsQueueFileName = "handle_ops_queue";
-static constexpr TStringBuf WriteBackCacheFileName = "write_back_cache";
-static constexpr TStringBuf DirectoryHandleStorageFileName = "directory_handles_storage";
-
-NProto::TError CreateAndLockFile(
-    const TString& dir,
-    const TStringBuf& fileName,
-    THolder<TFileLock>& fileLock)
-{
-    if (!NFs::MakeDirectoryRecursive(dir)) {
-        return MakeError(
-            E_FAIL,
-            TStringBuilder() << "Failed to create directories, path: " << dir);
-    }
-
-    auto filePath = TFsPath(dir) / fileName;
-    filePath.Touch();
-    fileLock = MakeHolder<TFileLock>(filePath);
-    if (!fileLock->TryAcquire()) {
-        return MakeError(
-            E_FAIL,
-            TStringBuilder() << "Failed to lock file, path: %s " << filePath);
-    }
-    return {};
-}
-
-NProto::TError UnlockAndDeleteFile(
-    const TString& dir,
-    THolder<TFileLock>& fileLock)
-{
-    fileLock->Release();
-
-    try {
-        NFs::RemoveRecursive(dir);
-    } catch (const TSystemError& err) {
-        return MakeError(
-            E_FAIL,
-            TStringBuilder() << "Failed to remove dir"
-                             << ", reason: " << err.AsStrBuf());
-    }
-    return {};
-}
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -711,6 +666,7 @@ private:
     const IProfileLogPtr ProfileLog;
     const ISessionPtr Session;
     const IFileMapMemoryLimiterPtr FileMapMemoryLimiter;
+    const TPersistentStateManagerPtr PersistentState;
 
     TLog Log;
 
@@ -725,14 +681,7 @@ private:
     TDirectoryHandleModuleStatsPtr DirectoryHandleStats;
     TFileSystemConfigPtr FileSystemConfig;
 
-    THolder<TFileLock> HandleOpsQueueFileLock;
-    THolder<TFileLock> WriteBackCacheFileLock;
-    THolder<TFileLock> DirectoryHandleStorageFileLock;
-
     TWriteBackCache WriteBackCache;
-
-    bool HandleOpsQueueInitialized = false;
-    bool DirectoryHandleStorageInitialized = false;
 
 public:
     TFileSystemLoop(
@@ -745,7 +694,8 @@ public:
             ITimerPtr timer,
             IProfileLogPtr profileLog,
             ISessionPtr session,
-            IFileMapMemoryLimiterPtr fileMapMemoryLimiter)
+            IFileMapMemoryLimiterPtr fileMapMemoryLimiter,
+            TPersistentStateManagerPtr persistentState)
         : Config(std::move(config))
         , Logging(std::move(logging))
         , StatsRegistry(std::move(statsRegistry))
@@ -756,6 +706,7 @@ public:
         , ProfileLog(std::move(profileLog))
         , Session(std::move(session))
         , FileMapMemoryLimiter(std::move(fileMapMemoryLimiter))
+        , PersistentState(std::move(persistentState))
     {
         Log = Logging->CreateLog("NFS_FUSE");
     }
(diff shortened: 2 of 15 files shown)


cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
SvartMetal, line 1: Tests

cloud/filestore/libs/vfs_fuse/fs_persistent_state.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    auto state = std::move(it->second);
+    component.AcquiredBySession.erase(it);
+
+    state.Lock->Release();
+    state.Lock.Reset();
+
+    try {
+        NFs::RemoveRecursive(state.Dir);
```
SvartMetal: It's dangerous, since right now the write-back cache, the handle ops queue, and the directory handles share the same state dir

I'll rework this

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -0,0 +1,281 @@
...
+    const TString fileName(component.FileName);
+    const auto filePath = dir / fileName;
+
+    TGuard guard(Mutex);
+
+    // Release the lock if the state file is held. Stop referencing the
+    // session directory when it holds no more state files, no matter whether
+    // the removal below succeeds.
+    NProto::TError releaseError;
+    bool dirHoldsStateFiles = false;
```
neihar: dirWithoutLocks to avoid multimple '!' ?

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1109,19 +1059,16 @@ class TFileSystemLoop final
...
             if (Config->GetDirectoryHandlesStoragePath()) {
-                auto path = TFsPath(Config->GetDirectoryHandlesStoragePath()) /
-                            FileSystemConfig->GetFileSystemId() / SessionId;
-                auto filePath = path / DirectoryHandleStorageFileName;
-
                 if (FileSystemConfig->GetDirectoryHandlesStorageEnabled()) {
-                    auto error = CreateAndLockFile(
-                        path,
-                        DirectoryHandleStorageFileName,
-                        DirectoryHandleStorageFileLock);
```
neihar, line 1120: with AcquireDirectoryHandleStorageStateFile instead of direct lock in loop, if loop will be destroyed, will unlock happen? or next loop will faile, because lock already aquired?

neihar: especially interesting for WBC, where direct release instead of delete is necessary

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -1540,38 +1484,33 @@ class TFileSystemLoop final
...
-                ReportWriteBackCacheCreatingOrDeletingError(error.GetMessage());
-            }
+        error = PersistentState->DeleteWriteBackCacheStateFile(
+            fileSystemId,
+            SessionId);
+        if (HasError(error)) {
+            ReportWriteBackCacheCreatingOrDeletingError(error.GetMessage());
         }
 
-        if (DirectoryHandleStorageInitialized) {
```
neihar, line 1567: should we delete unconditionally? looks correct in current logic, but smells like potential problem in future

cloud/filestore/libs/diagnostics/critical_events.h
```diff
@@ -45,7 +45,7 @@ namespace NCloud::NFileStore{
     xxx(WriteBackCacheInitializationError)                                     \
     xxx(WriteBackCacheWritingNotAllowedInDrainingMode)                         \
     xxx(ErrorWasSentToTheGuest)                                                \
-    xxx(DirectoryHandlesStorageError)                                          \
+    xxx(DirectoryHandleStorageError)                                           \
```
neihar, line 48: :)

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -0,0 +1,281 @@
...
+NProto::TError TPersistentStateManager::DeleteDirectoryHandleStorageStateFile(
+    const TString& fileSystemId,
+    const TString& sessionId)
+{
+    return DeleteStateFile(DirectoryHandleStorage, fileSystemId, sessionId);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+TPersistentStateManagerPtr CreatePersistentStateManagerStub()
```
neihar: should we use interface an real stub?

cloud/filestore/libs/vfs_fuse/persistent_state_manager.h
```diff
@@ -0,0 +1,140 @@
...
+        const TString& sessionId);
+
+    NProto::TError DeleteStateFile(
+        const TComponentConfig& component,
+        const TString& fileSystemId,
+        const TString& sessionId);
+
+private:
+    // Guards SessionDirs and the filesystem operations on the state files.
+    mutable TMutex Mutex;
```
neihar: will it block other FSes if operation too long?

cloud/filestore/libs/vfs_fuse/persistent_state_manager.h
```diff
@@ -0,0 +1,140 @@
...
+// components (HandleOpsQueue, WriteBackCache and DirectoryHandleStorage):
+// the on-disk layout (<basePath>/<fileSystemId>/<sessionId>/<fileName>),
+// file creation, advisory locking and cleanup.
+//
+// A single instance is shared by all filesystem loops, so it holds state files
+// of any number of filesystems and sessions at the same time.
+class TPersistentStateManager
+{
+public:
+    struct TAcquireStateFileResult
```
neihar: TResultOrError ?

SvartMetal: I think TResultOrError looks cryptic at the call site. It's better to have a named result, such as FilePath in this case

cloud/filestore/libs/vfs_fuse/persistent_state_manager.h
SvartMetal: Yes, and I think it's acceptable, because contention on (Start|Stop)Endpoint is nearly zero, and even in the bad case IO under the mutex will only increase the latency of (Start|Stop)Endpoint requests by a few tens of milliseconds.

The data plane isn't affected at all

cloud/filestore/libs/vfs_fuse/loop.cpp
SvartMetal: I think even unconditional deletion is safe, because all the state files and the session directory are associated with a session and should die with it

SvartMetal: Additionally, there's a guard in TPersistentManager that doesn't allow deletion if the state file wasn't acquired

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -756,10 +706,24 @@ class TFileSystemLoop final
...
+    {
+        // The loop may be destroyed without being stopped, e.g. when its
+        // start has failed and the endpoint is dropped. Release the locks on
+        // the state files (keeping the files) so that a new loop for the
+        // same session is able to acquire them.
+        if (FileSystemConfig && SessionId) {
+            PersistentState->ReleaseStateFiles(
+                FileSystemConfig->GetFileSystemId(),
+                SessionId);
+        }
```
SvartMetal: Will reconsider this

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
SvartMetal: Thanks, fixed

cloud/filestore/libs/vfs_fuse/loop.cpp
SvartMetal: Thanks, fixed

But I don't like the current approach, still working on it

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
SvartMetal: Thanks, fixed

SvartMetal:
AI review:
```
Reviewed current head 77b0e3e60d94fb292ebbf50614cae78649f84522, rechecked before finishing, including the description and review comments.

[P1] Failed startup permanently retains acquired locks.
persistent_state_manager.cpp:129
Locks now belong to the shared manager, which survives destruction of individual loops. Consider: handle-ops queue acquisition succeeds, then write-back cache initialization fails. The failed endpoint is discarded, but its queue lock remains registered. Even after fixing the original failure, every retry returns E_INVALID_STATE: ... already acquired. Recovery requires restarting vhost.
Previously, destroying the loop released its THolder<TFileLock> members. There is now no release-without-deletion operation.
Fix: return a lease owned by the loop that releases its locks on destruction while preserving state files. Keep deletion explicit. Test a failed start followed by a successful retry using the same manager.
```
Thanks, fixed

```
[P1] Changing the filesystem ID used for paths can skip existing persistent data.
loop.cpp:983–992
Previously, paths used FileSystemConfig->GetFileSystemId(), obtained from the server response. They now use Config->GetFileSystemId(), the requested ID. These differ when using an alias: the service explicitly resolves aliases during CreateSession.
After an upgrade, existing state under <base>/<real-id>/<session>/ is ignored while the loop opens <base>/<alias>/<session>/. Pending writes are not recovered; if WBC is disabled, the old cache is not discovered for draining either. Handle-ops and directory-handle state have the same path change.
Fix: consistently use the server-returned filesystem ID for acquisition, existence checks, and deletion. Add a restoration test where the requested and returned IDs differ.
```
Thanks, fixed

```
[P2] Recursive cleanup still deletes sibling state files that this manager has not acquired.
persistent_state_manager.cpp:189–191
dirHoldsStateFiles describes this manager’s locks, not the directory’s contents. A concrete counterexample: leave a WBC file after destroying one manager, then create another manager and call DeleteDirectoryHandleStorageStateFile() without acquiring WBC. The entire directory—including WBC—is removed, even if the directory-handle file never existed.
This is reachable when WBC’s path becomes unconfigured while directory-handle cleanup still uses the shared directory. The shared-directory concern from the existing comment is therefore only partially addressed.
Fix and simplification: remove only the requested file, then attempt a nonrecursive rmdir, treating ENOTEMPTY as normal. That also eliminates the need for directory-level lock bookkeeping solely to decide when cleanup is safe.
The tests contain useful assertions, but miss these regressions. In particular, the loop fixture creates a fresh manager per loop, masking the production lifetime problem; sibling-preservation tests acquire the siblings first.
I found no introduced out-of-bounds memory access. This was a source review; I did not build or run the C++ tests.
```
Thanks, fixed

cloud/filestore/libs/vfs_fuse/persistent_state_manager.h
```diff
@@ -0,0 +1,120 @@
...
+    struct TImpl;
+
+private:
+    THolder<TImpl> Impl;
+
+public:
+    TAcquireStateFileGuard();
+    explicit TAcquireStateFileGuard(THolder<TImpl> impl);
+    TAcquireStateFileGuard(TAcquireStateFileGuard&& other) noexcept;
+    TAcquireStateFileGuard& operator=(TAcquireStateFileGuard&& other) noexcept;
```
neihar, line 35: AI confirmed, that this move assignment can be a problem if used. It is not used right now, but theoretically

From AI:

Defaulted move-assignment corrupts the registry when the target guard holds a file. TAcquireStateFileGuard::operator=(TAcquireStateFileGuard&&) = default ([persistent_state_manager.cpp:88](vscode-webview://10pcv1vh37g6amslh5iviao593euu4rfjll643468far5s7o2luh/cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp#L88)) reduces to Impl = std::move(other.Impl), and THolder's assignment destroys the old TImpl — which has no destructor logic. The old file's flock is dropped by fd close, but UnregisterLocked() never runs, and the release happens outside the registry mutex — the exact race the guard destructor's comment defends against. The filename stays registered forever, so every later acquire of that path returns E_INVALID_STATE "already acquired" with nothing actually holding it: the same permanent-failure mode C1 just eliminated, reintroduced through one line. No current call site assigns onto a held guard (the loop members are empty at = result.ExtractResult()), so it's latent — but this class exists precisely to make that misuse impossible, and a future re-entrant StartAsync or a second component author will hit it. Fix: factor the destructor body into a private Reset() and call it from both the destructor and a hand-written move-assign. (Don't move the cleanup into ~TImpl naively — DeleteStateFile()'s local impl would then unregister a second time after the mutex is dropped, potentially erasing a concurrent re-acquisition's registration.)

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -0,0 +1,513 @@
...
+        return MakeError(
+            E_FAIL,
+            TStringBuilder() << "Failed to create directories, path: " << dir);
+    }
+
+    // Touch(), the TFileLock constructor (which opens the file) and
+    // TryAcquire() all report failures by throwing.
+    THolder<TFileLock> lock;
+    try {
+        filePath.Touch();
```
neihar, line 356: if we fail, will we leave 0 byte file?

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -0,0 +1,513 @@
...
+        return MakeError(
+            E_INVALID_STATE,
+            TStringBuilder() << "State file " << filePath
+                             << " is already acquired");
+    }
+
+    if (!NFs::MakeDirectoryRecursive(dir)) {
+        return MakeError(
+            E_FAIL,
+            TStringBuilder() << "Failed to create directories, path: " << dir);
```
neihar: LastSystemErrorText ?

neihar (dismissed):

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
SvartMetal: Yes, but I think it's better to leave it as is.

There's no "ideal" solution: even if the file was created by the current Acquire* invocation, it might have been picked up by another Acquire straight away, and in that case we have no right to delete it

cloud/filestore/libs/vfs_fuse/persistent_state_manager.h
SvartMetal: Thanks, fixed

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
SvartMetal: Fixed

cloud/filestore/libs/vfs_fuse/loop.cpp
SvartMetal: Rewritten

cloud/filestore/libs/vfs_fuse/persistent_state_manager.h
SvartMetal: This approach was used in the end

neihar (approved):

qkrorlqr (commented):
Let's also add a doc which describes directory layout showing multiple session dirs, per-component dirs inside them, etc

SvartMetal:
> Let's also add a doc which describes directory layout showing multiple session dirs, per-component dirs inside them, etc

Will be done in a separate PR

e673 (approved):

agalibin (approved):

Issue #7057, opened 2026-09-07
[NBS] Fix compaction map counters corruption when concurrent writes occurs


vladstepanyuk:
The following race was possible:

1. Start range compaction.
2. Prepare, execute, and complete the compaction transaction.
3. Receive a new WriteBlocks request for the same range.
4. Execute AddBlobs for the WriteBlocks request.
5. Execute AddBlobs for the compaction request.

As a result, the compaction map would contain counters that did not account
for the write request from step 3. This could prevent garbage from being
selected for compaction.

Previously, the compaction transaction passed the numbers of skipped blobs
and blocks to the AddBlobs transaction. The transaction then set the counters
to the numbers of skipped blobs and blocks plus the blobs and blocks produced
by compaction. Consequently, all blocks written after compaction started were
excluded from these counters.

PR #7058, opened 2026-09-07
[NBS] Account for data newer than the compaction snapshot

vladstepanyuk:
### Notes
During compaction account blobs and blocks with commit id higher than compaction commit Id in skipped blobs and blocks counters.
### Issue
#7057

### DIFF
--- a/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
@@ -876,16 +876,9 @@ void TCompactionActor::AddBlobs(const TActorContext& ctx)
             auto [blobIt, inserted] =
                 affectedBlobs.try_emplace(blobId, std::move(blob));
             if (!inserted) {
+                // In AddBlobs transaction we only looking at block masks, so we
+                // need to merge them.
                 auto& affectedBlob = blobIt->second;
-                affectedBlob.Offsets.insert(
-                    affectedBlob.Offsets.end(),
-                    blob.Offsets.begin(),
-                    blob.Offsets.end());
-                affectedBlob.AffectedBlocks.insert(
-                    affectedBlob.AffectedBlocks.end(),
-                    blob.AffectedBlocks.begin(),
-                    blob.AffectedBlocks.end());
-
                 affectedBlob.BlockMask.GetRef() |= blockMask;
             }
         }
--- a/cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
+++ b/cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
@@ -230,7 +230,7 @@ public:
         for (ui32 i = 0; i < end - start + 1; ++i) {
             const ui32 blobOffset = start - blockRange.Start + i;
             if (!skipMask.Get(blobOffset)) {
-                ab.MergedBlobsSpecificInfo->BlocksInRange++;
+                ++ab.MergedBlobsSpecificInfo->BlocksInRange;
             }
         }
 
@@ -240,10 +240,9 @@ public:
 
     // Returns blobs that do not have any blocks in the compaction range with
     // an available commit ID.
-    THashMap<TPartialBlobId, TAffectedBlob, TPartialBlobIdHash> Finish()
+    TAffectedBlobs Finish()
     {
-        THashMap<TPartialBlobId, TAffectedBlob, TPartialBlobIdHash>
-            blobsToDelete;
+        TAffectedBlobs blobsToDelete;
 
         for (auto& [blobId, ab]: Args.AffectedBlobs) {
             if (ab.MinCommitIdInCompactionRange > MaxCommitId) {
@@ -671,7 +670,8 @@ void ApplyBlobsSkipping(
     const TStorageConfig& config,
     const ui32 maxSkippedBlobs,
     TPartitionState& state,
-    TTxPartition::TRangeCompaction& args)
+    TTxPartition::TRangeCompaction& args,
+    TAffectedBlobs& skippedBlobs)
 {
     THashMap<TPartialBlobId, ui32, TPartialBlobIdHash> blobsToSkip;
     for (const auto& m: args.BlockMarks) {
@@ -688,17 +688,6 @@ void ApplyBlobsSkipping(
 
     if (state.GetMixedBlocksFilter()) {
         ApplyMixedBlocksSkipping(blobsToSkip, args);
-    }
-
-    args.BlobsSkipped += blobsToSkip.size();
-    for (const auto& [blobId, skippedBlockCount]: blobsToSkip) {
-        args.BlocksSkipped += skippedBlockCount;
-
-        auto* ab = args.AffectedBlobs.FindPtr(blobId);
-        Y_ABORT_UNLESS(ab);
-        if (ab->IndexKind == EChannelDataKind::Mixed) {
-            args.MixedBlocksSkipped += skippedBlockCount;
-        }
     }
 
     TAffectedBlocks skippedBlocks;
@@ -719,6 +708,7 @@ void ApplyBlobsSkipping(
                     {affectedBlock.BlockIndex, affectedBlock.CommitId});
             }
         }
+        skippedBlobs.emplace(ab->first, std::move(ab->second));
         args.AffectedBlobs.erase(ab);
     }
 
@@ -745,7 +735,7 @@ void ApplyBlobsSkipping(
 
 ////////////////////////////////////////////////////////////////////////////////
 
-void RecreateBlobMetas(TTxPartition::TRangeCompaction& args, ui64 commitId)
+void RecreateBlobMetas(TTxPartition::TRangeCompaction& args, ui64 commitId, ui64 tabletId)
 {
     for (auto& [blobId, ab]: args.AffectedBlobs) {
         if (ab.MergedBlobsSpecificInfo) {
@@ -766,7 +756,10 @@ void RecreateBlobMetas(TTxPartition::TRangeCompaction& args, ui64 commitId)
 
         auto& meta = ab.RecreatedBlobMeta.ConstructInPlace();
         auto* mixedBlocks = meta.MutableMixedBlocks();
-        Y_ABORT_UNLESS(ab.MixedBlobsSpecificInfo);
+        STORAGE_VERIFY(
+            ab.MixedBlobsSpecificInfo,
+            TWellKnownEntityTypes::TABLET,
+            tabletId);
         for (const auto& affectedBlock:
              ab.MixedBlobsSpecificInfo->AllVisitedBlocks)
         {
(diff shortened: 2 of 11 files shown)


cloud/blockstore/libs/storage/partition/part_ut.cpp
```diff
@@ -16697,6 +16697,83 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
 
         partition.Drain();
     }
+
```
qkrorlqr, line 16701: this is not the most trivial test so it would be nice to add comments explaining what exactly is happening and why

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -733,13 +766,83 @@ void RecreateBlobMetas(TTxPartition::TRangeCompaction& args, ui64 commitId)
 
         auto& meta = ab.RecreatedBlobMeta.ConstructInPlace();
         auto* mixedBlocks = meta.MutableMixedBlocks();
-        for (const auto& affectedBlock: ab.AffectedBlocks) {
+        Y_ABORT_UNLESS(ab.MixedBlobsSpecificInfo);
```
qkrorlqr: STORAGE_VERIFY

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -214,25 +222,40 @@ class TCompactionBlockVisitor final
...
-        ab.MergedBlobsSpecificInfo->SkippedBlocksCount = skippedBlocksCount;
+        ab.MergedBlobsSpecificInfo->SkippedBlocksCount = skipMask.Count();
+
+        const ui32 start = Max(blockRange.Start, Args.BlockRange.Start);
+        const ui32 end = Min(blockRange.End, Args.BlockRange.End);
+
+        for (ui32 i = 0; i < end - start + 1; ++i) {
+            const ui32 blobOffset = start - blockRange.Start + i;
+            if (!skipMask.Get(blobOffset)) {
+                ab.MergedBlobsSpecificInfo->BlocksInRange++;
```
qkrorlqr: nit: let's use prefix increment :)

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -214,25 +222,40 @@ class TCompactionBlockVisitor final
...
+        }
 
         ab.IndexKind = EChannelDataKind::Merged;
         return true;
     }
 
-    void Finish()
+    // Returns blobs that do not have any blocks in the compaction range with
+    // an available commit ID.
+    THashMap<TPartialBlobId, TAffectedBlob, TPartialBlobIdHash> Finish()
```
qkrorlqr: nit: mb a typedef (using) for `THashMap<TPartialBlobId, TAffectedBlob, TPartialBlobIdHash>` would look better

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -876,16 +876,9 @@ void TCompactionActor::AddBlobs(const TActorContext& ctx)
             auto [blobIt, inserted] =
                 affectedBlobs.try_emplace(blobId, std::move(blob));
             if (!inserted) {
+                // In AddBlobs transaction we only looking at block masks, so we
+                // need to merge them.
                 auto& affectedBlob = blobIt->second;
-                affectedBlob.Offsets.insert(
```
qkrorlqr, line 880: was Offsets / AffectedBlocks population moved somewhere else?

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -678,25 +701,35 @@ void ApplyBlobsSkipping(
...
     }
 
+    Sort(skippedBlocks);
+
     if (blobsToSkip.size()) {
         TAffectedBlocks affectedBlocks;
         for (const auto& b: args.AffectedBlocks) {
-            if (!skippedBlockIndices.contains(b.BlockIndex)) {
+            const bool isSkipped =
+                BinarySearch(skippedBlocks.begin(), skippedBlocks.end(), b);
```
qkrorlqr: why can't we use hash-set? if we're looking for strict equivalence

btw are you confident that we actually need to check for equivalence - I mean, that we need to include CommitId in the comparison?

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
```diff
@@ -733,13 +766,83 @@ void RecreateBlobMetas(TTxPartition::TRangeCompaction& args, ui64 commitId)
...
+    // of the range compaction. So they are visible for compaction map.
+    blobsSkipped += blobsSkippedByCommitId.size();
+    for (const auto& [blobId, ab]: blobsSkippedByCommitId) {
+        if (IsDeletionMarker(blobId)) {
+            continue;
+        }
+
+        switch (*ab.IndexKind) {
+            case EChannelDataKind::Mixed:
+
```
qkrorlqr: nit: unneeded blank line?

qkrorlqr:
claude:

> 1. Blobs skipped by incremental compaction escape the new accounting entirely. ApplyBlobsSkipping runs before AccountSkippedBlobsAndBlocks and erases every garbage-skipped blob from args.AffectedBlobs. For such a blob it only adds the BlockMark winners to BlocksSkipped — all of which are ≤ the compaction snapshot. If that same blob also holds blocks with CommitId > commitId (the exact case this PR targets), those blocks are counted nowhere: they're not in blobsSkippedByCommitId (Min ≤ snapshot) and the blob is gone before the second loop in AccountSkippedBlobsAndBlocks runs. So whenever maxSkippedBlobs > 0 and a skipped blob straddles the snapshot, the restored compaction-map counter still under-reports — #7057 persists on that path. The data is already at hand (AllVisitedBlocks is harvested right there for skippedBlocks), so counting the > commitId entries during the erase loop would close the gap.
> 
> 2. BlocksInRange for merged blobs ignores the hole mask. The new loop in the merged Visit counts every non-skip-masked position in the intersection, but FindMergedBlocks additionally skips HoleMask positions when enumerating real blocks. For a blob-patched merged blob that lands in blobsSkippedByCommitId, holes get counted as live blocks and inflate BlocksSkipped.
> 
> 3. Deletion markers are counted as skipped blobs but not blocks — verify this is intended. In the first loop of AccountSkippedBlobsAndBlocks, blobsSkipped += blobsSkippedByCommitId.size() happens before the IsDeletionMarker continue, and the second loop adds Max > commitId per blob with no marker check at all. If the per-range compaction-map blob counter doesn't count zero markers when the zero write happens, ranges with recent zeroing will have their blob count restored too high after compaction and keep getting re-selected. At minimum the blob/block treatment of markers is asymmetric and deserves a comment or a fix.

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
vladstepanyuk: no, we just not use this fields in add blobs transaction. We need to pass only blockmasks and blob metas to add blobs tx. so i removed this unnecessary lines

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
vladstepanyuk: I thought we need to provide some hash function to THashSet, to store custom structs.

I added commit id to skip Affected blocks deletion only for skipped blobs. It seems safe to me because block will be overwitten after compaction and will not be visible.

vladstepanyuk: What is codex thinks about this change:
```
• Yes—the comparison change in 3a8a7cd663 appears safe. I found no data-loss path caused by adding CommitId.

  Previously, skipping (block 0, commit 20) excluded every version of block 0 from deletion candidates. Now it excludes only that
  exact version; (block 0, commit 10) can remain a candidate.

  That is safe because:

  - Every visited version belonging to a skipped mixed blob is recorded in AllVisitedBlocks, so its exact pair remains protected.
  - Index deletion targets precisely (BlockIndex, CommitId), not all versions of a block (cloud/blockstore/libs/storage/partition/
    part_database.cpp:234).

  - Remaining candidates still pass through checkpoint and cleanup-barrier protection before deletion (cloud/blockstore/libs/
    storage/partition/part_actor_addblobs.cpp:714).

  So the change removes overly broad protection of other versions without removing protection from skipped data. This conclusion
  is based on code inspection; I haven’t run the tests.
  ```

**Comment by vladstepanyuk** [comment]
> 1. Blobs skipped by incremental compaction escape the new accounting entirely. ApplyBlobsSkipping runs before AccountSkippedBlobsAndBlocks and erases every garbage-skipped blob from args.AffectedBlobs. For such a blob it only adds the BlockMark winners to BlocksSkipped — all of which are ≤ the compaction snapshot. If that same blob also holds blocks with CommitId > commitId (the exact case this PR targets), those blocks are counted nowhere: they're not in blobsSkippedByCommitId (Min ≤ snapshot) and the blob is gone before the second loop in AccountSkippedBlobsAndBlocks runs. So whenever maxSkippedBlobs > 0 and a skipped blob straddles the snapshot, the restored compaction-map counter still under-reports — [[NBS] Fix compaction map counters corruption when concurrent writes occurs #7057](https://github.com/ydb-platform/nbs/issues/7057) persists on that path. The data is already at hand (AllVisitedBlocks is harvested right there for skippedBlocks), so counting the > commitId entries during the erase loop would close the gap.

Fixed.

> 2. BlocksInRange for merged blobs ignores the hole mask. The new loop in the merged Visit counts every non-skip-masked position in the intersection, but FindMergedBlocks additionally skips HoleMask positions when enumerating real blocks. For a blob-patched merged blob that lands in blobsSkippedByCommitId, holes get counted as live blocks and inflate BlocksSkipped.

Hole mask is deprecated and doesn't used anywhere. We already rewrite all blobs for many times since it became deprecated (during full compactions). But I can add it if you think that this is needed. 

> 3. Deletion markers are counted as skipped blobs but not blocks — verify this is intended. In the first loop of AccountSkippedBlobsAndBlocks, blobsSkipped += blobsSkippedByCommitId.size() happens before the IsDeletionMarker continue, and the second loop adds Max > commitId per blob with no marker check at all. If the per-range compaction-map blob counter doesn't count zero markers when the zero write happens, ranges with recent zeroing will have their blob count restored too high after compaction and keep getting re-selected. At minimum the blob/block treatment of markers is asymmetric and deserves a comment or a fix.

While adding zero blob we doesn't increment block counters, so we doesnt need to account for them too
https://github.com/ydb-platform/nbs/blob/55356a0c7584255deb52333b016852d43140b069/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp#L580-L589

File: cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp (hunk 6 above)
Comment (qkrorlqr) [reply cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp]: >   So the change removes overly broad protection of other versions without removing protection from skipped data. This conclusion
>  is based on code inspection; I haven’t run the tests.

this part of the diff looks like something that can be reviewed and merged separately from the main change done in this PR

File: cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp (hunk 5 above)
Comment (qkrorlqr) [reply cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp]: this part of the diff looks like something that can be reviewed and merged separately from the main change done in this PR

and it would be good to provide a short description of the initial purpose of those fields - why did we put them into affectedBlob in the first place

File: cloud/blockstore/libs/storage/partition/part_ut.cpp [hunk 8]
```diff
@@ -16697,6 +16698,81 @@ Y_UNIT_TEST_SUITE(TPartitionTest)
...
+            {
+                if (event->GetTypeRewrite() ==
+                        TEvPartitionCommonPrivate::EvWriteBlobRequest &&
+                    interceptWriteBlobRequest)
+                {
+                    writeBlobRequest.reset(event.Release());
+                }
+            });
+
+        // Intercepring write blob request to slow down compaction (as it should
```
qkrorlqr: nit: Intercepring -> Intercepting

cloud/blockstore/libs/storage/partition/part_compaction_logic.cpp
qkrorlqr: > I thought we need to **provide some hash function** to THashSet, to store custom structs.

yes, we do, but it's not a problem

qkrorlqr (dismissed):

komarevtsev-d (dismissed):

qkrorlqr (approved):

komarevtsev-d (approved):

PR #7060, opened 2026-09-07
issue-6956: extract journalled device into cloud/storage/core/libs

arigachnyy:
### Notes
Split the journalled device into two layers:

* NJournalled::IJournalledDevice (cloud/storage/core/libs/journalled_device) owns the transport-agnostic part - the log sequence number validation and the not-yet-implemented journal tail / lsn low watermark handling;
* NJournalled::IDevice is the page IO backend it is built on top of.

The disk_agent-specific code becomes a thin TDeviceAdapter over TDeviceClient, and IServerBackend now derives from IJournalledDevice instead of redeclaring its methods.

The TInstant now argument is dropped from these interfaces: the adapter takes the timestamp itself when calling the storage adapter.

### Issue

### DIFF
--- a/cloud/blockstore/libs/storage/disk_agent/journalled_device_adapter.cpp
+++ b/cloud/blockstore/libs/storage/disk_agent/journalled_device_adapter.cpp
@@ -4,6 +4,7 @@
 #include <cloud/blockstore/libs/storage/disk_agent/model/device_client.h>
 
 #include <cloud/storage/core/libs/common/error.h>
+#include <cloud/storage/core/libs/common/timer.h>
 #include <cloud/storage/core/libs/journalled_device/device.h>
 
 #include <util/generic/hash_set.h>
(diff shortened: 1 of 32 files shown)


cloud/blockstore/libs/storage/disk_agent/journalled_device_adapter.cpp
```diff
@@ -167,6 +155,7 @@ class TJournalledDevice final
         TVector<TFuture<NProto::TReadBlocksResponse>> futures;
         futures.reserve(request.PageGroupRefsSize());
 
+        auto now = TInstant::Now();
```
qkrorlqr: this will be hard to test without the ability to inject a custom timer

cloud/storage/core/libs/journalled_device/journalled_device.cpp
```diff
@@ -0,0 +1,117 @@
...
+    : public IJournalledDevice
+    , public std::enable_shared_from_this<TJournalledDevice>
+{
+private:
+    const IDevicePtr DataStore;
+
+    std::atomic<ui64> LastLsn = 0;
+
+public:
+    TJournalledDevice(IDevicePtr dataStore)
```
qkrorlqr: explicit

qkrorlqr (dismissed):

asorotsky (approved):

qkrorlqr (approved):

agalibin (approved):

PR #7062, opened 2026-09-07
Add op-based latency counters (LatencyTotalOps/LatencyGoodOps)

apliev:
## Summary

This PR adds optional threshold-based latency accounting for completed logical Read and Write operations.

The implementation covers NBD, embedded vhost, and external vhost while preserving existing request metrics and endpoint contracts.

For each volume instance, the existing `component=sli_volume` monitoring tree receives three cumulative counters:

- `LatencyGoodOps` — judged operations completed within the configured threshold.
- `LatencyTotalOps` — all judged operations, including good and bad outcomes.
- `LatencyThresholdsSkippedOps` — operations intentionally excluded from latency evaluation.

The latency SLI for a monitoring window is calculated from reset-aware counter deltas:

`LatencyGoodOps / LatencyTotalOps`

A zero `LatencyTotalOps` value means that there is no latency data for the window, not that the SLI is 100%.

## Current state

Existing latency histograms describe aggregate latency distributions, but they cannot reliably reconstruct the relationship between:

- the original request size;
- the final result after retries;
- internal request splitting;
- the threshold applicable to that particular operation.

As a result, an exact operation-based latency SLI cannot be derived from the existing metrics alone.

## Goals

- Evaluate each completed guest-visible Read or Write operation against a size-dependent threshold.
- Count one logical operation even when it is internally split or retried.
- Use the final result after the complete retry chain.
- Support different threshold ladders for Read and Write operations and for different storage media kinds.
- Keep existing metrics, error handling, retry behavior, and endpoint contracts unchanged.
- Keep latency tracking disabled by default.

## Scope and non-goals

This PR does not:

- change client-visible I/O behavior;
- change existing availability accounting;
- change existing request counters, latency histograms, or `BatchCompleted` semantics;
- add a saturation guard that excludes operations during high load;
- support dynamic threshold reloads;
- create production dashboards, alerts, or SLO configuration;
- classify operations other than Read and Write;
- turn incomplete or permanently hung operations into bad latency outcomes.

Incomplete operations require a separate in-flight or request-stall signal.

## Accounting model

Each completed logical Read or Write operation produces exactly one of the following outcomes.

### Good

An operation is good when:

- its final result is successful; and
- its cleaned execution time is less than or equal to the selected threshold.

It increments both `LatencyGoodOps` and `LatencyTotalOps`.

An operation completed exactly at the threshold is considered good.

### Bad

An operation is bad when:

- it succeeds but exceeds the threshold; or
- it finishes with a service failure that is not explicitly excluded.

This includes failures after all retries have been exhausted, including final retriable, session, and transport failures.

A bad operation increments only `LatencyTotalOps`.

There is no separate exported `LatencyBadOps` counter:

`Bad = Total - Good`

### Skipped

An operation is skipped when it cannot or should not be evaluated against a latency threshold. This includes:

- no threshold ladder configured for the volume's media kind;
- explicit throttling or checkpoint rejection;
- invalid input (`E_ARGUMENT`);
- cancellation (`E_CANCELLED`).

A skipped operation increments only `LatencyThresholdsSkippedOps` and is outside both the SLI numerator and denominator.

Operators must monitor skipped operations separately: a large skipped count means that the SLI covers only part of the traffic.

## Latency definition

The measured value is cleaned endpoint execution time, not full network or wire latency.

Where available, the implementation subtracts time recorded as:

- postponed wait;
- retry backoff;
- shaping wait.

The result is clamped at zero.

Response delivery time is not included. For external-vhost AIO, measurement starts after the request is dequeued, while existing legacy timestamps keep their previous semantics.

The original logical request size selects the threshold bucket. Internal request fragments do not select independent buckets and do not become separate SLI operations.

Read and Write operations use independent threshold ladders, but their outcomes are currently aggregated into the same three exported counters.

## Configuration

The diagnostics configuration receives two optional fields:

- `LatencyThresholdsEnabled`;
- `LatencyThresholds`.

The feature is disabled by default.

Thresholds are configured by:

- storage media kind;
- operation direction: Read or Write;
- original logical request size.

Configuration validation requires:

- an explicit media kind;
- no duplicate media-kind entries;
- at least one size bucket;
- the first size bucket to start at zero;
- strictly increasing size bounds;
- positive latency thresholds;
- no more than 1024 buckets per media kind.

The bucket limit also keeps the external-vhost environment payload within a predictable size.

Thresholds are expected to be non-decreasing as request size grows. A decreasing threshold produces a warning but is accepted.

Invalid configuration does not prevent the service from starting. Latency tracking remains disabled, and the server-level `LatencyThresholdsConfigInvalid` gauge is set.

The validated threshold table is immutable and shared for the process lifetime. Runtime reload is intentionally not implemented because threshold changes are expected to be rare. Applying a new threshold profile requires a controlled process restart.

### NBD

NBD records one logical operation after the final backend or session future completes and before the response is delivered.

The original request size is retained across retries, and unregister or cleanup paths do not record the same request twice.

### Embedded vhost

Embedded vhost records one operation at its final guest-visible completion boundary.

Shutdown cancellation is classified as cancellation for the new latency accounting without changing legacy request-statistics semantics.

### External vhost

External vhost performs classification in the child process, where the final guest-visible operation outcome is available.

The parent passes the selected media-specific threshold ladder through the versioned `NBS_VHOST_LATENCY_THRESHOLDS_CONFIG_V1` environment value.

The child collects separate Read and Write counts for:

- good;
- bad;
- skipped.

It exports them in a separate versioned `latency_counters` JSON object. The parent validates the complete payload, including its version, value types, monotonicity, and overflow bounds, before applying any values.

The parent calculates safe deltas and aggregates Read and Write results into the three public per-volume counters.

Latency outcomes are never reconstructed from legacy external-vhost request counts, error counts, or histograms. Existing JSON fields and legacy batch accounting remain unchanged.

## Compatibility and rollout

- Latency tracking is opt-in and disabled by default.
- Existing request metrics, availability counters, latency histograms, and `BatchCompleted` behavior are unchanged.
- Client-visible I/O, retries, and error codes are unchanged.
- Existing public factory entry points remain available.
- Invalid configuration fails safely by disabling latency tracking.
- Mixed parent and external-vhost child versions remain compatible for I/O and legacy metrics. A new parent with an older child simply receives no latency payload.
- Rollback consists of disabling `LatencyThresholdsEnabled` and restarting the affected processes.

Threshold profile identifiers are not included in metric labels. During a rolling restart, one time series may temporarily contain results produced with different threshold profiles. SLO windows overlapping a threshold change must therefore be excluded or treated separately.

A recommended rollout is:

1. Configure monitoring queries using reset-aware counter deltas.
2. Add monitoring for invalid configuration and skipped operations.
3. Enable the feature for a limited NBD or embedded-vhost scope.
4. Validate counter coverage and threshold behavior.
5. Expand the rollout after every process uses the same threshold profile.
6. Keep external-vhost data in shadow or pilot mode until its telemetry completeness and freshness limitations are addressed.

## External-vhost telemetry limitation

External-vhost latency telemetry is not currently lossless:

- counters are transferred as periodically collected interval deltas;
- a failed or malformed read can lose an interval;
- the current reader may stop collecting after an error;
- the final interval may be lost when the child process stops;
- there is no freshness or readiness metric;
- missing or unsupported latency payloads are ignored for compatibility.

For this reason, external-vhost counters are suitable for comparison and shadow validation but must not yet be used as the source of a contractual SLO.

### Derive the SLI from existing histograms

Rejected because existing histograms cannot preserve the exact relationship between request size, final result, retries, and internal splitting.

### Account latency in the generic request-completion path

Rejected because that path may observe internal attempts or fragments instead of the final logical operation. The new accounting uses explicit endpoint-owned completion boundaries.

### Reuse legacy external-vhost counters

Rejected because their semantics are insufficient for exact good, bad, and skipped classification. A separate versioned payload preserves the existing contract.

### Dynamically reload thresholds

Not implemented because threshold changes are expected to be rare. An immutable startup snapshot reduces complexity and avoids synchronization in the request path.

## End-to-end flow

1. The service validates the threshold configuration at startup.
2. A Read or Write operation enters its endpoint.
3. Internal fragments and retries remain part of the same logical operation.
4. The endpoint receives the final result and cleaned execution time.
5. The original request size selects the configured threshold.
6. The operation is classified as good, bad, or skipped.
7. The corresponding cumulative per-volume counters are updated.
8. Monitoring calculates the windowed ratio from reset-aware counter deltas.

The three new derivative counters add three monitoring series per active volume instance when the feature is enabled.

<details>
<summary>File summaries</summary>

| File or group | Summary |
| --- | --- |
| `cloud/blockstore/libs/diagnostics/latency_thresholds.h` *(new)* | Defines the immutable threshold-table model, size ladders, validation result, and good/total/skipped classification contract. |
| `cloud/blockstore/libs/diagnostics/latency_thresholds.cpp` *(new)* | Validates threshold tables, selects size buckets, and classifies final operations. |
| `cloud/blockstore/libs/diagnostics/latency_thresholds_config.h` *(new)* | Defines the versioned environment contract used to configure an external-vhost process. |
| `cloud/blockstore/libs/diagnostics/latency_thresholds_config.cpp` *(new)* | Serializes and parses the external-vhost threshold configuration. |
| `cloud/blockstore/vhost-server/latency_tracker.h` *(new)* | Tracks final Read and Write outcomes inside the external-vhost process. |
| `cloud/blockstore/config/diagnostics.proto`, `cloud/blockstore/libs/diagnostics/config.{h,cpp}` | Add the optional threshold configuration and expose it through the diagnostics configuration wrapper. |
| `cloud/blockstore/libs/daemon/{common,ydb}/bootstrap.cpp` | Pass validated configuration into volume statistics and external-vhost startup. |
| `cloud/blockstore/libs/diagnostics/server_stats.{h,cpp}` | Add opt-in methods for recording individual and aggregated latency outcomes while preserving existing completion methods. |
| `cloud/blockstore/libs/diagnostics/volume_stats.{h,cpp}` | Build the immutable threshold snapshot, calculate cleaned latency, and publish the per-volume counters. |
| `cloud/blockstore/libs/endpoints_vhost/external_vhost_server.{h,cpp}` | Select the volume's threshold ladder and pass it to the external child process. |
| `cloud/blockstore/libs/endpoints_vhost/external_endpoint_stats.cpp` | Validate child counters, calculate safe deltas, and import aggregated latency outcomes. |
| `cloud/blockstore/libs/nbd/server_handler.{h,cpp}` | Record one final logical NBD Read or Write using the original request size. |
| `cloud/blockstore/libs/vhost/endpoint.cpp` | Record embedded-vhost operations at the final guest-visible completion boundary. |
| `cloud/blockstore/vhost-server/backend_{aio,null,rdma}.cpp` | Add backend-specific final-completion hooks while preserving one outcome across splits and retries. |
| `cloud/blockstore/vhost-server/request_aio.{h,cpp}` | Preserve the original logical-operation timestamp through AIO preparation, splitting, rejection, and completion. |
| `cloud/blockstore/vhost-server/options.{h,cpp}` | Read the versioned threshold configuration during child startup. |
| `cloud/blockstore/vhost-server/server.cpp`, `cloud/blockstore/vhost-server/stats.{h,cpp}` | Aggregate per-queue outcomes and export the versioned latency payload. |
| `cloud/blockstore/libs/diagnostics/latency_thresholds_ut.cpp`, `cloud/blockstore/libs/diagnostics/volume_stats_ut.cpp`, `cloud/blockstore/libs/diagnostics/server_stats_test.h` | Cover validation, bucket boundaries, classification, cleaned latency, counter lifecycle, and test hooks. |
| `cloud/blockstore/libs/endpoints_vhost/external_endpoint_stats_ut.cpp`, `cloud/blockstore/libs/nbd/server_handler_ut.cpp`, `cloud/blockstore/libs/vhost/server_ut.cpp` | Cover external aggregation and NBD/embedded-vhost integration. |
| `cloud/blockstore/vhost-server/{options,request_aio,stats}_ut.cpp` | Cover child configuration parsing, logical AIO accounting, and statistics export. |
| `cloud/blockstore/libs/diagnostics/{ya.make,ut/ya.make}`, `cloud/blockstore/vhost-server/{ya.make,ut/ya.make,gtest/ya.make}` | Register the new implementation, dependencies, and tests. |

</details>

## Verification

The affected targets were rebuilt and rerun on Linux with a RelWithDebInfo build:

- `cloud/blockstore/libs/diagnostics/ut`
- `cloud/blockstore/libs/nbd/ut`
- `cloud/blockstore/libs/vhost/ut`
- `cloud/blockstore/libs/endpoints_vhost/ut`
- `cloud/blockstore/vhost-server/ut`
- `cloud/blockstore/vhost-server/gtest`

Results:

- unit tests: 221/221 passed;
- vhost-server gtests: 662/662 passed;
- total: 883/883 passed.

### DIFF
--- a/cloud/blockstore/config/diagnostics.proto
+++ b/cloud/blockstore/config/diagnostics.proto
@@ -6,6 +6,7 @@ option go_package = "github.com/ydb-platform/nbs/cloud/blockstore/config";
 
 import "cloud/storage/core/config/opentelemetry_client.proto";
 import "cloud/storage/core/protos/diagnostics.proto";
+import "cloud/storage/core/protos/media.proto";
 import "cloud/storage/core/protos/trace.proto";
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -117,6 +118,36 @@ message TMonitoringUrlData
     // - diskId
     optional string MonitoringUrlTemplate = 10;
 
+};
+
+////////////////////////////////////////////////////////////////////////////////
+// Operation latency thresholds (op-based latency metric, see
+// TDiagnosticsConfig.LatencyThresholdsEnabled/LatencyThresholds below).
+
+// A single operation-size bucket: the lower bound (inclusive) of the
+// operation size range together with the execution time thresholds for read
+// and write operations of that size. An operation is considered timely if
+// it completes within the threshold configured for its size class. The
+// upper bound of a bucket is the next bucket's MinRequestBytes (open-ended
+// for the last bucket of a media kind).
+message TLatencyThresholdBucket
+{
+    // Lower bound of the operation size range, in bytes, inclusive. The
+    // first bucket of every media kind must have MinRequestBytes == 0, so
+    // that no operation size falls outside of every bucket.
+    optional uint64 MinRequestBytes = 1;
+
+    // Execution time thresholds, in milliseconds. Must be > 0.
+    optional uint32 ReadThresholdMs = 2;
+    optional uint32 WriteThresholdMs = 3;
+};
+
+// Latency threshold ladder for a single storage media kind. Buckets must be
+// sorted strictly ascending by MinRequestBytes, with no duplicates.
+message TMediaKindLatencyThresholds
+{
+    optional NCloud.NProto.EStorageMediaKind MediaKind = 1;
+    repeated TLatencyThresholdBucket Buckets = 2;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -297,4 +328,16 @@ message TDiagnosticsConfig
     // Selects the critical event metrics to report.
     optional EVolumeCriticalEventsReportingMode
         VolumeCriticalEventsReportingMode = 60;
+
+    // Enables the operation latency threshold mechanism (LatencyTotalOps/
+    // LatencyGoodOps counters, see volume_stats.cpp). Disabled by default.
+    // When enabled with an empty or invalid LatencyThresholds table, the
+    // mechanism stays disabled (see the LatencyThresholdsConfigInvalid
+    // diagnostic gauge) instead of crashing on a config mistake.
+    optional bool LatencyThresholdsEnabled = 61;
+
+    // Per-media-kind operation size -> execution time threshold ladders.
+    // Empty by default. See TMediaKindLatencyThresholds/
+    // TLatencyThresholdBucket above for the validation rules.
+    repeated TMediaKindLatencyThresholds LatencyThresholds = 62;
 }
--- a/cloud/blockstore/libs/daemon/common/bootstrap.cpp
+++ b/cloud/blockstore/libs/daemon/common/bootstrap.cpp
@@ -385,7 +385,8 @@ void TBootstrapBase::Init()
             Configs->DiagnosticsConfig,
             inactiveClientsTimeout,
             EVolumeStatsType::EServerStats,
-            Timer);
+            Timer,
+            Log);
     }
 
     ServerStats = CreateServerStats(
--- a/cloud/blockstore/libs/daemon/common/bootstrap.cpp
+++ b/cloud/blockstore/libs/daemon/common/bootstrap.cpp
@@ -550,16 +550,18 @@ void TBootstrapBase::Init()
         if (Configs->ServerConfig->GetVhostServerPath()
                 && !Configs->Options->TemporaryServer)
         {
-            vhostEndpointListener = CreateExternalVhostEndpointListener(
-                Configs->ServerConfig,
-                Logging,
-                ServerStats,
-                Executor,
-                Configs->Options->SkipDeviceLocalityValidation
-                    ? TString {}
-                    : FQDNHostName(),
-                RdmaClient && RdmaClient->IsAlignedDataEnabled(),
-                std::move(vhostEndpointListener));
+            vhostEndpointListener =
+                CreateExternalVhostEndpointListenerWithDiagnostics(
+                    Configs->ServerConfig,
+                    Logging,
+                    ServerStats,
+                    Executor,
+                    Configs->Options->SkipDeviceLocalityValidation
+                        ? TString {}
+                        : FQDNHostName(),
+                    RdmaClient && RdmaClient->IsAlignedDataEnabled(),
+                    std::move(vhostEndpointListener),
+                    Configs->DiagnosticsConfig);
 
             STORAGE_INFO("VHOST External Vhost EndpointListener initialized");
         }
--- a/cloud/blockstore/libs/daemon/common/bootstrap.cpp
+++ b/cloud/blockstore/libs/daemon/common/bootstrap.cpp
@@ -388,6 +388,10 @@ void TBootstrapBase::Init()
             Timer,
             Log);
     }
+    // Some server variants construct VolumeStats while their monitoring
+    // proxy is still deferred. At this point monitoring is ready, so publish
+    // server-level diagnostics before the first volume is mounted.
+    VolumeStats->InitializeMonitoringCounters();
 
     ServerStats = CreateServerStats(
         Configs->ServerConfig,
--- a/cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
+++ b/cloud/blockstore/libs/daemon/ydb/bootstrap.cpp
@@ -615,7 +615,8 @@ void TBootstrapYdb::InitKikimrService()
         Configs->DiagnosticsConfig,
         Configs->StorageConfig->GetInactiveClientsTimeout(),
         EVolumeStatsType::EServerStats,
-        Timer);
+        Timer,
+        Log);
 
     ClientPercentiles = CreateClientPercentileCalculator(logging);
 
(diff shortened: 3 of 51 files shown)


nisqatsi:
> ## Latency definition
> 
> The measured value is cleaned endpoint execution time, not full network or wire latency.

Тут самый большой вопрос (либо не совсем понятно написано). Выше в **Goals** было указано:
> Evaluate each completed guest-visible Read or Write operation against a size-dependent threshold.

Т.е. необходимо рассчитывать latency, видимую для пользователя, а вот это - "not full network or wire latency" - вроде как противоречит этой цели (рассчитываем только latency внутри процессов NBS, без учета влияния сети и пр.?).

> Where available, the implementation subtracts time recorded as:
> 
> - postponed wait;
> - retry backoff;
> - shaping wait.

Тут бы пояснить немного, что они конкретно означают. По-хорошему, мы должны выбрасывать только время throttling, если запрос тротлится (т.е. пользователь превышает выделенную ему квоту).

> Response delivery time is not included. For external-vhost AIO, measurement starts after the request is dequeued, while existing legacy timestamps keep their previous semantics.

Не совсем понятно, что входит в "Response delivery time". Применительно к Write воспринимается как время, за которое NBS выполнил запрошенную операцию, но еще не вернул (не доставил) ответ об успешном выполнении пользователю.

> ## Configuration
> 
> Thresholds are configured by:
> 
> - storage media kind;
> - operation direction: Read or Write;
> - original logical request size.

Тут бы сразу немного пояснить, что "original logical request size" - это интервалы (в бакетах гистограммы).

PR #7064, opened 2026-09-07
issue-1559: [Disk Manager] Add GetStorageStats() to filestore client

jkuradobery:
### Notes
Need storage stats for metrics

### Issue
#1559

### DIFF
--- a/cloud/filestore/private/api/protos/ya.make
+++ b/cloud/filestore/private/api/protos/ya.make
@@ -1,6 +1,5 @@
 PROTO_LIBRARY(filestore-private-api-protos)
 
-EXCLUDE_TAGS(GO_PROTO)
 EXCLUDE_TAGS(JAVA_PROTO)
 
 PEERDIR(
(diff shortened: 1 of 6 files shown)


cloud/filestore/private/api/protos/ya.make
```diff
@@ -1,6 +1,5 @@
 PROTO_LIBRARY(filestore-private-api-protos)
 
-EXCLUDE_TAGS(GO_PROTO)
```
gy2411, line 3: Why do we need to remove this?

jkuradobery: Because we want to use the protobufs in go code.
Before that we used unsafe protos file, but @qkrorlqr removed ydbd dependency from tablet.proto and we can no use private/tablet.proto.

jkuradobery: I will migrate the remaining protobufs later.

gy2411 (approved):

debnatkh (approved):

PR #7065, opened 2026-09-07
Automatic certificates update for dm

yegorskii:
### Notes
- Add automatic TLS certificate refresh for Disk Manager gRPC clients and server, controlled by a new `RefreshCertsPeriod` config field (string duration, default `"0s"` = disabled) in the NBS client config, the NFS client config and the server `GrpcConfig`. The field name matches the blockstore/filestore server configs.

- Client TLS provider re-reads the root certificates file periodically. On success the `tls.Config` is swapped and the root CA `fingerprint` gauge is updated. New connections pick up the new roots automatically because the client transport credentials request a config snapshot on every handshake.

- Server TLS provider re-reads every cert/key pair independently. A refreshed chain is accepted only if every certificate in it is valid at refresh time and the chain can be built from the leaf up to the last certificate in the file via `x509.Verify` (issuer names, signatures, CA and name constraints). Whether the chain ends at a trusted root is not checked: the server has no trust store, and that is the client's job. Initial load does not validate chains so that the service is able to start, there is no last known good certificate to fall back to. `expireTs` and `certificateValidity` gauges are recomputed right after a successful refresh; the existing 24h `certificateValidity` ticker is kept.

- Stable-read: files are checked every `RefreshCertsPeriod`. New content is re-checked after half a period and applied only if it is unchanged, a read error restarts the count, so a change takes effect within one and a half periods. This is a heuristic that reduces the chance of picking up an intermediate state of a non-atomic rewrite, not a guarantee.

- Raw cert and key bytes are compared before any parsing, so an unchanged file costs a read and a comparison per tick, and an unchanged expired certificate does not produce a warning on every tick. New content that fails parsing or validation is reported on every tick until the file changes; the last successfully loaded certificate is kept in all failure cases.

- Certificate loading errors are plain `fmt.Errorf`, not `NonRetriableError`: the latter embeds a stack trace into `Error()` and would be logged on every failed tick.

- Reject negative refresh periods at provider construction.

### Issue


yegorskii:
Codex Review:

> Reviewed the latest version of the PR.
> 
> The current implementation looks good to me and I don't see any blocking correctness issues.
> 
> I specifically re-checked the following parts:
> 
> * The stable-read logic is implemented consistently: new content must be observed twice, read failures reset the pending state, and the mechanism is correctly described as a heuristic rather than a consistency guarantee.
> * Server certificate chains are validated with `x509.Verify`, including issuer relationships, signatures, CA constraints, and validity intervals, while deliberately treating the last certificate from the supplied chain as the trust anchor.
> * Both client and server preserve the last known good TLS configuration when refreshed material cannot be read, parsed, or validated.
> * Comparing raw contents before parsing avoids unnecessary parsing and metric updates for unchanged files.
> * The adaptive polling scheme (`RefreshCertsPeriod` normally, `RefreshCertsPeriod / 2` while waiting for the second stable read) avoids reading unchanged files twice as often while still applying detected changes within one and a half periods.
> * Using plain errors for refresh failures avoids repeatedly logging embedded stack traces.
> * I don't see strict validation of trailing malformed PEM blocks as a requirement for this change. The previous implementation already relied on the standard Go TLS/X.509 parsing semantics, while stable-read addresses the actual new problem introduced by runtime refresh: avoiding transient states during certificate-file rewrites. Persistently malformed certificate material should be treated as an error in the certificate distribution pipeline.
> 
> LGTM.

Issue #6957, opened 2026-08-29
[Filestore] fastshard - storage group


qkrorlqr:
Main issue: https://github.com/ydb-platform/nbs/issues/5895

Implement the storage group layer described here: https://github.com/ydb-platform/nbs/blob/main/doc/filestore/design/storage/fastshard/storage-group.md

The first version can be tested with 2/3 write + 2/3 read quorum or 3/4 write + 2/4 read quorum

PR #7071, opened 2026-09-07
issue-6957: proper storage group initialization and journal restoration

asorotsky:
### Notes
- added init/deinit routines for storage group
- resyncing journal across devices upon init of a storage group,  
- async wait for fastdatashard init upon tablet bootstrap
- properly bootstrapping fastdatashard across stack.

### Issue
https://github.com/ydb-platform/nbs/issues/6957

### DIFF
--- a/cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_quorum.cpp
+++ b/cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_quorum.cpp
@@ -262,7 +262,8 @@ int WriteDispatchFiberMain(TWriteDispatchParams* params) noexcept
 
 ////////////////////////////////////////////////////////////////////////////////
 
-using TJournalRecords = google::protobuf::RepeatedPtrField<NProto::TJournalRecord>;
+using TJournalRecords =
+    google::protobuf::RepeatedPtrField<NProto::TJournalRecord>;
 
 ui64 GetLastLsn(const NProto::TReadJournalTailResponse& journal)
 {
@@ -302,21 +303,21 @@ int ReplayFiberMain(TReplayParams* params) noexcept
 
 NProto::TError ReplayOnDevicesBehind(
     const TGroupStatePtr& state,
-    const TVector<ui64>& lsns,
+    const TVector<ui64>& maxLsnPerDevice,
     ui64 maxKnownLsn,
     const TJournalRecords& records)
 {
     TVector<ui32> proxyIndexesToReplay;
     TVector<ui32> journalPositionsToStartReplay;
-    for (ui32 i = 0; i < lsns.size(); ++i) {
-        if (lsns[i] == maxKnownLsn) {
+    for (ui32 i = 0; i < maxLsnPerDevice.size(); ++i) {
+        if (maxLsnPerDevice[i] == maxKnownLsn) {
             continue;
         }
 
         const auto it = std::upper_bound(
             records.begin(),
             records.end(),
-            lsns[i],
+            maxLsnPerDevice[i],
             [](ui64 lsn, const NProto::TJournalRecord& record)
             {
                 return lsn < record.GetLogSequenceNumber();
@@ -330,7 +331,7 @@ NProto::TError ReplayOnDevicesBehind(
                 TStringBuilder()
                     << "journal tail does not continue "
                     << state->Proxies[i]->DeviceUUID << " from lsn "
-                    << lsns[i]);
+                    << maxLsnPerDevice[i]);
         }
 
         proxyIndexesToReplay.push_back(i);
@@ -360,7 +361,8 @@ NProto::TError ReplayOnDevicesBehind(
             error = MakeError(
                 errors[i].GetCode(),
                 TStringBuilder()
-                    << "replay onto " << state->Proxies[proxyIndexesToReplay[i]]->DeviceUUID
+                    << "replay onto "
+                    << state->Proxies[proxyIndexesToReplay[i]]->DeviceUUID
                     << " failed: " << FormatError(errors[i]));
         }
     }
@@ -389,7 +391,8 @@ NProto::TError ReadRecordsAbove(
             E_INVALID_STATE,
             TStringBuilder()
                 << "journal of " << source.DeviceUUID
-                << " does not reach from lsn " << afterLsn << " to " << maxKnownLsn);
+                << " does not reach from lsn " << afterLsn << " to "
+                << maxKnownLsn);
     }
 
     return {};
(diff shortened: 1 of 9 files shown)


asorotsky:
claude review

Review of the combined commit, all files reread against the rebased base.

  Correctness and invariants

  - Watermark future lifecycle is sound. Set in the constructor, reset immediately before run, completed by the fiber; Deinit waits unconditionally and returns at once for a group whose Init failed or never ran. reset is called with no waiter and an unconsumed zero, within its contract.
  - Arithmetic cannot underflow. pushed is always a previous acked minus the history, so acked - pushed is non-negative, and acked - history only runs when acked > pushed + history.
  - The fiber bridge's Init and Deinit mirror the other forwarded methods: promise set from the fiber, spawn failure reported through the same future, Deinit detached with the implementation kept alive by the params. The naive shard is a constructor again.
  - No out-of-bounds risk found. No memcpy, memmove or memset in the change; every indexed access into recorded calls is guarded by a size check first; the rebuild's parallel vectors are sized together; rbegin is guarded by the empty check.
  - No exceptions introduced. Promises are set exactly once on every path; everything else returns TError. No custom move assignment anywhere; the non-movable futures live in vectors sized once and are joined before their frame ends.

  Threading

  - Shared state is reached through one shared_ptr; mutable fields are atomics or sequencers; the health object serialises under a fiber mutex. The test shard's plain bool and the factory's vector are touched only on the single-threaded actor test runtime.
  - Known window, unchanged: the shard's Deinit does not wait for a pending Init, so a tablet dying mid-init can release before the acquire lands. Bounded and accepted earlier; noting it once more because the bridge now owns both spawns.

  Tests

  - The tick timer handshake is deterministic: each TickOnce returns only when the fiber is parked again, so every assertion after it sees a completed iteration. TickUntilPushed absorbs the one real race, the third ack being counted a moment after the fake records the write; the zero-push assertions use a single
    tick because nothing can move in those states.
  - Stop-before-Deinit ordering is right and needed. Stopping the timer first frees the parked fiber; Deinit then flips Stopped and joins. Between the two calls the woken fiber can spin for a few instructions on another worker if its Sleep now returns at once; the test fiber does not yield in that gap, so it is
    bounded and harmless. The reverse order would deadlock, and the comment says so.
  - Each test reproduces what it claims: the slowest device holding the watermark, then the push at 1233 and no repeat; history of three withheld at exactly three and trailing by three above it; zero history pushing the acked LSN as is. The adapter tests observe the wait, the restart with a second shard, and the
    release on death.
  - Minor: if Init failed inside TQuorumFixture, the EXPECT fires but a later TickOnce would wait forever for a fiber that never started, turning a failure into a hang. Acceptable for a fixture that asserts success, worth knowing.

  Complexity

  Nothing left to take out that would not cost clarity: the test shard implements the interface through the same macro list the stub uses, the factory owns implementation selection, the tablet holds an interface pointer, and the group's watermark path is one loop and one future.

  Nits: the constructor comment "Will reset upon init call;" has a stray semicolon; the TODO: set client id and the untracked recovery.md still stand from earlier reviews.

  Nothing here needs a code change before merging, apart from running sn/quorum/ut and impl/naive_mirrored/ut where the fiber scheduler is allowed.

arigachnyy (dismissed):

cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_ut.cpp
```diff
@@ -643,14 +772,31 @@ void WaitUntilServes(TStorageFixture& fx, ui32 i)
...
+
 }   // namespace
 
 TEST(QuorumGroupTest, InitAcquiresEveryDeviceWithTheGeneration)
 {
     const int r = FiberScheduler::run(
         +[](int*) noexcept -> int
         {
-            TStorageFixture fx(CreateQuorumMirroredStorageGroup);
+            TQuorumFixture fx(false);
```
qkrorlqr: nit: `false /* init */`

cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_ut.cpp
```diff
@@ -107,25 +113,148 @@ struct TStorageFixture
...
+
+    TInstant Now() override
+    {
+        return TInstant::Now();
+    }
+
+    void Sleep(TDuration duration) override
+    {
+        Y_UNUSED(duration);
+        (void)Ticks.wait(Sleeps.increment());
```
qkrorlqr: nit:
```
int r = Ticks.wait(Sleeps.increment());
Y_UNUSED(r);
```

cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_ut.cpp
```diff
@@ -699,6 +851,445 @@ TEST(QuorumGroupTest, InitFailsIfAnyDeviceRefusesAcquire)
...
+            fx.StorageNodes[2]->ReadJournalTailResp = JournalTail({7});
+
+            auto error = fx.Group->Init();
+            EXPECT_EQ(S_OK, error.GetCode()) << error.GetMessage();
+
+            EXPECT_EQ((TVector<ui64>{8, 9, 10}), WrittenLsns(fx, 2));
+            if (WrittenLsns(fx, 2).size() != 3) {
+                return 1;
+            }
+            const auto& replay = fx.StorageNodes[2]->WriteCalls[0];
```
qkrorlqr: why do we check only `WriteCalls[0]`? let's check `WriteCalls[1, 2]` as well

cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_quorum.cpp
```diff
@@ -185,15 +244,331 @@ int WriteDispatchFiberMain(TWriteDispatchParams* params) noexcept
...
+    if (params->Op->Acks.increment() == state.Proxies.size()) {
+        state.LowWatermarkLsn.advance(params->Op->Lsn);
+    }
+
+    return 0;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+using TJournalRecords = google::protobuf::RepeatedPtrField<NProto::TJournalRecord>;
```
qkrorlqr: nit: > 80 symbols

cloud/filestore/libs/storage/fastshard/sn/quorum/storage_group_quorum.cpp
```diff
@@ -185,15 +244,331 @@ int WriteDispatchFiberMain(TWriteDispatchParams* params) noexcept
...
+            break;
+        }
+    }
+
+    return 0;
+}
+
+NProto::TError ReplayOnDevicesBehind(
+    const TGroupStatePtr& state,
+    const TVector<ui64>& lsns,
```
qkrorlqr: let's rename this to something like `maxLsnPerDevice`

qkrorlqr (commented):
overall LGTM

arigachnyy (approved):

qkrorlqr (approved):

yegorskii (approved):

Issue #7053, opened 2026-09-05
[Filestore] create control namespace for managing filestore settings from inside the filestore


debnatkh:
The main idea is to have a special directory under the root (`/`) that will expose some special files to manage the filesystem settings/read its settings directly

Will be useful in #6608, when we will want to expose some kind of inside-of-filestore API for quota management

PR #7072, opened 2026-09-07
issue-7053: hidden subdirectory with fsid special file

debnatkh:
### Notes
Example of it working:

```
$ mkdir sanity-dir && echo hi > sanity-dir/file.txt && cat sanity-dir/file.txt
hi

$ ls .filestore-ctl
fsid

$ cat .filestore-ctl/fsid
nfs

$ echo x > .filestore-ctl/fsid
permission denied

$ mkdir .filestore-ctl/newdir
Permission denied

$ mv sanity-dir/file.txt .filestore-ctl/file.txt
Permission denied

$ ln .filestore-ctl/fsid sanity-dir/link
Permission denied

$ cat .filestore-ctl
Is a directory

$ rm -rf sanity-dir
```

### Issue
#7053
#6608

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
@@ -1,19 +1,202 @@
 #include "service_actor.h"
 
+#include "service_actor_control_namespace.h"
+
+#include <cloud/filestore/libs/service/error.h>
+#include <cloud/filestore/libs/service/filestore.h>
+
+#include <cloud/storage/core/libs/common/error.h>
+
 namespace NCloud::NFileStore::NStorage {
 
 using namespace NActors;
 
 ////////////////////////////////////////////////////////////////////////////////
-// Stubs - no control namespace logic yet, every hook is a no-op.
+
+EControlNamespaceEntry ClassifyControlNamespaceEntry(ui64 nodeId)
+{
+    if (nodeId == ControlDirIno) {
+        return EControlNamespaceEntry::ControlDir;
+    }
+    if (nodeId == ControlFsIdFileIno) {
+        return EControlNamespaceEntry::FsId;
+    }
+    return EControlNamespaceEntry::None;
+}
+
+EControlNamespaceEntry ClassifyControlNamespaceEntry(
+    ui64 parentId,
+    TStringBuf name,
+    TStringBuf controlNamespaceDirName)
+{
+    if (parentId == RootNodeId && name == controlNamespaceDirName) {
+        return EControlNamespaceEntry::ControlDir;
+    }
+    if (parentId == ControlDirIno) {
+        return name == ControlFsIdFileName ? EControlNamespaceEntry::FsId
+                                           : EControlNamespaceEntry::Unknown;
+    }
+    if (parentId == ControlFsIdFileIno) {
+        // fsid is a file - it has no children, but be defensive.
+        return EControlNamespaceEntry::Unknown;
+    }
+    return EControlNamespaceEntry::None;
+}
+
+void FillControlDirAttr(NProto::TNodeAttr& attr)
+{
+    attr.SetId(ControlDirIno);
+    attr.SetType(NProto::E_DIRECTORY_NODE);
+    attr.SetMode(0555);
+    attr.SetLinks(2);
+    attr.SetSize(0);
+}
+
+void FillControlFsIdAttr(NProto::TNodeAttr& attr, const TString& fileSystemId)
+{
+    attr.SetId(ControlFsIdFileIno);
+    attr.SetType(NProto::E_REGULAR_NODE);
+    attr.SetMode(0444);
+    attr.SetLinks(1);
+    attr.SetSize(fileSystemId.size());
+}
+
+NProto::TError ControlNamespaceReadOnlyError()
+{
+    return MakeError(E_FS_ACCESS, "control namespace files are read-only");
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// BuildControlNamespaceResponse specializations, for the self-lookup-by-ino
+// forms reached via ForwardRequestToShard rather than a dedicated
+// TryHandleControlNamespaceXxx above.
+
+template <>
+std::unique_ptr<TEvService::TGetNodeAttrMethod::TResponse>
+BuildControlNamespaceResponse<TEvService::TGetNodeAttrMethod>(
+    const TEvService::TGetNodeAttrMethod::TRequest::TPtr& ev,
+    const TString& fileSystemId)
+{
+    const auto& record = ev->Get()->Record;
+    auto response =
+        std::make_unique<TEvService::TGetNodeAttrMethod::TResponse>();
+    switch (ClassifyControlNamespaceEntry(record.GetNodeId())) {
+        case EControlNamespaceEntry::ControlDir:
+            FillControlDirAttr(*response->Record.MutableNode());
+            break;
+        case EControlNamespaceEntry::FsId:
+            FillControlFsIdAttr(*response->Record.MutableNode(), fileSystemId);
+            break;
+        default:
+            *response->Record.MutableError() =
+                MakeError(E_FS_NOENT, "not found");
+            break;
+    }
+    return response;
+}
+
+template <>
+std::unique_ptr<TEvService::TCreateHandleMethod::TResponse>
+BuildControlNamespaceResponse<TEvService::TCreateHandleMethod>(
+    const TEvService::TCreateHandleMethod::TRequest::TPtr& ev,
+    const TString& fileSystemId)
+{
+    const auto& record = ev->Get()->Record;
+    auto response =
+        std::make_unique<TEvService::TCreateHandleMethod::TResponse>();
+
+    const auto entry = ClassifyControlNamespaceEntry(record.GetNodeId());
+    if (entry != EControlNamespaceEntry::ControlDir &&
+        entry != EControlNamespaceEntry::FsId)
+    {
+        *response->Record.MutableError() = MakeError(E_FS_NOENT, "not found");
+        return response;
+    }
+
+    const bool wantsWrite = HasFlag(
+        record.GetFlags(),
+        ProtoFlag(NProto::TCreateHandleRequest::E_WRITE));
+    if (wantsWrite) {
+        *response->Record.MutableError() = ControlNamespaceReadOnlyError();
+        return response;
+    }
+
+    if (entry == EControlNamespaceEntry::ControlDir) {
+        response->Record.SetHandle(ControlDirIno);
+        FillControlDirAttr(*response->Record.MutableNodeAttr());
+    } else {
+        response->Record.SetHandle(ControlFsIdFileIno);
+        FillControlFsIdAttr(*response->Record.MutableNodeAttr(), fileSystemId);
+    }
+    return response;
+}
+
+template <>
+std::unique_ptr<TEvService::TConfirmCreateHandleMethod::TResponse>
+BuildControlNamespaceResponse<TEvService::TConfirmCreateHandleMethod>(
+    const TEvService::TConfirmCreateHandleMethod::TRequest::TPtr& ev,
+    const TString& fileSystemId)
+{
+    Y_UNUSED(ev);
+    Y_UNUSED(fileSystemId);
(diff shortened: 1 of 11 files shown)


debnatkh:
_**Model: Claude Sonnet 5.**_

---

I got the PR description but not the actual code diff. Let me pull the raw diff to review the implementation.Got it. Let me get the full diff.Now the main file — the largest and most important one for review.Good, this confirms the wiring already existed (stubs) and this PR fills in the real logic. Let me check the `ShardedId`/`ExtractShardNo` implementation and skim the test file for coverage quality.Good coverage. Let me check the trickier tests to confirm expected behaviors (fstat/self-lookup via handle, disabled config, invalid session, offset reads) and skim key snippets.I pulled the actual diff (the GitHub UI doesn't render it via fetch, so I cloned the repo and diffed `main...pr-7072` directly). Here's my review of #7072.

## What it does

This fills in the previously-stubbed `TryHandleControlNamespaceXxx` hooks in the filestore service actor to implement a synthetic, read-only `.filestore-ctl/fsid` path: a hidden directory containing one file that reports the current filesystem's ID. Six files touched, ~900 lines added, mostly in `service_actor_control_namespace.cpp` plus a solid new test suite (`service_ut_control_namespace.cpp`).

## Design — the good parts

- **Clever collision-free ino scheme.** It reserves the shard number one past `MaxShardCount` (`ControlNamespaceShardNo = MaxShardCount + 1`, which equals `Max<ui16>()`, so it fits the 16-bit shard field exactly) and encodes the control dir/file as `ShardedId(1|2, ControlNamespaceShardNo)`. Since real shards are capped at `MaxShardCount`, these inos can never collide with real nodes — backed by `static_assert`s. Nice.
- **Piggybacking on `ForwardRequestToShard`.** Because these fake inos look shard-like to the generic dispatcher, they'd otherwise get routed toward a nonexistent shard tablet. The PR intercepts them right at the top of `ForwardRequestToShard`, before any shard-selection work happens, via `ClassifyControlNamespaceEntry(entityId)`. That one interception point, combined with the existing `FILESTORE_SERVICE_REQUESTS_FWD_TO_SHARD_BY_HANDLE`/`_BY_NODE_ID` macros, means `AccessNode`, `SetNodeAttr`, `UnlinkNode`, `ReadLink`, `AllocateData`, lock RPCs, etc. all get covered "for free" through the generic `BuildControlNamespaceResponse` primary template (defaults to `ControlNamespaceReadOnlyError`), even though only `GetNodeAttr`/`CreateHandle`/`ConfirmCreateHandle`/`DestroyHandle` got explicit specializations. That's a good use of a single choke point rather than patching every RPC handler.
- **"Hidden" is structural, not filtered.** The control dir has no real backing node in the directory table, so it simply never shows up in a normal `ListNodes` on the root — no special-case suppression code needed.
- **Spoofing protection.** `fsid` content is read from `session->FileStore.GetFileSystemId()` (the session's real filesystem), not from the request-supplied `FileSystemId` field, and there's a dedicated test (`ShouldIgnoreRequestSuppliedFileSystemIdForFsIdContent`) proving a spoofed request field doesn't change the answer.
- **Test coverage is genuinely thorough**: lookup, read, write-rejection via both `CreateHandle(WRNLY)` and direct `WriteData`, create/rename/link targeting the namespace, listing + pagination-cookie robustness, reading the dir as a file (`EISDIR`), invalid-session propagation, and the disabled-config no-op path.

## Things worth a second look

1. **`ReadData` accepts the raw ino as a handle without ever calling `CreateHandle`.** `TryHandleControlNamespaceReadData` classifies based on `msg->Record.GetHandle()` alone — there's no check that the handle was actually produced by a prior `CreateHandle` on this session. The PR is explicit about this being intentional (`ShouldReadViaAForgedHandleWithoutCreateHandle`, comment: "handles in this namespace are just the target inodes ... by design"). It's benign today since `fsid` isn't secret and the operation is stateless/idempotent, but it's a precedent worth flagging explicitly in the PR description/review thread if this control-namespace mechanism is meant to host less-benign files later — a reviewer coming from normal POSIX handle semantics could easily miss that CreateHandle-based permission checks are bypassable here by construction.

2. **`AccessNode` always returns `E_FS_ACCESS` via the generic fallback**, even for read-mode checks (`access(path, R_OK)`), since there's no `TAccessNodeMethod` specialization and the primary template unconditionally rejects. That could make `test -r .filestore-ctl/fsid`-style checks report "not accessible" even though reads actually succeed. Minor, but worth confirming it's intended vs. an oversight, given the file is meant to be readable.

3. **Fair bit of near-identical duplication** between the by-name `TryHandleControlNamespaceGetNodeAttr`/`CreateHandle` and the by-ino `BuildControlNamespaceResponse<TGetNodeAttrMethod>`/`<TCreateHandleMethod>` specializations — same classify → switch → fill-attr pattern written twice. Not a correctness issue (I traced both paths and they agree), but could be factored into one shared helper to reduce the risk of the two forms drifting apart later.

4. Nit: `ReadLink` also falls through to the generic "read-only" error rather than something like ENOTSUP/EINVAL for "not a symlink" — cosmetic, since it's not user-facing in the demonstrated flows.

## Verdict

The example transcript in the PR description checks out against the code for every case I traced (write rejected, mkdir rejected, mv into the dir rejected, hardlink rejected, `cat` on the dir gives `EISDIR`). Design is solid and the reserved-shard-number trick is a genuinely elegant way to get broad RPC coverage almost for free. The two things I'd actually ask about before merging are #1 (forged-handle read, mostly a documentation/precedent question) and #2 (AccessNode always denying); the rest are nice-to-haves.

---

1. iIntentional
2. Also intentional
3. Duplication is fine since, with time, those will be very different handlers
4. The readlink trade-off is also acceptable since it reduces the volume of code and just impacts the error code the client sees

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
```diff
@@ -1,100 +1,475 @@
...
+        return response;
+    }
+
+    if (entry == EControlNamespaceEntry::ControlDir) {
+        response->Record.SetHandle(ControlDirIno);
+        FillControlDirAttr(*response->Record.MutableNodeAttr());
+    } else {
+        response->Record.SetHandle(ControlFsIdFileIno);
+        FillControlFsIdAttr(*response->Record.MutableNodeAttr(), fileSystemId);
+    }
```
SvartMetal: Let's use a switch here - it's better to fail compilation when a new value is added to EControlNamespaceEntry than to silently take the else branch

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
```diff
@@ -1,100 +1,475 @@
...
+    auto response =
+        std::make_unique<TEvService::TGetNodeAttrMethod::TResponse>();
+    switch (ClassifyControlNamespaceEntry(record.GetNodeId())) {
+        case EControlNamespaceEntry::ControlDir:
+            FillControlDirAttr(*response->Record.MutableNode());
+            break;
+        case EControlNamespaceEntry::FsId:
+            FillControlFsIdAttr(*response->Record.MutableNode(), fileSystemId);
+            break;
+        default:
```
SvartMetal: If we don't use a default branch, the compiler will fail if we forget to handle some value

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
```diff
@@ -1,100 +1,475 @@
...
+    switch (entry) {
+        case EControlNamespaceEntry::ControlDir:
+            FillControlDirAttr(*response->Record.MutableNode());
+            break;
+        case EControlNamespaceEntry::FsId:
+            FillControlFsIdAttr(
+                *response->Record.MutableNode(),
+                session->FileStore.GetFileSystemId());
+            break;
+        default:
```
SvartMetal: Same

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
```diff
@@ -1,100 +1,475 @@
...
+        ProtoFlag(NProto::TCreateHandleRequest::E_WRITE));
+
+    if (entry == EControlNamespaceEntry::Unknown) {
+        *response->Record.MutableError() = MakeError(E_FS_NOENT, "not found");
+    } else if (wantsWrite) {
+        *response->Record.MutableError() = ControlNamespaceReadOnlyError();
+    } else if (entry == EControlNamespaceEntry::ControlDir) {
+        response->Record.SetHandle(ControlDirIno);
+        FillControlDirAttr(*response->Record.MutableNodeAttr());
+    } else {
```
SvartMetal: Also switch?

cloud/filestore/libs/storage/service/service_actor_control_namespace.h
```diff
@@ -0,0 +1,96 @@
...
+constexpr ui64 ControlDirIno = ShardedId(1, ControlNamespaceShardNo);
+constexpr ui64 ControlFsIdFileIno = ShardedId(2, ControlNamespaceShardNo);
+
+static_assert(ExtractShardNo(ControlDirIno) == ControlNamespaceShardNo);
+static_assert(ExtractShardNo(ControlFsIdFileIno) == ControlNamespaceShardNo);
+static_assert(ControlNamespaceShardNo > MaxShardCount);
+
+constexpr TStringBuf ControlFsIdFileName = "fsid";
+
+enum class EControlNamespaceEntry
```
SvartMetal, line 33: Let's add a comment explaining what this is, why it's needed, and how it's used

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
```diff
@@ -1,100 +1,475 @@
...
+        msg->Record.GetNewParentId(),
+        msg->Record.GetNewName(),
+        controlNamespaceDirName);
+
+    if (Y_LIKELY(
+            srcEntry == EControlNamespaceEntry::None &&
+            dstEntry == EControlNamespaceEntry::None))
+    {
+        return false;
+    }
```
SvartMetal: Why not simply return a read-only error?

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
```diff
@@ -1,100 +1,475 @@
...
 {
-    Y_UNUSED(ctx, ev, session);
-    return false;
+    if (StorageConfig->GetControlNamespaceDirName().empty() ||
+        Y_LIKELY(
+            ClassifyControlNamespaceEntry(ev->Get()->Record.GetNodeId()) ==
+            EControlNamespaceEntry::None))
+    {
+        return false;
+    }
```
SvartMetal: Looks like boilerplate that can be encapsulated in a method

SvartMetal (changes requested):

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
debnatkh: I decided to ALWAYS return EPERM

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
debnatkh: fixed

cloud/filestore/libs/storage/service/service_actor_control_namespace.h
debnatkh: comment added to entries

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
debnatkh: done

SvartMetal (approved):

agalibin (approved):

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
debnatkh: fixed

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
debnatkh: fixed

cloud/filestore/libs/storage/service/service_actor_control_namespace.cpp
debnatkh: fixed

Issue #6958, opened 2026-08-29
[Filestore] fastshard - shard


qkrorlqr:
Main issue: https://github.com/ydb-platform/nbs/issues/5895

Implement the missing pieces from the shard layer described here: https://github.com/ydb-platform/nbs/blob/main/doc/filestore/design/storage/fastshard/shard.md

Main pieces:
* proper page allocator that doesn't have insane contention on a handful of pages (the current allocator bitmap) **not sure whether it's actually a problem - we'll get rid of the dirty page check in PageStore after we implement proper lsn ordering in the layers below**
* proper data structures - a radix tree/an extent tree (described in the doc)

PR #7074, opened 2026-09-07
issue-6958: fastshard - open handles now prevent Node destruction

qkrorlqr:
### Notes
* tracking open handle count per node id in handle table
* a node is destroyed only when it has 0 links and 0 open handles
* implemented proper links tracking upon `UnlinkNode` (but hardlinks are currently not supported anyway so it's not really impactful) 

### Issue
https://github.com/ydb-platform/nbs/issues/6958

### DIFF
--- a/cloud/filestore/libs/storage/fastshard/impl/model/handle_table.cpp
+++ b/cloud/filestore/libs/storage/fastshard/impl/model/handle_table.cpp
@@ -62,7 +62,10 @@ NProto::TError THandleTable::Put(THandleSlot v, TWriteContext& writeContext)
     return Slots->Put(writeContext.Lsn, v, writeContext.PageGroups);
 }
 
-NProto::TError THandleTable::Delete(ui64 handle, TWriteContext& writeContext)
+NProto::TError THandleTable::Delete(
+    ui64 handle,
+    TWriteContext& writeContext,
+    ui64* nodeId)
 {
     THandleSlot slot{};
     auto error = Slots->Delete(
@@ -74,6 +77,8 @@ NProto::TError THandleTable::Delete(ui64 handle, TWriteContext& writeContext)
     if (error.GetCode() == E_FS_NOENT) {
         error = ErrorInvalidHandle(handle);
     }
+
+    *nodeId = slot.NodeId;
 
     return error;
 }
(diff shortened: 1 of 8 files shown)


qkrorlqr:
claude (final comment after a couple review iterations):
> With that, all the serious findings are resolved or explicitly accepted: Links is name-count only with handles tracked separately (matching tablet semantics), the error-path writes in CreateHandle are gone, the fail-stop-on-inconsistency behavior in Delete is your stated policy, and the layout change is covered by wiping test filesystems. Nothing serious left from my side — this looks ready to merge. The only open follow-ups, both outside this PR: a format-version marker before first real deployment, and eventually some handle GC/session-teardown story so unlinked-but-open nodes can't leak forever when clients vanish.

qkrorlqr:
claude - when told about the bug fixed in the last diff:

> You're right, and it's a fair catch on your side — worse, it's the exact pattern I was auditing. When I reviewed the Put bug I wrote that Delete "does check HasError", but I was looking at the check after the counter Get and overlooked that the check after Handles->Delete itself only special-cased E_FS_NOENT. Same family, one statement earlier in the same function. That one's on me.
> 
> For the record, the impact of the bug you fixed: a non-NOENT error from Handles->Delete (a page-store read/write failure) left slot zero-initialized, so the code proceeded to look up nodeId = 0 in the counter table — almost certainly NOENT — and returned that error, masking the real I/O failure with a misleading one. Since the caller still saw an error, the response errored and the guard rolled back, so unlike the Put case nothing could commit; the harm was error masking (and, pathologically, a decrement of a counter row keyed 0 inside a rolled-back transaction). Your fix — if (HasError(error)) return error; between the NOENT check and the counter lookup — is the right one and mirrors the Put fix.
> 
> Having been burned once, I swept both files for every remaining NOENT-special-case this time: in handle_table.cpp, AllocateHandle, Get, GetNodeId, and GetNodeHandleCount all have explicit non-NOENT arms; in shard.cpp, all five E_FS_NOENT sites are clean — Names.Get in CreateHandle has an else if (HasError(error)) arm before the open-existing branch, and the PageIndex sites in SetNodeAttr, DestroyNode, WriteData, and ReadData each handle the general error case explicitly. No further instances of the pattern remain in this PR's code.

asorotsky (approved):

arigachnyy (approved):

PR #7081, opened 2026-09-08
[Disk Manager] Allow variable chunk size

jkuradobery:
### Notes
We have high lattencies and would like to increase upload bandwidth by increasing the size of request.


jkuradobery:
I will separate this MR into
1) Support disk sizes not divisable by 4MiB chunk size.
2) Support variable chunk size for snapshots.
3) Support variable chunk size for images.

PR #7082, opened 2026-09-08
issue-6956: add the journal stores and an in-memory device

arigachnyy:
### Notes
Add IKeyBufferStore for the journal metadata, with an in-memory implementation, and IPageStore for the journal pages.

The page store keeps the pages on an IDevice and owns the page allocation: it hands out free pages, takes the given ones out of the free space when a record is restored, and returns them to it when a record is dropped. Every page is of the page size the store is created with, and the device holds a fixed number of them.

Add an in-memory IDevice as well: it overwrites a page that is written again and reads a page that has never been written as a zeroed one.

### Issue

### DIFF
--- a/cloud/storage/core/libs/journalled_device/device_page_store.cpp
+++ b/cloud/storage/core/libs/journalled_device/device_page_store.cpp
@@ -52,10 +52,10 @@ private:
 
 public:
     TDevicePageStore(
-            IDevicePtr device,
-            ui64 pageCount,
-            ui32 pageSize,
-            EDevicePageStoreMode mode)
+        IDevicePtr device,
+        ui64 pageCount,
+        ui32 pageSize,
+        EDevicePageStoreMode mode)
         : Device(std::move(device))
         , PageCount(pageCount)
         , PageSize(pageSize)
@@ -66,7 +66,7 @@ public:
         }
     }
 
-    TVector<TPageGroupRef> Allocate(ui64 pageCount) override
+    TVector<TPageRange> Allocate(ui64 pageCount) override
     {
         with_lock (Lock) {
             return AllocateImpl(pageCount);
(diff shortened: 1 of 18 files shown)


qkrorlqr (dismissed):
consider using https://github.com/ydb-platform/nbs/blob/main/cloud/storage/core/libs/common/disjoint_interval_map.h for disjoint interval storage

asorotsky (dismissed):

cloud/storage/core/libs/journalled_device/device_page_store.cpp
```diff
@@ -117,304 +131,280 @@ class TDevicePageStore final
...
+            {
+                if (it->second.Begin <= pageNo) {
+                    pageNo = Max(pageNo, it->second.End);
+                }
+            });
 
-        const ui64 endPageNo = ref.FirstPageNo + ref.PageCount;
-        if (freeEnd < endPageNo) {
-            return freeEnd;
+        if (pageNo < endPageNo) {
```
qkrorlqr, line 325: why can pageNo be >= than endPageNo? we initialize it to pageRange.FirstPageNo and then we update it with overlapping page nos - seems that by construction the result should lie within pageRange

cloud/storage/core/libs/journalled_device/device_page_store.cpp
```diff
@@ -117,304 +131,280 @@ class TDevicePageStore final
...
+                    FreeRanges.Add(rangeBegin, begin, {});
+                }
+                if (end < rangeEnd) {
+                    FreeRanges.Add(end, rangeEnd, {});
+                }
+            });
     }
 
-    std::optional<ui64> FindBusyPage(const TPageGroupRef& ref) const
+    std::optional<ui64> FindBusyPage(const TPageRange& pageRange) const
```
qkrorlqr, line 309: nit: Busy sounds weird in this context, probably Occupied or Allocated would sound more natural

cloud/storage/core/libs/journalled_device/device_page_store.cpp
arigachnyy: example:
FreeRanges = [0-4]
pageRange = [1-2]
So pageNo will be equal 4 (=Max(2,4)) => no allocated pages.

cloud/storage/core/libs/journalled_device/device_page_store.cpp
arigachnyy: ok, in the next PR if there are no more problems.

asorotsky (approved):

qkrorlqr (approved):

PR #7086, opened 2026-09-08
switch between rdma and grpc based on link state

yegorskii:
### Notes

[Cells] Move data back to gRPC while RDMA is down

Setting up RDMA stopped holding a cell connection back in #7035, but the
switch was one-way: once data moved onto RDMA nobody looked at it again.
A break then cost the whole reconnect window, because every retry went
into the transport that had just died while a healthy gRPC path sat
unused beside it.

Building on the endpoint state handler that landed with the synchronous
StartEndpoint, the switcher now owns the transport choice for as long as
the connection lives. A break moves data straight back onto gRPC; a
reconnect moves it onto RDMA again only once it has stayed up for
RdmaSettleTime, so a flapping link does not keep pulling traffic onto a
transport about to drop. The default settle time is 10 seconds and zero
switches over immediately.

The first connect never waits. The settle time guards a return to a link
that has already proved it can drop; a link that has never carried our
data has proved nothing, and making it wait would only keep the data on
the slower transport.

Taking the endpoint is synchronous now: it is handed back before it has
connected and reconnects on its own from then on, so the switcher asks
for it once and has nothing to retry. A failure there means the rdma
client itself cannot give us an endpoint, and the connection stays on
gRPC.

The win is in the retries rather than in the failing requests: the ones
already in flight are aborted by the rdma client itself, and what this
buys is that TDurableClient's next attempt lands somewhere that works.

Gated by the existing GrpcDataFallbackEnabled, off by default. The path
with the flag off is untouched and still waits for the first connect
through the asynchronous form: without a fallback transport, handing out
a connection early would trade a failed mount for IO that fails
retriably forever.

### Issue

### DIFF
--- a/cloud/blockstore/libs/cells/iface/config.cpp
+++ b/cloud/blockstore/libs/cells/iface/config.cpp
@@ -23,7 +23,7 @@ namespace {
         NProto::ECellDataTransport,                                            \
         NProto::CELL_DATA_TRANSPORT_GRPC                                      )\
     xxx(GrpcDataFallbackEnabled,     bool,                   false            )\
-    xxx(RdmaSettleTimeMs,            ui32,                   30000            )\
+    xxx(RdmaSettleTime,              TDuration,   TDuration::Seconds(10)     )\
 // BLOCKSTORE_CELL_DEFAULT_CONFIG
 
 #define BLOCKSTORE_CELL_DECLARE_CONFIG(name, type, value)                      \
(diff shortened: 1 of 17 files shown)


yegorskii:
Codex review

> Reviewed the transport switching logic, including RDMA state transitions, settle-time handling, concurrency, endpoint lifetime, and configuration propagation. I found no blocking issues.
> Local verification:
> - cloud/blockstore/libs/cells/impl/ut
> - cloud/blockstore/libs/client_rdma/ut
> - 42/42 tests passed
> - git diff --check passed
> LGTM

.

sharpeye (approved):

cloud/blockstore/libs/cells/iface/config.cpp
```diff
@@ -23,6 +23,7 @@ namespace {
         NProto::ECellDataTransport,                                            \
         NProto::CELL_DATA_TRANSPORT_GRPC                                      )\
     xxx(GrpcDataFallbackEnabled,     bool,                   false            )\
+    xxx(RdmaSettleTime,              TDuration,   TDuration::Seconds(10)     )\
```
agalibin, line 26: ')\' is not align

agalibin (approved):

PR #7089, opened 2026-09-08
issue-6956: add an lsn barrier between reads and the flush cycle

arigachnyy:
### Notes
Add TWatermarkTracker: a monotonic watermark with reference-counted pins. Pin() takes the current watermark and holds it, GetPinnedWatermark() returns the lowest pinned value.

Track the last indexed lsn with it in journalled device v2. A read pins the watermark for the whole request, and the flush cycle samples GetPinnedWatermark() once per cycle and passes it to the journal as maxAllowedLsn, so records a concurrent read may still need are not flushed and cleaned up from under it.

Interface changes required for this:
* Restore() now returns the lsn of the last indexed record instead of just an error, so the tracker starts from the restored state.
* GetFirstRecordToFlush() becomes GetRecordToFlush(maxAllowedLsn).

### Issue


asorotsky (dismissed):

cloud/storage/core/libs/journalled_device/journalled_device_v2.cpp
```diff
@@ -293,7 +303,22 @@ class TJournalledDeviceV2 final
         return response;
     }
 
-    void ScheduleFlushCycle() {
+    NCloud::NProto::TWriteLogRecordResponse DoWriteLogRecord(
+        NCloud::NProto::TWriteLogRecordRequest request)
+    {
+        auto lsn = request.GetLogSequenceNumber();
```
qkrorlqr: nit: no `auto` for primitives plz

cloud/storage/core/libs/journalled_device/journalled_device_v2.cpp
```diff
@@ -264,6 +269,11 @@ class TJournalledDeviceV2 final
     NCloud::NProto::TReadPagesResponse DoReadPages(
         NCloud::NProto::TReadPagesRequest request)
     {
+        auto pinnedLsn = IndexedLsnTracker.Pin();
```
qkrorlqr: nit: no `auto` for primitives plz

cloud/storage/core/libs/journalled_device/watermark_tracker.cpp
```diff
@@ -0,0 +1,64 @@
...
+void TWatermarkTracker::Advance(ui64 watermark)
+{
+    with_lock (Lock) {
+        if (Watermark < watermark) {
+            Watermark = watermark;
+        }
+    }
+}
+
+ui64 TWatermarkTracker::Pin()
```
qkrorlqr: let's use "barrier" terminology instead of "pin"

"pin" usually refers to eviction protection in caches

and this logic actually resembles the barrier logic present in filestore and blockstore tablets

cloud/storage/core/libs/journalled_device/watermark_tracker_ut.cpp
```diff
@@ -0,0 +1,278 @@
+#include "watermark_tracker.h"
+
+#include <library/cpp/testing/unittest/registar.h>
+
+#include <atomic>
+#include <thread>
+
+namespace NCloud::NJournalled {
+
+////////////////////////////////////////////////////////////////////////////////
```
qkrorlqr: nit: after `namespace {` plz

qkrorlqr (dismissed):

asorotsky (dismissed):

asorotsky (approved):

qkrorlqr (approved):

PR #7098, opened 2026-09-09
issue-1751: [Filestore] Add support for concurrent allocations in persistent storage by TWriteBackCache

e673:
### Notes
Pending request serialization into persistent storage inside a lock section is a bottleneck that limits the maximal throughput when using write-back cache. The PR address this issue and implements concurrent serialization by moving serialization outside lock secions.

### Issue


cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -209,19 +184,26 @@ NProto::TError TWriteDataRequestManager::Init(
 
 bool TWriteDataRequestManager::HasPendingRequests() const
 {
-    return !PendingRequests.Empty();
+    return !PendingRequests.Empty() || !AllocatedRequests.Empty() ||
+           !SerializingRequests.Empty();
```
SvartMetal, line 188: HasPendingRequests, HasAllocatedRequests, HasSerializedRequests

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -209,19 +184,26 @@ NProto::TError TWriteDataRequestManager::Init(
...
 }
 
 bool TWriteDataRequestManager::HasPendingOrUnflushedRequests() const
 {
-    return !UnflushedRequests.Empty() || !PendingRequests.Empty();
+    return HasPendingRequests() || !UnflushedRequests.Empty();
```
SvartMetal, line 193: HasUnflushedRequests

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -246,74 +234,198 @@ ui64 TWriteDataRequestManager::GetMaxUnflushedSequenceId() const
...
+        pendingRequest->Status = EPendingWriteDataRequestStatus::Removed;
+
+        while (!CancelSerializingRequests.empty() &&
+               CancelSerializingRequests.front()->Status ==
+                   EPendingWriteDataRequestStatus::Removed)
+        {
+            CancelSerializingRequests.pop_front();
         }
+
+        return AllocPendingRequestsInPersistentStorage();
```
SvartMetal, line 315: Let's add a comment

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
```diff
@@ -246,74 +234,198 @@ ui64 TWriteDataRequestManager::GetMaxUnflushedSequenceId() const
...
+                return false;
+            }
+
+            request->Status = EPendingWriteDataRequestStatus::Removed;
+            SerializingRequestsRemove(request.get());
+            return AllocPendingRequestsInPersistentStorage();
+        }
+        case EPendingWriteDataRequestStatus::CancelSerializing:
+        case EPendingWriteDataRequestStatus::Removed:
+            return true;
```
SvartMetal, line 425: Impossible event and return false?

SvartMetal (changes requested):

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: Added

cloud/filestore/libs/vfs_fuse/write_back_cache/write_data_request_manager.cpp
e673: This code was removed

e673:
AI review

> I'll pull the PR, its description and comments, then go through the diff for the issues you listed.Now the state, cache and storage code around the changed paths.I reviewed the branch tip (`848bb59`, 7 commits — the PR page still showed only `e6b5f3e`; later commits changed things like making `SerializeToAllocation` noexcept and reverting the metric split, so the PR description is now partly stale).
> 
> ## Verdict
> 
> No memory-safety bug or broken invariant found. The core design holds up:
> 
> - **Serialization outside the lock is safe.** The `TPendingWriteDataRequest` being serialized is pinned: it's owned by `TNodeCache::PendingRequests` (deque of `unique_ptr`, stable address), and every destruction path is now gated — `PopBackUnallocatedPendingRequest` checks `HasAllocation()`, `DequeuePendingRequest` only fires for the committed front (which must be `Serialized`, set under the lock by the same thread), `DropCachedData` bails on pending requests, and `DeleteNodeState` requires `Cache.Empty()`.
> - **memcpy is bounded.** `TMemoryOutput::DoWrite` does `Y_ENSURE(end <= End_)` before the `memcpy`, and the `Exhausted()` check catches under-fill; the `catch(...)` converts overflow into `false`. `Crc32c` covers exactly `[AllocationPtr, +AllocationByteCount)`, matching what the ring buffer computes in `Commit(ptr)`.
> - **Ring buffer tolerates multiple incomplete allocations.** `EraseFreeEntriesFromFront` stops at entries still in `EntryMap`, migration/pos-reset require `Empty()`, and no code path touches the data region of an incomplete entry under the lock while another thread writes it.
> - **Back-popping invariant is correct.** Allocation always takes `UnallocatedPendingRequests.Front()`, so all allocated requests have lower sequence ids than all unallocated ones; hence the node's back pending request is the global back unallocated one (the `Y_ABORT_UNLESS` in `FailAllUnallocatedPendingRequests` holds) and popping from the back until an allocated one is hit removes exactly the unallocated set.
> - Dropping `TriggerFlushAll(false)` for the backpressure case is fine: backpressure implies `>1` flush batch, which already forces `FlushRequested`.
> 
> ## Problems worth fixing
> 
> **1. Manual unlock/relock instead of RAII** — `write_back_cache_state.cpp:909-913`
> ```cpp
> guard.GetMutex()->ReleaseWithoutProcessingQueuedOperations();
> bool serializationSucceeded = pendingRequest->SerializeToAllocation();
> guard.GetMutex()->Acquire();
> ```
> This only works because `SerializeToAllocation` is `noexcept`; if anything in that window ever throws, `~TGuard` releases an unheld `TAdaptiveLock`. `util/system/guard.h` already has `TInverseGuard`/`TInverseLockOps` for exactly this. A small ops struct (`Acquire → q->Acquire()`, `Release → q->ReleaseWithoutProcessingQueuedOperations()`) and `TInverseGuard<TQueuedOperations, TPostponingOps> unguard(guard.GetMutex());` in a nested scope makes it exception-safe and self-documenting. Passing `TGuard<TQueuedOperations>&` into `ProcessPendingRequests` just to reach `QueuedOperations` (a member) is then unnecessary.

I've removed `ReleaseWithoutProcessingQueuedOperations` and replaced it with `Release` because it is worth running scheduled operations before serialization. Now `Unguard` is used.

Also I've removed `noexcept` from `SerializeToAllocation`. 

> **2. New abort path via `Deserialize`** — `write_data_request_manager.cpp:283-289`
> `GetNextReadyCachedRequest` round-trips a freshly serialized in-memory request through `TCachedWriteDataRequest::Deserialize` (a function whose contract is "return nullptr on bad on-disk data") and then `Y_ABORT_UNLESS(cachedRequest != nullptr)`. The old code constructed `TCachedWriteDataRequest` directly and had no failure mode. Today this is only unreachable because `ValidateWriteDataRequest` rejects zero-length writes upstream; `TWriteBackCacheState::AddWriteDataRequest` itself does not validate. Construct directly from `AllocationPtr`/`AllocationByteCount` and keep `Deserialize` for `Init()`.

I don't see the reason for this change.

> **3. `break` in the ReleaseHandle loop skips unrelated handles** — `write_back_cache_state.cpp:502-516`
> When one handle still has (allocated) pending requests, `break` exits the whole `while (!HandlesWithReleaseRequests.empty())` loop, so ReleaseHandle requests on *other* handles of the node are not failed in this round either. If the state is truly "impossible" that's tolerable, but the intent reads as "skip this handle", not "stop processing". Iterate with an explicit iterator and skip only the offending handle.

I've added a comment there.

> **4. Dead code** — `write_back_cache_state.cpp:101-105`
> `RequestManager.AddRequest` always returns a `make_unique` result; the `if (!pendingRequest) { SetFailedFlag(); ... }` branch can't execute.

Removed.

> **5. `noexcept` catch handler can itself throw** — `write_data_request.cpp:1075-1080`
> `"..." + CurrentExceptionMessage()` allocates inside a `noexcept` function's catch block; on `bad_alloc` that's `std::terminate`. Given the user-facing contract ("returns false on failure"), either report a static message or build the string before the `Report*` call in a `try`. Minor.

I've removed `noexcept`.

> **6. Metric semantics silently changed**
> Every request now transits the pending queue, so `PendingQueue_Count`/`ProcessedCount`/`MaxTime`/`ProcessedTime` include allocation+serialization of *every* write, not only writes blocked by full storage/backpressure. That's exactly why the original `fs_ut` failed in CI (it inferred backpressure from `pendingQueueCount == 1`) and why all the `ProcessedCount` expectations in `write_back_cache_state_ut.cpp` changed. Dashboards/alerts on `PendingQueue_*` will see the same shift; the "unnecessary MaxTime alerts" comment in `FlushFailed` now applies to a different population. The description still says a separate combined metric was added — it was reverted in `b27103e9`; update the description or reinstate the split.

I will update the PR description.

> **7. Failure order is now newest-first**
> `FailNodeUnallocatedPendingRequests`/`FailAllUnallocatedPendingRequests` pop from the back, so the guest sees errors in reverse sequence order. Harmless for correctness, but a behaviour change worth a sentence in the description.

It doesn't affect the logic.
I will update the PR description.

> ## Tests
> 
> - **`MultiThreadedReadWriteTest` hangs instead of failing.** All waits are `GetValueSync()` and the destructor does `Drain().Wait()`. If the cache ever enters the failed state (whose design is "hang everything"), the test blocks until the harness kills it, with no diagnostic. `stopRequested` can't help a thread parked in `GetValueSync()`. Use `Wait(timeout)` + `UNIT_ASSERT`, and bound the drain.

Done.

> - **The risky path (iovecs) isn't covered.** The whole point of the lock release is reading guest memory (`iovec.GetBase()`) outside the lock; the MT test only uses `SetBuffer`. A `ZeroCopyWriteEnabled = true` variant of the same test would exercise the lifetime argument in the `FlushFailed` comment.

Done.

> - **`ShouldDropBackpressuredRequestOnFlushFailure`** only has an unallocated pending request, so it doesn't verify the new claim that allocated requests *survive* `FlushFailed`. `write_data_request_manager_ut` can drive allocation step-by-step; a state-level test would need to interleave, but at minimum the manager-level survival could be asserted.

I've rewritten the test - `ShouldNotDropAllocatedRequestsOnFlushFailure`

> - `ShouldCommitRequestsInSequenceWhenSerializationCompletesOutOfOrder` is good and reproduces the intended ordering; nit: the comment line "requests are committed in global FIFO order." is duplicated (`write_data_request_manager_ut.cpp:~470`).

Done.

> - The rewritten `fs_ut` test is deterministic as far as I can trace (backpressure is set before the promise event is queued; the first flush is blocked so `unflushed == i` holds), so the earlier CI failure should be resolved — but the second CI run for the branch tip wasn't finished when I looked; confirm it's green.

Ack.

> ## Nits / design notes
> 
> - Typo: "Pending requests cannot exist **it** this case" (`write_back_cache_state.cpp:1059`).

Fixed.

> - `GetNextAllocatedPendingRequest()` actually *performs* the allocation; `TryAllocNextPendingRequest()` reads better.

Renamed.

> - `UpdateStats`: `Max(maxPending, maxAllocated)` is always the allocated front when it exists (it's older by construction); a one-liner picking `Allocated.Front()` else `Unallocated.Front()` is clearer.

Changed.

> - `friend class TWriteDataRequestManager` pokes four private fields of `TPendingWriteDataRequest`; a `SetAllocation(char*)` plus const getters would keep the "set under lock" rule visible in the interface.

I don't like the code bloat after this change.

> - `HasPendingRequests()` is now test-only.

No, it is called from `HasPendingOrUnflushedRequests()`

> - With `ReleaseWithoutProcessingQueuedOperations`, queued events (promise completions, `ScheduleFlushNode`) are drained by whichever thread next calls `Release()`. Nothing depends on the queueing thread, but note that a `ReadData` completion thread (via `UnpinCachedData → ProcessPendingRequests`) can now both serialize other threads' writes and run flushes inline. That's inherent to the work-stealing design; just worth stating in the PR so latency changes on the read path aren't a surprise.

Queued operations are now executed before serialization so it will not affect the latency.

SvartMetal (approved):

PR #7110, opened 2026-09-09
issue-2667: add LinkNodeInShard tablet op

debnatkh:
### Notes
Shard-side step of a cross-shard hard link: given a node id, bump that node's hard link count by one. It creates no NodeRef in the shard.

Exactly-once is guaranteed by the `ResponseLog` (same mechanism as `RenameNodeInDestination`)

Nothing calls this yet - the directory tablet's OpLog entry and worker actor will come in follow-up changes.

### Issue
#2667

### DIFF
--- a/cloud/filestore/libs/diagnostics/profile_log_events.cpp
+++ b/cloud/filestore/libs/diagnostics/profile_log_events.cpp
@@ -456,6 +456,14 @@ void InitProfileLogRequestInfo(
 template <>
 void InitProfileLogRequestInfo(
     NProto::TProfileLogRequestInfo& profileLogRequest,
+    const NProtoPrivate::TLinkNodeInShardRequest& request)
+{
+    profileLogRequest.MutableNodeInfo()->SetNodeId(request.GetNodeId());
+}
+
+template <>
+void InitProfileLogRequestInfo(
+    NProto::TProfileLogRequestInfo& profileLogRequest,
     const NProto::TAccessNodeRequest& request)
 {
     auto* nodeInfo = profileLogRequest.MutableNodeInfo();
--- a/cloud/filestore/libs/storage/api/tablet.h
+++ b/cloud/filestore/libs/storage/api/tablet.h
@@ -57,6 +57,7 @@ namespace NCloud::NFileStore::NStorage {
     xxx(RenameNodeInDestination,            __VA_ARGS__)                       \
     xxx(PrepareUnlinkDirectoryNodeInShard,  __VA_ARGS__)                       \
     xxx(AbortUnlinkDirectoryNodeInShard,    __VA_ARGS__)                       \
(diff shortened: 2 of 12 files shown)


debnatkh:
**Model used for reasoning:** Claude Sonnet 5.

---

**Remaining issue (1):**

- **Missing read-ahead cache invalidation on success.** The "fix review issue" commit deleted `InvalidateReadAheadCache(args.Request.GetNodeId())` entirely from `CompleteTx_LinkNodeInShard`, rather than gating it on success. Every sibling op that mutates node attrs (`SetNodeAttr`, `UnlinkNode`, `RenameNode`) invalidates the cache — unconditionally, even on error, in those cases — because attrs (link count, ctime/mtime) change. `LinkNodeInShard` also mutates attrs on success via `E_CM_CMTIME | E_CM_REF`, so as it stands now a successful link-count bump leaves stale cached data/attrs for that node. Needs either:
  ```cpp
  if (!HasError(args.Error)) {
      InvalidateReadAheadCache(args.Request.GetNodeId());
      CommitResponseLogEntry(std::move(args.ResponseLogEntry));
      ...
  }
  ```
  or an explanation for why this op is exempt from the convention.

That's the only open item — the two issues I raised earlier (misleading debug log, missing `MaxLink` test) are both resolved as of commit `1ca2bd1`.

---

This tx does not mutate the actual data in any way, this read-ahead cache invalidation is redundant here

---

Good push — I looked at the actual implementation rather than going by analogy, and you're right.

`InvalidateReadAheadCache` is defined in `tablet_state_data.cpp`:

```cpp
void TIndexTabletState::InvalidateReadAheadCache(ui64 nodeId)
{
    Impl->ReadAheadCache.InvalidateCache(nodeId);
}
```

and the read-ahead cache it wraps only stores `DescribeData` results — cached blob-range lookups for actual file *content*, registered via `RegisterReadAheadResult(nodeId, handle, range, describeDataResult)`. It has nothing to do with node attrs.

`LinkNodeInShard` only touches `Attrs.Links` and `ctime`/`mtime` (`E_CM_CMTIME | E_CM_REF`) — it never touches size or blob/byte-range data. So there's no stale-data risk to invalidate against. `SetNodeAttr` invalidates because it can change `Size` (truncate), which directly affects which byte ranges are valid to cache — that's not analogous here. So the removal in the fix commit was correct, not an oversight.

**Remaining issues: none.** With that resolved, I don't see any open items on this PR — the two earlier gaps (misleading debug log, missing `MaxLink` test) were fixed, and the cache-invalidation removal is justified by the actual cache semantics rather than a regression.

Model used for reasoning: **Claude Sonnet 5**.

qkrorlqr (dismissed):

qkrorlqr (approved):

yegorskii (approved):

PR #7119, opened 2026-09-10
support for down policy in drain request

yegorskii:
### Notes
- Added a local EDrainDownPolicy protobuf enum matching the Hive drain policies.
- Added DownPolicy to the private BlockStore and FileStore drain requests.
- Marked the legacy KeepDown field as deprecated and removed its use from the application code.
- Propagated DownPolicy through the service actors and Hive proxy.
- Added explicit conversion from the NBS enum to NKikimrHive::EDrainDownPolicy, avoiding a dependency on the YDB protobuf in the private APIs.
- Added compile-time checks to detect changes to the local policy enum.
- An omitted DownPolicy defaults to DRAIN_POLICY_NO_DOWN.

### Motivation
The existing KeepDown flag is a deprecated compatibility field in the YDB Hive API. As a boolean, it can represent only two policies:
- false → DRAIN_POLICY_NO_DOWN
- true → DRAIN_POLICY_KEEP_DOWN
KeepDown cannot represent DRAIN_POLICY_KEEP_DOWN_UNTIL_RESTART, where the drained node remains down until the node itself restarts and registers with Hive again.

### Issue

### DIFF
--- a/cloud/filestore/libs/storage/service/service_actor_actions_drain_tablets.cpp
+++ b/cloud/filestore/libs/storage/service/service_actor_actions_drain_tablets.cpp
@@ -35,11 +35,12 @@ struct TDrainTabletsActionActor final
             return;
         }
 
-        LOG_INFO(
+        LOG_INFO_S(
             ctx,
             TFileStoreComponents::SERVICE,
-            "Draining tablets started, DownPolicy=%d",
-            request.GetDownPolicy());
+            "Draining tablets started, DownPolicy="
+                << NCloud::NProto::EDrainDownPolicy_Name(
+                       request.GetDownPolicy()));
 
         DrainTablets(ctx, request);
         Become(&TThis::StateWork);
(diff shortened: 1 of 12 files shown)


cloud/filestore/libs/storage/service/service_actor_actions_drain_tablets.cpp
```diff
@@ -38,10 +38,10 @@ struct TDrainTabletsActionActor final
         LOG_INFO(
             ctx,
             TFileStoreComponents::SERVICE,
-            "Draining tablets started, KeepDown=%d",
-            request.GetKeepDown());
+            "Draining tablets started, DownPolicy=%d",
```
debnatkh: nit: printing the raw integer isn't very readable in logs

Consider using `NProto::EDrainDownPolicy_Name(downPolicy)`, swapping to `"%s"` with that would make the logs easier to grep/debug

debnatkh (dismissed):

debnatkh (dismissed):

debnatkh (approved):

kobzonega (approved):

agalibin (approved):

PR #7122, opened 2026-09-10
[NBS][vhost] Fix deadlock during endpoint shutdown

agalibin:
## Summary

Fix a deadlock during vhost endpoint shutdown that prevents `StopEndpoint` from completing and causes subsequent requests to time out. The fix covers both individual `StopEndpoint` calls and server shutdown through `StopAllEndpoints`.

## Root cause

`StopEndpoint` calls `TEndpoint::Stop()` while holding the server mutex. Inside libvhost, this call synchronously waits for the control thread.
If the control thread is completing another endpoint's shutdown, its callback attempts to acquire the same mutex through `HandleStoppedEndpoint`. This creates a circular wait that blocks further endpoint operations.
`StopAllEndpoints` also calls `TEndpoint::Stop()` while holding the mutex.

### DIFF
--- a/cloud/blockstore/libs/vhost/server.cpp
+++ b/cloud/blockstore/libs/vhost/server.cpp
@@ -41,6 +41,28 @@ TVector<TExecutor*> NormalizeExecutors(
     return normalized;
 }
 
+void RequestEndpointStop(
+    const TEndpointPtr& endpoint,
+    bool deleteSocket,
+    TPromise<NProto::TError> stopPromise)
+{
+    const auto error = SafeExecute<NProto::TError>(
+        [&]
+        {
+            endpoint->Stop(deleteSocket).Subscribe(
+                [stopPromise](const auto& future) mutable
+                {
+                    stopPromise.SetValue(
+                        SafeExecute<NProto::TError>(
+                            [&] { return future.GetValue(); }));
+                });
+            return NProto::TError();
+        });
+    if (HasError(error)) {
+        stopPromise.SetValue(error);
+    }
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
 class TServer final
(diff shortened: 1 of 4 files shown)


cloud/blockstore/libs/vhost/server.cpp
```diff
@@ -288,17 +297,19 @@ TFuture<NProto::TError> TServer::StartEndpoint(
...
-        error.SetMessage("Vhost server is stopped");
-        return MakeFuture(error);
-    }
-
     TEndpointPtr endpoint;
-    TFuture<NProto::TError> stopFuture;
+    auto promise = NewPromise<NProto::TError>();
 
     with_lock (Lock) {
+        if (ShouldStop.test()) {
```
nisqatsi: Может оставить в начеле функции как было? (не под Lock)

cloud/blockstore/libs/vhost/server.cpp
```diff
@@ -41,6 +41,28 @@ TVector<TExecutor*> NormalizeExecutors(
     return normalized;
 }
 
+void StartStoppingEndpoint(
```
komarevtsev-d: Имя путающее. Мб лучше `RequestEndpointStop()`?

cloud/blockstore/libs/vhost/server.cpp
komarevtsev-d, line 469: Давай вот тут вынесем удаление `TStoppingEndpoint` из под лока?

Вне лока: `TStoppingEndpoint stopped;`
И внутри if'a:
`stopped = std::move(it->second);`

komarevtsev-d (approved):

cloud/blockstore/libs/vhost/server.cpp
```diff
@@ -286,6 +325,33 @@ TFuture<NProto::TError> TServer::StartEndpoint(
...
 
+TServer::TStopRequest TServer::RegisterStoppingEndpoint(
+    const TString& socketPath,
+    TEndpointPtr endpoint)
+{
+    auto promise = NewPromise<NProto::TError>();
+
+    // Keeps the server alive until the stop completes. Empty when called from
+    // ~TServer() via Stop().
+    auto self = weak_from_this().lock();
```
sharpeye, line 336: shared_from_this() ?

sharpeye (approved):

PR #7123, opened 2026-09-10
issue-6956: add a device-backed key buffer store

arigachnyy:
### Notes
The store keeps its buffers in the pages of a device. Two superblock slots at the start of the device carry the erased bound and are written in turns. The rest of the pages hold the buffers, each page carrying a header with the key, the chunk position, a write sequence number and a checksum. Restore scans the device, drops torn or erased entries and keeps the newest intact copy of each key.

### Issue

### DIFF
--- a/cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
+++ b/cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
@@ -1,8 +1,24 @@
 #include "key_buffer_store.h"
 
-#include <util/generic/map.h>
+#include "device.h"
+#include "device_page_store.h"
+
+#include <cloud/storage/core/libs/common/future_helper.h>
+
+#include <library/cpp/digest/crc32c/crc32c.h>
+
+#include <util/generic/algorithm.h>
+#include <util/generic/utility.h>
+#include <util/generic/ylimits.h>
+#include <util/stream/buffer.h>
+#include <util/stream/mem.h>
 #include <util/string/builder.h>
 #include <util/system/spinlock.h>
+#include <util/system/yassert.h>
+#include <util/ysaveload.h>
+
+#include <optional>
+#include <utility>
 
 namespace NCloud::NJournalled {
 
@@ -12,38 +28,46 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+using TRestoreResult = TResultOrError<TMap<ui64, TBuffer>>;
+
+////////////////////////////////////////////////////////////////////////////////
+
 class TInMemoryKeyBufferStore final: public IKeyBufferStore
 {
 private:
-    mutable TAdaptiveLock Lock;
+    TAdaptiveLock Lock;
     TMap<ui64, TBuffer> Buffers;
+    std::optional<ui64> ErasedUpToKey;
 
 public:
+    TFuture<TRestoreResult> Restore() override
+    {
+        with_lock (Lock) {
+            return MakeFuture<TRestoreResult>(Buffers);
+        }
+    }
+
     TFuture<NCloud::NProto::TError> Write(ui64 key, TBuffer buffer) override
     {
         with_lock (Lock) {
+            if (ErasedUpToKey && key <= *ErasedUpToKey) {
+                return MakeFuture(MakeError(
+                    E_ARGUMENT,
+                    TStringBuilder() << "key " << key << " is erased"));
+            }
+
             Buffers[key] = std::move(buffer);
         }
         return MakeFuture(MakeError(S_OK));
     }
 
-    TFuture<TResultOrError<TBuffer>> Read(ui64 key) const override
-    {
-        with_lock (Lock) {
-            auto it = Buffers.find(key);
-            if (it == Buffers.end()) {
-                return MakeFuture<TResultOrError<TBuffer>>(MakeError(
-                    E_NOT_FOUND,
-                    TStringBuilder() << "no buffer for key " << key));
-            }
-
-            return MakeFuture<TResultOrError<TBuffer>>(it->second);
-        }
-    }
-
     TFuture<NCloud::NProto::TError> EraseUpTo(ui64 lastKey) override
     {
         with_lock (Lock) {
+            if (!ErasedUpToKey || *ErasedUpToKey < lastKey) {
+                ErasedUpToKey = lastKey;
+            }
+
             auto end = Buffers.upper_bound(lastKey);
             if (end == Buffers.begin()) {
                 return MakeFuture(MakeError(S_FALSE));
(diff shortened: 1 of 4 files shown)


cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +77,648 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+
+    if (magic != SuperblockMagic || version != StoreFormatVersion ||
+        crc != Crc32c(page.data(), SuperblockSize - sizeof(ui32)))
+    {
+        return std::nullopt;
+    }
+
+    if (hasErasedUpToKey) {
+        superblock.ErasedUpToKey = erasedUpToKey;
+    }
```
asorotsky: Maybe it can be just erasedUpToKey either 0 or non 0 instead of separate hasXXX?

cloud/storage/core/libs/journalled_device/key_buffer_store.h
```diff
@@ -19,20 +19,23 @@ struct IKeyBufferStore
 {
     virtual ~IKeyBufferStore() = default;
 
+    [[nodiscard]] virtual auto Restore()
+        -> NThreading::TFuture<TResultOrError<TMap<ui64, TBuffer>>> = 0;
```
asorotsky: maybe sorted vector of pairs <ui64, buffer> would be more cache friendly, but depends on how often this method would be called

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +75,636 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+
+        futures.push_back(Device->ReadPages(std::move(request)));
+    }
+
+    return WaitAll(futures).Apply(
+        [self = shared_from_this(),
+         futures = std::move(futures)](const TFuture<void>&)
+        {
+            TVector<TString> pages;
+            pages.reserve(self->PageCount);
```
asorotsky, line 379: nit: with TVector you can write pages(self->PageCount)

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: fixed

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
asorotsky: actually pages(Reserve(count)).... never mind

asorotsky (dismissed):

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+
+constexpr ui64 SuperblockMagic = 0x4B4253'53555042ULL;   // KBS'SUPB
+constexpr ui64 EntryMagic = 0x4B4253'454E5452ULL;        // KBS'ENTR
+constexpr ui32 StoreFormatVersion = 1;
+
+constexpr ui64 SuperblockSlotCount = 2;
+
+// Magic, Seq, Key, PayloadSize, Version, PageIndex, PageCount, Crc. The
+// checksum covers the fields before it and the chunk that follows.
+constexpr ui32 EntryHeaderSize = 4 * sizeof(ui64) + 4 * sizeof(ui32);
```
qkrorlqr: better to put those fields into a struct and use `sizeof(TThatStruct)`

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+
+struct TSuperblock
+{
+    ui64 Seq = 0;
+    ui64 ErasedBelowKey = 0;
+};
+
+ui64 EntryPageCountFor(ui64 payloadSize, ui64 chunkCapacity)
+{
+    return Max<ui64>(1, (payloadSize + chunkCapacity - 1) / chunkCapacity);
```
qkrorlqr: let's extract the rounding logic into a separate helper

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+}
+
+ui64 ChunkSizeOf(const TEntryHeader& header, ui64 chunkCapacity)
+{
+    if (header.PageIndex + 1 < header.PageCount) {
+        return chunkCapacity;
+    }
+    return header.PayloadSize - chunkCapacity * header.PageIndex;
```
qkrorlqr: looks like PayloadSize minus some offset

I don't understand the logic

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+}
+
+TBuffer
+MakeEntryPage(const TEntryHeader& header, TStringBuf chunk, ui32 pageSize)
+{
+    TBuffer page(pageSize);
+    TBufferOutput out(page);
+
+    Save(&out, EntryMagic);
```
qkrorlqr: I don't see any value in ysaveload

Let's either go with some standard extensible format like flatbuffers / protobuf or just reinterpret_cast and memcpy structs directly

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+
+    out.Write(chunk.data(), chunk.size());
+    page.Fill('\0', pageSize - page.Size());
+
+    return page;
+}
+
+// Returns the header and the chunk of a valid entry page, nothing for a page
+// that holds anything else.
+std::optional<std::pair<TEntryHeader, TStringBuf>> ParseEntryPage(
```
qkrorlqr: no `std::pair`s please, let's use structs with properly named fields

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+// checksum covers the fields before it and the chunk that follows.
+constexpr ui32 EntryHeaderSize = 4 * sizeof(ui64) + 4 * sizeof(ui32);
+
+// Magic, Seq, ErasedBelowKey, Version, Crc. The checksum covers the fields
+// before it.
+constexpr ui32 SuperblockSize = 3 * sizeof(ui64) + 2 * sizeof(ui32);
+
+constexpr ui64 MaxPagesPerReadRequest = 1024;
+
+struct TEntryHeader
```
qkrorlqr, line 97: this struct needs comments, currently the meaning of the fields is not clear

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+
+    return std::make_pair(header, chunk);
+}
+
+TBuffer MakeSuperblockPage(const TSuperblock& superblock, ui32 pageSize)
+{
+    TBuffer page(pageSize);
+    TBufferOutput out(page);
+
+    Save(&out, SuperblockMagic);
```
qkrorlqr: `struct TSuperBlock { ... fields ... }; reinterpret_cast / memcpy, etc`

qkrorlqr: or flatbuffers

qkrorlqr: ysaveload - это ни туда, ни сюда

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+    const ui64 PageCount;
+    const ui32 PageSize;
+    const ui64 ChunkCapacity;
+    const IDevicePageStorePtr Pages;
+
+    TAdaptiveLock Lock;
+
+    bool RestoreStarted = false;
+    bool Restored = false;
+    ui64 NextSeq = 1;
```
qkrorlqr, line 327: what is Seq / NextSeq? comments required

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+            return MakeFuture(MakeError(
+                E_ARGUMENT,
+                TStringBuilder() << "key " << key << " is erased"));
+        }
+
+        seq = NextSeq++;
+    }
+
+    const ui64 pageCount = EntryPageCountFor(buffer.Size(), ChunkCapacity);
+    if (pageCount > Max<ui32>()) {
```
qkrorlqr: Max<ui32>() sounds like way too much

2^32 x PageSize (4KiB) == 16TiB - that's way above a reasonable limit

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+    pages.reserve(pageCount);
+
+    TStringBuf payload(buffer.Data(), buffer.Size());
+    for (ui64 i = 0; i < pageCount; ++i) {
+        TEntryHeader header = {
+            .Seq = seq,
+            .Key = key,
+            .PayloadSize = buffer.Size(),
+            .PageIndex = static_cast<ui32>(i),
+            .PageCount = static_cast<ui32>(pageCount),
```
qkrorlqr: why do we need to store pageCount in each page's header?

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +80,641 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+                entry.Seq = seq;
+            } else {
+                // a newer write of the key has landed already
+                stalePages = locations;
+            }
+        }
+    }
+
+    if (!stalePages.empty()) {
+        Y_UNUSED(Pages->Free(stalePages));
```
qkrorlqr: let's use the following pattern:
```
auto r = DoSomething();
Y_SOME_MACRO(r);
```
instead of
```
Y_SOME_MACRO(DoSomething());
```

to avoid the case like:
```
Y_DEBUG_ABORT_UNLESS(DoSomething());
```
// works in debug mode and in relwithdebinfo mode but not in release mode

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: added comment explaining how the class works, there is also information about Seq.

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: added comment explaining how the class works. Renamed PageIndex,PageCount -> ChunkIndex,ChunkCount.

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: used reinterpret_cast and memcpy structs directly

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: removed

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: removed; right after that, we have another check

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: Because of the buffer might not have fit onto a single page, so when we restore data, we need to understand how many pages contain data for this buffer, and the index of each pages.
Renamed pageCount -> chunkCount.

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +83,678 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+            continue;
+        }
+
+        maxSeq = Max(maxSeq, parsed->Seq);
+        if (!superblock || superblock->Seq < parsed->Seq) {
+            superblock = *parsed;
+            superblockSlot = slot;
+        }
+    }
+
```
qkrorlqr, line 578: Shouldn't we verify that we successfully parsed at least one superblock page? Or, if we haven't, then that what we read is an empty page?

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +83,678 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+    };
+
+    THashMap<ui64, TWinner> winners;
+
+    for (const auto& [id, candidate]: candidates) {
+        if (superblock && id.Key < superblock->ErasedBelowKey) {
+            continue;
+        }
+
+        if (candidate.Broken) {
```
qkrorlqr, line 641: again probably some kind of verification is missing - for most of the metadata we can be sure that at least for one of the seq nos the entry should not be broken - at least up to the last acked lsn

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +83,678 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+                // a newer write of the key has landed already
+                stalePages = locations;
+            }
+        }
+    }
+
+    if (!stalePages.empty()) {
+        auto freeError = Pages->Free(stalePages);
+        if (HasError(freeError)) {
+            STORAGE_ERROR(
```
qkrorlqr, line 741: we'll need metrics in the future - this should be reported to some metrics

qkrorlqr, line 741: or maybe reported via the returned error

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -53,19 +83,678 @@ class TInMemoryKeyBufferStore final: public IKeyBufferStore
...
+        }
+
+        erasedAny = end != Entries.begin();
+        Entries.erase(Entries.begin(), end);
+    }
+
+    if (!pagesToFree.empty()) {
+        auto freeError = Pages->Free(pagesToFree);
+        if (HasError(freeError)) {
+            STORAGE_ERROR(
```
qkrorlqr, line 783: same

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -75,4 +764,26 @@ IKeyBufferStorePtr CreateInMemoryKeyBufferStore()
     return std::make_shared<TInMemoryKeyBufferStore>();
 }
 
+IKeyBufferStorePtr CreateDeviceKeyBufferStore(
+    ILoggingServicePtr logging,
+    IDevicePtr device,
+    ui64 pageCount,
+    ui32 pageSize)
+{
+    Y_ABORT_UNLESS(
```
qkrorlqr, line 807: better to return TError

qkrorlqr (dismissed):

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: Both superblock pages can be empty. We write the superblock only when advance lsn. If app crashes while writing first superblock, the page will not remain empty, and this is normal.

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: added ReadOnly mode

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: Added a check: if the superblock is missing, at least one page must be empty.

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
arigachnyy: we have some checks on higher levels.

asorotsky (dismissed):

cloud/storage/core/libs/journalled_device/key_buffer_store.cpp
```diff
@@ -250,6 +250,12 @@ std::optional<TSuperblock> ParseSuperblockPage(TStringBuf page, ui32 pageSize)
     return header;
 }
 
+bool IsEmptyPage(TStringBuf page, ui32 pageSize)
+{
+    return page.size() == pageSize &&
+           AllOf(page, [](char c) { return c == '\0'; });
```
qkrorlqr: not sure whether perf is important specifically for the usecases of this func, but the idiomatic and more performant approach is this:
```
*p == 0 && memcmp(p, p + 1, len - 1) == 0
```

qkrorlqr (approved):

asorotsky (approved):

PR #7135, opened 2026-09-11
issue-6956: implement the journal on top of the record chain and page index

arigachnyy:
### Notes
TJournal now restores its state from the meta store, writes records through the page store, serves reads and tail reads from the page index, advances the last acked lsn and drives the flush cycle.

### Issue


asorotsky (dismissed):

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+`TDevicePageStore` owns the page allocation of a device: `Allocate`,
+`AllocateAt` (used by restore to re-claim the pages of the restored records),
+`Free`, and page-granular `Read`/`Write`. Allocation is guarded by a lock and
+may be called from anywhere; the data methods validate against the allocation
+state under that lock but issue the device request outside of it, so the caller
+must not free pages it is still reading or writing. In `Trusted` mode the
+validation is skipped everywhere but `AllocateAt`.
+
+### `key_buffer_store.h` - the log metadata store
```
qkrorlqr, line 155: this needs to have a diagram showing how everything is structured - for example, how multi-page values are stored

and currently it's not clear how to restore the whole thing if the pages don't have to be contiguous - it means that they should be somehow linked and I don't see a description for how it's done

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
...
+on a write it would be mapped twice, leaving the winner to the order the groups
+happen to be in and handing the tail two groups claiming the same page. Ranges
+covering no pages are ignored by the check.
+
+### Write
+
+1. The page groups must sit inside the data device and must not intersect
+   (`E_ARGUMENT`).
+2. `Lsn` must not be the reserved metadata key (`Max<ui64>`); a record at or
+   below `LastIndexedLsn` is answered `S_ALREADY`.
```
qkrorlqr, line 204: IMO `LastIndexedLsn` is not the best name because it doesn't fully reflect its meaning from the client's standpoint

it's actually the last "complete" lsn in the sense that everything before it represents a durable chain without gaps

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
+# journalled/impl
```
qkrorlqr, line 1: this needs to be referenced from https://github.com/ydb-platform/nbs/blob/main/doc/filestore/design/storage/fastshard/storage-node.md

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
...
+### Read
+
+`TJournal::Read` pins the flushed barrier, looks the ranges up in the page
+index, reads the mapped pages out of the log data store, and returns them as
+page groups - only the pages the journal actually has.
+
+`TJournalledDeviceV2::DoReadPages` pins the indexed barrier, asks the journal,
+computes the gaps (`MakeMissingRequest`), reads those from the data device, and
+stitches the two responses together in the shape of the request
+(`MergeResponses`). A page missing from both is an `E_INVALID_STATE`.
```
qkrorlqr, line 229: why `E_INVALID_STATE` and not `E_FS_NOENT`?

qkrorlqr, line 229: or `E_NOT_FOUND` or another specialized code, e.g. you can introduce a new one - `E_PAGE_NOT_FOUND`

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+A record says "these pages now hold this content", and chains to its
+predecessor through `PrevLsn`. The lsns are the writer's, the journal never
+invents them; it only requires `PrevLsn < Lsn`. The `PageMappings` are filled in
+by the journal once it knows where in the log the payload landed, and they are
+what gets serialized into the metadata store - the payload itself stays on the
+log data device.
+
+The file also carries the (de)serialization of a record and of the journal
+metadata (`TJournalMetadata`: a format version and `LastAckedLsn`).
```
qkrorlqr, line 105: what is `LastAckedLsn`? it's the first place in the doc where it's mentioned - should be defined at this point

qkrorlqr: if it's actually our low watermark, let's call it `LsnLowWatermark` explicitly

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
...
+   client request behind a flush, so the write carries the `deviceUUID` and
+   `backgroundClientId` the device was built with - the disk agent refuses a
+   write without a device UUID and checks access by the client id. A record
+   that maps no pages is a chain link and nothing more, so it is marked flushed
+   without a device request at all.
+3. When the loop ends - nothing is left to flush, or a flush failed and will be
+   retried on the next cycle - `CleanupFlushedRecords` erases the
+   metadata entries below the barrier (`EraseBelow`, keyed by `PrevLsn`), drops
+   the index mappings at or below it, removes the records from the chain and
+   frees their log pages. Records stranded below the new watermark - ready, but
```
qkrorlqr, line 266: how can we end up having stranded records?

cloud/fastshard/journal/impl/README.md
```diff
@@ -0,0 +1,319 @@
...
+that work is done. The stores are free-threaded: their state is under a lock
+and the device requests are issued outside of it. The chain, the index and the
+barriers each guard themselves with a spinlock.
+
+## Errors worth knowing
+
+| code | when |
+| --- | --- |
+| `S_ALREADY` | the record is already indexed, or the watermark is not new |
+| `E_REJECTED` | the log data or metadata store is out of pages, an erase or a watermark advance is already in progress, a key erased while it was being written |
```
qkrorlqr, line 304: > a key erased while it was being written

why is it `E_REJECTED`? what will the client get upon a retry attempt?

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+        for (const auto& range: ranges) {
+            total += range.PageCount;
+        }
+        return total;
+    };
+
+    const ui64 pageCount = totalPageCount(requestPageRanges);
+    STORAGE_VERIFY(
+        pageCount == totalPageCount(locations),
+        "PageCount",
```
qkrorlqr, line 103: this argument is not about the name of the parameter, it's about the type of the affected entity - in this case it should be something like "Device" (and should probably be added to `TWellKnownEntityTypes` or how's that called) and the next argument should contain deviceUUID

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+        pageCount);
+
+    TVector<TPageMapping> mappings;
+
+    size_t requestIndex = 0;
```
qkrorlqr, line 108: why not ui64 as well

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+    ui64 requestOffset = 0;
+    ui64 locationOffset = 0;
+
+    while (requestIndex < requestPageRanges.size() &&
+           locationIndex < locations.size())
+    {
+        const auto& requestRange = requestPageRanges[requestIndex];
+        const auto& location = locations[locationIndex];
+
+        const ui64 requestLeft = requestRange.PageCount - requestOffset;
```
qkrorlqr, line 119: you mean request**s**Left and location**s**Left?

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
-    TFuture<TResultOrError<ui64>> Restore() override
-    {
-        return MakeFuture<TResultOrError<ui64>>(
-            MakeError(E_NOT_IMPLEMENTED, "Restore"));
+    // Restoring
+
+    // Restores the journal state and returns lsn of the last indexed record
+    [[nodiscard]] NThreading::TFuture<TResultOrError<ui64>> Restore() override;
+
+    // Device API
```
qkrorlqr, line 193: nit: let's use the following style for the comments delineating large sections of the code or just blocks of code that have different purposes:
```
[[nodiscard]] NThreading::TFuture<TResultOrError<ui64>> Restore() override;

//
// Device API
//

[[nodiscard]] auto Write(NCloud::NProto::TWriteLogRecordRequest request)
    -> NThreading::TFuture<
        NCloud::NProto::TWriteLogRecordResponse> override;
```

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+                        << "log record with key " << key << " maps page range "
+                        << pageNo << "x" << location.PageCount
+                        << " which is outside the device of "
+                        << DevicePageCount << " pages");
+            }
+        }
+
+        if (!initialized) {
+            initialized = true;
+            auto lsn = Min(record->PrevLsn, LastAckedLsn.load());
```
qkrorlqr, line 321: why record->PrevLsn and not record->Lsn?

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+            return error;
+        }
+
+        auto insertResult = LogRecordChain.Insert(record);
+        if (HasError(insertResult)) {
+            return insertResult.GetError();
+        }
+
+        bool marked = LogRecordChain.MarkAsReady(record->PrevLsn);
+        STORAGE_VERIFY(marked, "MarkAsReady", record->PrevLsn);
```
qkrorlqr, line 338: incorrect argument values (see other similar comment)

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+        return MakeFuture<TResponse>(TErrorResponse(
+            E_ARGUMENT,
+            TStringBuilder()
+                << "lsn " << lsn << " is reserved for journal metadata"));
     }
 
-    auto AdvanceLastAckedLsn(
-        NCloud::NProto::TAdvanceLsnLowWatermarkRequest request)
-        -> TFuture<NCloud::NProto::TAdvanceLsnLowWatermarkResponse> override
+    if (lsn <= LogPageIndex.GetLastIndexedLsn()) {
```
qkrorlqr, line 384: probably makes sense to do a lookup and return S_ALREADY only if this lsn is actually present in the metadata

otherwise it looks more like a non-retriable error

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
 
-    auto GetRecordToFlush(ui64 maxAllowedLsn) const
-        -> TFuture<TResultOrError<NCloud::NProto::TJournalRecord>> override
+    Y_DEFER
     {
-        Y_UNUSED(maxAllowedLsn);
+        if (!success) {
+            auto error = DataStore->Free(locations);
+            if (HasError(error)) {
+                STORAGE_ERROR(
```
qkrorlqr, line 448: an AppCriticalEvent would've helped

qkrorlqr: can be done in a separate PR, we can leave a TODO(#issue-no) here

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
-    void MarkRecordAsFlushed(ui64 lsn) override
+    record.PageMappings = CreatePageMappings(requestPageRanges, locations);
+
+    auto metaFuture = MetaStore->Write(record.PrevLsn, SerializeRecord(record));
+    if (const auto& error = Executor->WaitFor(metaFuture); HasError(error)) {
+        return error;
+    }
+
+    bool marked = LogRecordChain.MarkAsReady(record.PrevLsn);
+    STORAGE_VERIFY(marked, "MarkAsReady", record.Lsn);
```
qkrorlqr, line 468: same

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+            {.FirstPageNo = ref.GetFirstPageNo(),
+             .PageCount = ref.GetPageCount()});
     }
 
-    TFuture<NCloud::NProto::TError> CleanupFlushedRecords() override
+    auto lookup = LogPageIndex.Lookup(ranges, lsnBarrierGuard.GetLsn());
+
+    TResponse response;
+
+    // TODO: rename proto field into LastIndexedLogSequenceNumber
```
qkrorlqr, line 499: TODO(#issue-no)

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+
+TFuture<NCloud::NProto::TReadJournalTailResponse> TJournal::ReadTail(
+    NCloud::NProto::TReadJournalTailRequest request) const
+{
+    using TResponse = NCloud::NProto::TReadJournalTailResponse;
+
+    auto lsnBarrierGuard =
+        FlushedLsnBarrier.AcquireAtLeast(request.GetAfterLogSequenceNumber());
+
+    auto lastAckedLsn = LastAckedLsn.load();
```
qkrorlqr, line 518: nit: no `auto` for primitives plz

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+    auto afterLsn = Max(lsnBarrierGuard.GetLsn(), lastAckedLsn);
```
qkrorlqr, line 519: too many lsns used here - we need a comment explaining why we first do AcquireAtLeast (and not just Acquire) and even after that we still need to Max with lastAckedLsn

`LastAckedLsn.load()` also implies that we're concerned about thread-safety here but I thought that we actually run all these methods via a coroutine executor - do we use multiple executor threads? if yes, then we need thread-safety and I don't understand why a load() + a-group-of-other-calls pattern doesn't introduce a race - why can't some other thread change LastAckedLsn after we load() it?

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+    return MakeFuture(std::move(response));
+}
+
+TFuture<NCloud::NProto::TAdvanceLsnLowWatermarkResponse>
+TJournal::AdvanceLastAckedLsn(
+    NCloud::NProto::TAdvanceLsnLowWatermarkRequest request)
+{
+    using TResponse = NCloud::NProto::TAdvanceLsnLowWatermarkResponse;
+
+    const auto lastAckedLsn = request.GetLsnLowWatermark();
```
qkrorlqr, line 550: nit: no `auto` for primitives

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+        if (HasError(error)) {
+            return MakeFuture<TResponse>(TErrorResponse(error));
+        }
+    }
+
+    return MakeFuture(std::move(response));
+}
+
+TFuture<NCloud::NProto::TAdvanceLsnLowWatermarkResponse>
+TJournal::AdvanceLastAckedLsn(
```
qkrorlqr, line 545: looks more like AdvanceLsnLowWatermark()

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+
+    if (lastAckedLsn > lastIndexedLsn) {
+        return MakeFuture<TResponse>(TErrorResponse(
+            E_ARGUMENT,
+            TStringBuilder()
+                << "lsn low watermark " << lastAckedLsn
+                << " reaches past the last indexed lsn " << lastIndexedLsn));
+    }
+
+    if (AdvancingLastAckedLsn.exchange(true) == true) {
```
qkrorlqr, line 561: I assume we don't actually need memory_order_seq_cst everywhere

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+TFuture<TResultOrError<NCloud::NProto::TJournalRecord>>
+TJournal::GetRecordToFlush(ui64 maxAllowedLsn) const
+{
+    using TResult = TResultOrError<NCloud::NProto::TJournalRecord>;
+
+    auto lsnBarrierGuard = FlushedLsnBarrier.Acquire();
+
+    NCloud::NProto::TJournalRecord response;
+
+    auto record = LogRecordChain.GetNext(lsnBarrierGuard.GetLsn());
```
qkrorlqr, line 599: a comment would be very useful, otherwise it's very hard to maintain all the "where we use prev lsn, where we use actual lsn, how we use barriers, etc" context while reading this

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+TFuture<NCloud::NProto::TError> TJournal::CleanupFlushedRecords()
+{
+    auto eraseUpToLsn = FlushedLsnBarrier.GetBarrierLsn();
+
+    auto future = MetaStore->EraseBelow(eraseUpToLsn);
+    if (const auto& error = Executor->WaitFor(future); HasError(error)) {
+        return MakeFuture(error);
+    }
+
+    auto recordsOrError = LogRecordChain.EraseUpTo(eraseUpToLsn);
```
qkrorlqr, line 631: a comment once again would be very useful

it's hard to read the code having to continuously track the logic like "meta-store maps prev-lsn to meta, chain and index EraseUpTo calls accept lsn, so if we erase-below from meta-store then we actually keep the record whose PrevLsn == eraseUpToLsn so we should erase all the records whose Lsn == eraseUpToLsn from the chain and the index to maintain consistency"

or maybe it would make sense to actually make the methods of all these components consistent to be able to call either EraseBelow for all of them or EraseUpTo for all of them

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+    STORAGE_VERIFY_C(
+        !HasError(recordsOrError),
+        "Lsn",
```
qkrorlqr, line 634: same

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+        eraseUpToLsn,
+        FormatError(recordsOrError.GetError()));
+
+    auto records = recordsOrError.ExtractResult();
+    LogPageIndex.EraseUpTo(eraseUpToLsn);
+
+    auto result = MakeError(S_OK);
+    for (const auto& record: records) {
+        // a flushed record has its promise set, a ready record stranded
```
qkrorlqr, line 643: why can we end up having "stranded" records?

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -11,94 +26,734 @@ namespace {
...
+    }
+
+    // one buffer per page, in the order the ranges were asked for
+    const auto& buffers = response.GetResult();
+
+    size_t bufferIndex = 0;
+    for (const auto& [pageNo, location]: mappings) {
+        STORAGE_VERIFY(
+            bufferIndex + location.PageCount <= buffers.size(),
+            "PageNo",
```
qkrorlqr, line 681: same

cloud/fastshard/journal/impl/log_record.cpp
```diff
@@ -1 +1,128 @@
 #include "log_record.h"
+
+#include <cstring>
+#include <type_traits>
+
+namespace NCloud::NJournalled {
+
+using namespace NThreading;
+
+////////////////////////////////////////////////////////////////////////////////
```
qkrorlqr, line 10: nit: this should go after `namespace {`

cloud/fastshard/journal/impl/journal.cpp
qkrorlqr: I understand why it's the last erased lsn but I don't understand why it's passed here: LogPageIndex.InitLastIndexedLsn(lsn);

qkrorlqr: Ok, I understood it, but a comment is required anyway

Issue #7136, opened 2026-09-11
[Filestore] cross-shard exchange rename leaves source ref ShardId/ShardNodeName empty


qkrorlqr:
Basically cross-shard exchange rename (i.e. rename op with RENAME_EXCHANGE) doesn't properly work - in most of the cases when the destination node ref exists the values of ShardId and ShardNodeName of the destination node ref are not transported back to the source shard and thus the source node ref is not filled.

Nodes aren't lost but they become orphaned.

PR #7137, opened 2026-09-11
issue-7136: fixing cross-shard RenameNode with EXCHANGE flag - properly returning OldTargetNodeShard{Id,Name} to the source shard

qkrorlqr:
### Notes
The main problem: `args.Response` was filled but not copied to the actual response - fixed it, added a ut.

Other changes:
* `RenameNodeInDestination` with `EXCHANGE` flag returns `OldTargetNode` ref upon `S_ALREADY`
* `ListNodes` doesn't crash upon missing local node - reports a critical event instead
* `ResponseLogEntryTTL` is now 1 day by default
* `OpLogEntryCount` and `ResponseLogEntryCount` sensors are now aggregatable sums

### Issue
https://github.com/ydb-platform/nbs/issues/7136
https://github.com/ydb-platform/nbs/issues/2667

### DIFF
--- a/cloud/filestore/libs/diagnostics/critical_events.h
+++ b/cloud/filestore/libs/diagnostics/critical_events.h
@@ -73,6 +73,7 @@ namespace NCloud::NFileStore{
     xxx(AvailabilityCountersUnavailableInterval)                               \
     xxx(AvailabilityCountersMissingIntervals)                                  \
     xxx(PersistentStateSessionDirNotEmpty)                                     \
+    xxx(ListNodesLocalNodeNotFound)                                            \
 // FILESTORE_CRITICAL_EVENTS
 
 #define FILESTORE_CRITICAL_EVENTS_WITHOUT_LOGGING(xxx)                         \
@@ -98,6 +99,7 @@ namespace NCloud::NFileStore{
     xxx(UnexpectedLocalNode)                                                   \
     xxx(NoRenameNodeInDestinationRequest)                                      \
     xxx(BadChildRefUponCommitRenameNodeInSource)                               \
+    xxx(EmptyOldTargetNodeInExchangeRename)                                    \
     xxx(FailedToLockNodeRef)                                                   \
     xxx(InvalidNodeRefUponCompleteUnlinkNode)                                  \
     xxx(UnknownOpLogEntry)                                                     \
--- a/cloud/filestore/libs/storage/core/config.cpp
+++ b/cloud/filestore/libs/storage/core/config.cpp
@@ -356,7 +356,7 @@ using TAliases = NProto::TStorageConfig::TFilestoreAliases;
         NProto::LNSM_NAME_ONLY                                                )\
     xxx(UseListNodesInternal,              bool,      false                   )\
                                                                                \
-    xxx(ResponseLogEntryTTL,                TDuration,  TDuration::Hours(1)   )\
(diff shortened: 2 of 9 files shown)


debnatkh (dismissed):

yegorskii (dismissed):

debnatkh (approved):

yegorskii (approved):

qkrorlqr:
Claude (Fable 5):

> I read the full diff (2 commits, 9 files). Overall the fix is sound — args.Response is now actually moved into the RPC response in CompleteTx_RenameNodeInDestination, the S_ALREADY path fills the old-target fields, and the tests cover both the happy path and the same-request-id replay. A few things I'd consider serious enough to raise:
> 
> 1. Source side detects the broken state but proceeds anyway (tablet_actor_renamenode_source.cpp). In ExecuteTx_CommitRenameNodeInSource, when OldTargetNodeShardId/ShardNodeName are empty you report EmptyOldTargetNodeInExchangeRename — and then fall through to CreateNodeRef regardless, persistently writing the very dangling ref the comment warns about. If aborting the commit isn't possible at this point in the distributed op, that's defensible, but then it deserves a comment explaining why writing a known-bad ref is preferable to skipping the ref creation (which would "just" leak the old target node instead of also leaving a broken ref that ListNodes will later trip on). Also worth double-checking there's no legitimate case where the old target is a purely local node in the destination (identified by node id, with empty shard fields) — if such a case exists, this critical event will false-fire on every exchange.

That's deliberate, keeping the old node ref after it's been transferred to the destination is worse

> 
> 2. Does the S_ALREADY path survive a retry? The old-target fields for isSameExternalNode are set in PrepareTx, where (as far as I can tell) no tx executes and nothing is written to the response log. That's fine if the same-node condition is stable and re-derivable on every retry — which it should be, since the ref state doesn't change — but if any interleaved operation changes the ref between retries, the retry gets a different answer than the original. Given your own test comment says this path only becomes reachable after #2667, it would be good to confirm that assumption holds under the future hardlink implementation too.

Yes, this condition is stable

> 
> 3. UnsafeGetNodeRef behavior change is a silent bug fix bundled in (tablet_actor_unsafe_node_ops.cpp). NodeId → ChildNodeId is clearly correct (the old code returned the parent's id as ChildId), but it changes the output of an existing admin endpoint. If any ops tooling or scripts consume ChildId today, they were consuming garbage — still, this deserves its own mention in the PR description/changelog rather than riding along unremarked.

There's no code that relies on that, it's safe to change

> 
> 4. ResponseLogEntryTTL 1h → 1d is a 24× retention increase. On a busy tablet this multiplies response-log storage and the cost of whatever cleanup scans it. You've made the count sensors aggregatable in the same PR, which suggests you're planning to watch this — but it's worth stating whether there's a size cap or only TTL-based cleanup, since TTL alone with 24× retention can bite on hot filesystems.

Checked in real clusters - currently the monitorings show that all ResponseLogEntries are deleted on the happy path, so this TTL either doesn't fire at all or almost never fires
So even in 24h we won't accumulate too many entries
But I changed ResponseLogEntryCount sensor specifically to be able to make a proper alert for suspiciously large numbers of such entries

> 
> Not an issue, just confirming intent: replacing TABLET_VERIFY(childNode) in ListNodes with a crit event + E_INVALID_STATE is the right call for prod resilience, and the corrupted-ref test covers it.

Issue #7070, opened 2026-09-07
[NBS] NBD: a stalled write overwrites newer data after the connection is restored


sharpeye:
After a prolonged loss of connectivity to the storage nodes, a block silently reverts to content written before the outage — after the connection was restored and after newer writes to that same block were acknowledged. No I/O errors are reported; every write returns success.

The in-flight write gets stuck in the durable client, which grows its backoff. The kernel nbd driver times the request out, reconnects and re-sends it; to NBS that is an independent request, so it completes and is acknowledged. Later writes to the same block complete too. Then the original request's backoff expires, it executes, and it overwrites the newer data with its stale payload. This would mean in-flight requests are not cancelled when the connection that carried them dies. Needs verification against the durable client's retry path and its backoff settings — the observed delay is around 6s from the re-sent request's ack to the revert.

PR #7139, opened 2026-09-11
issue-7070: prevent stale NBD requests from overwriting data after reconnect

sharpeye:
Fixes a race during NBD reconnection after a temporary YDB outage. Before starting a new connection, the server waits for all requests from the previous connection to complete, preventing delayed durable requests from overwriting newer data.

Adds an NBD endpoint drain mechanism and tests covering request serialization across connections.

### DIFF
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -702,6 +702,12 @@ private:
         const NProto::TStartEndpointRequest& request,
         const TSessionInfo& sessionInfo);
 
+    NProto::TError DrainAllEndpointSockets(
+        const NProto::TStartEndpointRequest& request);
+
+    NProto::TError DrainEndpointSocket(
+        const NProto::TStartEndpointRequest& request);
+
     void CloseAllEndpointSockets(const NProto::TStartEndpointRequest& request);
     void CloseEndpointSocket(const NProto::TStartEndpointRequest& request);
 
@@ -1498,10 +1504,27 @@ void TEndpointManager::DoProcessException(
         endpoint->Device.reset();
     }
 
+    const auto& socketPath = endpoint->Request->GetUnixSocketPath();
+
+    STORAGE_INFO(prefix << " drain socket");
+    if (auto error = DrainAllEndpointSockets(*endpoint->Request);
+        HasError(error))
+    {
+        STORAGE_ERROR(
+            prefix << " failed to drain socket: " << FormatError(error));
+        context->Generation++;
+        ProcessException(std::move(context), std::move(prefix));
+        return;
+    }
+
+    auto endpointIt = Endpoints.find(socketPath);
+    if (endpointIt == Endpoints.end() || endpointIt->second != endpoint) {
+        STORAGE_WARN(prefix << " endpoint is down, cancel restart");
+        return;
+    }
+
     STORAGE_INFO(prefix << " close socket");
     CloseAllEndpointSockets(*endpoint->Request);
-
-    auto socketPath = endpoint->Request->GetUnixSocketPath();
 
     STORAGE_INFO(prefix << " update error handler");
     NbdErrorHandlerMap->Erase(socketPath);
@@ -1587,6 +1610,41 @@ NProto::TError TEndpointManager::OpenEndpointSocket(
         sessionInfo.Volume,
         sessionInfo.Session);
 
+    return Executor->WaitFor(future);
+}
+
+// waits for requests accepted through the endpoint sockets to complete
+NProto::TError TEndpointManager::DrainAllEndpointSockets(
+    const NProto::TStartEndpointRequest& request)
+{
+    auto error = DrainEndpointSocket(request);
+    if (HasError(error)) {
+        return error;
+    }
+
+    auto nbdRequest = CreateNbdStartEndpointRequest(request);
+    if (nbdRequest) {
+        STORAGE_INFO("Drain additional endpoint: "
+            << nbdRequest->GetUnixSocketPath().Quote());
+        error = DrainEndpointSocket(*nbdRequest);
+    }
+
+    return error;
+}
+
+NProto::TError TEndpointManager::DrainEndpointSocket(
+    const NProto::TStartEndpointRequest& request)
+{
+    auto ipcType = request.GetIpcType();
+    const auto& socketPath = request.GetUnixSocketPath();
+
+    auto listenerIt = EndpointListeners.find(ipcType);
+    STORAGE_VERIFY(
+        listenerIt != EndpointListeners.end(),
+        TWellKnownEntityTypes::ENDPOINT,
+        socketPath);
+
+    auto future = listenerIt->second->DrainEndpoint(socketPath);
     return Executor->WaitFor(future);
 }
 
--- a/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
+++ b/cloud/blockstore/libs/endpoints/endpoint_manager.cpp
@@ -1517,6 +1517,8 @@ void TEndpointManager::DoProcessException(
         return;
     }
 
+    // Draining may yield to another coroutine, which can stop or replace the
+    // endpoint. Do not resume a stale restart.
     auto endpointIt = Endpoints.find(socketPath);
     if (endpointIt == Endpoints.end() || endpointIt->second != endpoint) {
         STORAGE_WARN(prefix << " endpoint is down, cancel restart");
@@ -1544,7 +1546,7 @@ void TEndpointManager::DoProcessException(
     }
 
     if (hasDevice) {
-        STORAGE_INFO(prefix << "start device");
+        STORAGE_INFO(prefix << " start device");
         auto device = NbdDeviceFactory->Create(
             TNetworkAddress(TUnixSocketPath(socketPath)),
             endpoint->Request->GetNbdDeviceFile(),
@@ -1553,7 +1555,7 @@ void TEndpointManager::DoProcessException(
         auto startDeviceFuture = device->Start();
         error = Executor->WaitFor(startDeviceFuture);
         if (HasError(error)) {
-            STORAGE_ERROR(prefix << "failed to start device: "
+            STORAGE_ERROR(prefix << " failed to start device: "
                 << FormatError(error));
             context->Generation++;
             ProcessException(std::move(context), std::move(prefix));
--- a/cloud/blockstore/libs/nbd/server.cpp
+++ b/cloud/blockstore/libs/nbd/server.cpp
@@ -66,11 +66,16 @@ private:
     ILimiterPtr Limiter;
     IServerHandlerPtr Handler;
     TSocketHolder Socket;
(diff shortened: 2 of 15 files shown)


cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -1498,11 +1504,30 @@ void TEndpointManager::DoProcessException(
         endpoint->Device.reset();
     }
 
+    const auto& socketPath = endpoint->Request->GetUnixSocketPath();
+
+    STORAGE_INFO(prefix << " drain socket");
+    if (auto error = DrainAllEndpointSockets(*endpoint->Request);
```
tpashkin, line 1510: wouldn't it block the whole thread?

sharpeye: DrainAllEndpointSockets calls DrainEndpointSocket, which in turn calls:
```
auto future = listenerIt->second->DrainEndpoint(socketPath);
return Executor->WaitFor(future);
```
This doesn't block the thread, because WaitFor simply switches execution to another coroutine.

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -140,17 +154,32 @@ class TConnection final
...
         if (Limiter) {
             if (!Limiter->Acquire(requestBytes)) {
+                CompleteRequest();
                 return false;
             }
 
             InFlightBytes += requestBytes;
         }
+
+        if (IsShuttingDown()) {
```
drbasic, line 179: зачем тут повторная проверка?
Добавь комментарий что мы убираем гонку между ShuttingDown и ActiveRequests

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -66,25 +66,32 @@ class TConnection final
...
     TSocketHolder Socket;
+    const TFuture<void> Ready;
 
     TContLockFreeQueue<TServerResponsePtr> ResponseQueue;
 
     size_t InFlightBytes = 0;
+    std::atomic<size_t> ActiveRequests = 0;
+    std::atomic<bool> ReceiveFinished = false;
     std::atomic_flag ShuttingDown = false;
+    std::atomic_flag DrainCompleted = false;
```
drbasic: кажется DrainCompleted не нужен, просто делай DrainResult.TrySetValue();

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -66,25 +66,32 @@ class TConnection final
     ILimiterPtr Limiter;
     IServerHandlerPtr Handler;
     TSocketHolder Socket;
+    const TFuture<void> Ready;
 
     TContLockFreeQueue<TServerResponsePtr> ResponseQueue;
 
     size_t InFlightBytes = 0;
+    std::atomic<size_t> ActiveRequests = 0;
```
drbasic, line 77: можно комментарий в каком порядке эти поля выставляются, какие гарантии, что например DrainCompleted может переходить только в одном направлении и т.д.

cloud/blockstore/libs/nbd/server.cpp
drbasic, line 206: метод Receive(TCont* c) вызывается один раз за всю жизнь?

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -349,6 +408,25 @@ class TEndpoint final
...
+                Listener->Stop();
+            }
+            return NProto::TError();
+        });
+
+        if (HasError(error) || !Connection) {
+            return MakeFuture(std::move(error));
+        }
+
+        return Connection->Drain().Apply([] (const auto& future) {
```
drbasic: если использовать явный тип const TFuture<xxx>& future
то clangd перестает сходить с ума и может подсказывать полезное. 
плюс читателю проще.

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -516,6 +596,30 @@ class TExecutorThread final
         });
     }
 
+    TFuture<NProto::TError> DrainEndpoint(const TString& address)
+    {
+        return Executor->Execute([this, address] {
```
drbasic: ты передал в другой поток this. кто гарантирует что он живой будет?

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -708,6 +812,36 @@ class TServer final
...
+    {
+        if (AtomicGet(ShouldStop) == 1) {
+            NProto::TError error;
+            error.SetCode(E_REJECTED);
+            error.SetMessage("NBD server is stopped");
+            return MakeFuture(error);
+        }
+
+        auto address = PrintHostAndPort(listenAddress);
+        TExecutorThread* executorThread;
```
drbasic: = nullptr;

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -708,6 +812,36 @@ class TServer final
...
+                error.SetMessage(TStringBuilder()
+                    << "endpoint " << address.Quote()
+                    << " has already been stopped");
+                return MakeFuture(error);
+            }
+
+            executorThread = it->second;
+        }
+
+        return executorThread->DrainEndpoint(address);
```
drbasic: есть гарантия что TExecutorThread остается живой?

sharpeye: Теперь такая гарантия не нужна - внутри DrainEndpoint ничего не прикапывается

cloud/blockstore/libs/nbd/server.cpp
sharpeye: Да, но почему нет?

drbasic (dismissed):

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -66,25 +66,38 @@ class TConnection final
...
     TSocketHolder Socket;
+    const TFuture<void> Ready;
 
     TContLockFreeQueue<TServerResponsePtr> ResponseQueue;
 
     size_t InFlightBytes = 0;
+
+    // Requests read from this connection that may still reach or be executing
+    // in the backend. Includes requests waiting in Limiter::Acquire.
+    std::atomic<size_t> ActiveRequests = 0;
```
tpashkin, line 77: Why not use `InFlightBytes` for this?

sharpeye: `InFlightBytes` tracks limiter quota and can reach zero before backend requests finish. `ActiveRequests` tracks request completion directly.

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
```diff
@@ -1498,11 +1504,30 @@ void TEndpointManager::DoProcessException(
...
+            prefix << " failed to drain socket: " << FormatError(error));
+        context->Generation++;
+        ProcessException(std::move(context), std::move(prefix));
+        return;
+    }
+
+    // Draining may yield to another coroutine, which can stop or replace the
+    // endpoint. Do not resume a stale restart.
+    auto endpointIt = Endpoints.find(socketPath);
+    if (endpointIt == Endpoints.end() || endpointIt->second != endpoint) {
```
tpashkin: This is actually true for a whole bunch of the following actions...

Guess I will have to rewrite the whole handler using `ProcessingSockets`

cloud/blockstore/libs/endpoints/endpoint_manager.cpp
tpashkin: I'd argue that extending API to explicitly drain NBD endpoints is completely unnecessary. The point at which this handler gets called is exactly the place where NBD endpoint could be drained automatically

cloud/blockstore/libs/nbd/server.cpp
tpashkin: How so? I see just a single place where it could get to 0 prematurely (after Send ends it's loop) and it's easily fixable

sharpeye: Yes, we could make it work, but we would need to keep `Send` running until `InFlightBytes` reaches zero and handle zero-length [requests](https://github.com/ydb-platform/nbs/blob/main/cloud/blockstore/libs/nbd/server_handler_ut.cpp#L522) separately. I'd rather not mix two concerns: `ActiveRequests` tracks exactly what drain waits for, while `InFlightBytes` remains limiter accounting.

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -66,25 +66,38 @@ class TConnection final
...
+
+    // Set after the receive loop exits, so ActiveRequests can no longer grow.
+    // Drain completes when this is set and ActiveRequests reaches zero.
+    std::atomic<bool> ReceiveFinished = false;
```
tpashkin: Do we actually need this though? `Recv` will insert `nullptr` into the `ResponseQueue` after exiting it's main loop, resulting in `ShutDown` being called from `Send` and `ShuttingDown` flag being set. They are indistinguishable for all intents and purposes

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -379,7 +455,12 @@ class TEndpoint final
             Executor,
             Limiter,
             HandlerFactory->CreateHandler(),
-            std::move(socket));
+            std::move(socket),
+            ready);
+
+        // Keep the whole chain even if this connection closes before it starts
+        // processing requests.
+        DrainResult = WaitAll(ready, Connection->GetDrainResult());
```
tpashkin, line 454: I don't think it works the way you intended it to work

So the idea behind this is to let `StopEndpoint` finish right away and get on with it. But if we look at the code of `TEndpointManager`, it immediately dismantles the session
https://github.com/ydb-platform/nbs/blob/991c41d748313831eaf34e18cfdf9cafe6096563/cloud/blockstore/libs/endpoints/endpoint_manager.cpp#L1227
So it looks like we are essentially cancelling requests in a very roundabout way

cloud/blockstore/libs/nbd/server.cpp
tpashkin: Oh well, I've tried

drbasic (dismissed):

cloud/blockstore/libs/nbd/server.cpp
```diff
@@ -330,12 +378,15 @@ class TEndpoint final
...
-    NProto::TError Stop(bool deleteSocket)
+    TFuture<NProto::TError> Stop(bool deleteSocket)
     {
-        return SafeExecute<NProto::TError>([&] {
+        TFuture<void> drainResult = MakeFuture();
+
+        auto error = SafeExecute<NProto::TError>([&] {
             if (Connection) {
+                drainResult = Connection->GetDrainResult();
                 Connection->Stop();
```
tpashkin, line 388: Why not return drain future from Stop?

tpashkin (approved):

drbasic (approved):

sharpeye:
#7070

PR #7141, opened 2026-09-11
issue-1751: [Filestore] CreateHandle with E_TRUNCATE should be executed under barrier when using WriteBackCache

e673:
### Notes
Resolve returning incorrect file size and corrupting the file by flush when opening it with `O_TRUNC` while having unflushed data in write-back cache.

Scenario to reproduce a problem:
1. Open a file.
2. Write some data
3. Open a file again with `O_TRUNC`
4. Read the data using the first handle.

The problem is resolved the same way as SetNoteAttr: CreateHandle is executed under a barrier that ensures that all prior writes are flushed and evicted from cache and no new flushes can take place.

### Issue


SvartMetal (dismissed):

e673:
> I'll pull the PR diff via the GitHub API to review the actual code.The clone worked. Now let me examine the full source files for context — the `Open`/`Create` paths, the barrier machinery, and `ResetMaxWrittenOffset` semantics.I fetched the PR, cloned the branch, and traced the change through the barrier machinery (`ExecuteRequestUnderBarrier` → `AcquireBarrier`/`CheckAndAcquireBarriers`/`ReleaseBarrier` → `ResetMaxWrittenOffset`), the FUSE `Open`/`Create` paths, and both test files. Overall verdict: the fix is correct and closely mirrors the proven `SetNodeAttr` pattern, tests are deterministic, and no memory-safety issues are introduced (there's no `memcpy`/`memmove`/`memset` or any raw buffer manipulation in the diff at all). Findings below, roughly by severity.
> 
> ## Real gap: the `Create()` path still truncates without a barrier
> 
> `TFileSystem::Create` (FUSE_CREATE, i.e. `open(O_CREAT|...)`) at fs_impl_data.cpp:302 still calls `Session->CreateHandle` directly, and `O_TRUNC` maps to `E_TRUNCATE` there too (`filestore.cpp:30`). So a truncating create that lands on an existing node bypasses the barrier and reproduces the original bug.
> 
> In practice this is nearly unreachable from a single mount: having unflushed data in the WBC implies the guest kernel holds the inode, so a positive dentry lookup routes `open(O_CREAT|O_TRUNC)` through FUSE_OPEN, not FUSE_CREATE. And the barrier approach can't be applied naively in `Create()` anyway, because the request carries the *parent* nodeId — the target nodeId is unknown until the response. Still, I'd add a comment in `Create()` explaining why it's exempt, and optionally a defensive check: after a truncating `Create` response, if `WriteBackCache && GetMaxWrittenOffset(response.GetNodeAttr().GetId()) > 0`, report a critical event, since that means the invariant was violated.
 
I've added a comment and a defensive check
 
> ## Minor correctness nit: `ResetMaxWrittenOffset` on an async-created handle
> 
> The response processor resets on `!HasError(response)`. If a buggy server returns `HandleCreatedAsync` for a truncating open (a contract violation — the client never sets `AllowAsyncCreateHandle` for `E_TRUNCATE`, since `IsAsyncCreateHandleEligible` excludes both `E_TRUNCATE` and writable opens), the truncation isn't persisted yet, but the callback would still reset `MaxWrittenOffset`, and the subsequent handle confirmation runs outside the barrier. It's an already-reported anomalous path (`ReportUnexpectedAsyncCreateHandleResponse`), but the guard is one condition: reset only when `!HasError(response) && !response.GetHandleCreatedAsync()`.

This check is excessive in WriteBackCache.

> ## Correctness aspects I verified as sound
> 
> The subtle invariants all hold. `ResetMaxWrittenOffset` recomputes from remaining cached data (`CachedData.rbegin()->second.End`) rather than zeroing, so writes accepted *during* the barrier (like write2 in the fs test) keep contributing to the reported size — that's what makes the test's `nodeSize == 3` correct. The `Y_ABORT_UNLESS` preconditions inside `ResetMaxWrittenOffset` (front barrier acquired, `FlushedRequests.empty()`) are guaranteed by call order in `ExecuteRequestUnderBarrier` (reset happens before `ReleaseBarrier`, and no flush can run while the front barrier is held) — no new crash paths. There's no deadlock either: barrier acquisition only waits for requests with `SequenceId <= BarrierId` to be flushed and evicted, so newer cached writes don't block it.
> 
> ## Duplication worth reducing
> 
> `TImpl::CreateHandle` is a near-verbatim copy of `SetNodeAttr`: flag check → bypass → executor → identical reset-on-success callback → `ExecuteRequestUnderBarrier`. The comment "Apply the same logic as in SetNodeAttr" is a hint that the logic should literally be shared. A small helper like
> 
> ```cpp
> template <class TResponse>
> TFuture<TResponse> ExecuteSizeChangingRequestUnderBarrier(
>     ui64 nodeId, TRequestExecutor<TResponse> auto executor);
> ```

I thought about it and already tried a couple of variants — the code became less readable.

> owning the `ResetMaxWrittenOffset` callback would keep the two paths from drifting (e.g., the async-handle guard above would then only need fixing once). In `Open()`, the refactored callback still contains the redundant shadowed re-declaration `const auto& response = future.GetValue();` inside the `if` — it was in the old code, but since the function is being restructured anyway, drop the inner one. The `if (WriteBackCache) ... else ...` duplication of `.Subscribe(std::move(callback))` could collapse into a ternary selecting the future, though the current form matches `SetAttr`'s established pattern, so that's a wash.
> 
> ## Tests
> 
> Both new tests genuinely reproduce the intended behavior and I found no flakiness after checking the details that could cause it:
> 
> - The final `writeDataCalled == 1` in `ShouldUseWriteBackCacheBarrierForTruncatingOpen` looks racy at first (barrier release calls `UpdateFlushStatus`), but with `automaticFlushPeriodMs = 0` the node only reaches `ReadyToFlush`, and `TriggerFlushAll` never runs because the periodic flush is disabled (`if (!FlushConfig.AutomaticFlushPeriod) return;`). The `0` survives config defaulting because the proto field is `optional` with explicit presence, so `Set...(0)` doesn't fall back to the 100ms default. Write2 also can't sneak into the in-flight flush batch (dispatched batches are fixed; the post-flush status check sees the barrier and yields `NothingToFlush`). So the second `WriteDataHandler` branch (`call == 1`, offset 0) fires only during `bootstrap.Stop()` in `Y_DEFER` — after all asserts. Deterministic.
> - The test correctly pins the key property: CreateHandle must wait for the WriteData *response*, not just its dispatch (`!createHandleCalledFuture.Wait(1s)` before setting the promise). The negative 1-second wait adds fixed wall time, but the neighboring pre-existing test already does `Sleep(1s)` for the same reason — acceptable per suite conventions.
> - `ShouldKeepMaxWrittenOffsetOnFailedTruncatingCreateHandle` nicely asserts flush-before-dispatch (`SessionWriteDataHandlerCalled == 1` inside the handler) and no reset on error. Writing to an unopened handle 456 is consistent with the rest of the suite.
> - Small coverage gap: there's no direct WBC unit test for the *success* path resetting `MaxWrittenOffset` (e.g., 103 → 0 with an empty cache, or → end-of-newer-write with a non-empty one); it's only covered end-to-end via `nodeSize == 3` in fs_ut. A symmetric third unit test would be cheap and would pin the reset semantics where they live.
> 
> ## RAII, move semantics, exceptions
> 
> Nothing concerning. `TWriteBackCache` is a pimpl value type over `shared_ptr<TImpl>`; the PR doesn't touch special members. Its user-declared destructor suppresses implicit moves, so `TFileSystem`'s ctor init `WriteBackCache(std::move(writeBackCache))` actually copy-constructs (a refcount bump) — harmless, and there is no move-assignment anywhere in the object's life, hence no destroy-source hazard. Lifetime in callbacks is handled idiomatically: `weak_from_this()` + `lock()` throughout, `responseProcessor` capturing `this` is only invoked under a successful `ptr.lock()` of the same object, and the barrier is released before the promise is fulfilled, so the fs-level callback never observes a held barrier. No new throwing surfaces beyond the codebase-standard `future.GetValue()` in subscribers, which is safe here by convention (session futures carry errors in `response.Error`, never exceptions); the new WBC method itself only allocates.
> 
> In short: approve with minor comments — the `Create()` exemption deserves at least a comment (ideally a defensive critical event), the reset callback should exclude `HandleCreatedAsync`, and the `SetNodeAttr`/`CreateHandle` duplication should be factored before a third size-changing operation appears.

SvartMetal (dismissed):

SvartMetal (approved):

yegorskii (approved):

PR #7145, opened 2026-09-12
issue-6608: forbid hard links that cross quota domains

debnatkh:
### Notes
A hard link whose target's QuotaId differs from the new parent's is now rejected outright (E_FS_XDEV). PrepareTx_CreateNode checks both hard-link paths (same-shard and cross-shard). TLinkActor now issues a compensating UnlinkNode when the leader rejects the nodeRef after the shard has already created its side of the link

This also fixes partly #2667 by removing danglink lnks in case of a failed leader response

### Issue
#6608 #2667


debnatkh:
Just one, carried over from earlier and not touched by any of the pushes:

**Unbounded retry in `UndoShardLink`.** In `HandleUndoShardLinkResponse`, when the compensating `UnlinkNode` fails with a retriable error, it calls `UndoShardLink(ctx)` again immediately — no backoff, no attempt cap. If the shard tablet is transiently unavailable, this can hot-loop resending `UnlinkNode` until it eventually succeeds or the actor is torn down some other way. As you pointed out, this mirrors an existing pattern elsewhere in the codebase (`TRenameNodeInDestinationActor`), so it's not a defect unique to this PR — but it's still a real gap in the code as written, just a pre-existing one you're extending rather than introducing. Worth a comment on the PR either way, framed as "matches existing precedent, fix separately if at all" rather than a blocker.

Everything else raised across this review is closed out:
- Silent leak on non-retriable undo failure → fixed with `ReportHardLinkUndoFailed` + `critical_events.h` entry, now covered by `ShouldNotLeaveDanglingLinksForInvalidHardLinks`.
- QuotaId==0 / both-unquota'd cross-shard case → already correct and already covered by the pre-existing `ShouldLinkExternalNodes` test (my earlier flag was a misread).
- Comment placement on the `behaveAsShard` branches → already documented correctly at line 648; my earlier suggestion pointed at the wrong branch.

The one disclosed, out-of-scope item is the general cross-shard hardlink atomicity gap (#2667 — a crash mid-flight between the shard-side create and the leader-side create can still leak a ref), which the PR description explicitly calls out as not addressed here. That's a known, accepted limitation rather than an outstanding issue with this PR.

---

Unbounded retry is acceptable, same as done in other cross-shard places of the codebase

SvartMetal (approved):

agalibin (approved):

PR #7146, opened 2026-09-13
[Filestore] vfs_fuse: TPersistentStateManager should have the state file registry as the single source of truth about them

SvartMetal:
### Notes

Preparation for limiting the total size of the state files (follow-up PR). `TPersistentStateManager` learns what is on disk once, from a listing of the base paths, and answers every question about the state files from that in-memory knowledge afterwards, instead of probing the filesystem at each operation.

**Guard.** `TAcquireStateFileGuard` no longer does any bookkeeping itself: it holds the lock and a callback capturing the manager by `shared_ptr`, and hands the file back through `ReleaseStateFile(impl, deleteFile)` on destruction / `DeleteStateFile()`. All lock release, file removal, registry updates and the `rmdir` happen there, under the one mutex.

**No filesystem existence probes.** `Has*State` and "is this a new file" are answered by the registry. The base path is not pre-checked with `Exists()` either: the listing itself reports `ENOENT` for an absent base path (fine, nothing was ever created), and any other failure - `EIO`, `EACCES`, `ENOTDIR` — is an error rather than "no state".

### Behavior differences

1. **`Has*State` returns `TResultOrError<bool>`. It may perform the initial listing, and a listing failure must not read as "no state" for a restore check.
2. **A session directory is removed only once no state file is known to be in it**, rather than when no file is *held* in it. Consequently `PersistentStateSessionDirNotEmpty` now fires only for files the registry has never heard of, i.e. of a component which is not configured anymore - a leftover of a configured component is known and simply left alone.
3. **A base path which exists but cannot be listed is an error**, where `Exists()` used to turn `EIO`/`EACCES` into an empty listing and every state file on disk would go unnoticed.
4. A state file appearing on disk without going through the manager is invisible until the next restart's listing (before, `Has*State` would have seen it). This is the "manager is the only writer" assumption made total.

### Issue
#7048


SvartMetal:
AI review: https://chatgpt.com/share/6aa6b3c5-7ba8-83ed-b677-b1e8d08cbb02

```
Reviewed b61b4264b263cb55e6a676253f3aa96c82fd91d0, the head fetched for this review. The visible conversation contained the description and an automated CI comment; no human review comments were visible. I treated the documented behavior changes—particularly ignoring out-of-band file creation and retaining known sibling state files—as intentional.
I found two issues worth fixing. The first can cause existing state to be permanently omitted from recovery until the manager restarts.
```

```
1. [P1] IsDirectory() still converts filesystem errors into a successful, incomplete registry
Location: persistent_state_manager.cpp, ListStateFilesLocked(), around lines 323–343.
The base-path List() errors are handled, but both directory-filtering checks still use TFsPath::IsDirectory(). That helper uses TFileStat(...).IsDir() without checking whether obtaining the stat succeeded. The same utility file’s ForceDelete() explicitly checks TFileStat::IsNull() and handles the saved system error—the distinction missing here. Consequently, a failed stat can cause an existing filesystem or session directory to be skipped.
A concrete failure sequence is:
Existing write-back state is present before manager initialization.
The base directory is readable but lacks search permission. Listing its entry names succeeds, but statting its children fails with EACCES.
The directory checks skip the children, and initialization completes successfully.
Permissions are restored. Subsequent queries still use the incomplete registry rather than retrying discovery.
This does not violate the “manager is the only writer” assumption: nobody creates or removes a state file in this sequence.
The dangerous case is restoring a session after write-back caching has been disabled. The loop relies on HasWriteBackCacheState() to decide whether the old cache still needs to be opened. An incorrectly cached false skips that recovery path even after filesystem access has recovered.
Fix: use a checked stat operation. Skip an entry only after successfully establishing that it is not a directory. A failure to establish its type should return an error, allowing the existing initialization rollback/retry path to operate.
Regression test: inject an EACCES or EIO while statting an existing filesystem/session directory. The first query must return an error; after removing the fault, the next query must discover the state.
The new ShouldReportUnlistableBasePathInsteadOfIgnoringIt test is valid, but its regular-file base path exercises ENOTDIR from listing, not failures in these intermediate directory checks.
```
Fixed, but reported a critical event rather than a failure

```
2. [P2] Failed acquisition can leave a manager-created file outside the registry
Location: persistent_state_manager.cpp, AcquireStateFile(), around lines 507–526.
The ordering is:
Touch the file
Open it for locking
Try to acquire the lock
Register it as present and acquired
If creation succeeds but the subsequent open or lock operation fails, acquisition returns an error without registering or removing the file. The filesystem and the supposed authoritative registry now disagree.
For example, a newly created file can receive permissions that prevent reopening it. The creating open succeeds, the locking open fails, and the file remains on disk. A lock-operation failure after successful creation produces the same bookkeeping problem.
This also stays within the single-writer assumption: the manager itself created the untracked file.
The immediate consequence is an incorrect Has*State() result. Deleting another component’s last registered file can also classify this manager-created leftover as unknown state. It would additionally undermine the forthcoming capacity accounting, although that accounting is explicitly outside this PR’s scope.
Fix: separate “present” from “acquired” during acquisition. Once touching/opening the file succeeds, record it as present but unacquired; mark it acquired only after locking succeeds. Alternatively, implement rollback, but remove only a file provably created by this attempt—never pre-existing state or a file whose lock belongs to another owner.
Regression test: fail the locking open or lock operation after successful creation. After the error, assert that either the file was safely rolled back or the registry still reports it as present and unacquired.
ShouldReportErrorInsteadOfThrowingOnAcquireFailure does not cover this: it creates a directory at the target beforehand, rather than exercising a newly created regular file followed by a later failure.
```
Fixed

SvartMetal:
One more conversation with AI:
```
Question: What impact will be observed if, for example, base path is silently becomes corrupted (for example if it is not a directory anymore)? How our error would be seen to the guest?

Let me trace it precisely from the loop's stop-to-start path rather than reason abstractly — including what the error message will literally say and who sees it.Here's the full chain, and then the honest answer about the guest.

**What happens, step by step.** Say `WriteBackCachePath` has become a regular file (or hit `EIO`). The first session start after that:

1. `Has*State` → `EnsureRegistryInitializedLocked` → `basePath.List()` throws `TSystemError(ENOTDIR)`; the code sees a non-`ENOENT` errno and returns `E_FAIL`: `"Failed to list write_back_cache state files under /…/WriteBackCache, reason: (Not a directory) can not open '/…/WriteBackCache'"`. The registry is cleared and stays uninitialized.
2. `StartWithSessionState` reports `WriteBackCacheCreatingOrDeletingError` (critical-event counter + `ERROR` log with the `[f:][c:]` tags and that message) and returns the error.
3. The `StartAsync` future resolves to the error; `DoStartEndpoint` puts it into the `StartEndpointResponse`, **does not store the endpoint**, and the loop is destroyed (guards release nothing — nothing was acquired).

So the failure surfaces in three places, all on the **host**: the `AppCriticalEvents/…` counter, the daemon log, and the `StartEndpoint` API response to whoever drives the vhost daemon (the compute/orchestration side).

**What the guest sees: nothing specific.** The guest is behind a virtio-fs device; it has no channel through which a filestore error string can reach it. The endpoint simply never comes up, so:

- **New attach / VM boot:** the vhost-user backend socket for that device isn't served. Depending on the hypervisor that's a failed device setup on the host side, or a guest-side `mount` of the virtiofs tag that fails with a generic `ENODEV`/`ENOENT`-class error, or a mount that hangs waiting for a backend.
- **Daemon restart with the guest already running:** the endpoint isn't restored, the vhost-user reconnect never completes, and the guest's already-mounted filesystem **hangs** — I/O queues with no backend to drain it. That's the nastier case, because the guest is running and blocked, not failed. Since the state files live on tmpfs, this is exactly the daemon-restart-without-reboot scenario.

**Two things about the behavior that are worth deciding on explicitly:**

- **Retry is cheap and automatic.** Because the failed listing leaves the registry uninitialized, every subsequent `StartEndpoint` re-lists. Fix the path, retry the endpoint, no daemon restart needed. Good.
- **One corrupted path blocks every component and every session on the host.** `EnsureRegistryInitializedLocked` lists *all* configured base paths before any operation, so a broken `DirectoryHandlesStoragePath` — an optional feature — fails the write-back cache and handle-ops checks too, and therefore every session start. That's the direct consequence of "list everything before the first operation," and it's a fail-fast-vs-availability trade-off: it reliably surfaces the misconfiguration, but it also means a guest that doesn't use directory-handle storage at all can't be attached. If you'd rather degrade — e.g. list per component lazily so an unrelated path's corruption only affects sessions that need it — that's the earlier per-component listing design, and I can restore it in PR 1. I'd want that to be a deliberate choice rather than an accident of the refactor, so I'm flagging it now.

One pre-existing detail, unchanged by this work: the server-side session was already created by the time the listing fails, and the loop returns without destroying it, so it lingers until the server times it out. That's true of every post-`CreateSession` start failure in this code, not something introduced here.
```

So overall, I think it's better to return an error if something is wrong with the component base path configuration, instead of silently ignoring it. In the case of an error, existing IO will hang and the duty engineer will have a chance to fix the situation, while new sessions will fail on start, leading to subsequent StartEndpoint retries by a Compute node component (kubelet or something else)

SvartMetal:
image

Not relevant to this change

Data race in qemu-kikimr-sidechannel-test: https://github-actions-s3.storage.eu-north2.nebius.cloud/ydb-platform/nbs/PR-check/34764704820/1/nebius-x86-64-tsan/logs/1/cloud/filestore/tests/fastshard/fmdtest/qemu-kikimr-sidechannel-test/test-results/py3test/testing_out_stuff/recipe_stop_vhost-recipe-2.err

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -299,20 +311,219 @@ TFsPath TPersistentStateManager::GetSessionDir(
...
+                    }
+
+                    Registry.Register(
+                        sessionDir.GetPath(),
+                        fileName,
+                        false /* fileAcquired */);
+                }
+            }
+        }
+    } catch (const yexception& e) {
```
neihar, line 395: // Whether the path is a directory to descend into. We should report weird
// (non-directory) entries, but must not fail every session start.

shouldn't we apply the same rule here and make catch more granular?

neihar (approved):

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
SvartMetal: Will do in a separate PR

agalibin (approved):

PR #7148, opened 2026-09-14
[Filestore] return ESTALE if operation by inode failed with E_FS_NOENT

neihar:
### Notes
Problem: after a cross-client os.replace() the reader kernel still holds a
dentry for the destroyed node and sends OPEN/GETATTR/... by nodeid; the tablet
answers E_FS_NOENT and vfs_fuse returned ENOENT, which does not drop the dentry,
so the app kept failing until READDIRPLUS rewrote it.

Fix: for ino-addressed requests map E_FS_NOENT to ESTALE, so the VFS retries the
path walk with LOOKUP_REVAL and picks up the new nodeid. Name-addressed requests
still return ENOENT.

**Testing**
Added UTs

Added  manual test cross_client_atomic_replace (2-guest qemu): writer on guest 0 does 150 os.replace every 100 ms, reader on guest 1 opens and reads the target every 5 ms. Asserts that the vhost log contains at least one failed on a stale node (the stale-dentry open actually happened) and that the reader never observed ENOENT (only ESTALE is tolerated). Before the fix: 137 of 1995 reads failed with ENOENT. After: 148 stale opens recovered transparently, 0 errors in 2133 reads.

### Issue
#7151


cloud/filestore/tests/cross_client_atomic_replace/test.py
```diff
@@ -0,0 +1,48 @@
...
+def test():
+    root = os.path.join(
+        os.getenv("NFS_MOUNT_PATH"), "cross_client_atomic_replace")
+
+    writer = run(0, "writer", root, wait=False)
+    reader = run(1, "reader", root, wait=True)
+    writer.wait(timeout=TIMEOUT_SECONDS)
+
+    summary = json.loads(reader.stdout.decode("utf8").splitlines()[-1])
+    with open(common.output_path("filestore-vhost.err")) as log:
```
qkrorlqr: it's a fragile hack, let's at least rely on profile-log and not on stderr which is intended for humans to read, not for scripts

there're multiple examples of profile-log usage in other tests

qkrorlqr (changes requested):
the code is fine apart from the hack in the test

cloud/filestore/tests/cross_client_atomic_replace/test.py
```diff
@@ -0,0 +1,48 @@
...
+        stale_opens = log.read().count("failed on a stale node")
+    logging.info(f"reader: {summary}, stale opens: {stale_opens}")
+
+    # the reader opened the replaced node through its cached dentry at least
+    # once and the kernel recovered via ESTALE instead of reporting ENOENT
+    assert stale_opens > 0
+    # a replace landing inside the kernel's single ESTALE retry surfaces
+    # ESTALE to the application; ENOENT must never be reported
+    assert set(summary["errors"]) <= {str(errno.ESTALE)}, summary
```
qkrorlqr: this line is pretty cryptic

neihar: fixed

cloud/filestore/tests/cross_client_atomic_replace/test.py
neihar: fixed

qkrorlqr (approved):

yegorskii (approved):

PR #7150, opened 2026-09-14
Move a connection off a host that stopped answering

yegorskii:
### Notes
A cell connection used to be pinned to one host for life: if that host
died, IO returned errors until the client tore the session down and built
a new one. Two hooks for doing better were already in the code and both
were dead - SetHostAlive was never called, and OnTabletHostChanged had no
implementation.
    
The host pool now pings the hosts of a cell on a timer and tells the
connections sitting on a host when it stops answering. A connection
rebuilds itself on another host - new control channel, new data endpoint -
behind two long-lived routers, so the session holding it sees nothing. A
mount response naming a different tablet host moves the connection the
same way, which saves a proxy hop.
    
Nothing in the session layer had to learn about this: a request arriving
at a host that does not know the session gets E_BS_INVALID_SESSION, and
TSession already turns that into a remount and a retry.
    
Two smaller things come with it. MinCellConnections now means what its
comment always said - the minimum is maintained at all times, not only at
start - and slots that configured hosts cannot fill are held by hosts
discovered from the tablet host of a mount, so a cell whose configured
hosts are all down still has a way in. And cells stopped asking the rdma
client for endpoints asynchronously: the endpoint is handed back before it
has connected, which removes a phase from the migration path. Without a
gRPC fallback, a host whose rdma never comes up now reports that through
HandleUnavailable and the connection leaves, instead of the mount failing.
    
All of it is gated by HostMigrationEnabled, off by default.

### Issue


yegorskii:
Codex review:

LGTM.

I went through the migration/liveness state machine in detail, including the less obvious asynchronous interleavings around host pings, mount responses, RDMA callbacks, and migration completion.

In particular, I checked:

* stale events from bindings that are no longer current;
* mount responses racing with host-failure driven migrations;
* RDMA `HandleUnavailable` / `HandleConnected` interactions with pinger verdicts;
* failures reported by a binding while it is still being prepared;
* successful and aborted migrations and the corresponding settle paths;
* ordering of chained/pending tablet-host migrations;
* move-away requests arriving while another migration is already in flight;
* move-away requests surviving an unsuccessful concurrent migration;
* cancellation of a pending move when the underlying RDMA failure disappears;
* expiration of host-avoidance cooldowns while a request is waiting;
* requests associated with a newer binding that is later abandoned, without losing a still-relevant request for the current binding;
* avoiding immediate migration back to hosts known to be dead or temporarily unusable;
* discovered hosts and the interaction with the host pool / channel lifetime;
* initial connection setup, where callbacks may arrive before the first binding is fully installed;
* cleanup/hand-off between bindings so that old switchers, handlers, or control channels cannot affect the new binding.

The final `MoveAwayRequests` model looks much cleaner: a move-away request remains an outstanding intent while host selection happens outside the connection lock, and is consumed only at the commit point. This makes recovery/cancellation and concurrent migration cases much easier to reason about and avoids the

agalibin (approved):

apliev (approved):

PR #7157, opened 2026-09-14
issue-6956: add a file device

arigachnyy:
### Notes
Add an IDevice that keeps its pages in a file. The reads and writes go through IFileIOService, one vectored request per page group.

The file is created if it does not exist and grown to the device size if it is shorter, a longer one is left as is. Every page must be of the page size and lie within the device, otherwise the whole request is rejected before any IO is issued.

### Issue


qkrorlqr (approved):

qkrorlqr:
But the IDevice interface requires changes - it shouldn't talk in terms of page groups because it doesn't provide read/write atomicity for a group of page ranges.

asorotsky (approved):

Issue #6996, opened 2026-09-01
[Filestore] TCPSideChannel


qkrorlqr:
We already have a version of side channel implementation here: https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/libs/service_kikimr/side_channel.cpp

Currently it works only for ReadData/WriteData and has bugs.

This issue is to track:
1. bugfixing
2. possibly - supporting other request types

PR #7160, opened 2026-09-14
issue-6996: fixed a profile-log-vs-extract race in tcp-side-channel

qkrorlqr:
### Notes
If response future gets set before we subscribe the logging callback, then this logging callback would suffer from 2 problems:
* it can race with other callbacks
* it can even access the future value after it's been extracted by another callback

Fixing it by calling the logging callback before the future is set. 

### Issue
https://github.com/ydb-platform/nbs/issues/6996


qkrorlqr:
Claude:

> The fix itself is sound. The old code subscribed the profile-log callback on the future after Dispatch could have already set it, so the callback could run concurrently with (or after) the consumer's destructive UnsafeExtractValue — moving the hook to run synchronously inside complete before SetValue genuinely eliminates that race, and it's applied consistently to both Read and Write paths. TPromise is a shared handle, so the caller can still fall back when Dispatch returns false. I found one serious issue and one thing worth double-checking:
> 
> 1. An exception in onComplete will leave the promise unfulfilled — request hangs forever. This is a new failure mode introduced by the reordering. Previously the profile-log callback ran as a future subscriber after SetValue, so if it threw, the response was already published. Now onComplete(resp) runs before response.SetValue(...) on the completion path; if FinalizeProfileLogRequestInfo or the profile-log write throws (allocation failure, a bug in the log backend, etc.), SetValue is never reached and the client's read/write blocks indefinitely — and the endpoint is also never pushed back to the pool. Either guarantee the hook is noexcept, or wrap it so SetValue still executes on the exception path.
> 

We try our best to avoid exception usage in futures

> 2. Verify the hook can never be invoked twice. MakeProfileLogHook's lambda is mutable and does std::move(profileLogRequest) on invocation, and complete is also single-shot by construction today (exactly one of the three completion paths runs). That invariant is now load-bearing for correctness of the moved-from state; a future edit that retries or invokes complete on two paths would silently log a moved-from profileLogRequest. A one-line comment or a debug assert (e.g. a Y_DEBUG_ABORT_UNLESS once-flag) would make the contract explicit. This one is preventative rather than a current bug — take it or leave it.
> 

This invariant is enforced locally inside tcp-side-channel, no need to add extra precautions here

> No other serious problems: null-endpoint handling in the error path is fine (Push tolerates null), the ReadData hook observes the response after MoveBufferToIovecs exactly as the old subscriber did, and dropping profile logging when Dispatch returns false matches the old if (dispatched) behavior since the main channel logs via TStorageServiceActor.

asorotsky (approved):

arigachnyy (approved):

yegorskii (approved):

PR #7161, opened 2026-09-14
[NBS] Track compaction map updates during in-flight compaction

vladstepanyuk:
### Notes
Added class that tracks compaction map counters during compaction execution and on compaction end applies tracked counters to primary compaction map.

This changes needed to correctly account writes  in compaction map that happens between two moments in time: compaction tx executed, add blobs tx executed.

### Issue
 #7057


cloud/blockstore/libs/storage/core/config.cpp
```diff
@@ -743,6 +743,7 @@ NProto::TLinkedDiskFillBandwidth GetBandwidth(
     xxx(MixedBytesCountCompactionThresholdHDD,      ui32,       0             )\
     xxx(MixedBytesCountCompactionThresholdSSD,      ui32,       0             )\
     xxx(MixedBlocksCountCompactionRangeCountPerRun, ui32,       1             )\
+    xxx(CompactionStatsTrackerEnabled,             bool,       false         )\
```
qkrorlqr: nit: something's wrong with the indentation of `bool`, `false` and `)\`

cloud/blockstore/libs/storage/partition/model/compaction_stats_tracker.cpp
```diff
@@ -0,0 +1,166 @@
...
+bool TCompactionStatsTracker::HasCompaction() const
+{
+    return !CommitIdToCompaction.empty();
+}
+
+void TCompactionStatsTracker::CompactionStarted(
+    ui64 commitId,
+    TVector<ui32> rangeIndices)
+{
+    if (!IsSorted(rangeIndices.begin(), rangeIndices.end())) {
```
qkrorlqr: can't we just make sure that all the calls use a sorted vector? `IsSorted()` call itself is not free

and if we don't care about perf here, why not sort the input every time without any checks

cloud/blockstore/libs/storage/partition/model/compaction_stats_tracker.h
```diff
@@ -0,0 +1,56 @@
+
```
qkrorlqr: nit: 2 unneeded blank lines

cloud/blockstore/libs/storage/partition/model/compaction_stats_tracker.h
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+#include <cloud/blockstore/libs/storage/core/compaction_map.h>
+
+#include <util/generic/hash.h>
+#include <util/generic/vector.h>
+#include <util/system/types.h>
+
+namespace NCloud::NBlockStore::NStorage::NPartition {
+
```
qkrorlqr, line 10: ////////////////

cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -563,38 +575,66 @@ class TAddBlobsExecutor
...
 
             if (!IsDeletionMarker(blob.BlobId)) {
-                TCompactionMap::UpdateCompactionCounter(
-                    rangeStat->BlockCount + 1,
-                    &rangeStat->BlockCount
-                );
-                TCompactionMap::UpdateCompactionCounter(
-                    rangeStat->MixedBlockCount + 1,
-                    &rangeStat->MixedBlockCount);
+                blockCountForCurrentRange++;
```
qkrorlqr: nit: please use prefix ++

cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -563,38 +575,66 @@ class TAddBlobsExecutor
...
 
-    void UpdateCompactionMap(TPartitionDatabase& db)
+    void RegularUpdateCompactionMap(TPartitionDatabase& db)
     {
         for (const auto& kv: CompactionCounters) {
+            auto rangeStat = kv.second.Stat;
+
+            if (Args.Mode == ADD_COMPACTION_RESULT) {
+                STORAGE_VERIFY(
+                    !State.AccessCompactionStatsTracker(),
```
qkrorlqr, line 625: why? comment needed

cloud/blockstore/libs/storage/partition/part_state.h
```diff
@@ -337,7 +338,8 @@ class TPartitionState
         ui64 tabletId,
         const std::optional<TMixedBlocksFilterConfig> mixedBlocksFilterConfig,
         bool checkpointAwareCleanupEnabled,
-        bool useBlobChannelDataKindForCounters);
+        bool useBlobChannelDataKindForCounters,
+        bool compactionStatsTrackerEnabled = false);
```
qkrorlqr: no defaults in non-test code please

cloud/blockstore/libs/storage/partition/model/compaction_stats_tracker.cpp
```diff
@@ -0,0 +1,166 @@
...
+        counter,
+        TWellKnownEntityTypes::TABLET,
+        TabletId,
+        TStringBuilder() << "Counter for range index " << rangeIdx
+                         << " not found in compaction with commit id "
+                         << commitId);
+    return counter;
+}
+
+TVector<TCompactionCounter*> TCompactionStatsTracker::AccessCompactionCounters(
```
qkrorlqr: how can we have multiple concurrent compactions for the same `rangeIdx`?

vladstepanyuk: We can have two compactions enqueued for the same `rangeIdx`. But they all executing in sequential order. Moved StartCompaction call to prepare compaction function and simplified compaction stats tracker interface so now it supports only one compaction at the time

qkrorlqr (approved):

qkrorlqr:
AI review plz

komarevtsev-d (approved):

vladstepanyuk:
> AI review plz

Found two blocking correctness issues in [PR #7161](https://github.com/ydb-platform/nbs/pull/7161):

- **[P1] Concurrent writes leave the range incorrectly marked as compacted** — [compaction_stats_tracker.cpp:86](/home/iajwjw/github/nbs/cloud/blockstore/libs/storage/partition/model/compaction_stats_tracker.cpp:86)

  `FinishCompaction()` always passes `true` for `compacted`. A write newer than the compaction snapshot temporarily marks the range non-compacted, but this call overwrites that state. In the new test’s scenario, the final range has two blobs and one garbage block but remains `Compacted=true`; garbage-compaction selection explicitly ignores compacted ranges. Thus the main symptom from [issue #7057](https://github.com/ydb-platform/nbs/issues/7057)—garbage not being selected—can remain despite the numeric counters being corrected. Track whether a post-snapshot write touched each range and finish such ranges with `compacted=false`. The integration test should assert this flag or attempt garbage selection.

- **[P1] Concurrent zeroes lose `NewlyZeroedBlocks` accounting** — [part_actor_addblobs.cpp:725](/home/iajwjw/github/nbs/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp:725)

  A `ZeroBlocks` completed after the compaction snapshot updates the primary range’s `NewlyZeroedBlocks`, but `AddCompactionStats()` never transfers that field into the tracker. Compaction completion then subtracts the entire current value from the partition-wide counter, and `FinishCompaction()` replaces the range with the tracker’s zero value. That zero was not processed by this compaction, so both per-range and global counters become incorrect, affecting `UsedBlocksIgnoringZeroed` and subsequent compaction decisions. The tracker needs to preserve post-snapshot newly-zeroed deltas. Add a test that pauses compaction `AddBlobs`, performs `ZeroBlocks`, then checks both counters.

Focused tests passed:

- `TCompactionStatsTrackerTest::*`: 6/6
- `ShouldAccountForConcurrentWritesInCompactionMapCounters`: 1/1

The current integration test reproduces the original numeric-counter race correctly, but its three assertions do not expose either invariant violation above. No new raw-memory operations or move-assignment implementations were present in the diff.

---

This issues seems minor to me and can be fixed in separate pr. Ignoring zeros counter was set to zero explicitly after Compaction, so I just kept this logic as it was. Fixing it would need to take into account skipped blobs and other things beyond this pull request.

PR #7162, opened 2026-09-14
issue-6996: fixed a side-channel-update-vs-extract race

qkrorlqr:
### Notes
This race is similar to https://github.com/ydb-platform/nbs/pull/7160 but in a different place in the code

In this case if a subscribe call higher in the call chain (e.g. in durable client) is triggered when the value is already set, a race will occur

The root problem is that Subscribe+Extract mechanics are a bit broken (or, in other words, "require some extra care")

### Issue
https://github.com/ydb-platform/nbs/issues/6996


qkrorlqr:
Claude:

> The fix is correct and cleanly resolves the race described. In the old code, sc->Update was a second subscriber on the same future that the upstream caller (e.g. durable client) subscribes to and extracts destructively — so callback ordering decided whether f.GetValue() read a valid value or a moved-from husk. The new structure is sound: the main channel now fulfills a private inner promise whose only consumer is the forwarding lambda, so UnsafeExtractValue(inner) is safe by construction; the caller's promise gets exactly one setter and the caller remains its sole extractor. Attaching the subscription before ExecuteRequestImpl dispatches means there's no window where inner can be set-and-extracted out from under anyone.
> 
> I checked the interactions that could make this wrong and found no serious issues:
> 
> The correctness hinges on one cross-file contract: when SideChannel->ExecuteRequest returns false, the side channel must never touch the promise copy it received, otherwise you'd get a double SetValue (once from the side channel, once from the forwarding lambda). I verified this holds in the current Dispatch — the false return happens synchronously before any async subscription that could invoke complete, and the lambda holding the promise copy is simply destroyed. Combined with #7160 removing the profile-log subscriber, the caller's future ends up with exactly one setter and one consumer on every path, which is the invariant both PRs are converging on. The synchronous-completion path (WriteData's E_ARGUMENT early return, where the promise is set before ExecuteRequest even returns true) is also fine, since the caller subscribing after the fact just runs its callback inline as the sole extractor.
> 
> One adjacent observation, pre-existing and out of scope but sitting exactly on the fallback branch this PR routes through: for ReadData, side_channel.cpp swaps the iovecs out of the original request before Dispatch, so when ExecuteRequest returns false, the main channel serves a request that has permanently lost its iovecs and responds via Buffer instead. If the fuse-side response handler copes with a Buffer response to an iovec request that's fine — worth a ten-second sanity check, nothing more.

I'll address this adjacent observation separately

asorotsky (approved):

arigachnyy (approved):

yegorskii (approved):

PR #7167, opened 2026-09-14
[NBS] Reduce Disk Registry migration completion redo size

Nik-Pavlov:
### Notes

`TFinishMigration` previously rewrote the disk configuration and placement group and added a reallocation request for each migration, amplifying commit redo size for large disks.

`FinishDeviceMigrations` now applies the migrations in memory, then persists the disk and placement group once and adds one reallocation request if any succeed. Every migration is still attempted and logged. The actor still returns the first failure.

Successful migrations share a `SeqNo` because their mapping changes commit together. Cleanup already releases all migrations at or below the acknowledged number, and reboot recovery already assigns a shared number to restored migrations. Separate batches receive fresh numbers, preserving protection against stale acknowledgements.

### Testing

- Added `ShouldRestoreBatchedFinishedMigrationsAfterReboot`: completes two migrations in one request, checks that the placement-group version increments once, and withholds the volume acknowledgement across a Disk Registry reboot. It checks the target mapping, retained source IDs, and pending notification in both memory and the local database, then verifies source cleanup and reuse after acknowledgement.
- Updated a state test to finish two migrations as a batch and assert that their sequence numbers match.

### Redo-size comparison

Controlled E2E benchmark using real copying on main and this branch, with fresh disks, 128 MiB devices, matching pending metadata, and one forced migration batch per completion request:

| Migrations | Main redo bytes | Branch redo bytes | Reduction |
| ---: | ---: | ---: | ---: |
| 8 | 10,162 | 1,419 | 86.04% |
| 32 | 136,542 | 4,947 | 96.38% |
| 64 | 529,009 | 9,667 | 98.17% |

Results show the reduction for this workload; production savings depend on disk metadata size, history, and batch size. These are uncompressed redo bytes for the completion transaction, measured with `tx.DB.GetCommitRedoBytes()`, the metric used by the tablet executor's commit-size limit. Setup, acknowledgement, and secure-erase transactions are excluded.

For measurement, temporary logging was added at the end of `TDiskRegistryActor::ExecuteFinishMigration` in [cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_migration.cpp](https://github.com/ydb-platform/nbs/blob/1530ce8696a3d5fe41d657858b7dac60ef8abc55/cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_migration.cpp#L252-L296), after processing all migrations:

```cpp
LOG_INFO(ctx, TBlockStoreComponents::DISK_REGISTRY,
    "REDO_BENCH DiskId=%s Migrations=%zu RedoBytes=%zu",
    args.DiskId.c_str(), args.Migrations.size(),
    tx.DB.GetCommitRedoBytes());
```


sharpeye (dismissed):

cloud/blockstore/libs/storage/disk_registry/disk_registry_state.h
```diff
@@ -736,13 +738,15 @@ class TDiskRegistryState
         const TDeviceId& sourceDeviceId,
         const TDeviceId& targetDeviceId);
 
-    NProto::TError FinishDeviceMigration(
+    using TFinishDeviceMigrationHandler = std::function<
+        void(const NProto::TDeviceMigrationIds& ids, NProto::TError error)>;
```
komarevtsev-d:
```suggestion
        void(const NProto::TDeviceMigrationIds& ids, const NProto::TError& error)>;
```

cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_migration.cpp
```diff
@@ -302,7 +297,7 @@ void TDiskRegistryActor::CompleteFinishMigration(
     LOG_INFO(
         ctx,
         TBlockStoreComponents::DISK_REGISTRY,
-        "%s FinishMigration complete. DiskId=%s Migrations=%d",
+        "%s FinishMigration complete. DiskId=%s Migrations=%zu",
         LogTitle.GetWithTime().c_str(),
         args.DiskId.c_str(),
         args.Migrations.size());
```
komarevtsev-d: 1) I would log the error here.
2) Change the priority of the log to ERROR when HasError(args.Error) is true: 

```cpp
LOG_LOG(
        ctx,
        HasError(args.Error) ? NLog::EPriority::PRI_ERROR
                             : NLog::EPriority::PRI_INFO,
        TBlockStoreComponents::DISK_REGISTRY,
         ...
```

cloud/blockstore/libs/storage/disk_registry/disk_registry_state.h
Nik-Pavlov: Added const ref

cloud/blockstore/libs/storage/disk_registry/disk_registry_actor_migration.cpp
Nik-Pavlov: Added error and priority based on error

komarevtsev-d (approved):

sharpeye (approved):

PR #7169, opened 2026-09-14
[Storage] file_backed_containers: check only the header for zeros when opening a file ring buffer

SvartMetal:
### Notes

`TFileRingBufferAccessor::DoValidateAndInitialize()` used to scan the **whole mapped file** to decide between "not initialized" (all zeros) and "corrupted" (header invalid). Now it checks only the header.

A zero header means the file has never been initialized, whatever the rest of it contains: the buffer never reads beyond the header except between the positions the header holds, and the slack space marker is written explicitly rather than relied upon to be zero, so stale bytes beyond the header are unreachable and harmless. The full scan therefore added nothing to correctness. It did add a real hazard: the file is mapped shared, and reading a hole through such a mapping makes the kernel materialize a zero page in RSS for it. Scanning a large sparse file - one preallocated with `ftruncate`, or one left by a crash between `ResizeAndRemap` and the header write - commits the file's whole size to memory just to find out that it is empty. On tmpfs, where the state files live, that memory is the file itself and stays allocated after the process exits.

This is preparation for the vhost persistent state manager creating state files with their configured size (follow-up PR), but it stands on its own: today's crash window can already produce a full-sized zero file that the next open would scan in full.

### Behavior differences

1. **A zero header over a non-zero body is now "not initialized" rather than "corrupted".** Previously such a file failed validation with "Unsupported file ring buffer version 0", now it is re-initialized as a fresh, empty buffer. The data was unrecoverable either way (no `ReadPos`/`WritePos` without a header), so what changes is only that the event is silent instead of reported. A non-zero byte *inside* the header still fails validation exactly as before (`ShouldNotValidateFileWithZeroVersionAndNonZeroContents` is unchanged).
2. Opening a large sparse, never-initialized file no longer commits it to memory.

### Issue
#7048


e673 (approved):

neihar (approved):

SvartMetal:
AI review: https://claude.ai/share/243e61b3-327b-4b81-91ef-d58a785d4753
```
I pulled the diff and the full sources at commit d25f89d and traced every read path that could touch bytes beyond the header. Summary up front: the core change is correct — I could not find a crash, out-of-bounds access, or broken invariant, and all three new tests genuinely fail against the old code. The findings below are mostly hardening/observability notes and nits.
Correctness of the production change
No OOB. RawData.subspan(0, sizeof(TFileRingBufferHeader)) is guarded by the RawData.size() < sizeof(TFileRingBufferHeader) check a few lines above, which returns Failed first. No memcpy/memmove/memset is added anywhere in the PR. AllOf over a std::span is fine and <util/generic/algorithm.h> was already included in the accessor.
The claimed invariant holds. I verified "nothing beyond the header is read until the header says so" in code, for the only path whose behavior changes (zero header → reinit as empty):
ValidateData() returns immediately when readPos == writePos — the fresh header has ReadPos = WritePos = 0, so validation reads zero bytes of the data area.
GetEntry(pos) short-circuits to CreateEmpty when pos == WritePos before touching the data region, so GetFrontEntry/EraseFreeEntriesFromFront/VisitEntries read nothing on an empty buffer (CreateEmpty has DataSize == 0, so HasValue() is false and the loop never dereferences data).
Metadata CRC is computed over MetadataSize bytes, which is 0 after reinit, so stale metadata-region bytes are also dead.
The slack marker is indeed written explicitly (WriteSlackSpaceMarker → WriteEntryHeader(pos, {})) in SetReadAndWritePosToZeroForEmptyBuffer and in the Alloc wrap path; nothing assumes the data area is zero.
Validate() at the second call site still treats NotInitialized the same as Failed (both SetCorrupted + return false), so a header zeroed mid-life behaves as before, just with a different message.
One bonus the description undersells: the old code actually had a recovery hole. A crash after ResizeAndRemap but before the header write, on a file with stale non-zero body, produced zero-header + non-zero-body → old code returned Failed ("version 0") and the buffer stayed bricked until manual intervention. The new code self-heals that window.
Findings, in decreasing order of importance
Silent reinit destroys forensic evidence (disclosed trade-off, but worth one more thought). Previously a zero header over a non-zero body produced an error and left the file intact for post-mortem; now the header is rewritten immediately and the first PushBack starts overwriting the stale body. If this state ever arises from a bug elsewhere (something zeroing the first 64 bytes of the wrong file), it's now invisible — no log line, no counter. Full detection is impossible without the scan you're removing, but sampling a handful of pages, or at least logging "reinitialized a non-empty-sized file" (file size > header is cheap to check via RawData.size()), would cost nothing and preserve a signal. Non-blocking; the data was unrecoverable either way.
Zero-check boundary vs. the format's reserved header area. The check covers sizeof(TFileRingBufferHeader) = 64 bytes, but the format reserves HeaderReserveSize = 256 bytes (MetadataOffset of fresh buffers), and reinit writes only the 64-byte struct. After a stale-body reinit, bytes [64, 256) keep garbage — the only stale bytes that end up inside the reserved header region. Nothing reads them today, and ValidateHeader pins HeaderSize == sizeof(THeader) so a future extended header must go through explicit migration anyway, but zeroing [0, 256) on reinit is one page of writes and would eliminate the asymmetry with a truly fresh file. Minor.
Test environment sensitivity in ShouldNotReadWholePreallocatedSparseFileOnOpen — checked carefully, and it's not flaky in the failing direction. The important properties all check out: TFileMap::Ptr() returns MappedData() (aligned base + Head), i.e. the pointer to the requested tailOffset, so mincore examines exactly [48 MB, 64 MB); tailSize is page-aligned so InCoreMemory's vecLen * pageSize < len throw can't fire; the second read-only mapping sees page-cache residency of pages materialized through the buffer's own mapping, which is exactly what a scan regression would produce; and against the old code the scan faults in 100% of the tail vs. the < 50% threshold, so it reliably reproduces the bug with a wide margin. The caveats all make the test vacuously pass, never spuriously fail: on non-Linux InCoreMemory memsets the vector to zero, and post-CVE-2019-5489 kernels would hide residency only if the process didn't own the file (it does — it created it). The one way it can spuriously fail is UNIT_ASSERT_VALUES_EQUAL(0, TFileStat(...).AllocationSize) on a filesystem without sparse-file support; fine on the Linux CI this targets, but if the test suite ever runs on an exotic FS, consider turning that precondition into a skip rather than a failure.
All three tests reproduce the pre-fix behavior, which the task asked me to verify explicitly: the mincore test fails on old code (full residency), ShouldValidateFileWithZeroHeaderRegardlessOfBody fails on old code (status Failed, message "Unsupported file ring buffer version 0"), and ShouldReinitializeFileWithZeroHeaderAndStaleBody fails on old code at UNIT_ASSERT(rb.Validate()). The comment in the mincore test explaining why capacity (128 MB) deliberately exceeds the preallocation (64 MB) — so reinit grows instead of truncating and the evidence pages survive to be counted — is correct and a nice touch; a smaller capacity would have silently weakened the test.
Requested angles that don't apply, confirmed clean: no move assignment is touched; RAII in the tests is idiomatic (scoped TFiles close before reopen, TTempFileHandle owns cleanup, TFileMap destructs); and the production change adds no throwing calls to library code — AllOf + subspan only. The throwing utilities (InCoreMemory, TFileMap) live in test code, where an exception is just a test failure.
Nits. RawData.first(sizeof(TFileRingBufferHeader)) reads slightly better than subspan(0, ...). In file_ring_buffer_ut.cpp the new #include <util/generic/algorithm.h> landed after util/random/random.h, breaking the alphabetical grouping of the existing block. TVector<char> zeros(sizeof(THeader), 0) could be const char zeros[sizeof(TFileRingBufferHeader)] = {}. All cosmetic.
CI: the affected target (cloud/storage, where all three tests live) is fully green; the single red is one failed test in a cloud/blockstore workload whose parallel blockstore run passed — almost certainly a pre-existing flake, but worth a one-line confirmation before merge.
Complexity-wise there's nothing left to remove: the fix is minimal (a bounded subspan instead of a whole-file scan), simpler than any alternative that preserves the "corrupted vs. uninitialized" distinction, and doesn't sacrifice any check that mattered — the body scan protected bytes that are provably unreachable.
```

agalibin (approved):

PR #7174, opened 2026-09-15
issue-1541: drain DelayedReleaseQueue from the idle tick too

WilyTiger:
### Notes
follow-up https://github.com/ydb-platform/nbs/pull/7047#discussion_r3980997568

Close request postponed due to HandleOpsQueue overflow is normally moves to HandleOpsQueue  after completing a HandleOpsQueue batch (CompleteHandleOpsQueueBatch). That only happens while HandleOpsQueue has entries of its own to process - once it goes empty, nothing schedules another attempt at DelayedReleaseQueue, so a postponed release could in theory sit there until unrelated queue activity happened to refill HandleOpsQueue and trigger a batch completion.

This closes that gap cheaply: the idle tick (AsyncHandleOperationIdlePeriod, 50ms by default) now retries DelayedReleaseQueue itself, so a postponed release is retried at least once per idle period regardless of whether anything else is going on.

Note: this doesn't fix a race between HandleOpsQueueLock and DelayedReleaseQueueLock that could in theory produce this exact situation. Reaching this race in practice requires HandleOpsQueueSize configured much smaller than its 1 GB default, small enough that a single batch (32 entries by default) can drain it down to completely empty - and that draining has to happen on one thread at the exact same moment another thread's release overflows the queue on the other side. 

No test: this bug only shows up when two threads hit the exact same moment at the same time, and our tests don't run things on separate threads like that - everything in a test happens one step after another, in order. So there's no way to write a test that actually recreates two things happening at once. Every time I tried to fake this situation with a single-threaded test, the existing retry code in CompleteHandleOpsQueueBatch just fixed it right away on its own, before the new idle-tick code even got a chance to matter

### Issue
#1541


cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -554,6 +554,19 @@ void TFileSystem::ProcessHandleOpsQueue()
...
+        // HandleOpsQueue has nothing of its own to process right now, but a
+        // release postponed earlier (due to overflow) may still be waiting
+        // in DelayedReleaseQueue - this is normally drained as a side effect
+        // of completing a HandleOpsQueue batch, which does not happen while
+        // HandleOpsQueue stays empty. Retry it here too, so it is not stuck
+        // until unrelated queue activity happens to pick it up. Capped at
+        // batchSize, same as a regular batch, so a large backlog cannot make
+        // a single tick process large number of entries.
+        for (ui32 i = 0; i < batchSize; ++i) {
+            if (!ProcessDelayedRelease()) {
```
budevg: let's refactor to `DrainDelayedReleaseQueue(ui32 maxCount)` and call it from different places.
Also inside `DrainDelayedReleaseQueue` it should stop if HandleOpsQueue is full and avoid keep going

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -554,6 +554,19 @@ void TFileSystem::ProcessHandleOpsQueue()
...
+                break;
+            }
+        }
         ScheduleProcessHandleOpsQueue(
```
budevg, line 573: If we were able to process delayed release there is no reason to schedule with idle period since there is work inside handle ops queue
```
    drained = DrainDelayedReleaseQueue(batchSize)
     ScheduleProcessHandleOpsQueue(                                                                                           
         drained ? Config->GetAsyncHandleOperationDrainPeriod()                                                               
                 : Config->GetAsyncHandleOperationIdlePeriod());                                                              
```

cloud/filestore/libs/vfs_fuse/fs_impl.cpp
```diff
@@ -554,6 +554,19 @@ void TFileSystem::ProcessHandleOpsQueue()
 
     auto& entries = frontResult.Entries;
     if (entries.empty()) {
+        // HandleOpsQueue has nothing of its own to process right now, but a
+        // release postponed earlier (due to overflow) may still be waiting
+        // in DelayedReleaseQueue - this is normally drained as a side effect
+        // of completing a HandleOpsQueue batch, which does not happen while
```
budevg, line 569: let's add unit test for this case

budevg (changes requested):

budevg (approved):

yegorskii (approved):

PR #7185, opened 2026-09-15
[Filestore] vfs_fuse: limit the total size of the write-back cache and handle ops queue state files

SvartMetal:
### Notes

Builds on the registry refactoring PR (#7146). The registry now records each state file's size (the actual one for a listed file, the configured one for a file created by this process), so the total per component is a sum over the in-memory registry.

**Configuration** (`TVhostServiceConfig`): `WriteBackCacheTotalSizeLimit` (default **10 GiB**) and `HandleOpsQueueTotalSizeLimit` (default **0**). `0` disables the limiting entirely. The size a new state file is created with, and what it is assumed to add to the total, is the existing per-session `WriteBackCacheCapacity` / `HandleOpsQueueSize`; no new knob for that. `DirectoryHandleStorage` is neither sized nor limited.

### Behavior differences

1. **Fresh deployments get a 10 GiB cap on write-back cache state files**: with the default 1 GiB capacity that admits ten cached sessions per host, the eleventh runs without a cache. Raise `WriteBackCacheTotalSizeLimit` or set it to `0` where more are needed.
2. **The limit applies to creating a new state file only.** Existing state is always restored, so pending writes are never dropped.
3. **New state files are created pre-sized (sparse, `ftruncate`).** On tmpfs this reserves nothing physically, the ring buffer resizes the file to its own header + capacity on first open. (Note: the ring buffer's all-zero scan of a pre-sized file and its effect on RSS/tmpfs is addressed separately in #7169)

### Issue
#7048


cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -331,6 +369,34 @@ bool IsDirectory(const TFsPath& path)
...
+// Lists the children of a directory found under a base path. Such a
+// directory being unlistable is reported but, just like an unstatable entry
+// (see IsDirectory()), must not fail every session start: only what is
+// under it goes unnoticed. A directory which has just disappeared is not
+// worth reporting though.
+bool ListDirectory(const TFsPath& dir, TVector<TFsPath>& children)
+{
+    const auto report = [&](const yexception& e)
+    {
+        ReportPersistentStateUnlistableDir(
```
SvartMetal, line 381: Fix for https://github.com/ydb-platform/nbs/pull/7146#discussion_r4000678286

cloud/filestore/libs/vfs_fuse/loop.cpp
```diff
@@ -992,9 +992,19 @@ class TFileSystemLoop final
...
                     HandleOpsQueueStateFileGuard = result.ExtractResult();
-                    handleOpsQueue = CreateHandleOpsQueue(
-                        HandleOpsQueueStateFileGuard.GetFilePath(),
-                        Config->GetHandleOpsQueueSize());
+                    if (HandleOpsQueueStateFileGuard) {
+                        handleOpsQueue = CreateHandleOpsQueue(
+                            HandleOpsQueueStateFileGuard.GetFilePath(),
+                            Config->GetHandleOpsQueueSize());
+                    } else {
+                        // The total size limit of the state files is reached
```
neihar: low: it is not clear that else clause means exactly this

neihar (dismissed):

cloud/filestore/libs/vfs_fuse/persistent_state_manager.cpp
```diff
@@ -34,17 +35,28 @@ struct TComponentConfig
     const TString BasePath;
     // State file name. Points to a static string.
     const TStringBuf FileName;
```
e673: How can you guarantee that this `TStringBuf` will point to a valid memory?
Let's use `TString`.

cloud/filestore/libs/vfs_fuse/loop.cpp
SvartMetal: Made the comment more verbose

e673 (approved):

yegorskii (approved):

PR #7191, opened 2026-09-16
[NBS] Persist mixed-block compaction statistics

vladstepanyuk:
### Notes
Store each range’s mixed-block count in a backward-compatible compaction-map payload and restore it after tablet restart; rows written by older versions default to zero.

Keep persisted counts accurate when compaction skips mixed blobs, exclude deletion markers, count all affected blocks, and let restored counts trigger compaction independently of transient compacted state. Add database, reboot, and skipped-blob coverage.

### Issue
#6930


cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp
```diff
@@ -740,7 +747,10 @@ class TAddBlobsExecutor
             db.WriteCompactionMap(
                 blockIndex,
                 rangeStat.BlobCount,
-                rangeStat.BlockCount);
+                rangeStat.BlockCount,
+                MixedBlocksCountCompactionEnabled
```
SvartMetal: Maybe write it unconditionally but only use it under the flag?

cloud/blockstore/libs/storage/partition/part_actor_compaction.cpp
```diff
@@ -1332,9 +1333,11 @@ class TCompactionTriggerer
 public:
     TCompactionTriggerer(
         const TStorageConfigConstPtr config,
+        const bool mixedBlocksCountCompactionEnabled,
```
SvartMetal: Can we get this from the config passed above?

vladstepanyuk:
https://claude.ai/share/3e14e45d-6206-492f-9240-7e6f0afb5f0d

1. Fixed
2. Added checks to ShouldNotEraseSkippedBlocksFromIndexDuringIncrementalCompactionToMixed test
3. If we persist in this new field, there may be some rollback problems. A new binary with a feature disabled will write counters, after that the old binary will compact and not update these values. After enabling the feature, we may get stale counters that don't match reality.

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
```diff
@@ -324,6 +324,13 @@ void TPartitionActor::CompleteLoadState(
     if (CompactionMapLoadState) {
         LoadNextCompactionMapChunk(ctx);
     } else {
+        // If mixed blocks count compaction is disabled, we need to reset
+        // potentially stale mixed block counts.
+        if (!IsMixedBlocksCountCompactionEnabled()) {
+            for (auto& counter: args.CompactionMap) {
+                counter.Stat.MixedBlockCount = 0;
+            }
+        }
```
SvartMetal, line 333: Why? Maybe we can just keep the counters up to date even if compaction is disabled

cloud/blockstore/libs/storage/partition/part_schema.h
```diff
@@ -231,11 +231,18 @@ struct TPartitionSchema
         {
         };
 
+        struct AdditionalData
+            : public Column<4, NKikimr::NScheme::NTypeIds::String>
+        {
+            using Type = NProto::TCompactionMapAdditionalData;
+        };
```
SvartMetal, line 238: Did we measure the performance impact of adding this?

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
SvartMetal: Is it because you don't want to pay for mixed block counts on disks that don't need mixed blocks compaction?

But it looks like right now you'll always write additional data to the compaction map
https://github.com/ydb-platform/nbs/pull/7191/changes#diff-fa807cd21a530a93ca4f4bf1af8de71c5f168aab434e5bb54a3eed1b5b1c4ac6R1189

cloud/blockstore/libs/storage/partition/part_schema.h
```diff
@@ -231,11 +231,18 @@ struct TPartitionSchema
...
+
         using TKey = TableKey<BlockIndex>;
         using TColumns = TableColumns<
             BlockIndex,
             BlobCount,
-            BlockCount>;
+            BlockCount,
+            AdditionalData>;
```
SvartMetal, line 245: Maybe add the mixed block count by analogy with the previous columns? That would be more consistent

Or do you have plans for an additional data structure, maybe adding more fields to it?

vladstepanyuk: I dont have any plans right now, but it just more flexible with protobuff fields, we can easily add new fields, remove old ones etc. I don't even know can we delete ordinary column from local db.

cloud/blockstore/libs/storage/partition/part_schema.h
vladstepanyuk: I measured execution CPU time of AddBlobs tx under eternal test. 
image
left - with feature enabled, center - old binary, right - feature disabled. 
As we can see performance difference is quite small

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
vladstepanyuk: Changed code so no we are not even filling this fields in compaction map if feature is disabled (no persist no filling)

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
```diff
@@ -324,6 +324,13 @@ void TPartitionActor::CompleteLoadState(
     if (CompactionMapLoadState) {
         LoadNextCompactionMapChunk(ctx);
     } else {
+        // If mixed blocks count compaction is disabled, we need to reset
+        // potentially stale mixed block counts.
+        if (!IsMixedBlocksCountCompactionEnabled(Config, PartitionConfig)) {
+            for (auto& counter: args.CompactionMap) {
+                counter.Stat.MixedBlockCount = 0;
+            }
+        }
```
SvartMetal, line 333: Can we move this code to CompactionMap stored in PartitionState?

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
SvartMetal: Strange, I have a feeling that the additional compaction map is being written unconditionally

https://github.com/ydb-platform/nbs/pull/7191/changes#diff-fa807cd21a530a93ca4f4bf1af8de71c5f168aab434e5bb54a3eed1b5b1c4ac6R1189

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
SvartMetal: Something like this https://pastebin.com/0pCuLGgr

SvartMetal: Like this https://pastebin.com/1HqYQvKQ

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
vladstepanyuk: but we are setting this field to zero here
https://github.com/ydb-platform/nbs/blob/c51bde3d2d1ca5029d5116f4c8cfdc3149723a63/cloud/blockstore/libs/storage/partition/part_actor_addblobs.cpp#L718-L722

We are writing empty Protobufs to the local database if the feature is disabled. We need to update an additional field with an empty Protobuf, because otherwise the old field will remain in the local database and we will receive stale mixed block counts after a restart.

For example, we have a range, and the feature enabled

| blockIndex  | blobCount | blockCount | mixedBlockCount
|-------|-----|------------|-----------|
| 0 | 4  | 2048   | 2048 |

we rolled back feature and ran compaction in this range (we don't update mixedBlockCount field if mixed blocks count compaction disabled)

| blockIndex  | blobCount | blockCount | mixedBlockCount
|-------|-----|------------|-----------|
| 0 | 1  | 1024   | 2048 |

then after next feature enabling we will read that stale mixedBlockCount, which now doesn't have any sense

cloud/blockstore/libs/storage/partition/part_actor_loadstate.cpp
vladstepanyuk: Added commit based on this diff

SvartMetal (approved):

komarevtsev-d (approved):

PR #7204, opened 2026-09-17
[NBS] Don't mark ranges with concurrent writes as compacted when compaction finishes; Fix arithmetic issues in used bytes count calculation

vladstepanyuk:


cloud/blockstore/libs/storage/partition/model/compaction_stats_tracker.cpp
```diff
@@ -77,10 +77,10 @@ TVector<ui32> TCompactionStatsTracker::FinishCompaction()
     VerifyCompactionIsActive();
 
     for (auto& counter: CurrentCompaction->CountersForRangeIndices) {
-        ui32 usedBlockCount = UsedBlocks.Count(
+        const auto usedBlockCount = UsedBlocks.Count(
```
qkrorlqr, line 80: nit: const ui32

qkrorlqr (approved):

komarevtsev-d (approved):

PR #7217, opened 2026-09-17
issue-6956: check the client session in the journalled device tcp handler

arigachnyy:
### Notes
Move the session check from the device adapter to the TCP handler: every
request is validated against the client id from its headers and the
access mode it needs (read-only for reads, read-write for writes), and
requests without a client id are rejected. The adapter accesses the
device directly, so the journal flush no longer needs a client id of its
own. Only devices from the "journalled" pool are exposed, and the TCP
server is not started when there are none.

### Issue


asorotsky (dismissed):

asorotsky:
// The devices are acquired on Start, so these requests are no-ops kept for
// the protocol compatibility.

They are definitely not no op, but can live with that for now

yegorskii (dismissed):

sharpeye (approved):

yegorskii (approved):

PR #7224, opened 2026-09-18
issue-1751: [Filestore] Allow TFileRingBuffer data capacity resize

e673:
### Notes
Allocated by `TWriteBackCache` state files consume space and may prevent newer clients from using write-back cache if total size of the state files hits the limit. After decreasing the default value of its capacity (https://github.com/ydb-platform/nbs/pull/7203) it is needed to reduce the size of existing state files.

This PR add support for `TFileRingBuffer` resize and its use by `TWriteBackCache`.

Resize is implemented as follows:
1. Block adding new entries to the buffer (it reports as it is full);
2. Wait until the buffer is empty;
3. Perform resize;
4. Unlock adding new entries.

`TWriteBackCache` uses state file resize in the following scenarios:
- Initialization: new size is passed at session recreation;
- Draining: state file is resized to zero capacity after drain is finished.

Note: when the resize is needed, new WriteData requests will remain pending until the cache is emptied and resized.

### Issue


e673:
AI review

> I'll pull the PR description, comments and diff, then review the code and tests.I'll fetch the PR diff, description, and comments now.The shell fetch came back empty — let me check the proxy and retry.The diff needs fetch approval. Retrying https://github.com/ydb-platform/nbs/pull/7224.diff and also trying raw file access.raw.githubusercontent works. Fetching the changed files from both the PR branch and main.Now let me read the full implementation files in the PR head.I have the full diff, the surrounding code (ring buffer, accessor, data processors, write-back cache state/request manager), and the PR page. There are no review comments on the PR yet; the description says the resize is applied at session re-creation and after draining.
> 
> ## Findings, most severe first
> 
> ### 1. `SetTargetDataCapacity(0)` is called outside the state lock on a "Non-thread safe" object — concurrent remap
> `write_back_cache.cpp:193-210`: `State.IsDrained()` takes the state mutex and releases it; then `PersistentStorage->SetTargetDataCapacity(0)` runs with no lock. Every other access to `PersistentStorage` goes through `LockStateAndPostponeQueuedOperations()`. `TFileRingBuffer` is documented `// Non-thread safe`, and `IsDrained()` sits in `GetMode()`, which the comment says is on the hot path (called from fuse worker threads).
> 
> Concrete race: thread A (`GetMode()` → `IsDrained()`) writes `Args.DataCapacity = 0` and calls `TryResizeDataCapacity()`; thread B, under the lock, is in `ReleaseHandle`/`UnpinCachedData` → `EvictUnpinnedFlushedEntries` → `Free` → `EraseFreeEntriesFromFront`, which after making the buffer empty checks `IsDataCapacityResizeNeeded()` — it now observes the 0 written by A and resizes too. Two threads doing `ftruncate` + `munmap`/`mmap` on the same `TFileMap`, plus `SetCounters()` on B reading `Header()` through a mapping A is unmapping → SIGSEGV/SIGBUS or torn accessor state. `IsDrained()` doesn't require flushed-but-unevicted entries to be gone, so B's eviction after A's `State.IsDrained()` returned true is exactly the plausible schedule. Fix: perform the resize inside `TWriteBackCacheState` under the guard (e.g. `State.IsDrained()` completes the drain and calls `RequestManager`/storage there), keep the `DrainCompleted` atomic only as the fast path.

I've removed resize at Drain.

> ### 2. `ShouldDownsizeStateFileAfterDrainingCompleted` does not test the feature and passes on `main`
> `write_back_cache_ut.cpp:2251-2263`: the resize is triggered only from `TImpl::IsDrained()`, and the test never calls `IsDrained()`/`GetMode()`, so `SetTargetDataCapacity(0)` is never invoked; the file stays at its original size and `UNIT_ASSERT_LE(newSize, prevSize)` passes because they are equal. It should call `UNIT_ASSERT(b.Cache.IsDrained())` and assert the exact size (`DataOffset` = 256 with `MetadataCapacity = 0`) or at least `UNIT_ASSERT_LT`.

I've removed resize at Drain.
 
> ### 3. Rollback hazard: a state file shrunk to capacity 0 breaks the previous code version
> Old `TFileRingBuffer` ignores the requested capacity for an existing file, so after a rollback it opens the drained file with `DataCapacity = 0`, `GetMaxSupportedAllocationByteCount()` returns 0, `TWriteDataRequestManager::Init` fails (`< 1 MiB + 1016`), the state goes to `IsFailed`, and every WriteData hangs in `HangingRequests`. Consider shrinking to the configured capacity instead of 0, or deleting the file when the session is really gone.

I've removed resize at Drain.

> ### 4. Zero capacity does not "behave like it is full" — `Alloc` returns an error the cache treats as corruption
> `file_ring_buffer.cpp:593`: with `DataCapacity == 0`, `Capabilities().MaxAllocationByteCount == 0`, so `Alloc(n)` returns `E_ARGUMENT`, not `nullptr`. `TFileRingBufferStorage::Alloc` then fires `ReportWriteBackCacheCorruptionError` and the manager sets the failed flag. Today it's unreachable only because `AddWriteDataRequest` rejects in draining mode first; the ring-buffer contract in the header (`behaves like it is full`) is not what the code does. The ut's `!rb.PushBack("12345678").Pushed` in the "Zero size" block passes for either reason — assert `!HasError(...Error)` for the "full" cases so the test pins the contract.

I've removed resize at Drain.

> ### 5. Test double inverts the semantics
> `test_persistent_storage.cpp:89-92`: `SetTargetDataCapacity(0)` → `SetCapacity(0)`, but in `TTestStorage` `Capacity` is an entry *count* and `0` means unlimited. So the double makes storage unlimited where the real one makes it unusable. Better to model "resize when empty" or just ignore the call.

I've removed this functionality.

> ### 6. Overflow guard silently rewrites the requested capacity and still corrupts
> `file_ring_buffer.cpp:324-327`: on overflow `Args.DataCapacity` is clamped to `Max<ui64>() - DataOffset`, then `ResizeAndRemap(Max<ui64>())` throws inside the accessor → `SetCorrupted`. The guard does keep the file from being truncated to a wrapped small size (good), but mutating the caller's target is surprising; reject the value in `SetTargetDataCapacity`/ctor instead (and there's a double space in `=  Max`).

Done

> ### 7. Wrapper drops failure and stats
> `persistent_storage.cpp:188-191`: unlike every other `IPersistentStorage` method, `SetTargetDataCapacity` is `void`, doesn't call `SetCounters()` (so `RawCapacityBytesCount` stays stale after the shrink), and can't report the corruption that `TryResizeDataCapacity` may cause (ENOSPC on grow, EIO). Make it return `NProto::TError`/`bool` and handle it like `Free`.

1. Method `SetTargetDataCapacity` has been removed from `IPersistentStorage`.
2. Added the return value to `TFileRingBuffer::SetTargetDataCapacity`.

> ### 8. Design note: blocking on grow stalls writes unnecessarily
> `Alloc` returns "full" whenever `IsDataCapacityResizeNeeded()`, including when the target is *larger*. A restart with a bigger configured capacity therefore stalls all cached writes until the entire old backlog is flushed (pending queue is global FIFO). Blocking is only required for shrink; for grow you could keep allocating and resize opportunistically when the buffer happens to be empty. Also note the constructor semantics change applies to every `TFileRingBuffer` user in the repo, not only filestore — worth checking blockstore callers that open with a different `dataCapacity`.

This is acceptable and will block operations no more than 2-3 seconds.
And resize will be a rare event.

> ## Correctness of the resize itself (checked, OK)
> `TryResizeDataCapacity` is sound for memory bounds: it only runs when `ReadPos == WritePos` (and `EntryMap` empty, so no outstanding `Alloc`s), zeroes positions first, writes the header capacity *before* truncating on shrink and *after* extending on grow, so `ValidateHeader`'s `DataCapacity <= fileSize - DataOffset` holds at every instant and across a crash. `Validate()` after `ResizeAndRemap` is required (remap resets `Header()` to null), and the final `Validate()` rebuilds the data processor/capabilities. No new throwing paths (`ResizeAndRemap` catches). Nothing in the PR touches move assignment/RAII; `TFileRingBuffer` remains non-copyable/non-movable via `unique_ptr<TImpl>` + user dtor.
> 
> ## Simplifications
> - `SetTargetDataCapacity` and `EraseFreeEntriesFromFront` wrap `TryResizeDataCapacity()`/`TryMigrate()` in `if (Is…Needed())` although the `Try*` functions already check it — drop the outer `if`s.

Not needed - I want the intentions to be more clear.

> - `return Impl->SetTargetDataCapacity(...)` in a void function — drop `return`.

Fixed.

> - Add a one-line comment on the `Min(...)` header write ("header must never claim more capacity than the file has"); it's the crux and currently unexplained.

Done.

> - ut: the "Zero size is valid" block relies on the entry left over from the previous block; push it explicitly, and add `UNIT_ASSERT(!rb.IsCorrupted())` at the end of each block. The ring-buffer test is otherwise deterministic (temp file, exact lengths, V5/V6 both consistent) — no flakiness.

No, I don't want to bloat the code.

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
```diff
@@ -2948,6 +2951,56 @@ Y_UNIT_TEST_SUITE(TWriteBackCacheTest)
...
+            b.Metrics.Storage.RawCapacityByteCount->Get());
+
+        auto flush = b.Cache.FlushNodeData(1);
+        b.RunAllScheduledTasks();
+        write.ProceedAll();
+        UNIT_ASSERT(flush.HasValue());
+
+        UNIT_ASSERT_VALUES_EQUAL(
+            DefaultCacheCapacityBytes,
+            b.Metrics.Storage.RawCapacityByteCount->Get());
```
SvartMetal, line 3017: Let's read all data that we ever write in this test to validate its contents

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -397,6 +438,10 @@ class TFileRingBuffer::TImpl
         if (IsMigrationNeeded()) {
             TryMigrate();
         }
+
+        if (IsDataCapacityResizeNeeded()) {
+            TryResizeDataCapacity();
+        }
```
SvartMetal: TryResizeDataCapacity without guarding if

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -1040,6 +1086,21 @@ class TFileRingBuffer::TImpl
...
+    {
+        if (!ValidateAccess("SetTargetDataCapacity")) {
+            return MakeBufferIsCorruptError();
+        }
+
+        Args.DataCapacity = dataCapacity;
+
+        if (IsDataCapacityResizeNeeded()) {
+            TryResizeDataCapacity();
+        }
```
SvartMetal: Same

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -979,7 +1025,7 @@ class TFileRingBuffer::TImpl
             return 0;
         }
 
-        if (IsMigrationNeeded()) {
+        if (IsMigrationNeeded() || IsDataCapacityResizeNeeded()) {
```
SvartMetal: Consider merging migration and resize logic together to reduce complexity

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
```diff
@@ -617,16 +617,107 @@ Y_UNIT_TEST_SUITE(TFileRingBufferTest)
...
+            UNIT_ASSERT(rb.PushBack("12345678").Pushed);
+
+            UNIT_ASSERT(!HasError(rb.SetTargetDataCapacity(len)));
+            UNIT_ASSERT_VALUES_EQUAL(
+                maxDataSize,
+                rb.GetMaxSupportedAllocationByteCount());
+            UNIT_ASSERT_VALUES_EQUAL(f.GetLength(), len + 248);
+            UNIT_ASSERT_VALUES_EQUAL(0, rb.GetAvailableByteCount());
+            UNIT_ASSERT(rb.PopFront().Removed);
+            UNIT_ASSERT_VALUES_EQUAL(f.GetLength(), len + 256);
```
SvartMetal: expected vs actual?

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
```diff
@@ -301,6 +306,42 @@ class TFileRingBuffer::TImpl
...
+                Args.DataCapacity));
+            return;
+        }
+
+        const ui64 newFileSize = Header()->DataOffset + Args.DataCapacity;
+
+        // Header must never claim more capacity than the file has
+        Header()->DataCapacity = Min(Header()->DataCapacity, Args.DataCapacity);
+
+        if (!ResizeAndRemap(newFileSize) || !Validate()) {
```
SvartMetal, line 328: Do we want to resize file also in constructor in case if we crashed right before this point?

SvartMetal (changes requested):

cloud/filestore/libs/vfs_fuse/write_back_cache/write_back_cache_ut.cpp
e673: Added validation

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Done

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Done

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Merged

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
e673: Done

cloud/storage/core/libs/file_backed_containers/file_ring_buffer.cpp
e673: Yes, fixed + added a test

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
```diff
@@ -683,6 +791,10 @@ Y_UNIT_TEST_SUITE(TFileRingBufferTest)
 
         UNIT_ASSERT(HasError(rb.SetMetadata("x").Error));
         UNIT_ASSERT_STRINGS_EQUAL(dump, Dump(f));
+
+        UNIT_ASSERT(HasError(rb.SetTargetDataCapacity(40)));
+        UNIT_ASSERT_STRINGS_EQUAL(dump, Dump(f));
+
```
SvartMetal: excessive \n

SvartMetal (dismissed):

cloud/storage/core/libs/file_backed_containers/file_ring_buffer_ut.cpp
e673: Removed

SvartMetal (approved):

yegorskii (approved):

PR #7227, opened 2026-09-18
issue-1541: stabilize async close test after queue batching was implemented

WilyTiger:
### Notes
The async-close test became flaky after HandleOpsQueue batching was enabled by default. It may fail with:
```
AssertionError: Expected non-zero handles count after script run, got 0
```
The test expected some server handles to remain after the guest script closed all files and SSH returned. With batches of 32 operations, the queue could drain before this check, causing a failure even when asynchronous close worked correctly.

Observe HandleOpsQueue while the workload runs to verify that asynchronous close is used, then wait for both the queue and server handles to drain. Configure both async-close and async-open tests to process one operation every 50 ms so pending operations remain observable

### Issue
#1541


cloud/filestore/tests/async_close_test/nfs-storage-patch.txt
```diff
@@ -0,0 +1,4 @@
+# Keep the async queue observable while the guest closes files and vhost/tablets
+# restart. Process one operation per interval regardless of the default batch size.
```
SvartMetal: `Process one operation per interval`

Looks like now there are 2 operations?

cloud/filestore/tests/async_close_test/test.py
```diff
@@ -48,48 +74,77 @@ def test():
...
+            if not workload.running:
+                break
+            assert monotonic() - started_at < MAX_WAIT_SECONDS, (
+                f"Guest workload did not finish within {MAX_WAIT_SECONDS} seconds")
+            sleep(1)
 
-    # Check that test script successfully finished.
-    assert 0 == res.returncode
+        workload.wait(timeout=10)
+        assert workload.returncode == 0, (
```
SvartMetal: Looks like this will never fire, because the wait above raises an exception in case of a non-zero return code

SvartMetal (changes requested):

SvartMetal (approved):

yegorskii (approved):

Issue #6684, opened 2026-08-04
[DM] Need to fix potential race in optimizeBaseDisksTask


agalibin:
# Issue was found by Claude

Let me verify the last few links in the chain before writing it out.All line references below are at head `1b36d90`.

## Setup

Config: `MaxActiveSlots=640`, `MaxBaseDiskUnits=640`, `BaseDiskIdleTTL="24h"`.

Pool P = (image `I`, zone `Z`), one base disk, quiet for a day:

```
configs:    image_id=I, zone_id=Z, capacity=640, image_size=0
pools:      image_id=I, zone_id=Z, status=ready, size=640, free_units=640, acquired_units=0
base_disks: id=B1, from_pool=true, status=ready(2), units=640, max_active_slots=640,
            active_slots=0, active_units=0, idle_since=T-30h
free:       (I, Z, B1)
```

`capacity == 640 == one disk's slots` is the load-bearing detail. It's also the natural config for a small pool.

---

## Step 1 — T+0s, collect

`optimize_base_disks_task.go:102` → `GetIdleBaseDisks(ctx, I, Z, 24h, 100)` → `storage_ydb_impl.go:3606`. B1 matches: it's in `free`, `active_units=0`, `status=ready`, `idle_since` non-null and older than `T-24h`. Returns `[B1]`.

Line 119: `reduction += min(uint32(640), 640)` → `reduction = 640`.
Line 120: `IdleBaseDiskIds = ["B1"]`.
Lines 126-129: `poolInfo.Capacity(640) > reduction(640)` is **false** → `newCapacity = 0`.
Line 131: `PoolReductionRequests = [{Zone:Z, Image:I, Capacity:0, UseImageSize:false}]`.

State saved.

## Step 2 — T+5s, capacity zeroed

Line 246 loop schedules `pools.ConfigurePool` with `Capacity: 0` and waits. `configure_pool_task.go:77` → `storage.ConfigurePool` → `configurePool` (`storage_ydb_impl.go:2462`) upserts:

```
configs: image_id=I, zone_id=Z, capacity=0
```

Note the ordering: **all** `PoolReductionRequests` complete before **any** retire (line 273 is a separate loop). Capacity is deterministically 0 before the retire, not racily.

## Step 3 — T+40s, an overlay disk lands on B1

Nothing marks B1 as pending-retirement. It's still `ready`, still `from_pool`, still in `free`.

[...]

Issue #6139, opened 2026-06-05
[Disk Manager] Speed up overlay disk creation


jkuradobery:
Base disks acquire waits for the free disks. Free disks are created by schedule base disks. 
Schedule base disks runs once a minute and schedules disks based on difference between pool size and pool capacity, which rounds to 1, since the default pool capacity is 1. 
If we have many parallel acquire base disks tasks, the base disk creation can starve for free base disks for a long time since base disks are created once a minute.

Possible solutions:
1) Increase default pool capacity
2) Add field for desired pools which is to be increased then no pools are present. (Dynamically extending pool).

SvartMetal:
We also can create base disks using other base disks as source, it will reduce load on YDB/S3 and speed up base disk creation.
We can do even better: introduce new CopyBlocks request which copies data from one disk and write to another, and use this request for TransferFromDiskToDisk task, it will eliminate overhead on transferring data from NBS to Disk Manager dataplane and back

jkuradobery:
Faster new base disk creation want solve us not meeting the overlay disks demand, yes we will create base disks faster, but we still will schedule 1 base disks per minute. data path is not the bottleneck here.

SvartMetal:
We have a known bug: the target base disk can be deleted while the `RetireBaseDisk` operation is in progress. 
As a result, overlay disks that were rebased onto the target base disk fail to be created.

It looks like this bug may occur more frequently after we optimize base disk creation by using existing base disks as the source for creating new base disks.

Issue #1577, opened 2024-07-10
[Disk Manager] race and deadlock in retire_base_disk task


gy2411:
[retire_base_disk](https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/services/pools/retire_base_disk_task.go) task schedules child [rebase_overlay_disk](https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/services/pools/rebase_overlay_disk_task.go) tasks for each overlay disk over the retiring base disk. Such rebase_overlay_disk task might fail if the corresponding overlay disk is created concurrently with retire_base_disk task:

1) [create_overlay_disk](https://github.com/ydb-platform/nbs/blob/main/cloud/disk_manager/internal/pkg/services/disks/create_overlay_disk_task.go) aquires slot for the overlay disk. But it hasn't created the overlay disk in nbs yet.
2) The retire_base_disk task sees this slot and therefore wants to rebase this overlay disk.
3) But the overlay disk is not created in nbs yet. Therefore, rebase fails.
4) The overlay disk is created in nbs.

Moreover, if some of the child rebase_overlay_disk falis, then the parent retire_base_disk task hangs due to deadlock:
1) retire_base_disk waits for all child tasks. But it doesn't check either all child tasks finished successfully or some of them finished with error. 
2) retire_base_disk waits until the base disk is retired.
3) But one of rebase_overlay_disk failed, therefore the base disk still has an overlay disks, therefore the base disk never retires.

Maybe retire_base_disk should check the status of child tasks and fail if one of them fails.

Need to fix these two problems.

gy2411:
Desided that It is ok if retire_base_disk hangs in case of failure of rebase_overlay_disk.

PR #7234, opened 2026-09-19
[Disk Manager] issue-6139: hold source base disk while replacement base disk is being created from it

SvartMetal:
### Notes

`pools.RetireBaseDisk` with `UseBaseDiskAsSrc` (image deletion, `OptimizeBaseDisks`, private `RetireBaseDisks`) generates a replacement base disk `T` with `src_disk_id = A` and creates it via `TransferFromDiskToDisk(A -> T)`. Retiring sets `from_pool=false` on `A`, so `A` becomes deletable the moment its last slot is released, and nothing checks whether a base disk is still being created from it.

Race:
1. `RetireBaseDisk(A, srcDisk=A)` generates `T` (`scheduling`), the overlay slots on `A` target `T`.
2. `ScheduleBaseDisks` -> `CreateBaseDisk(T)` starts the transfer from `A`.
3. The overlay disks on `A` are deleted while the transfer is running -> `A.active_units == 0` -> `A` goes to `deleting` -> `DeleteBaseDisks` destroys `A` in NBS.
4. The transfer fails -> `T` becomes `creation_failed`.
5. Any new overlay disk that acquired a slot on `T` while it was `creating` (`T` is in `free` since step 1) fails `AcquireBaseDisk` -> `CreateDisk` fails.

Fix: a base disk must not be deleted while another base disk is being created from it.

This is also preparatory work for a later optimization that allows base disks to be created from each other (using a base disk as the source) instead of using an image as the source. This might reduce the load on the YDB/S3 that holds the image data and speed up base disk creation

### Issue

#1577, #6684, #6139


SvartMetal:
AI review: https://chatgpt.com/share/6ab26875-9d0c-83eb-aed2-376d99476569

```
Reviewed the displayed head af50a581bfb7293c17917403b31fd31e29b31c47, including the PR description, visible discussion, dependency-accounting helper, and added tests. The main problem I found is rollout safety of the new persistent counter. There are two failure modes, one of which can persist after the rollout finishes.

1. Mixed-version workers can leave a permanent hold on a retired disk

Location: storage_ydb_impl.go, applyInflightDependents().

Priority: P1 if this is deployed with old and new task workers running concurrently.

The accounting is driven by transitions: a child entering the inflight state increments its source’s counter; leaving that state decrements it. There is no reconciliation when the child was already completed by another implementation.

A problematic execution is:

An upgraded worker retires A and generates replacement T, recording A.inflightDependents = 1.
An old worker handles BaseDiskCreated(T). It makes T ready without executing the newly added source-counter update.
An upgraded worker releases the last overlay slot on A.
A now has no active units and is outside the pool, but its counter remains 1, preventing deletion.

Replaying completion on an upgraded worker does not repair this: the persisted state of T is already ready, so both oldHolds and holds are false. No decrement is generated. This follows directly from the transition-based accounting and the new deletion condition.

The result is a retained base disk with no actual inflight dependents. The new consistency check detects the mismatch, but only returns an error; it does not repair the counter. The added tests also establish that retaining the source keeps IsBaseDiskRetired() false, so this affects retirement completion, not just storage reclamation.

Required change: either document and enforce a deployment procedure that excludes mixed-version writers, or introduce a coordinated rollout: deploy counter maintenance, ensure all writers understand it, reconcile existing counts, and only then rely on the counter for deletion decisions. Adding the column before deployment alone is insufficient.
```
Fixed, added a feature flag

```
2. Existing inflight replacements are not protected after upgrading

Location: common.go, the new inflight_dependents column and scanBaseDisk(); storage_ydb_impl.go, the underflow handling.

Priority: P2 — the advertised fix does not cover work already in progress.

Existing source rows read the newly added nullable field as zero. Accounting only increments the counter when a child transitions from not holding to holding; reading an existing scheduling or creating child does not register its dependency. The underflow comment explicitly anticipates children that were already being created when the column was introduced.

Consequently, this execution remains possible even after every worker has been upgraded:

Before upgrade:
    A is retired.
    T is creating from A.
    A still has one overlay slot.

After upgrade:
    A.inflightDependents reads as 0.
    Release the last overlay slot on A.
    A becomes deleting while T is still creating.

That preserves the source-deletion race described in the PR. Clamping a subsequent decrement to zero does not help: protection was needed before the child finished.

Required change: bootstrap counts from existing inflight children before enabling the deletion guard, with synchronization against concurrent transitions. Alternatively, explicitly require draining this work before the cutover. The warning-and-clamp branch is arithmetic protection, not a migration mechanism.

These two findings share a deployment concern, but their effects differ: missing counts permit premature deletion; stale positive counts prevent eventual deletion.
```
It is OK

doc/disk_manager/pools/README.md
jkuradobery, line 176: A bit hard to comprehend. A base disk which used as a source for creating another base disk (used as `src_disk_id` of replacement base disk) must be locked until the base disk creation is finished.

doc/disk_manager/pools/README.md
```diff
@@ -171,6 +173,20 @@ contribute capacity to the pool.
 units if an overlay acquired a slot while the base disk was `creating`. When
 those active units are released, invariants move it to `deleting`.
 
+A base disk that is the `src_disk_id` of a replacement base disk holds that
+source until the replacement leaves `scheduling`/`creating` (it becomes
+`ready`, `creation_failed`, or is deleted before creation). The hold is the
+`inflight_dependents` counter on the source. Every base disk transition that
+makes a disk with `src_disk_id` inflight increments the source's counter, and
+every transition that takes it out of the inflight state decrements it
```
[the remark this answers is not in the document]
SvartMetal: This feature is temporary - we'll delete it after the first deployment

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -648,32 +649,162 @@ func (s *storageYDB) getPoolOrDefault(
...
+
+			deltas[t.state.srcDiskID]--
+		}
+	}
+
+	// Sort for determinism.
+	srcDiskIDs := make([]string, 0, len(deltas))
+	for srcDiskID := range deltas {
+		srcDiskIDs = append(srcDiskIDs, srcDiskID)
+	}
```
jkuradobery, line 703: Line break

cloud/disk_manager/internal/pkg/services/pools/storage/storage_ydb_impl.go
```diff
@@ -648,32 +649,162 @@ func (s *storageYDB) getPoolOrDefault(
...
+			// holdBaseDisksWithInflightDependents was disabled.
+			logging.Info(
+				ctx,
+				"inflight dependents underflow for source base disk %+v, delta %v",
+				srcDisk,
+				delta,
+			)
+			srcDisk.inflightDependents = 0
+		} else {
+			srcDisk.inflightDependents = uint64(
```
jkuradobery, line 756: Why not just store int64 in ydb?

SvartMetal: Reasonable, will fix it

BarkovBG (approved):

PR #7235, opened 2026-09-21
[NBS] Back up boot info for tablets started through Hive Local

agalibin:
Capture boot information for Volume and DiskRegistry tablets started through Hive Local, extending the existing external boot capture path. This prepares fallback recovery for ordinary and base disks whose Volume tablets may start without an endpoint or mount request.

- Report storage metadata and the actual executor generation when a leader tablet activates.
- Reuse the existing HiveProxy backup writer and file formats. Capture requires a configured `TabletBootInfoBackupFile`.
- Prevent delayed lower-generation updates from replacing newer entries collected by the running process, including their channel history. Equal-generation updates remain accepted.
- Ignore the new notifications in fallback mode. Existing external boot and fallback startup behavior remain unchanged.
- Add unit tests for activation and reboot, capture before mount, both backup formats, fallback reads, missing backup configuration, and updates arriving out of order across both capture paths.
- Preserve tablet boot info backups across restarts.


agalibin:
> Claude's review:
> 
> > Summary
> > The core change is sound: the new TEvTabletBootInfo notification, the HandleTabletBootInfo forwarding, the fallback IgnoreFuncs, and the "drop lower generation" guard are all correct and the tests do reproduce the two behaviors they claim to (the guard test fails without the guard; the retain-across-restart test fails without commit 2). No memcpy/memset/OOB anywhere, no new move-assignment, no throwing in library code. I did not find a crash bug or flaky test. What I found is one design risk introduced by commit 2 that I think needs a decision, one crash-surface hardening item, and several simplifications.
> > Findings
> > 
> > 1. Commit 2 makes the backup file grow without bound and serves ever-staler entries in fallback — needs a deliberate decision
> > 
> > tablet_boot_info_backup.cpp:44-90. Before this PR, normal mode started BackupProto empty and the first Backup() overwrote the file with only the tablets booted by this process — an implicit GC on every restart. Now entries are loaded into the live proto and nothing ever removes one. Consequences:
> > Deleted volumes and tablets that moved to other nodes stay in the file forever; the file (and the text-format serialization that runs on the IO pool every 10s when dirty) grows monotonically with the lifetime of the node.
> > In fallback mode the node hands out channel history that can be years stale for tablets that long since moved. The generation guard does not help here: no newer update ever arrives for a tablet that left this node. Booting a tablet with an outdated channel history is exactly the case where it can miss the newest group's log and come up on an old snapshot. This risk existed before but was bounded by one process lifetime; now it is unbounded.
> > Over time every node's file accumulates overlapping entries for the same tablets (A→B→A moves), so "whatever is in my backup" becomes a worse signal for "what this node should try to boot in fallback."
> > I'm not saying revert it — the motivation (volumes that never re-booted in this process) is legitimate — but the PR should at least state the trade-off, and ideally add a per-entry timestamp with an age cutoff on load, or a max-entries cap. The description bullet "Preserve tablet boot info backups across restarts" undersells the behavior change.
> > 
> > 2. Y_ABORT_UNLESS(msg->StorageInfo) is now reachable from a public notification
> > 
> > tablet_boot_info_backup.cpp:283, reached via hive_proxy_actor.cpp:673-689. Previously the only sender of TEvUpdateTabletBootInfoBackupRequest was HiveProxy's own external-boot code. Now any tablet in the process can send TEvTabletBootInfo, and HandleTabletBootInfo forwards msg->StorageInfo unchecked. Today's two callers pass Info() which is never null in OnActivateExecutor, but a public API that aborts the whole nbs process on a null pointer is the wrong contract for library code. Validate in HandleTabletBootInfo (drop + LOG_ERROR/critical event) and leave the abort for the internal path, or better, convert the internal one too.
> > 
> > 3. LoadFromBinaryFormat leaves partial state on non-throwing failure — now in the live proto
> > 
> > tablet_boot_info_backup.cpp:206. MergeFromString returning false leaves whatever it parsed in backupProto, and unlike the catch branch there is no reset. The caller then tries LoadFromTextFormat, which merges on top. Pre-PR this only polluted InitialBackupProto; now it's BackupProto, which gets persisted. In practice you're saved because a text parse of binary garbage throws and resets, and a binary parse of a text file fails on the first tag — but that's accidental. Simpler and safer: parse each attempt into a local TTabletBootInfoBackup and BackupProto = std::move(tmp) on success. That also deletes the BackupProto.Clear() at line 80 and the reset-in-catch in both loaders.
> > 
> > 4. Test helper TTabletBootInfoObserver — RAII is right, but it's implicitly copyable
> > 
> > testlib/tablet_boot_info.h. The constructor installs a filter capturing this; the destructor restores the previous one. Good. But the class has a reference member and no explicitly deleted copy/move, so the implicit copy constructor exists: a copy would leave the runtime's lambda pointing at the original this and both destructors would "restore" the filter. Add = delete for copy/move. Also, the lambda guards against an empty PreviousEventFilter but the destructor doesn't — SetEventFilter(std::move(empty)) would make the next Send throw bad_function_call. Not reachable today (default filter is non-empty), but make the two consistent.
> > Observe runs inside the runtime's Send hook under Runtime->Mutex and throws via UNIT_ASSERT — that's how other tests in the repo do it, so acceptable.
> > 
> > 5. Complexity that can go
> >    HandleListTabletBootInfoBackups (:328) and HandleGetTabletBootInfos (:345) are now byte-identical after InitialBackupProto removal. One CollectTabletBootInfos() helper.
> >    HandleUpdateTabletBootInfoBackup does data.find then data[id] = ... — assign through it when found.
> >    IgnoreFunc(TEvHiveProxy::TEvTabletBootInfo) in THiveProxyActor::StateFallback (hive_proxy_actor.cpp:651) is redundant — that state already has default: break;. The one in THiveProxyFallbackActor::StateWork is needed (config-fallback mode registers that actor as the service directly). Keep it if you want it as documentation, but say so.
> >    cloud/storage/core/libs/api gains a PEERDIR on libs/actors for a five-line NCloud::Send wrapper. ctx.Send(MakeHiveProxyServiceId(), new TEvTabletBootInfo(...)) inline in the header avoids the new dependency on the api library.
> >    hive_proxy_ut.cpp:1595 — the sendUpdate lambda duplicates TBackupTestEnv::Update; that test could just use TBackupTestEnv.
> >    volume_ut.cpp:11764-11772 — PrepareTestActorRuntime({}, {}, {}, {}, {}, false, nullptr, MOUNTED) with five positional {}s. TTestRuntimeBuilder already exists; add .With(EVolumeStartMode).
> > 6. Questions, not defects
> >    Why skip reporting for MOUNTED volumes? (volume_actor.cpp:316). The external-boot path stores Hive's suggested generation; the tablet's actual generation is max(suggested, stateStorage+1) and can be higher. Reporting from OnActivateExecutor in both modes would be accepted by the guard (actual ≥ suggested) and would make the backup carry the true generation for mounted volumes too. If the intent is only to avoid a redundant message, that's a weaker reason than the accuracy gain.
> >    The guard compares two different notions (Hive's suggestion on the external path vs. actual executor generation on the new path). I traced the sequences and they're monotonic in practice because Hive learns the real generation via Local's tablet-status before suggesting again. Worth a one-line comment in HandleUpdateTabletBootInfoBackup so the next reader doesn't have to re-derive it.
> >    Tests — verified OK
> >    All timing is virtual; dispatch is deterministic; the AdvanceCurrentTime(120s) pattern is copied from the existing runtime-fallback test. No flakiness I could find.
> >    Ordering between fire-and-forget SendTabletBootInfo and the subsequent request/response calls holds because everything traverses the same edge→HiveProxy→backup mailboxes in FIFO order.
> >    File names are distinct per test and per loop iteration; DeleteIfExists or truncating write at the start of each.
> >    TBackupTestEnv member init order (Runtime → HiveProxy → Sender) is correct.
> >    The removed InitialBackup test's coverage is subsumed by ShouldRetainBackupEntriesAcrossNormalRestarts.
> >    Fallback "ignore reports" is exercised in both config-fallback and runtime-fallback modes, and the in-memory +1 bump is asserted to survive ignored reports.
> > 
> > If you'd like, I can turn items 2, 3 and 5 into a concrete patch on top of the branch.

1. так и предполагалось, чтобы сделать механизм в следующих PR, но решил пока убрать совсем и добавить уже полностью позже.
2. Добавлена проверка в HandleTabletBootInfo: некорректное уведомление отбрасывается с LOG_ERROR.
3. поправлено.
4. поправлено.

5. Complexity that can go
HandleListTabletBootInfoBackups (:328) and HandleGetTabletBootInfos (:345) общий код вынесен в CollectTabletBootInfos(), обработчики идентичн и возможно в будущем стоит удалить один из них.
HandleUpdateTabletBootInfoBackup does data.find then data[id] - поправлено.
IgnoreFunc(TEvHiveProxy::TEvTabletBootInfo) in THiveProxyActor::StateFallback (hive_proxy_actor.cpp:651) is redundant — оставлен комментарий.
cloud/storage/core/libs/api gains a PEERDIR on libs/actors for a five-line NCloud::Send wrapper. - поправлено.
hive_proxy_ut.cpp:1595 — стало неактуально.
volume_ut.cpp:11764-11772 — поправлено.

6. Questions, not defects
Why skip reporting for MOUNTED volumes? (volume_actor.cpp:316) - поправлено.

cloud/blockstore/libs/storage/testlib/tablet_boot_info.h
```diff
@@ -0,0 +1,128 @@
...
+    const ui64 TabletId;
+    TString ExpectedStorageInfo;
+    THashMap<NActors::TActorId, ui32> RestoredGenerations;
+    THashMap<NActors::TActorId, ui32> ReportedGenerations;
+    ui32 LastGeneration = 0;
+
+public:
+    TTabletBootInfoObserver(
+        NActors::TTestActorRuntimeBase& runtime,
+        const NKikimr::TTabletStorageInfo& storageInfo)
```
SvartMetal: expectedStorageInfo?

cloud/storage/core/libs/api/hive_proxy.cpp
```diff
@@ -11,4 +13,16 @@ TActorId MakeHiveProxyServiceId()
     return TActorId(0, "sto-hproxy");
 }
 
+void ReportTabletBootInfo(
```
SvartMetal: SendTabletBootInfoToHiveProxy

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -41,6 +42,90 @@ static constexpr ui64 FakeMissingTablet = 0x0000000000840104;
...
+    ui64 tabletId,
+    TTabletTypes::EType tabletType,
+    ui32 generation)
+{
+    TTabletStorageInfoPtr storageInfo =
+        CreateTestTabletInfo(tabletId, tabletType);
+    for (auto& channel: storageInfo->Channels) {
+        channel.History.emplace_back(
+            generation,
+            channel.History.back().GroupID + generation);
```
SvartMetal: channel.History.back().GroupID + generation

Looks like a weird thing - what is it supposed to represent?

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -41,6 +42,90 @@ static constexpr ui64 FakeMissingTablet = 0x0000000000840104;
...
+    }
+    return storageInfo;
+}
+
+void AssertStorageInfo(
+    const NKikimrTabletBase::TTabletStorageInfo& actual,
+    const TTabletStorageInfoPtr& expected)
```
SvartMetal: Let's use the conventional expected vs actual order

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -41,6 +42,90 @@ static constexpr ui64 FakeMissingTablet = 0x0000000000840104;
...
+{
+    NKikimrTabletBase::TTabletStorageInfo expectedProto;
+    TabletStorageInfoToProto(*expected, &expectedProto);
+    UNIT_ASSERT_VALUES_EQUAL(actual.DebugString(), expectedProto.DebugString());
```
SvartMetal: expected vs actual

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -41,6 +42,90 @@ static constexpr ui64 FakeMissingTablet = 0x0000000000840104;
...
+}
+
+void AssertStorageInfo(
+    const TTabletStorageInfoPtr& actual,
+    const TTabletStorageInfoPtr& expected)
```
SvartMetal: same

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -41,6 +42,90 @@ static constexpr ui64 FakeMissingTablet = 0x0000000000840104;
...
+    } else {
+        MergeFromTextFormat(path, backup);
+    }
+    return backup;
+}
+
+void AssertBackupEntry(
+    const NHiveProxy::NProto::TTabletBootInfoBackup& backup,
+    const TTabletStorageInfoPtr& storageInfo,
+    ui32 generation)
```
SvartMetal: expectedStorageInfo

expectedGeneration

?

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -41,6 +42,90 @@ static constexpr ui64 FakeMissingTablet = 0x0000000000840104;
...
+{
+    const auto it = backup.GetData().find(storageInfo->TabletID);
+    UNIT_ASSERT(it != backup.GetData().end());
+    AssertStorageInfo(it->second.GetStorageInfo(), storageInfo);
+    UNIT_ASSERT_VALUES_EQUAL(it->second.GetSuggestedGeneration(), generation);
```
SvartMetal: expected vs actual?

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -1243,6 +1356,377 @@ Y_UNIT_TEST_SUITE(THiveProxyTest)
...
+            10);
+        env.SendTabletBootInfo(sender, storageInfo, 10);
+        env.SendBackupTabletBootInfos(sender, S_OK);
+        const auto savedFile = TFileInput(backupFilePath).ReadAll();
+
+        env.SendTabletBootInfo(sender, {}, 11);
+        env.SendBackupTabletBootInfos(sender, S_FALSE);
+        UNIT_ASSERT_VALUES_EQUAL(
+            TFileInput(backupFilePath).ReadAll(),
+            savedFile);
```
SvartMetal: expected vs actual?

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -1243,6 +1356,377 @@ Y_UNIT_TEST_SUITE(THiveProxyTest)
...
+        const auto result = env.SendGetTabletBootInfos(sender, S_OK);
+        UNIT_ASSERT_VALUES_EQUAL(result.TabletBootInfos.size(), 1);
+        AssertStorageInfo(
+            result.TabletBootInfos[0].StorageInfoProto,
+            storageInfo);
+        UNIT_ASSERT_VALUES_EQUAL(
+            result.TabletBootInfos[0].SuggestedGeneration,
+            10);
```
SvartMetal: same


expected vs actual

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -1243,6 +1356,377 @@ Y_UNIT_TEST_SUITE(THiveProxyTest)
...
+        const auto original = CreateStorageInfoWithHistory(
+            FakeTablet2,
+            TTabletTypes::BlockStoreVolume,
+            10);
+        env.SendTabletBootInfo(sender, original, 10);
+        env.SendTabletBootInfo(sender, original, 10);
+        env.SendBackupTabletBootInfos(sender, S_OK);
+
+        auto backup = ReadTabletBootInfoBackup(backupFilePath);
+        UNIT_ASSERT_VALUES_EQUAL(backup.GetData().size(), 1);
```
SvartMetal: expected vs actual?

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
```diff
@@ -1243,6 +1356,377 @@ Y_UNIT_TEST_SUITE(THiveProxyTest)
...
+
+        const auto updated = CreateStorageInfoWithHistory(
+            FakeTablet2,
+            TTabletTypes::BlockStoreVolume,
+            11);
+        env.SendTabletBootInfo(sender, updated, 11);
+        env.SendBackupTabletBootInfos(sender, S_OK);
+
+        backup = ReadTabletBootInfoBackup(backupFilePath);
+        UNIT_ASSERT_VALUES_EQUAL(backup.GetData().size(), 1);
```
SvartMetal: same

cloud/storage/core/libs/hive_proxy/hive_proxy_ut.cpp
agalibin: It just generates different test group IDs, but that isn't necessary because FromGeneration already distinguishes the histories. We can simply reuse the existing GroupID.

cloud/blockstore/libs/storage/volume/volume_actor.cpp
```diff
@@ -312,6 +313,7 @@ void TVolumeActor::OnActivateExecutor(const TActorContext& ctx)
     ScheduleRegularUpdates(ctx);
 
     if (!Executor()->GetStats().IsFollower()) {
+        UpdateTabletBootInfoBackup(ctx, Info(), Executor()->Generation());
```
SvartMetal: Let's add a comment about why it's needed

cloud/blockstore/libs/storage/disk_registry/disk_registry_actor.cpp
```diff
@@ -163,6 +165,10 @@ void TDiskRegistryActor::OnActivateExecutor(const TActorContext& ctx)
     RegisterCounters(ctx);
 
     if (!Executor()->GetStats().IsFollower()) {
+        NCloud::NStorage::UpdateTabletBootInfoBackup(
+            ctx,
+            Info(),
+            Executor()->Generation());
```
SvartMetal: Let's add a comment about why it's needed

SvartMetal (approved):

Issue #7240, opened 2026-09-21
[Filestore ] Move ShouldCreateALotOfShards and ShouldCreateALotOfShardsThrottled to a separate test suite


ruslan-gar:
As these tests are heavy, it is better to move them to a separate test suite and change 
FORK_SUBTESTS to MODULO in order to guarantee that heavy adjacent tests go to different chunks.

PR #7241, opened 2026-09-21
issue-7240: [Filestore] Separate suite for heavy tests that create a lot of shards

ruslan-gar:
### Notes
Heavy test was moved to a separate test suite

### Issue
#7240


ruslan-gar:
GPT-5.6 Sol:

### Findings
No confirmed bugs

I did not find a blocker, important, or minor correctness defect introduced by the latest revision.

The change is essentially a test-suite split plus extraction of helpers/counters. I checked the moved heavy tests, the newly shared helper implementation, and all three affected ya.make files. The latest CI result is also green for 293e3b5.

A few areas that looked suspicious on first pass but did not survive surrounding-code inspection:

TShardRequestCounter installs an event filter and restores the previous filter in its destructor; the counter is used with a lifetime enclosing the request activity, so I did not find a dangling-reference/lifetime bug there.
The request filter intentionally chains to the previous filter rather than replacing its semantics.
CreateOrResizeFilesystem waits for the post-ConfigureShards tablet startup event to identify the main tablet; the source itself documents that this depends on the “first tablet to start” invariant, so I did not promote that to a speculative bug.
The new heavy suite still compiles the common helper implementation it uses, and the original suite retains the same helper source; there is no missing-definition/linkage issue in the resulting targets.
Question / invariant worth confirming

### Question, not a confirmed bug: cookie uniqueness for TShardRequestCounter.

Location: cloud/filestore/libs/storage/service/service_ut_helpers.h, TShardRequestCounter::CountEvent, specifically the EvConfigureAsShardRequest / EvConfigureAsShardResponse and EvDescribeFileStoreRequest / response bookkeeping. The implementation stores the originating sender in hash maps keyed only by ev->Cookie.

Concrete scenario: two matching requests are simultaneously in flight with the same cookie. The second request overwrites the first sender entry; when the first response arrives, its sender lookup can no longer match the original request, so the response counter is understated.

Expected: each in-flight request has an independently matchable request/response pair.

Actual under that scenario: bookkeeping is keyed by a single ui64 cookie and can collapse concurrent entries.

I am not calling this a bug because it depends on an implicit invariant about cookie uniqueness for these events. The current tests pass, and I did not find evidence in the surrounding code available to this review that the invariant is violated. If that uniqueness guarantee is contractual, it would be useful to make it explicit near the helper.

### Review-history check

There were no existing human review comments to repeat or mark as fixed in the accessible PR history. The only earlier revision shown is 3fd8be9 (“fix eof space”), followed by merge commit 293e3b5; the latter is the tested HEAD and its Filestore checks passed.

### Conclusion: 
no actionable code-review finding at the requested severity levels for HEAD; one non-blocking invariant question around cookie uniqueness.

ruslan-gar:
About cookie uniqueness.
Cookie uniqueness is of the counted events is guaranteed by TAlterFileStoreActor.
Cookie is a shard number and requests of a particular type are sent to a particular shard once.

asorotsky (approved):

nisqatsi (approved):

PR #7242, opened 2026-09-21
[Filestore] add ShardsCount tablet sensors

debnatkh:
### Notes
can be useful for dashboards displaying the number of shards

### Issue
Small change – no issue here


e673 (approved):

debnatkh:
**_Fable 5.1 review:_**

I need the actual diff. Let me pull it via the GitHub API.The PR is small and clean: it adds a `ShardsCount` tablet sensor (header field, `REGISTER_LOCAL`, one `store()` in `RegisterStatCounters`) plus a unit test. Style matches the neighbouring `HasOverrides` sensor exactly, CI is green, and it already has one approval. I'd approve it too, with one substantive suggestion.

**Main point: the value only refreshes on tablet load.** `Metrics->ShardsCount.store(...)` lives in `RegisterStatCounters`, which runs once at startup. The periodic `HandleUpdateCounters` path calls `UpdateMetrics(...)` and never touches `ShardsCount`, so after `ConfigureShards` (or a shard being added later) the sensor stays stale until the tablet reboots. The test actually encodes this: it has to call `tablet.RebootTablet()` after `ConfigureShards` to see `2`. For a "number of shards" dashboard that's misleading, since the whole point is to watch that number change. `UpdateMetrics` already receives `const NProto::TFileSystem& fileSystem` and stores things like `fileSystem.GetNodesCount()` and `GetDirectoryCreationInShardsEnabled()`, so the natural fix is to move the store there:

```cpp
Store(Metrics->ShardsCount, fileSystem.GetShardFileSystemIds().size());
```

Then `RegisterStatCounters` gets it for free (it calls `UpdateMetrics` before `Register`), and the test could drop the second reboot in favour of advancing the runtime past `UpdateCountersInterval`. Even if the author prefers to keep it startup-only (like `HasOverrides`, which is also config-ish), it's worth a comment or a follow-up, because this sensor is far more likely to change at runtime than `HasOverrides` is.

Smaller notes:

- `REGISTER_LOCAL` with `MT_ABSOLUTE` means no aggregation across filesystems, which is right for a per-filesystem count.
- `.size()` returns `int` from the protobuf repeated field, stored into `std::atomic<i64>`; fine, no narrowing.
- The test itself is reasonable: checks 0 before, 2 after, on the `filesystem=test` label, and the extra entry in the existing counters list guards the default. The `InitSession` call is unnecessary for what's being verified, but harmless.

Verdict: approve, ideally with the store moved into `UpdateMetrics` so the sensor tracks reality rather than boot-time state.

---

The number of shards is not intended to change during normal execution – we intentionally kill the tablet after changing the the shard count

agalibin (approved):

PR #7259, opened 2026-09-21
[NBS, Filestore] drain in-flight gRPC requests on shutdown

yegorskii:
### Notes
Drain in-flight gRPC requests on shutdown. Resused TRequestsInFlight from storage/core/libs/grpc everywhere. Replaced Y_ABORT_UNLESS with STORAGE_VERIFY

### Issue


cloud/storage/core/libs/grpc/request.h
```diff
@@ -24,7 +26,20 @@ class TRequestHandlerBase
 
 using TRequestHandlerPtr = std::unique_ptr<TRequestHandlerBase>;
 
-template<typename TRequestHandler >
+template <const TStringBuf& EntityTypeName>
```
qkrorlqr: const ref as template arg? why not just plain `TStringBuf`?

cloud/storage/core/libs/grpc/request.h
```diff
@@ -24,7 +26,20 @@ class TRequestHandlerBase
...
+struct TNoIdRequestsInFlightDiag
```
qkrorlqr: let's not introduce such a helper

the whole point of `STORAGE_VERIFY` was to enforce explicit passing of entity ids into verify statements

it's like the whole idea, the main feature

cloud/storage/core/libs/grpc/request_ut.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+        while (!handler.Cancelled.load()) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        }
+        // A correct Shutdown() spins while the handler is registered, so it
+        // cannot have returned here whatever the timing - this assert never
+        // flakes. There is no way to directly observe "still inside Shutdown()",
+        // so the 1s just gives a regressed, non-draining Shutdown() ample time
+        // to wrongly return and be caught.
+        std::this_thread::sleep_for(std::chrono::seconds(1));
```
qkrorlqr: no, no sleeps in unit tests please, it makes tests run longer and at the same time makes them unstable

qkrorlqr (changes requested):

cloud/storage/core/libs/grpc/request_ut.cpp
yegorskii: Ok, I will remove it, but we decrease probability to catch that draining is broken

cloud/storage/core/libs/grpc/request.h
yegorskii: TStringBuf cannot be Non-Type Template Parameter

```
error: type 'TStringBuf' (aka 'TBasicStringBuf<char>') of non-type template
parameter is not a structural type
```

qkrorlqr (approved):

apliev (approved):

PR #7261, opened 2026-09-22
[NBS] Reduce Disk Registry migration start redo size

Nik-Pavlov:
### Notes

`TStartMigration` previously rewrote the disk configuration and placement group and added a reallocation request for each migration, amplifying commit redo size for large disks. This extends the batching approach from [#7167](https://github.com/ydb-platform/nbs/pull/7167) to automatic migration starts.

`StartDeviceMigrations` starts the whole migration list in one call and returns one result per source device. It persists each changed disk once and adds one reallocation request per changed disk. Each affected placement group is persisted once per batch, with one configuration-version increment.

Placement-group rack information is updated in memory after each successful start, so subsequent allocations respect the newly occupied racks. Migration ordering, device-based limits, per-migration logging, and continued processing after individual failures are preserved. Forced migrations keep the single-migration API as `StartForceMigration`, which persists the disk and placement group itself.

### Measured redo reduction

These are **synthetic** numbers from a controlled comparison of this branch against its parent (which already includes the merged finish-migration optimization). They measure redo bytes for **the single `TStartMigration` transaction** only, in the worst case where all of a disk's devices sit on one agent and are migrated at once - up to 64 devices on one disk.

| Migrations | Parent redo bytes | Branch redo bytes | Reduction |
|---|---|---|---|
| 8 | 5,964 | 1,029 | 82.75% |
| 32 | 63,536 | 3,241 | 94.90% |
| 64 | 232,544 | 6,185 | 97.34% |


komarevtsev-d (approved):

SvartMetal:
AI review: https://claude.ai/share/d32885c6-8df7-4df2-a30f-0545b0ef2098

PR #7262, opened 2026-09-22
issue-6956: serialize journal log records and metadata

arigachnyy:
### Notes
Add the on-disk format the journal will keep in its meta store: a log record is a fixed header (lsn, prev lsn, page mapping count) followed by its page mappings, the journal metadata is a versioned header holding the last acked lsn. Deserialization rejects truncated buffers, trailing bytes, mismatched counts and unknown format versions.

The log record promise now carries a TWriteLogRecordResponse, which is what the journal hands back to the writer.

### Issue


qkrorlqr (approved):

cloud/fastshard/journal/impl/log_record.cpp
```diff
@@ -1 +1,124 @@
...
+
+    return buffer;
+}
+
+std::optional<TJournalMetadata> DeserializeMetadata(const TBuffer& buffer)
+{
+    if (buffer.Size() != sizeof(TMetadataHeader)) {
+        return std::nullopt;
+    }
+
```
asorotsky, line 56: Shouldn't it be >=? E.g. you read header, move buffer read pos, then you can parse next structure?

asorotsky (approved):

PR #7267, opened 2026-09-22
issue-6956: write and read pages through the journal

arigachnyy:
### Notes
TJournal now writes a record by storing its pages in the page store and its page mappings in the meta store, chains it and, once every record before it is in place, adds it to the page index. Reads are served from the page index, returning only the pages the journal holds. Both reject page ranges that are outside the device or intersect each other.

Restore, the tail, the last acked lsn and the flush cycle are still not implemented.

### Issue


cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -96,6 +95,139 @@ NCloud::NProto::TError ValidatePageRanges(
...
+    // Verify that the locations hold exactly as many pages as the request -
+    // the page store allocates the page count it is asked for
+    STORAGE_VERIFY(
+        pageCount == totalPageCount(locations),
+        "PageCount",
+        pageCount);
+
+    TVector<TPageMapping> mappings;
+
+    size_t requestIndex = 0;
```
qkrorlqr, line 177: probably makes sense to implement something like `TPageGroupIterator` which can be initialized from a const ref to a vector of `TPageRangeRef`s

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -96,6 +95,139 @@ NCloud::NProto::TError ValidatePageRanges(
...
+    for (const auto& group: request.GetPageGroups()) {
+        ranges.push_back(
+            {.FirstPageNo = group.GetFirstPageNo(),
+             .PageCount = group.ContentSize()});
+    }
+
+    return ranges;
+}
+
+TVector<TBuffer> GetPages(const NCloud::NProto::TWriteLogRecordRequest& request)
```
qkrorlqr, line 125: better to use `TArrayRef` or something similar instead of `TBuffer` in the parts of the code where owning the pages is not needed

cloud/fastshard/journal/impl/journal.cpp
```diff
@@ -211,7 +475,36 @@ TFuture<NCloud::NProto::TReadPagesResponse> TJournal::Read(
...
+
+    TVector<TPageRangeRef> ranges;
+    ranges.reserve(request.PageGroupRefsSize());
+    for (const auto& ref: request.GetPageGroupRefs()) {
+        ranges.push_back(
+            {.FirstPageNo = ref.GetFirstPageNo(),
+             .PageCount = ref.GetPageCount()});
+    }
+
+    ui64 lastFlushedLsn = 0; // TODO: implement with MarkRecordAsFlushed
```
qkrorlqr, line 491: TODO(#issue-no)

qkrorlqr (approved):

asorotsky (approved):

cloud/fastshard/journal/impl/journal.cpp
arigachnyy: will do in the next PR

cloud/fastshard/journal/impl/journal.cpp
arigachnyy: will do in another PR

Issue #7199, opened 2026-09-16
[Filestore] Executor DB tables are not queryable


szumal:
We do not have a convenient way to run queries against tables defined https://github.com/ydb-platform/nbs/blob/main/cloud/filestore/libs/storage/core/tablet_schema.h

We would like to have a SQL-like parser in the client that would allow us to run queries like:
`SELECT * FROM NodeRefs WHERE child_id in {}`.

For now, we only want basic select queries with column matching in the `WHERE` clause and limits.

As a bonus, it would be nice to have the query support in the monitoring page.

PR #7269, opened 2026-09-22
issue-7199: Add custom query parsing

szumal:
### Issue
https://github.com/ydb-platform/nbs/issues/7199


cloud/filestore/apps/client/lib/query/parser.cpp
```diff
@@ -0,0 +1,318 @@
...
+                            Input[Position] == '_'))
+                    {
+                        ++Position;
+                    }
+                    return {
+                        EToken::Word,
+                        TString(Input.substr(offset, Position - offset)),
+                        offset};
+                }
+                return {EToken::End, {}, offset};
```
szumal: ## [P1] Unknown characters silently terminate parsing.

[parser.cpp:112] returns End for every unrecognized character. For example, SELECT * FROM NodeRefs @ WHERE child_id = 1 is accepted as an unfiltered SELECT * FROM NodeRefs, silently discarding the intended condition. Unknown characters need a distinct invalid token that causes parsing to fail.

cloud/filestore/apps/client/lib/query/parser.cpp
```diff
@@ -0,0 +1,318 @@
...
+                const char quote = ch;
+                TString value;
+                while (Position < Input.size() && Input[Position] != quote) {
+                    if (Input[Position] == '\\' && Position + 1 < Input.size())
+                    {
+                        ++Position;
+                    }
+                    value += Input[Position++];
+                }
+                if (Position == Input.size()) {
```
szumal, line 81: ## [P2] Unterminated strings become valid empty strings.
 At [parser.cpp:80], reaching EOF before the closing quote returns a normal String token with an empty value. Thus WHERE name = 'hello succeeds as name = ''. This should produce a lexical error.

cloud/filestore/apps/client/lib/query/parser.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+        if (AcceptWord("WHERE") && !ParseConditions(result.Conditions)) {
+            return Nothing();
+        }
+
+        if (AcceptWord("LIMIT")) {
+            if (Current.Type != EToken::Number) {
+                return Fail("expected non-negative LIMIT");
+            }
+            result.Limit = FromString<ui64>(Current.Text);
```
szumal: ## [P2] Numeric overflow escapes the parser contract as an exception. 
Both [LIMIT](https://github.com/ydb-platform/nbs/pull/7269/changes#diff-0681e3c91dd339b5b0e100ff643061ffea3717706c6be1ac1fb86ebf21df0f11R179) and [condition values](https://github.com/ydb-platform/nbs/pull/7269/changes#diff-0681e3c91dd339b5b0e100ff643061ffea3717706c6be1ac1fb86ebf21df0f11R283) use throwing FromString<ui64>. A sufficiently long numeric literal throws instead of returning Nothing() and filling TParseError. Use a checked/non-throwing conversion or catch the conversion error.

cloud/filestore/apps/client/lib/query/parser.cpp
```diff
@@ -0,0 +1,318 @@
...
+                    return false;
+                }
+            } else if (AcceptWord("IN")) {
+                condition.Predicate = EPredicate::In;
+                if (!Accept(EToken::OpenBrace)) {
+                    Fail("expected '{' after IN");
+                    return false;
+                }
+                if (Current.Type == EToken::CloseBrace) {
+                    Fail("IN list must not be empty");
```
szumal, line 264: ## [P2] The motivating IN {} form is explicitly rejected. 
[Issue #7199](https://github.com/ydb-platform/nbs/issues/7199) gives SELECT * FROM NodeRefs WHERE child_id in {} as its example, while this rejects an empty list and the unit test locks that behavior in. Either support the documented form or clarify/change the requirement.

szumal (commented):
CODEX REVIEW

cloud/filestore/apps/client/lib/query/parser.cpp
szumal: `SELECT * FROM NodeRefs WHERE child_id in {}` was given as an example. Don't expect anyone to query for an empty set.

cloud/filestore/apps/client/lib/query/ut/parser_ut.cpp
```diff
@@ -0,0 +1,77 @@
+#include <cloud/filestore/apps/client/lib/query/parser.h>
+
+#include <library/cpp/testing/unittest/registar.h>
+
+namespace NCloud::NFileStore::NClient::NQuery {
+
```
qkrorlqr, line 6: nit: ///////... before declarations/definitions plz (in this case - before Y_UNIT_TEST_SUITE)

cloud/filestore/apps/client/lib/query/ut/parser_ut.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    Y_UNIT_TEST(ShouldParseSelectAll)
+    {
+        auto query = Parse(
+            "SELECT * FROM NodeRefs WHERE child_id IN {1, 2, 3} LIMIT 10");
+
+        UNIT_ASSERT(query);
+        UNIT_ASSERT_VALUES_EQUAL(query->Table, "NodeRefs");
+        UNIT_ASSERT(query->Columns.empty());
+        UNIT_ASSERT_VALUES_EQUAL(query->Conditions.size(), 1);
```
qkrorlqr, line 17: nit: we usually put `expected` first and `actual` - second

cloud/filestore/apps/client/lib/query/ut/ya.make
```diff
@@ -0,0 +1,9 @@
+UNITTEST_FOR(cloud/filestore/apps/client/lib/query)
```
qkrorlqr, line 1: why in client/lib? I thought we would do query execution at the tablet side

cloud/filestore/apps/client/lib/query/parser.cpp
```diff
@@ -0,0 +1,334 @@
+#include "parser.h"
+
+#include <util/string/cast.h>
+
+#include <cctype>
+
+namespace NCloud::NFileStore::NClient::NQuery {
+
+namespace {
+
```
qkrorlqr, line 10: let's describe the grammar and generate the parser via something like lex + yacc/bison then

or ragel

